PR #2424, opened 2023-07-20
ENG-1703 Mostly Everything is a Node or an Edge

jhelwig:


lib/dal/src/workspace_snapshot/lamport_clock.rs
```diff
@@ -0,0 +1,69 @@
...
+    #[error("Change Set error: {0}")]
+    ChangeSet(#[from] ChangeSetError),
+}
+
+pub type LamportClockResult<T> = Result<T, LamportClockError>;
+
+#[derive(Clone, Copy, Deserialize, Serialize)]
+pub struct LamportClock {
+    #[serde(with = "chrono::serde::ts_nanoseconds")]
+    pub counter: DateTime<Utc>,
```
zacharyhamm, line 21: Curious question: my understanding of lamport clocks is that they use a logical clock (just a simple counter). We're incrementing by calling `::now()`, but there's no real guarantee that the clock is monotonically increasing, especially in cloud deployments (it's perfectly possible for ::now to actually be less than or the same as the last tick, even if it's extremely rare, even if rust tries to ensure that it is always greater than it). Is that a problem here? It might be a problem with my understanding of the use we're making of it!

jhelwig: Disclaimer: This is going to be a bit rambling/disorganized, and likely incomplete, since I'm trying to remember/relay several weeks of iterations on designs/behaviors into something that isn't baking an apple pie from scratch by first inventing the universe.

In a textbook Lamport Clock, you're correct that they're nothing more than a locally generated, monotonically increasing integer. Where things get a bit trickier is that with a textbook Vector/Lamport Clock (at least from my understanding) there is no inherent conflict detection, and it's always "last received event wins", but we definitely want to detect conflicting updates. There's also that trying to structure things so that all updates are more "pure" events (like Vector Clocks are designed for) gets a bit eldritch when trying to think through all of the implications. So, it's being called Vector/Lamport Clock, but it isn't quite (at least not according to the textbook definitions).

So, stepping back from that for a bit... Some of the things we're trying to achieve:
* Being able to do garbage collection without having to compare the entire graph of every workspace snapshot against every other (active/valid) workspace snapshot. I believe having to do this kind of comparison would be necessary if we were to use tombstone markers for deletion like we currently do in the DB. If we didn't want the snapshots to have to store the fact that something (or even a relationship) existed at some point for all of time, we'd need to be able to figure out when it's safe to remove the marker/tombstone by doing this kind of in-depth comparison to see if anything else still knows about the tombstoned thing/relation.
* When doing things like rebasing one change set (`A`) onto another one (`B`) (which is how both keeping "active/open" change sets up to date with "main"/`HEAD`, and merging a change set into another one would work), being able to tell the difference between `B` has something that `A` does not because it's new in `B`, and because `A` actively removed it. With the follow-on question of whether or not `B` incorporated changes to the thing since `A` was last "in-sync" with `B` (has it been modified since `A` would have known about it, which would be a conflict).

Originally, we were using ULIDs for the Lamport Clocks, since we mostly cared about "are these sortable for doing relative comparisons", even between machines & vector clock entries. While it is possible for clocks to move backwards, it should be within tolerances with real-world data propagation times and reasonably stable/synchronized clocks (the first 1/2 of a ULID is a timestamp accurate to milliseconds, and the second 1/2 is semi-random). Clock stability/synchronization is most critical for the multi-player aspects of making changes to the same workspace, and I believe that the tolerances should be much more generous for things like module import/export, and sharing data between organizations/workspaces.

The reason we switched from using ULID to a nanosecond timestamp was mostly due to non-determinism in tests (that are very similar scenarios to multiple people editing different change sets of the same workspace). ULID are only guaranteed sortable within the same millisecond if they come from the same generator, we can't guarantee that change sets use the same generator. This was causing some issues for us in tests where operations on multiple change sets were happening within the same millisecond. We would set up a snapshot in change set `A`, do some changes in `B` (based on `A`), do some changes in `A`, and the comparisons we were making to try to determine where (and more importantly why) there were differences between `A` & `B` weren't what we expected, because change sets `A` & `B` couldn't be expected to share the same ULID generator, and the entire sequence of events was happening in <= 1 millisecond. Using `DateTime<Utc>`, got things to be stable as that got the resolution down to 1 nanosecond.

It's entirely possible that we may end up not needing to do the comparisons the way we are. We've learned a lot already, but we're still learning things as the implementation(s) get more concrete.

nickgerace:
This branch has been closed in favor of https://github.com/systeminit/si/pull/2879, which will be merged into `main` directly instead of `eng-1703-mostly-everything-is-a-node-or-an-edge`.

I will always remember this branch...
https://www.youtube.com/watch?v=lpZiPZwwXhM

PR #2572, opened 2023-08-14
feat(dal): Adding the first of an AWS IAM pkg

stack72:
This asset was entired authored via the function editor, submitted to the module index and tested to ensure working for create, refresh, delete and 
qualifications. It is the first module who's qualifications are tests to ensure validity of the object being sent to AWS since no `dry-run` exists for the AWS IAM 
part of the CLI

Related: https://github.com/systeminit/si/issues/2647


stack72:
bors merge

PR #2599, opened 2023-08-16
feat(si): Ensure that the SI binary can update it's containers

stack72:
- fix(auth-api): Reenable the container updates from the github releases API
- fix(si): Ensure we are comparing the container version vs release digest from the auth-api for updates

there is an order of operations here - we need to release the SI binary *first* to get the good update behaviour in place then we can launch the auth-api


stack72:
bors merge

Issue #2609, opened 2023-08-16
Docker image exists qualification failure during tutorial


Octogonapus:
### What happened?

I'm following the tutorial and an unable to get the docker image qualification test to pass.
I get this qualification failure:
```
time="2023-08-16T16:33:50Z" level=fatal msg="Error parsing image name \"docker://systeminit/whiskers\": unable to retrieve auth token: invalid username/password: unauthorized: your account must log in with a Personal Access Token (PAT) - learn more at docs.docker.com/go/access-tokens"
```

I've supplied a PAT with all scopes. I've tested this PAT using `docker login -u <name> -p <PAT>`.

### How can we reproduce this?

The steps I took were:
1. Install and launch SI. Supply a Docker password instead of a PAT (erroneously, because I forgot)
2. Follow the tutorial until you are expected to get the Docker image qualification to pass
3. Observe the qualification fails with the above message
4. Run `si configure -f` and supply a Docker PAT instead
5. Run `si restart`
6. Kick the Docker image name and observe the qualification still fails with the same message

### Operating system

Linux

### Architecture

x86_64 / amd64

### What browsers are you seeing the problem on?

Firefox

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Octogonapus:
I managed to fix this by:

1. `si delete`
2. `si install`
3. `si launch`

It seems something in a container volume was hanging on to the old credentials

PR #2610, opened 2023-08-16
fix(si): Ensure that we can recreate a new veritech container

stack72:
Fixes: #2609

When the credentials are incorrect, we want to be able to stop
and then restart the container to get the new credentials passed
to the container


fnichol (approved):
Ok, that should do us, nice turnaround!

fnichol:
bors merge

PR #2618, opened 2023-08-16
fix(si): Ensure that we have a default email in place when using the launcher

stack72:
There was a chicken and egg scenario that the data_dir we store creds
in may not exist until we write the creds or the keys BUT we were trying
to read the SI email from it

We now default to sally@systeminit.com if it's not found


nickgerace (approved):

stack72:
bors merge

stack72:
bors retry

Issue #2627, opened 2023-08-17
local-sdf-1 container won't start when selinux is enforcing


rothgar:
### What happened?

I was getting an error when running `si start`

```
Error:
   0: docker api: error 500 Internal Server Error - Cannot link to a non running container: /local-sdf-1 AS /local-web-1/sdf
   1: error 500 Internal Server Error - Cannot link to a non running container: /local-sdf-1 AS /local-web-1/sdf
```

The container logs it says

```
Error:
   0: file not found: /run/sdf/jwt_signing_public_key.pem
Location:            
   <unknown>
```

I verified the container mount and saw that the mount exists 

```
"Type": "bind",
                "Source": "/home/jgarr/.local/share/SI/jwt_signing_public_key.pem",
                "Destination": "/run/sdf/jwt_signing_public_key.pem",
                "Mode": "",
                "RW": true,
                "Propagation": "rprivate"
```

The file exists in my home directory but I realized this machine has selinux in enforcing mode
```
ls -Zl /home/jgarr/.local/share/SI/jwt_signing_public_key.pem
-rw-r--r--. 1 jgarr jgarr unconfined_u:object_r:data_home_t:s0 799 Aug 16 21:43 /home/jgarr/.local/share/SI/jwt_signing_public_key.pem
```
I set selinux to permissive and restarted the containers and it worked

### How can we reproduce this?

Run `si start` on a system with selinux enabled.

### Operating system

Linux

### Architecture

x86_64 / amd64

### What browsers are you seeing the problem on?

Chrome

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Additional context

I know the fix is to add `:Z` to the volume here https://github.com/systeminit/si/blob/7c15b746cc16de0e39180915d8150a69fe5f285d/lib/si-cli/src/cmd/start.rs#L387-L389 but I'm not at a computer with dev tools installed so I can PR it later.

PR #2631, opened 2023-08-17
fix(si): Ensure we pass :Z to the sdf container mounts

stack72:
Fixes: #2627

:Z indicates the bind mount is private which is required for systems
with selinux enabled

I tested this on a non-selinux enabled system and all works as expected


stack72:
bors merge

Issue #2629, opened 2023-08-17
unable to connect to the docker engine


NickLarsenNZ:
### What happened?

I ran `si start`, entered credentials, then got an error:

```
System Initiative Launcher is in "local" mode


Unable to retrieve updates: docker api: error trying to connect: Connection refused (os error 61)
System Initiative needs some credentials in order to be able to interact with AWS and Docker.
The credentials are never sent back to System Initiative and can be inspected at the location:
/Users/nick/Library/Application Support/SI/si_credentials.toml

Checking that the system is able to interact with the docker engine to control System Initiative...
Error:
   0: unable to connect to the docker engine

Location:
   bin/si/src/main.rs:113

Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
```

When I check with `docker ps`, I can see docker is running:

```
❯ docker ps
CONTAINER ID   IMAGE     COMMAND   CREATED   STATUS    PORTS     NAMES
```

I am running Colima,

```
echo $DOCKER_HOST
unix:///Users/nick/.colima/default/docker.sock
```

### How can we reproduce this?

I'm unsure, but probably:

1. Use Colima, `colima start`
2. `si start`

### Operating system

macOS

### Architecture

x86_64 / amd64

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #2632, opened 2023-08-17
fix(si): Allow a user to pass a custom location to a docker.sock file

stack72:
Fixes: #2629

```
SI_DOCKER_SOCK=/var/run/docker.sock buck2 run //bin/si check
Checking for user supplied docker.sock


Launcher update found, please run `si update` to install it

Checking that the system is able to interact with the docker engine to control System Initiative...
```


bin/si/src/main.rs
```diff
@@ -41,14 +47,20 @@ async fn main() -> Result<()> {
...
-        .iter()
-        .find(|candidate| candidate.exists())
-        .ok_or(eyre!(
+    let docker: DockerClient;
+    if let "" = docker_sock.as_str() {
+        let socket = docker_socket_candidates
+            .iter()
+            .find(|candidate| candidate.exists())
+            .ok_or(eyre!(
             "failed to determine Docker socket location; candidates={docker_socket_candidates:?}"
```
zacharyhamm, line 56: Might make sense to add a note to this error about passing in --si-docker-sock or setting the environment variable

stack72:
```

Error:
   0: failed to determine Docker socket location. Set a custom location using `--docker_sock` or `SI_DOCKER_SOCK`; candidates=["/Users/stack72/.docker/run/docket.sock"]

Location:
   <unknown>
   ```

File: bin/si/src/main.rs [hunk 2]
```diff
@@ -37,18 +43,25 @@ async fn main() -> Result<()> {
...
-        .find(|candidate| candidate.exists())
-        .ok_or(eyre!(
-            "failed to determine Docker socket location; candidates={docker_socket_candidates:?}"
+    let docker: DockerClient;
+    if let "" = docker_sock.as_str() {
+        let socket = docker_socket_candidates
+            .iter()
+            .find(|candidate| candidate.exists())
+            .ok_or(eyre!(
+            "failed to determine Docker socket location. Set a custom location using `--docker_sock` \
```
zacharyhamm, line 56: Pretty sure clap will turn that into `--docker-sock` (underscores become dashes)

stack72: you are correct!

stack72:
bors merge

PR #2633, opened 2023-08-17
chore(build): Remove duplicate GHA bot from PRs

stack72:


fnichol (approved):
Seems reasonable!

stack72:
bors merge

.github/workflows/pull-request.yml
nickgerace, line 1: I like the friendliness to this message that I don't think the `systeminitbot` is doing yet. There's something here that would be nice to bring over. I'll think on it!

Issue #2626, opened 2023-08-17
crash when starting if $HOME/.local (and friends) don't exist


NeilHanlon:
### What happened?

When starting SI in a new WSL environment with an empty $HOME directory, I receive the following error and crash. I've included the -p option as a debugging step to see where the code is failing.

```
[neil@NEIL-DESKTOP SI]$ RUST_BACKTRACE=full COLORBT_SHOW_HIDDEN=1 si -p start
System Initiative Launcher is in "local" mode



Preview mode... System Initiative would have taken the following actions
Error:
   0: io: No such file or directory (os error 2)
   1: No such file or directory (os error 2)

Location:
   bin/si/src/main.rs:113

  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ BACKTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
   1: color_eyre::config::EyreHook::into_eyre_hook::{{closure}}::h41997158edf4f9a5
      at <unknown source file>:<unknown line>
   2: eyre::capture_handler::h08c762ed93877fd5
      at <unknown source file>:<unknown line>
   3: eyre::error::<impl core::convert::From<E> for eyre::Report>::from::h02239bbd19841cd7
      at <unknown source file>:<unknown line>
   4: si::main::{{closure}}::hc959437c33741098
      at <unknown source file>:<unknown line>
   5: tokio::runtime::park::CachedParkThread::block_on::h8657001296f26bdc
      at <unknown source file>:<unknown line>
   6: tokio::runtime::scheduler::multi_thread::MultiThread::block_on::hc81638d8fa067ff8
      at <unknown source file>:<unknown line>
   7: tokio::runtime::runtime::Runtime::block_on::haa13a5c81da961a0
      at <unknown source file>:<unknown line>
   8: si::main::h9c3529cb47f1a178
      at <unknown source file>:<unknown line>
   9: std::sys_common::backtrace::__rust_begin_short_backtrace::hb6319b36f3c9c391
      at <unknown source file>:<unknown line>
  10: std::rt::lang_start::{{closure}}::hadca2ee8395f889c
      at <unknown source file>:<unknown line>
  11: std::rt::lang_start_internal::h76f3e81e6b8f13f9
      at <unknown source file>:<unknown line>
  12: main<unknown>
      at <unknown source file>:<unknown line>
  13: __libc_start_call_main<unknown>
      at <unknown source file>:<unknown line>
[...]

PR #2634, opened 2023-08-17
fix(si): Ensure that the entire SI data directory path exists or create it

stack72:
Fixes: #2626


fnichol (approved):
This was a fun find, using `strace` as a debug buddy!

stack72:
bors merge

PR #2635, opened 2023-08-17
feat(si): Expose the versions of the running containers in the status command

stack72:
Screenshot 2023-08-17 at 20 51 14


stack72:
bors merge

PR #2637, opened 2023-08-17
New deploy modal for auth-portal

theoephraim:
This adds functionality which checks the currently running version (using the the script included in the rendered html file) against the latest (using the manifest.json produced during the build).

If the versions dont match, we pop up a modal prompting the user to please refresh.

This will likely need future work, but this is a good start.

image


theoephraim:
bors merge

PR #2638, opened 2023-08-17
Add arguments to set host and port for web container

NeilHanlon:
This change adds the `--host` and `--port` arguments to the `si` command which allow the user to set the host and port for the user-facing web container. This allows a user to run the portal on one machine, but access it from another one (e.g.).

This is (one of) my first rust contribution(s), so I'm more than happy to address any and all issues.

I considered adding this to the Start command's arguments (and state), but figured it rather belonged in the global AppState so it could be used by other commands, e.g., the Restart and Status commands, to show the proper information to the user.

Thank you in advance for your consideration.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -2,22 +2,29 @@
 
 We contributors to System Initiative:
 
-* License all our contributions to the project under the Apache License, Version 2.0
-* Have the legal rights to license our contributions ourselves, or get permission to license them from our employers, clients, or others who may have them
-* Add our names and GitHub handles to this CONTRIBUTORS.md file to create a permanent record that users, distributors, and other contributors can all rely on
+- License all our contributions to the project under the Apache License, Version
+  2.0
+- Have the legal rights to license our contributions ourselves, or get
+  permission to license them from our employers, clients, or others who may have
+  them
+- Add our names and GitHub handles to this CONTRIBUTORS.md file to create a
+  permanent record that users, distributors, and other contributors can all rely
+  on
 
------------
-* Adam Jacob (@adamhjk)
-* Anna Saulwick (@AnnaAtMax)
-* Brit Myers (@britmyerss)
-* Fletcher Nichol (@fnichol)
-* Jacob Helwig (@jhelwig)
-* Mahir Lupinacci (@mahirl)
-* Nick Gerace (@nickgerace)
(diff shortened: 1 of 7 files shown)


bin/si/src/args.rs
```diff
@@ -50,6 +50,15 @@ pub(crate) struct Args {
     /// Show a preview of what the System Initiative Launcher will do
     #[arg(long, short = 'p', default_value = "false")]
     pub is_preview: bool,
+    
+    /// Allows starting the service and binding to a specific IP
+    #[arg(long = "host", default_value = "127.0.0.1")]
+    pub bind_host: String,
+
+    /// Allows starting the service and binding to a specific port
+    #[arg(long = "port", default_value = "8080")]
```
stack72, line 59: Is this for the web port or the sdf port or both?

NeilHanlon: web port only -- probably can make that more clear by specifying it precisely, but I wasn't sure what (if any) strategy there should be more holistically for overriding/customizing all the ports and binds.

stack72: I think that would be a good distinction to make so that we can then expose the ability to do the same with SDF :) Maybe web_port would be a good name?

CONTRIBUTORS.md
```diff
@@ -2,22 +2,29 @@
...
+- Mahir Lupinacci (@mahirl)
+- Nick Gerace (@nickgerace)
+- Paul Stack (@stack72)
+- Paulo Cabral (@paulocsanz)
+- Theo Ephraim (@theoephraim)
+- Victor Bustamante (@vbustamante)
+- Wendy Bujalski (@wendybujalski)
+- Zack Hamm (@zacharyhamm)
+- Dan Miller (@jazzdan)
+- Neil Hanlon (@NeilHanlon)
```
zacharyhamm, line 30: Did changing the markdown from `*` to `-` break the contributors check here?

stack72:
ok, I am going to fix this up and merge it!

stack72:
I squashed this PR into https://github.com/systeminit/si/pull/2640 - closing this out now and will merge #2640

CONTRIBUTORS.md
NeilHanlon: oops - sorry about that. will check my linter/formatter

PR #2641, opened 2023-08-17
Clone func when cloning variant def (ENG-1821)

nickgerace:
This commit clones the scaffolding func when cloning the schema variant def to retain 1:1 relationship constraints.

Without this commit, the func would be re-used from the existing variant def. Accessor queries will fail when that happens, as observed when executing "exec_variant_def" after the clone. Specifically, this error is observed when cloning an asset and immediate creating the clone.

### DIFF
--- a/lib/dal/src/func.rs
+++ b/lib/dal/src/func.rs
@@ -10,9 +10,10 @@ use thiserror::Error;
 
 use crate::func::argument::FuncArgumentError;
 use crate::{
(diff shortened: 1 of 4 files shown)


zacharyhamm (approved):
Kudos for taking the time to add the test

nickgerace:
Kudos to you for sticking with me in agony as I forgot to await the futures 🤦‍♂️

nickgerace:
bors merge

PR #2645, opened 2023-08-18
[WIP] Add a Dockerfile for the si cli

Tensibai:
A first attempt at creating a docker image for the si cli so I could run it on Ubuntu20 where the glibc is a bit too old (3.21).

Running as root, it works using the following command line `docker run --rm --privileged -v /var/run/docker.sock:/var/run/docker.sock -v /root/.local/share/:/root/.local/share/ -it si:latest <command>`

In bash adding an alias `alias si='docker run --rm --privileged -v /var/run/docker.sock:/var/run/docker.sock -v /root/.local/share/:/root/.local/share/ -it si:latest'` works well.

I wasn't sure where to document it, so opening this PR as first step to discuss it.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -22,3 +22,4 @@ We contributors to System Initiative:
 * Zack Hamm (@zacharyhamm)
 * Dan Miller (@jazzdan)
 * Neil Hanlon (@NeilHanlon)
+* Tensibai Zhaoying (@Tensibai)
--- a/bin/si/Dockerfile
+++ b/bin/si/Dockerfile
@@ -1,0 +1,6 @@
+FROM bitnami/minideb:latest
+
+RUN install_packages curl ca-certificates
+RUN curl -sSfL https://auth.systeminit.com/install.sh | sh
+
+ENTRYPOINT [ "/usr/local/bin/si" ]


keeb:
This is a great start, I have verified it works on my machine.

Questions for the SI team: which base image is default for si-based docker images? 

Some ideas about running as root and for `~/.local/share` 

* `--uid` can be used to do user mapping so you can run as your normal user
* using VOLUME for ~/.local/share and having a mapping to that would be ideal.

Maybe also look at including a docker-compose or Tilt file.

Tensibai:
Fair points.

The source image is to have tooling around (like curl 1'd glibc). I've tried Google distroless base with debug tag and alpine and the setup wasn't smooth.

The UID part is something I thought about, I was unsure how it would play out and didn't tested that right away.
That would require a bit more complex alias/command too.

The volume part however would prevent the use of the binary on itself for 'normal' cases, I don't think that's desirable.

This PR is more to open the talk and provide a workaround in this early stage, I believe the si binary should be statically linked to avoid is requirements as much as possible.

adamhjk:
Yeah - I think we're probably unlikely to actually distribute the launcher in a container like this - especially given the fact that how it works today would require some hijinks to make work well (since it launchers more docker containers from within a docker container). 

@fnichol is looking in to how to bring some resolution to the broader issue (that we should be statically linking/using an older glibc.) 

Is there another use case for having the launcher wrapped in a container, other than isolating the libc we used?

fnichol:
Sorry about being a little late on this thread, yes to what @adamhjk was talking about. I was spending some time seeing if we can't statically link our launcher binary so there wouldn't be a libc dynamic load at runtime (similar to most Go binaries). In the Rust world we typically compile to the `x86_64-unknown-linux-musl` or `aarch64-unknown-linux-musl` target triple. The default behavior for this target is to statically link so as long as we're careful with our crate dependencies (and we were sensitive to this possibility), it's usually not a hard task to pull off.

The extra interesting challenge in our current setup is that we'll be building this in our Nix flake environment and as it's already setup to compile and link to glibc I think we'll be asking Nix to cross-compile to a musl libc derivation (or something close to that). At the moment, only our binary needs to build this way which is why I'm hopeful we can ship this in short order. If that proves trickier than expected we can still try to cross compile with a much older glibc and rely on lib's aggressive backwards compat to work on newer systems as well.

keeb:
Thanks for the contribution. Given the feedback from Adam and Fletcher, we are going to close this and address the glibc problem separately.

PR #2646, opened 2023-08-18
fix(sdf): Ensure that we pass through identity func when using valueFrom in socket definitions

stack72:


nickgerace (approved):

zacharyhamm (approved):
:+1: 

the si-cli stuff is a little extraneous but harmless. Are we not running the same checks (like the format check) on that as the other paths?

stack72:
bors merge

bin/si/src/args.rs
```diff
@@ -130,8 +129,7 @@ pub(crate) struct StatusArgs {
 }
 
 #[derive(Debug, clap::Args)]
-pub(crate) struct StartArgs {
-}
+pub(crate) struct StartArgs {}
```
fnichol, line 132: I noticed that the original change here didn't trigger a `check-format` for `bin/si`. I can look into this

PR #2649, opened 2023-08-18
fix(dal): simplify region asset

zacharyhamm:
Removes the region qualification (which does not fire since qualifications only run on a change of value). Replaces it with a validation that requires "region" to be set. However, this validation also does not fire until a value changes and the UX does not permit you to reset the region to a null value. For that reason as well, the "si:getRegion" function has been dropped, since we are setting the default value of /root/si/name on every schema to the be the name of the schema, and we can't unset region, so all that's required is the identity function mapping /root/domain/region onto /root/si/name.

Also removes the superfluous "si:awsRegionRefreshAction", which is never executed since regions do not have a resource (without a create action, refresh will never be triggered).

Closes ENG-1824 and ENG-1802 and kicks the can down the road on the issue of validations and qualifications on component creation.

NOTE: this is a breaking change. When SDF is restarted after this it will fail to migrate if the database is not also cleared out with `si delete`.


zacharyhamm:
bors try

PR #2650, opened 2023-08-18
feat(si): Enable the use of debug logs for veritech SI container

stack72:
```
si --with-function-debug-logs start
```


fnichol (approved):
Seems like a good approach for the time being and should get the job done!

fnichol:
bors merge

PR #2654, opened 2023-08-18
fix(buck2): when computing dependent targets, use `attrregexfilter`

fnichol:
This change fixes the dependent targets calculation used in CI/CD to look for target names *starting with* the appropriate phase names such as `check-format`, `check-lint`, etc.

Prior to this change, only an exact target name match of `check-format` would return a result. As we start to use more `test_suite` groups to group up similar tests to run under one convenient target we have started to generate such target names as `check-format-rust-bin` or `check-lint-rust-unit`. It appears as though the `test_suite` Buck2 rule type does *not* form a dependency  relationship with its inputs and therefore the rolled up target names such as `check-format` or `check-lint` were not being selected to run in a pipeline.


fnichol:
bors merge

PR #2655, opened 2023-08-18
fix(web): prevents anchor click from hard reloading

paulocsanz:
Prevents the anchor click from propagating when closing a function tab, as it was was forcing a hard reload of the page.


paulocsanz:
bors merge

PR #2664, opened 2023-08-21
fix(sdf): fix nightly ci

zacharyhamm:
Whiskers nightly test was checking for the dropped region qualification and failing.


stack72 (approved):

fnichol:
Oh that's perfect, I'm glad there was an explanation

PR #2666, opened 2023-08-21
fix(si): Ensure that we mount the SI data_dir with the correct mount options

stack72:


vwbusguy:
This fixed pinga for me.  I can log in now with SELinux enforcing.  Thanks!

stack72:
Thanks for testing this for us @vwbusguy

stack72:
bors merge

PR #2670, opened 2023-08-21
fix(si): Ensure SI containers restart on failure

stack72:
This will ensure any race conditions are fixed by a restart


fnichol (approved):
Looks like this should help until we make our services a touch more patient/resilient to the boot-up race

stack72:
bors merge

PR #2676, opened 2023-08-22
chore(build): Ensure that the logic regarding the community check is correct and avoid unnecessary labelling

stack72:


stack72:
bors merge

PR #2681, opened 2023-08-22
fix(si): Ensure that running an update when no containers installed doesn't trigger start

stack72:
ENG-1886
ENG-1887

- feat(si): Ensure we capture the name of the containe engine when tracking usage
- fix(si): Ensure that running an update when no containers installed doesn't trigger start

### DIFF
--- a/lib/si-cli/src/cmd/update.rs
+++ b/lib/si-cli/src/cmd/update.rs
@@ -114,21 +114,23 @@ impl AppState {
         let current_containers = self.container_engine().get_container_details().await?;
 
         let mut containers = Vec::new();
-        let latest_containers: Vec<LatestContainer> = req.json().await?;
-        'outer: for latest in latest_containers {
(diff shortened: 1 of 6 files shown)


lib/si-cli/src/cmd/update.rs
```diff
@@ -114,21 +114,23 @@ impl AppState {
         let current_containers = self.container_engine().get_container_details().await?;
 
         let mut containers = Vec::new();
-        let latest_containers: Vec<LatestContainer> = req.json().await?;
-        'outer: for latest in latest_containers {
-            for current in &current_containers {
-                if current.image != format!("{}/{}", latest.namespace, latest.repository) {
-                    continue;
-                }
+        if !current_containers.is_empty() {
```
stack72, line 117: We don't want to run through ANY updates if there are no containers currently installed

fnichol:
bors merge

PR #2682, opened 2023-08-22
feat(web) detect cached webapp running and prompt user to reload

theoephraim:
very similar to the new deploy notification on the auth portal, except this detects a cached version of the web app running compared to what is now available on disk via the si launcher. This could potentially happen if the user left their browser open while running an `si update`

image


stack72 (approved):

theoephraim:
bors merge

Issue #2606, opened 2023-08-16
[BUG-24] Can't copy text from an qualification failure


Octogonapus:
### What happened?

I am trying to copy the text from a qualification failure, but when I go to select it with my mouse cursor, the window showing the failure closes.

### How can we reproduce this?

1. Open a qualification failure using the diagram outline panel (not the bottom bar) (mine was `Docker Image Exists`)
2. Use your mouse to select the failure text
3. The window closes when you click

### Operating system

Linux

### Architecture

x86_64 / amd64

### What browsers are you seeing the problem on?

Firefox

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

<sub>[BUG-24](https://linear.app/system-initiative/issue/BUG-24/cant-copy-text-from-an-qualification-failure)</sub>

adamhjk:
Hi! We'll work on making it selectable from the outline panel - thanks!

In the meantime, if you open the qualification from the sub-panel at the bottom, you should be able to select it.

Octogonapus:
I get the feeling this might be related to https://github.com/tailwindlabs/headlessui/issues/2535 but I've been trying the listed fixes for a while and I can't get it working. Leaving this to someone more knowledgeable...

PR #2684, opened 2023-08-22
have modal capture mousedown events

theoephraim:
when the modal is open, it is totally in charge, so I think it makes sense as a general rule to capture the mousedown events. This means you must first dismiss the modal before something else that may be listening for clicks will fire.

specifically this fixes https://github.com/systeminit/si/issues/2606


stack72 (approved):
thank you!!

PR #2685, opened 2023-08-22
fix(sdf-server): Ensure standard name for Asset scaffold funcs

stack72:
When an asset is scaffolded via the UI, we give the func a name like
`new asset 2703`. When we package the asset, it doesn't make sense
to include a func with that name. Therefore, when the exec_variant_def
is called, we now generate the func name as:

myTestFuncScaffold_202308222128

We are camelcasing the assetname, adding `Scaffold` to it and appending
the date in year month day hour minute format. We need to ensure that funcs
are unique and generating to minute should be unique enough


zacharyhamm (approved):
👍 glad this was easy

stack72:
bors merge

PR #2689, opened 2023-08-23
fix(si): support self update through symbolic links on macOS

fnichol:
When the launcher (i.e. the `si` program) is installed from a non-root user and when they don't have either `$HOME/.local/bin` or `$HOME/bin` currently in their `$PATH`, the launcher is installed to `$HOME/.si/bin/si` and a symlink is created at `/usr/local/bin/si` with a sudo call (that is, the root user owns the symlink as it is a system directory).

Prior to this change, when a self update was attempted (via calling `si update --self) under the scenario above, the new binary would attempt to write itself to a temp file under the *symlink*'s  parent directory of `/usr/local/bin`. This would cause an permissions error that looks like:

```
Downloading new binary
Replacing '/usr/local/bin/si' with new binary
Error:
   0: io: Permission denied (os error 13) at path "/usr/local/bin/.si.__temp__6DqCUt"
   1: Permission denied (os error 13) at path "/usr/local/bin/.si.__temp__6DqCUt"

Location:
   <unknown>

```

Additionally, if the non-root user happened to have write permission on `/usr/local/bin`, then the new binary would overwrite the symlink and install itself to `/usr/local/bin/si`, thus eliminating the symlink.

The detection of a symlink and its resolution by calling `std::fs::read_link` is provided via a fix in the upstream [self-replace](https://crates.io/crates/self-replace) crate.

References: mitsuhiko/self-replace#18

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4517,8 +4517,7 @@ dependencies = [
 [[package]]
 name = "self-replace"
 version = "1.3.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a0e7c919783db74b5995f13506069227e4721d388bea4a8ac3055acac864ac16"
+source = "git+https://github.com/systeminit/self-replace.git?branch=unix-current-exe-read-link#702afa3e3ccc1d683bd7141a56f2dd4459f979f0"
 dependencies = [
  "fastrand",
  "tempfile",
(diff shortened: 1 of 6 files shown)


fnichol:
I made sure to test this out on macOS by building the binary and getting the path to it with `buck2 build bin/si --show-output`, then making a symlink to it owned by root (i.e. `chown 0:0`) in a directory owned by root (so my user wouldn't be able to write the tempfile in the symlink's parent directory). That's how I was able to reproduce the error message in the commit before making the code change.

fnichol:
bors merge

fnichol:
I'll add that for anyone affected by this issue, you'll want to re-run the shell download script with:

```sh
curl -sSfL https://auth.systeminit.com/install.sh | sh
```
 
Or re-downloading a new release from the [download page](https://auth.systeminit.com/download).

PR #2706, opened 2023-08-25
feat(web,sdf,auth-api) recover more gracefully from a reset database

theoephraim:
this change helps us recover more gracefully from the database being empty - a situation that is now very common due to the way our launcher is setup and how we trash our containers.

The user remains logged in with a valid token, but the workspace (and user) do not exist in this freshly migrated db yet. Previously that initialization happened only in the handoff between the auth system and sdf.

Now instead we have an additional endpoint which can use the existing auth token and just handle this re-initialization. Note that we could do it on every reload (rather than just fetching from sdf which we do now) but this would add extra overhead to that call. So instead we first try to reestablish auth with sdf only, and then if we detect the situation of the workspace not existing, we reach out to the auth api to try to reinitialize it.

This means the flow is now that the user will be thrust back into an empty workspace again without any explanation, which I believe may feel a bit jarring. If we don't like how it feels, we can interject an additional screen which explains what is going on, and gives them the opportunity to log out, restore from a backup, etc...

NOTE - needs a little cleanup before merging

### DIFF
--- a/app/web/src/App.vue
+++ b/app/web/src/App.vue
@@ -1,14 +1,26 @@
 <!-- eslint-disable vue/no-multiple-template-root -->
 <template>
   <div class="font-sans">
-    <RealtimeConnectionStatus />
-    <router-view :key="selectedWorkspace?.pk" />
-    <Teleport to="body">
-      <canvas
-        id="confetti"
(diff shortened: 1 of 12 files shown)


theoephraim:
bors merge

PR #2707, opened 2023-08-28
fix(auth-portal): Adding extra podman machine installation instructions

stack72:
Screenshot 2023-08-28 at 19 55 03


fnichol (approved):
Looking good. I'm guessing once we hit a third engine we may need to consider accordian sections, more tabs or some other way to focus the platform/engine matrix that will start to build up. For now though, we're golden!

stack72:
bors merge

PR #2711, opened 2023-08-28
chore(web): refactor asset and func id handling

paulocsanz:


app/web/src/store/asset.store.ts
```diff
@@ -155,24 +135,26 @@ export const useAssetStore = () => {
           this.openAssetFuncIds[assetId] = funcs;
         },
 
-        closeFunc(assetId: AssetId, funcId: FuncId) {
+        CLOSE_FUNC(assetId: AssetId, funcId: FuncId) {
```
theoephraim, line 138: the convention here (which has already been broken in a few places) is that the ALL_UPPER functions are api action (ie they make an api request, and so will have a request status available).

app/web/src/store/asset.store.ts
```diff
@@ -91,23 +90,22 @@ export const useAssetStore = () => {
...
-        selectedAssetId: null as AssetId | null,
         openAssetFuncIds: {} as { [key: AssetId]: FuncId[] },
       }),
       getters: {
         assets: (state) => _.values(state.assetsById),
         selectedAsset(): Asset | undefined {
           return this.assetsById[this.selectedAssetId || 0];
         },
-        urlSelectedAssetId(): AssetId | undefined {
+        selectedAssetId(): AssetId | undefined {
```
theoephraim, line 100: In the other stores, I've been careful to leave the name as `urlSelectedThingId` to make it clear that this is the one selected via the URL only. Then I usually have another `selectedThingId` which is the id of the thing once we have verified the id is good and loaded the thing.

This helps deal with the situation where there is an ID in the url but it isn't good (either totally bogus, or thing has been deleted, no access, etc...)

app/web/src/store/asset.store.ts
theoephraim: yes plz. lets try to keep the UPPER_CASE actions only things that call the api

paulocsanz:
bors merge

paulocsanz:
bors merge

PR #2712, opened 2023-08-29
feat(web): Secret store with only browser local data

vbustamante:
Functional version of secrets store with add and delete functionality, not hooked up to back end yet

Also implements a global secrets panel on the customize screen. On this iteration,  this is for development purposes only 

image


vbustamante:
bors merge

PR #2714, opened 2023-08-30
feat(si-pkg): Add ChangeSet node to SiPkg

zacharyhamm:
Adds a new package "category" for change sets. Adds the idea of a "default_change_set" to the SiPkg. ChangeSets are a name, a status and a pointer to the change set that they are "based on", along with child nodes for functions and schemas. In the future they will also include Components (with Node Positions), and Edges.


zacharyhamm:
bors try

zacharyhamm:
bors retry

PR #2716, opened 2023-08-30
fix(web): Auth store usage

vbustamante:


app/web/src/store/secrets.store.ts
```diff
@@ -39,7 +39,7 @@ export function useSecretsStore() {
         ) {
           const id = Math.floor(Math.random() * 899999 + 100000).toString();
 
-          const { id: userId, name: userName } = useAuthStore().user ?? {
+          const { pk: userId, name: userName } = useAuthStore().user ?? {
             id: "-1",
```
theoephraim, line 43: think you'll want this as `pk` also?

vbustamante: I just wanna fix the error, most of this code is going to the backend on my next PR 😄

vbustamante:
bors merge

PR #2718, opened 2023-08-31
workspace export/import UI

theoephraim:
adds a workspace settings cog, along with import/export modals. Currently mocking the requests with fake delays...

Some intermediate loading states as well, but these screenshots show the flow. Playing with the terms "backup" and "restore" a bit now, as this more accurately describes the flow as we will be implementing them for now. I suspect in the future that backup/restore may be some options within the broader category of export/import - ie you may be able to export/import subsets of a workspace without losing your local data, rather than only all or nothing.

new workspace settings menu
image

export modal
image
image

import modal
image
image

### DIFF
--- a/app/web/src/components/AssetDetailsPanel.vue
+++ b/app/web/src/components/AssetDetailsPanel.vue
@@ -3,7 +3,6 @@
     <RequestStatusMessage
       v-if="loadAssetReqStatus.isPending"
       :requestStatus="loadAssetReqStatus"
-      showLoaderWithoutMessage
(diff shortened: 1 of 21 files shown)


zacharyhamm:
My only note on this is we should hide it with a flag. I created a flag in posthog called `workspaceBackups`

theoephraim:
@zacharyhamm - yeah that was the plan. Will use your flag - thanks!

theoephraim:
bors merge

PR #2719, opened 2023-08-31
feat(dal,sdf): change_set aware export of workspaces

zacharyhamm:
Adds the ability for the nodes in our package graph to be empty of data
but still have child nodes. Adds the ability to track deleted objects
when exporting workspaces. Import is currently unchanged and only
supports "modules", which are a closure of the state of the active
change_set+head for specific schema variants.


zacharyhamm:
bors try

zacharyhamm:
bors try

zacharyhamm:
bors try

zacharyhamm:
bors try

PR #2720, opened 2023-08-31
AttributeValues in WorkspaceSnapshotGraphs

jhelwig:


jhelwig:
@nickgerace Want to take a look before I add this to the `eng-1703` branch? It's not "complete", but I think there's enough in place that we can have it in the shared project branch.

lib/dal/src/workspace_snapshot/graph.rs
```diff
@@ -172,37 +211,165 @@ impl WorkspaceSnapshotGraph {
...
+            // pre-existing data types.
+            let view_pointer = match view.pointer_mut(&write_location) {
+                Some(pointer) => {
+                    *pointer = current_node_content.clone();
+                    pointer
+                }
+                None => {
+                    // This is an error, and really shouldn't ever happen.
+                    dbg!(view, write_location, current_node_content);
+                    todo!();
```
nickgerace, line 238: Wonder if this should be `unreachable!` if it'll be here in the long term

nickgerace (approved):

PR #2722, opened 2023-09-01
fix(si): Infer the container engine based on args being passed

stack72:
Currently, if a user passes Podman as the engine and a docker-sock
then the CLI will error, We can make an assumption of the engine
based on the param passed and then let the user know we are doing so

```
si --podman-sock /var/run/podman.sock status
WARNING: A `podman_sock` location parameter was passed but no Podman engine was chosen. Continuing with the assumption you meant to choose Podman as the container engine...

Checking for user supplied docker.sock
```

```
SI_CONTAINER_ENGINE=podman si --docker-sock /var/run/docker.sock status
WARNING: A `docker_sock` location parameter was passed but no Docker engine was chosen. Continuing with the assumption you meant to choose Docker as the container engine...

Checking for user supplied docker.sock
```

### DIFF
--- a/bin/si/src/args.rs
+++ b/bin/si/src/args.rs
@@ -71,6 +71,11 @@ pub(crate) struct Args {
     #[arg(value_parser = PossibleValuesParser::new(Engine::variants()))]
     #[arg(long, short, env = "SI_CONTAINER_ENGINE", default_value = "docker")]
     engine: String,
+    /// A path to a podman.sock file. The default paths checked are `$XDG_RUNTIME_DIR/podman/podman.sock`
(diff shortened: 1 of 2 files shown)


stack72:
Feedback from @NeilHanlon in discord is to rename `--[docker|podman]-sock` to just `--container-sock` and then need to pass an engine

Otherwise, we would remove the need to pass an engine and thus determine the engine based on the type of sock passed, so `--docker-sock` always infers Docker engine etc

keeb:
👍

stack72:
Ok, we have talked internally and we are going to go with separate socket names so we can be explicit on the engines :)

stack72:
bors merge

Issue #2700, opened 2023-08-24
Collection of minor issues of the SI installer


keeb:
### What happened?

Now that `--web-host` flags are introduced, there's some subtle issues.

* Subsequent runs after containers have been created the first time do not respect the `--web-host flag`
* Using `--web-host` only affects the web. Does not affect SDF which is listening on all interfaces by default.

### How can we reproduce this?

Deploy SI using the installer without specifying `--web-host` and then run `si --web-host 0.0.0.0 start` and notice it does not change

### Operating system

Linux

### Architecture

x86_64 / amd64

### What browsers are you seeing the problem on?

Other (please specify)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Additional context

I will add any other minor nits as I find them.

keeb:
If you try to `si delete` while the stack is running it does not call stop and then delete, resulting in this error

```[keeb@system-init-installer ~]$ si delete
System Initiative Launcher is in "local" mode


Launcher update found, please run `si update` to install it

Deleting container: local-jaeger-1 (f6b19193f10c5e3521934b23efb2dcfb40c43f38e256f78e2f1529054c55b91d)
Error:
   0: docker api: error 409 Conflict - You cannot remove a running container f6b19193f10c5e3521934b23efb2dcfb40c43f38e25
6f78e2f1529054c55b91d. Stop the container before attempting removal or force remove
   1: error 409 Conflict - You cannot remove a running container f6b19193f10c5e3521934b23efb2dcfb40c43f38e256f78e2f15290
54c55b91d. Stop the container before attempting removal or force remove

Location:
   bin/si/src/main.rs:112```

keeb:
There's no indication as to when the service is actually available. 

For my system on a fresh install, I get auth errors (`Not found` - which doesn't mean anything to me really) when hitting the local SI homepage (`http://localhost:8080` equivalent) until ~9 minutes after I have started.

I believe this is all of the DB setup/migrations that happen on first start.

stack72:
@keeb this is fantastic! I can get this fixed up :) thank you!!

stack72:
The last issue to be tackled here is that when the cli args change for host ip and port, we need to stop and recreate the web container

PR #2723, opened 2023-09-01
feat(si): Recreate SDF or Web containers when port mappings change

stack72:
Closes: #2700

If you want to  change the port map for SDF or Web, then you'd currently
need to stop and delete the containers manually then re-run start

With this PR we can now do the following:

```
si --sdf-port 5157 start
```

This caused the Launcher to act as follows:

```
Container Port Mappings have changed for local-web-1 so recreating
Stopping container /local-web-1
Deleting container: local-web-1 (4d15095f701d8f2872a4e5a838f1187d1e0b980297bd23d15c21bc966eb41188)
Starting systeminit/web:stable as local-web-1
All system components running... System Initiative is alive!

You can now use the `si launch` command to open the System Initiative web portal:
```

### DIFF
--- a/lib/si-cli/src/cmd/start.rs
+++ b/lib/si-cli/src/cmd/start.rs
@@ -311,17 +311,53 @@ async fn invoke(app: &AppState, is_preview: bool) -> CliResult<()> {
                 .get_existing_container(container_name.clone())
                 .await?;
             if let Some(existing) = container_summary {
+                let mut needs_recreated = false;
+                if let Some(ports) = existing.ports {
+                    if !ports.is_empty() {
+                        let port = ports.first().unwrap().clone();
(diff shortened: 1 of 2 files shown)


keeb:
Need to check what happens when it's stopped (not `running`) but `created`. I don't think this is captured, but happens as a result of rebooting the machine in my case.

In the case where it's running and needs to be updated - we're 👍

stack72:
bors merge

stack72:
bors cancel

stack72:
@keeb I found the fix to the usecase where the machine wasn't started - works in that case now too

paulocsanz (approved):

stack72:
bors merge

PR #2724, opened 2023-09-01
build: fix typescript rule docstring

jazzdan:
Just a small copy pasta I think


stack72:
thank you so much for this @jazzdan - appreciate the fix here and love making sure our docs / comments are correct!!

stack72:
bors merge

PR #2725, opened 2023-09-04
chore(web,sdf,dal): add action to change-set

paulocsanz:


app/web/src/api/sdf/dal/change_set.ts
```diff
@@ -8,10 +8,32 @@ export enum ChangeSetStatus {
...
+  name: string;
+  componentId: ComponentId;
+}
+
+export type ChangeSetId = string;
+export interface ChangeSet {
+  id: ChangeSetId;
+  pk: ChangeSetId;
   name: string;
-  note?: string;
```
stack72, line 14: is this something we want to remove?

lib/dal/src/migrations/U2302__actions.sql
```diff
@@ -0,0 +1,50 @@
+CREATE TABLE actions
```
stack72, line 1: how will this fold into everything is a node or edge work? Do we need to make @nickgerace and @jhelwig aware?

stack72:
I think the only issue is the following - when I drag a component to the canvas, I get the following:

Screenshot 2023-09-05 at 17 42 30

Notice how it shows ALL actions associated with that component

stack72:
I also don't think this order is correct:

Screenshot 2023-09-05 at 17 46 53

The keypair will need to be created first so there may be a regression in how the ordering happens here

paulocsanz:
@stack72 you are right we have a regression on the sorting, we have to figure this out. If we want to suggest a confirmation and enforce the order or allow them to add custom orders.

About all actions showing, yes, that's by design, what's up? Only the chosen ones will appear in the right, but in the popover you can select anything

paulocsanz:
@stack72 why would it change? I thought the popover would always have all the actions so the user can pick any one they want

stack72:
> @stack72 why would it change? I thought the popover would always have all the actions so the user can pick any one they want

OOOHHHHH that's why they are on the left - it doesn't read that way but that makes WAY more sense :) It looks like the list of actions that have happened :)

paulocsanz:
Yeah, it definitely was not designed. Just something to enable the user to re-run creation actions and such. Definitely needs some love.

paulocsanz:
bors try

stack72:
@paulocsanz this is working a treat!! Merge when it's green :)

stack72 (approved):

paulocsanz:
bors merge

PR #2736, opened 2023-09-08
connect backup/restore to module index

theoephraim:
connects the backup/restore UI to the module index.

Need to swap in the calls to actually create a proper workspace backup (instead of a dummy module) and restore the workspace, but these are things @zacharyhamm is still working on.


lib/module-index-server/src/migrations/U0003__support_backups.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE modules
+    ADD COLUMN is_backup boolean DEFAULT false NOT NULL
```
theoephraim, line 2: will likely end up with a type field as we'll probably end up with modules, backups, and just partial exports, but this is simpler for now...

PR #2737, opened 2023-09-08
Graph logic in the Rebaser (ENG-1921)

nickgerace:


lib/dal/src/change_set_pointer.rs
```diff
@@ -0,0 +1,125 @@
...
+
+pk!(ChangeSetPointerId);
+
+#[derive(Clone, Serialize, Deserialize)]
+pub struct ChangeSetPointer {
+    pub id: ChangeSetPointerId,
+    #[serde(flatten)]
+    pub timestamp: Timestamp,
+    #[serde(skip)]
+    pub generator: Arc<Mutex<Generator>>,
```
jhelwig, line 44: This doesn't need to be marked default, or told which function to call to populate when deserializing from the DB?

nickgerace: There's two things going on here I believe. First, `skip` attribute indicates that the field should be neither serialized nor deserialized **and** that serde will use `Default::default()` when deserializing. Second, I believe that since `Generator` implements `Default` you do not need to specify a function to be called.

jhelwig: I had forgotten that skip implied default (didn't require explicitly marking to use default, or to mark how to default the value).

nickgerace: Yep! It's a bit confusing truthfully because `skip` means "don't do anything" followed by "oh yeah it actually uses `Default::default()` when deserializing".

nickgerace:
Cherry picked the sole commit in this PR into `eng-1703-mostly-everything-is-a-node-or-an-edge`. Closing.

PR #2749, opened 2023-09-13
feat(web) cleaning up actions ui and store

theoephraim:
cleaning up the ui (and store) around actions a bit. more to come, but this should hopefully lay a nicer foundation

image


stack72:
bors merge

PR #2750, opened 2023-09-14
feat(module-index-server): Allow rejecting modules

stack72:
When a module is rejected, it is no longer returned in the response to the modules page in the UI


stack72:
bors merge

PR #2751, opened 2023-09-14
fix(web) auth reconnect bug

theoephraim:
small bug in auth flow was causing reconnecting ui to fire even during login


theoephraim:
bors merge

PR #2752, opened 2023-09-14
fix(web) further fix to login flow

theoephraim:
last fix stopped the user facing bug, but we were still calling connect twice and had some other unexpected behaviour around logout.


theoephraim:
bors merge

PR #2753, opened 2023-09-14
fix(web) fix labels on actions in right hand panel

theoephraim:


theoephraim:
bors merge

stack72:
bors merge

PR #2754, opened 2023-09-14
chore(web) update import/export terminology

theoephraim:
I got rid of the "backup" and "restore" terminology in favour of sticking with "export" and "import"

I will say it still feels quite awkward as export/import work well as the verbs but I feel there is a missing noun, for example "backup" or "snapshot"...  "Workspace export" feels like a bit of a mouthful and using just "workspace" doesnt quite feel right as you are not importing a "workspace" you are importing a fixed snapshot of a workspace.


theoephraim:
bors merge

PR #2755, opened 2023-09-14
fix(auth) make auth whoami types in auth api client match auth api

theoephraim:
serde blowing up because types werent quite right. shouldn't break anything else though


theoephraim:
bors merge

PR #2756, opened 2023-09-15
feat(web): Introduce changes to the Diagram to help understand status

stack72:
We are introducing a new way of understanding some of the status. This includes some changes that allow us to be able to show that a resource actually exists in the real world

Screenshot 2023-09-18 at 22 55 57

Screenshot 2023-09-18 at 22 56 06


stack72:
bors merge

PR #2758, opened 2023-09-18
chore(web,dal): change action prototype kind setter UX

paulocsanz:
image

The error message is because I made the action delete, but there already is one, so it rolled back and showed the error message

### DIFF
--- a/app/web/src/components/FuncEditor/ActionDetails.vue
+++ b/app/web/src/components/FuncEditor/ActionDetails.vue
@@ -1,15 +1,33 @@
 <template>
   <div class="p-3 flex flex-col gap-2">
+    <ErrorMessage :requestStatus="props.requestStatus" />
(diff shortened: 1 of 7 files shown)


app/web/src/components/FuncEditor/ActionDetails.vue
```diff
@@ -1,15 +1,35 @@
 <template>
   <div class="p-3 flex flex-col gap-2">
+    <span v-if="props.validationMessage" class="text-red-400">{{
+      props.validationMessage
+    }}</span>
     <h2 class="pt-4 text-neutral-700 type-bold-sm dark:text-neutral-50">
-      Kind of Action:
+      <SiCheckBox
```
theoephraim, line 7: why not just use a radio or a dropdown? doesnt make sense IMO to have checkboxes if only one is selectable

paulocsanz:
bors merge

PR #2761, opened 2023-09-19
feat(object-tree,si-pkg): Support optional nodes in object tree

zacharyhamm:
We removed a node child kind (FuncDescriptions), but this breaks old modules that have func descriptions in them. We still want to be able to read those modules and we want the ability to make changes like this to the package format in the future without invalidating the old modules. This makes the read_bytes trait in object-tree return an Option, so if a node is not one we want to read any more, we can skip it (and its children!), while preserving the rest of the module.


zacharyhamm:
bors try

PR #2766, opened 2023-09-21
chore: add `--bind-all` flag to `buck2 run dev:up`

fnichol:
This change will run Tilt binding to all IPs as well as the `app/web` service.

This means you can run:

```
buck2 run dev:up -- --bind-all
```


fnichol:
bors merge

PR #2771, opened 2023-09-22
fix: handle special case functions that should be intrinsics

stack72:


zacharyhamm (approved):
:+1:  let's just remember to remove this in the future :)

stack72:
bors merge

PR #2772, opened 2023-09-22
fix: revert "chore: add `--bind-all` flag to `buck2 run dev:up`"

fnichol:
This reverts commit 665f89cade8b9d7a4ae2ce00088658d97f8d0b13.

We'll revisit this with a follow up change, sorry to those affected in the meantime!

PR #2775, opened 2023-09-22
un-revert #2772 and fix `buck2 run dev:down`

keeb:
This fix un-reverts #2772 and also fixed the underlying issue by detecting if the command is an `up` or `down` command.


stack72:
bors merge

PR #2779, opened 2023-09-22
feat(web): autoformat code on blur

paulocsanz:
This uses a very simple auto formatter, but it can be easily swapped. I tried prettier, there was some trickery with module installation so I decided to go with the simplest version that helps us test this concept and be tuned gradually.

It definitely will need some customization on the formatting patterns, but feels good enough.

All functions will endup being formatted, I can update all packages but it doesn't feel necessairy.


paulocsanz:
bors merge

PR #2791, opened 2023-09-26
fix(cyclone-server,veritech-server): Reducing the logging of shutdown events

stack72:
Currently the veritech process throws 100s of these (due to cyclone)

```
023-09-26T11:21:12.380324Z  INFO ThreadId(03) cyclone_server::server: received internal shutdown, performing graceful shutdown; source=Some(LimitRequest)
```

I have moved this to a trace level event


paulocsanz:
bors merge

Issue #2802, opened 2023-09-27
Some containers start at boot


Octogonapus:
### What happened?

After stopping all Docker containers and restarting my system, these containers were started at boot:
```
CONTAINER ID   IMAGE                        COMMAND                  CREATED       STATUS       PORTS                          NAMES
9e61c0f9b92f   systeminit/nats:stable       "/nats-server --conf…"   2 weeks ago   Up 5 hours   4222/tcp, 6222/tcp, 8222/tcp   local-nats-1
9e5add07dd8a   systeminit/postgres:stable   "/usr/local/bin/entr…"   2 weeks ago   Up 5 hours   5432/tcp                       local-postgres-1
```

Given that SI is stopped, I would expect none of the containers to be running.

### How can we reproduce this?

You should be able to just start and stop SI and reboot.

### Operating system

Linux

### Architecture

x86_64 / amd64

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #2803, opened 2023-09-27
fix(si): Change SI container restart policy

stack72:
Currently, we had a restart policy of on-failure BUT this causes us to
try and start containers on system boot as docker doesn't keep track
of the previous state - https://github.com/moby/moby/issues/35166#issuecomment-338211235

This means that we will continually hit this problem and the containers
may fail to start due to ordering issues

Fixes: #2802


keeb (approved):
👍 changes lgtm

stack72:
bors merge

stack72:
bors retry

fnichol:
bors retry

fnichol:
bors retry

PR #2806, opened 2023-09-28
fix(sdf): add json content-type to create_node response

zacharyhamm:
The lack of a content-type header on the create_node HTTP response causes Firefox to raise an XML Parsing Error on every call (since it interprets the type as XML by default), and likely will cause issues with other HTTP clients. This quiets those errors.


nickgerace (approved):
Yikes. Great catch.

PR #2809, opened 2023-09-28
Add content-store and content-store-test crates (ENG-1946)

nickgerace:
## Summary

This commit adds the content-store and content-store-test crates, which primarily revolve around the "Store" trait. They also provide the ability to perform migrations and database setup for "PgStore", a user of the aforementioned crate.

Why two crates? The test database setup and migration logic should never be used in production, but it should remain portable. Thus, the content-store-test crate exists.

## The `Store` Trait

The "PgStore" can have three different backing databases: "si_content_store", "si_test_content_store", and
"si_test_content_store_<id>". The first is the database used for production. The second is the global database used for integration tests and the final copies its contents on a test-by-test basis.

The new content-store crate brings over the "content" module contents from the dal. That includes the "ContentPair" concept as well as the primitives migration.

Currently, there are two users of the "Store" trait:

- "LocalStore": exists entirely in-memeory and can be used everywhere (currently used for unit tests, but may have other use cases in the future)
- "PgStore": leverages pull-through caching, offers persistent writes to Postgres and can be used everwhere so long as _someone_ creates the database and performs migrations

## `PgStore` and Testing

Let's talk about integration tests then since they are the sole users of the "PgStore" currently. The dal-test macro now includes setup for the global "si_test_content_store" database. It mirrors the "si_test_dal" database setup path, but is provided via a single client from the content-store-test crate.

There's a catch: the "DalTestPgStore" wrapper should be used in dal integration tests because dal-test does not provide a "PgStore" for every single test yet. Rather than performing surgery in dal-test and si-test-macros' expand module, we'll create "PgStores" with accompanying test-specific databases on a case-by-case basis for now.

## Everything Else

This commit contains secondary changes encountered during refactoring as well:

- Remove "si_test_rebaser" and rebaser integration tests since the rebaser mainly uses dal logic and few meaningful tests can currently be conducted with the rebaser on its own
- Reduce log level for the rebaser skip message during test setup
- ChangeSetPointer and WorkspaceSnapshot queries, structs and migrations have slightly changed to adapt to the new changes
- Where relevant, we implement "TryFrom<PgRow>" instead of using json serialization for speed
- Co-locate all integration tests for the "Mostly Everything is a Node or an Edge" work

## Additional Information

I'll cherry pick this and close once approved. This will not go through bors or the GitHub UI merge options, much like the other PRs for #2424.


nickgerace:
I'll cherry pick this and close once approved. This will not go through bors or the GitHub UI merge options, much like the other PRs for #2424.

lib/content-store/src/store/local.rs
```diff
@@ -0,0 +1,46 @@
...
+    fn len(&self) -> usize {
+        self.0.len()
+    }
+
+    fn add<T>(&mut self, object: &T) -> StoreResult<ContentHash>
+    where
+        T: Serialize + ?Sized,
+    {
+        let value = serde_json::to_vec(object)?;
+        let key = ContentHash::new(&serde_json::to_vec(&value)?);
```
jhelwig, line 26: I'm curious why the two layers of `serde_json::to_vec` here. Shouldn't the key be `ContentHash::new(&value)`, instead of `ContentHash::new(serde_json::to_vec(&value)?`), since `object` was already converted to a `Vec<u8>` of the serialized bytes?

lib/content-store/src/store/pg.rs
```diff
@@ -0,0 +1,91 @@
...
+    fn len(&self) -> usize {
+        self.inner.len()
+    }
+
+    fn add<T>(&mut self, object: &T) -> StoreResult<ContentHash>
+    where
+        T: Serialize + ?Sized,
+    {
+        let value = serde_json::to_vec(object)?;
+        let key = ContentHash::new(&serde_json::to_vec(&value)?);
```
jhelwig, line 58: Same question about the double `serde_json::to_vec` as with the local store.

lib/content-store/src/store/pg/migrations/U0001__setup_primitives.sql
```diff
@@ -0,0 +1,69 @@
+-- TODO: add constraints, plpgsql doesn't do domain constraints though:
```
jhelwig, line 1: Do we need this migration? I'm not sure where/why we'd be creating ULIDs from within the DB for the content store.

lib/dal/src/migrations/U3001__change_set_pointers.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE change_set_pointers
+(
+    id                    ident primary key        NOT NULL DEFAULT ident_create_v1(),
+    created_at            timestamp with time zone NOT NULL DEFAULT CLOCK_TIMESTAMP(),
+    updated_at            timestamp with time zone NOT NULL DEFAULT CLOCK_TIMESTAMP(),
+    name                  text                     NOT NULL,
+    -- TODO(nick): add once workspaces are added
```
jhelwig, line 7: Do we want to add the `base_change_set_id` field now?

lib/content-store/src/store/local.rs
nickgerace: That... is an accident.

nickgerace: It should be just one.

lib/content-store/src/store/pg.rs
nickgerace: Copy and paste error. Also an accident 🤦‍♂️

lib/content-store/src/store/pg/migrations/U0001__setup_primitives.sql
nickgerace: Hmmm... good point. We likely don't. I'll run through it again, but I think I originally wanted something from there. Now, I don't think so.

lib/dal/src/migrations/U3001__change_set_pointers.sql
nickgerace: I'm not using it right now, so I do not need it. I changed these migrations because how I accessed them and used them in integration tests also changed. I'll take a look though.

jhelwig: Not really a big deal either way, since it'll be a nullable field.

nickgerace:
Changes made from review and cherry-picked.

PR #2810, opened 2023-09-28
fix(auth) - RichText hljs plaintext support

wendybujalski:
Tutorial navigation bug was actually caused by an issue with the RichText markdown parser. Fixed!


stack72:
Remarkable fix!!

wendybujalski:
bors merge

PR #2811, opened 2023-09-29
Add BUCK2 docs file (ENG-1968)

nickgerace:
Add BUCK2 docs file for getting started with buck2, grokking high level
concepts and how to use buck2 and BUCK files.

The document contains a link to the "RUNNING_RUST_TESTS" document, which
describes similar, but more pointed, functionality.

_**[view the rendered file here](https://github.com/systeminit/si/blob/nick/e17b215/docs/BUCK2.md)**_

PR #2812, opened 2023-09-29
Ensure buck2 vsix hyperlink is clear (ENG-1969)

nickgerace:
Currently, the buck2 vsix hyperlink in the new BUCK2 doc implies that it goes to "VS Code", the editor. Now, it should imply that it links to the extension that can be locally built.

This commit also fixes the Discord badge in the same file to link to the server.

This is follow-up to https://github.com/systeminit/si/pull/2811


nickgerace:
bors merge

PR #2818, opened 2023-10-02
feat(module-index): Add ability to mark a package as a builtin

stack72:


lib/module-index-server/src/routes/list_modules_route.rs
```diff
@@ -84,6 +84,9 @@ pub async fn list_module_route(
         query
     };
 
+    // We want to filter out the builtins from the list as they will already be in our system
+    let query = query.filter(si_module::Column::IsBuiltinAt.is_null());
```
zacharyhamm, line 88: 👍

lib/module-index-server/src/routes/download_builtin_route.rs
```diff
@@ -0,0 +1,63 @@
...
+    };
+
+    if module.is_builtin_at.is_none() {
+        return Err(DownloadBuiltinError::NotBuiltin(module_id));
+    }
+
+    let download_url =
+        s3_bucket.presign_get(format!("{}.sipkg", module.latest_hash), 60 * 5, None)?;
+
+    Ok(Redirect::temporary(&download_url))
```
zacharyhamm, line 62: is this what we do for the normal download? i like the redirect instead of streaming bytes through the server

zacharyhamm (approved):
looks solid

lib/module-index-server/src/routes/download_builtin_route.rs
stack72: This is the same :)

stack72:
bors merge

PR #2825, opened 2023-10-03
feat(web,dal,sdf,lang-js): Integrate secret widget into PropertyEditor

vbustamante:
- Fail secret defining schema creation with more than one secret prop
- Show function exec errors asset creation
- Load secret Definitions from the backend
image


vbustamante:
bors merge

PR #2827, opened 2023-10-04
feat(editor): open tooltip on hotkey ctrl+m

sprutton1:
Pop tooltip for current source under cursor using `ctrl/cmd+m`. Hotkey chosen arbitrarily as one that didn't have a binding in Chrome.


paulocsanz (approved):
That refactor is awesome! Great job!

sprutton1:
> Did you manage to fix the styling issue?

Yep. The hover tooltip was putting all of its tooltips in a container div while the regular tooltip does not. We just needed to change a class name to style the right div

PR #2836, opened 2023-10-06
Introduce VectorClockId and use it in rebaser-server

nickgerace:
Introduce VectorClockId to separate ChangeSetPointers from VectorClocks. This will make the behavior of the rebaser-server easier to understand because, as it stands today, VectorClockIds are just ChangeSetPointerIds. They share the same ULID. That might not always be the case either, so this change has domain-driven benefits too.

Use the new VectorClockId in rebaser-server and change "updates and conflicts" detection to actually perform as intended. As a result, detection no longer requires ChangeSetPointers directly and only requires VectorClockIds.

Ensure the ChangeSetMessage reflects the new VectorClockId changes. The fields now use "to_rebase" and "onto" prefixes to help imply how to prepare the payload.

Add ChangeSetLoopError to differentiate errors in the rebaser-server outer loop to the inner loop(s).

### DIFF
--- a/lib/dal/src/change_set_pointer.rs
+++ b/lib/dal/src/change_set_pointer.rs
@@ -9,6 +9,7 @@ use telemetry::prelude::*;
 use thiserror::Error;
 use ulid::{Generator, Ulid};
 
+use crate::workspace_snapshot::vector_clock::VectorClockId;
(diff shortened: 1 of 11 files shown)


nickgerace:
Cherry-picking...

PR #2838, opened 2023-10-09
fix(module-index): Ensure that we download from the production module-index

stack72:
If the local module index returns a 404 then we should pull the builtin from the production module-index


zacharyhamm:
testing this now

zacharyhamm (approved):
Pushed up a commit that gets this working (we were returning on 404 with error_for_status and never hitting the pass through code)

PR #2839, opened 2023-10-09
fix(sdf): builtins on transactions, skip dependent values update for builtins

zacharyhamm:
Two changes:
  1. Install each builtin on its own Postgres connection so they run
     truly concurrently.
  2. Skip dependent values updates on builtin installation for
     near-instant builtin creation.

I can't figure out a reason to execute dependent value update jobs on
builtin creation. If there is a value that needs to exist on an
attribute for a schema variant in the default context, we can
pre-execute the function during authoring to produce a default value for
that attribute, or explicitly set the default value.  Otherwise, every
value computed by a function will be properly computed when the
component is created, and this is the context where we care about
computing the most up to date values in any case.

This makes builtin installation take about 30 seconds, which is a huge
win to me versus any desire we might have to execute the dependent value
update jobs for the schema variants themselves, before a component is
created.

### DIFF
--- a/lib/dal/src/context.rs
+++ b/lib/dal/src/context.rs
@@ -66,6 +66,7 @@ impl ServicesContext {
         DalContextBuilder {
             services_context: self,
             blocking,
+            no_dependent_values: false,
         }
     }
 
@@ -202,6 +203,9 @@ pub struct DalContext {
     /// This is useful to ensure child jobs of blocking jobs also block so there is no race-condition in the DAL.
     /// And also for SDF routes to block the HTTP request until the jobs get executed, so SDF tests don't race.
     blocking: bool,
+    /// Determines if we should not enqueue dependent value update jobs for attribute updates in
+    /// this context
+    no_dependent_values: bool,
 }
 
 impl DalContext {
@@ -211,6 +215,7 @@ impl DalContext {
         DalContextBuilder {
             services_context,
             blocking,
+            no_dependent_values: false,
         }
     }
 
@@ -228,6 +233,10 @@ impl DalContext {
 
     pub fn blocking(&self) -> bool {
         self.blocking
+    }
+
+    pub fn no_dependent_values(&self) -> bool {
+        self.no_dependent_values
     }
 
     pub fn services_context(&self) -> ServicesContext {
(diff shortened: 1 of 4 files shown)


paulocsanz (dismissed):

keeb:
Tested in 2 different environments

* beefy threadripper machine went from 1100 seconds to 300. 
* A low powered vm (6 cores, 16GB of RAM) finished in 720 seconds. I didn't have a baseline for the low powered VM.

In both cases, 35 out of 36 assets were installed successfully. 1 failed. 

Both UI's work and are operational.

stack72:
I just tested this locally and everything is working as expected - dependent values are flowing - I think this can get merged as some of this code will change with new DAG work anyway so :shipit: 

```
2023-10-10T14:47:39.750543Z  INFO ThreadId(02) sdf.init.migrate_database: sdf_server::server::server: migrating elapsed=25.001672744750977
2023-10-10T14:47:44.749710Z  INFO ThreadId(02) sdf.init.migrate_database: sdf_server::server::server: migrating elapsed=30.000812530517578
Pkg si-aws-ec2-instance-builtin Install finished successfully. 37 of 37 installed.
2023-10-10T14:47:46.351339Z  INFO ThreadId(02) sdf.init.migrate_database: sdf_server::server::server: migrating completed elapsed=31.60245132446289
2023-10-10T14:47:46.354079Z  INFO ThreadId(02) sdf_server::server::server: binding to HTTP socket; socket_addr=0.0.0.0:5156
2023-10-10T14:47:46.354219Z  INFO ThreadId(02) dal::tasks::status_receiver: starting status receiver for dependent values update jobs
[readiness probe: success] {"ok":true}
```

paulocsanz (approved):
We should inspect why the dependent values are running, but we can do that later. This PR is _insane_, wild how fast it is.

I think we propagate things like default values, for example region's default name. It should not take that long we are probably propagating a shitton of stuff by updating domain, since qualifications/codegen, etc all depend on domain.

You should double check that the default values are not gone. But I will think more if there are other cases where the dependent values would trigger. Interesting that it was only one line causing all this mess hahah.

lib/dal/src/context.rs
```diff
@@ -566,6 +575,9 @@ pub struct DalContextBuilder {
     /// This is useful to ensure child jobs of blocking jobs also block so there is no race-condition in the DAL.
     /// And also for SDF routes to block the HTTP request until the jobs get executed, so SDF tests don't race.
     blocking: bool,
+    /// Determines if we should not enqueue dependent value update jobs for attribute value
+    /// changes.
+    no_dependent_values: bool,
```
paulocsanz, line 580: I love a simple and hacky solution that makes things hundreds of times better!

PR #2840, opened 2023-10-09
feat(web): Allow rejection of builtins

stack72:
We could promote a module to be a builtin but we couldn’t reject a builtin when it was created and thus the situation of not being able to replace an existing builtin

Screenshot 2023-10-10 at 01 15 20

Screenshot 2023-10-10 at 01 15 29


vbustamante (approved):
Great!

stack72:
bors merge

PR #2841, opened 2023-10-10
feat: change interfaces to types for quickinfo

sprutton1:
Same as described [here](https://github.com/microsoft/TypeScript/issues/38040), the Typescript system does not extract deep information from `interface`s, so if we want the details on the hover/hotkey, we need them to be `type`s.

I have no idea what implications this might have elsewhere, so insight there would be good. Everything seems good in my tests still.


paulocsanz (approved):
Have you checked if this causes a lint breakage for any function? Because variable definition works differently with interfaces and types.

If not, I'm cool with it.

sprutton1:
> Have you checked if this causes a lint breakage for any function? Because variable definition works differently with interfaces and types.
> 
> If not, I'm cool with it.

I don't see any new linter errors on the existing functions I checked. I haven't checked all of them, but I imagine we'll catch them as @stack72 goes through and fixes them up.

PR #2845, opened 2023-10-10
Wendy/eng 1952 mock of the test a function panel

wendybujalski:
more than a mock, the panel is fully integrated!~


wendybujalski:
bors merge

PR #2853, opened 2023-10-12
(feat) (lang-js) add fetch to commandRunSandbox

keeb:
This updates the vm2 sandbox to allow `fetch` to be used from within asset actions


keeb:
bors merge

PR #2855, opened 2023-10-12
feature: watch in siExec

adamhjk:
This adds 'siExec.watch()' to lang-js. It allows you to watch the output of an siExec.waitUntilEnd over repeated calls, and should be handy any time you need to poll the state of something while you wait for it to succeed.

```typescript
// Get the date in seconds since the epoch
const getCurrentTime = await siExec.waitUntilEnd("date", ["+%s"]);
const startSeconds = getCurrentTime.all || "3";
// Wait until its at least 3 seconds later;
const waitUntil = startSeconds + 3;

// Call date, trying again every 2 seconds, a maximum of 10 times, and
// return true if more than 3 seconds have elapsed.
const r = siExec.watch({
      cmd: "date",
      args: ["+%s"],
      retryMs: 2000,
      maxRetryCount: 10,
      callback: async (r) => {
        const elapsed = r.all || "3";
        return elapsed > waitUntil;
      }
    });
// r.result == the SiExecResult
// r.failed? == "commandFailed" | "deadlineExceeded"
```

Should help DRY up any repeat calls for status.

### DIFF
--- a/bin/lang-js/src/sandbox/exec.ts
+++ b/bin/lang-js/src/sandbox/exec.ts
@@ -1,6 +1,22 @@
 import execa from "execa";
-import { ExecaReturnValue, Options } from "execa";
+import {ExecaReturnValue, Options} from "execa";
 import Debug from "debug";
+import {setTimeout} from "timers/promises";
+
+export interface WatchArgs {
+    cmd: string,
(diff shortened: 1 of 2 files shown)


keeb:
bors merge

PR #2863, opened 2023-10-13
Add si-cbor and use for WorkspaceSnapshot and ContentPair (ENG-1983)

nickgerace:
Add a new library, si-cbor, for quickly encoding and decoding CBOR serialized objects.

Use si-cbor for both the serialized snapshots in WorkspaceSnapshot and for the value in ContentPair. This should help with space savings and performance in the long term.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -660,6 +660,33 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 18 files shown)


nickgerace:
Cherry picked.

PR #2864, opened 2023-10-16
feat: implement symmetric encryption for secrets & keys

fnichol:


fnichol:
bors try

lib/si-crypto/src/symmetric.rs
```diff
@@ -0,0 +1,329 @@
+//! Symmetric key cryptography.
```
vbustamante, line 1: Probably a good call to move this to a new module

fnichol:
bors merge

PR #2865, opened 2023-10-16
Perform updates in the rebaser (ENG-1933)

nickgerace:
## Description

This PR contains two commits that enable us to perform updates in the rebaser.

## Commit 1/2: [Bail on conflicts in rebaser-server](https://github.com/systeminit/si/commit/9a0a4fd098631b6639f822aa9c3c795d6ad8661a) 

Bail on conflicts in rebaser-server by returning a new variant of the
ChangeSetReplyMessage. The new variant, ConflictsFound, indicates that
while no functions failed, there was at least one conflict found.
Therefore, no updates could be performed.

## Commit 2/2: [Perform updates in the rebaser](https://github.com/systeminit/si/commit/d3780aab4e8ab25a5fc99d8fb0267e63f84f7726) 

This commit adds the ability for the rebaser to perform the updates it
finds.

Two new tests have been added: one showcasing a "fork" rebasing
flow including a simple conflict. The other test is contained within
WorkspaceSnapshotGraph and helped fix a bug found while writing the
former test: when "to_rebase" has never seen anything from the "onto"
change set, all items are new. Before this commit, those updates were
not detected.

Primary dal changes:
- Add ability to update content, import subgraph, remove edge by
  EdgeIndex (only when performing updates), replace references, get node
  weight, find equivalent node, print "dot" output to stdout, get node
  index by id, and add edge from root for WorkspaceSnapshot
  - It is likely that many of these methods will not last the test of
    time and exist mainly for the commit's integration test
- Replace ChangeSetPointer function parameters with VectorClockId where
  appropriate (continuation of prior work with VectorClockId)
- Re-export NodeIndex for use outside of WorkspaceSnapshotGraph
  - This should be rarely used and we may be able to abstract it away in
    the future
- Add ability to get the root NodeIndex, find an equivalent node, and
  remove edge by EdgeIndex for WorkspaceSnapshotGraph

Primary rebaser changes:
- Renamed client methods to closer match intent
- Centralize stream name generation into core
- Add debug logging to both the client and server to aid in debugging
  timeouts
- Expand ChangeSetReplyMessage options to aid in testing and debugging
  in production
- Move DalContext building into the listener loop rather than re-using
  the same DalContext for every delivery
- Convert Producer from "Dedup" to "Nodedup" since a Producer can be
  re-created for the same stream and process
  - Without this change, you could produce a message to a stream, see
    that it would not appear as "ready" in the RabbitMQ Dashboard and
    observe that no error or failure occurred (read: this was a pain in
    the ass)
  - From this change, the number of constructors have been reduced from
    two to one

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3925,6 +3925,7 @@ dependencies = [
 name = "rebaser-client"
 version = "0.1.0"
 dependencies = [
+ "log",
  "rebaser-core",
  "remain",
  "serde",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3941,6 +3941,7 @@ name = "rebaser-core"
 version = "0.1.0"
 dependencies = [
  "serde",
+ "serde_json",
  "ulid",
 ]
 
--- a/lib/dal/src/change_set_pointer.rs
+++ b/lib/dal/src/change_set_pointer.rs
@@ -86,13 +86,12 @@ impl ChangeSetPointer {
                 &[&name],
             )
             .await?;
-        Ok(Self::try_from(row)?)
+        Self::try_from(row)
     }
 
     /// Create a [`VectorClockId`] from the [`ChangeSetPointer`].
     pub fn vector_clock_id(&self) -> VectorClockId {
-        let ulid: Ulid = self.id.into();
-        VectorClockId::from(ulid)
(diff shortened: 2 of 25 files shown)


nickgerace:
Cherry picked.

PR #2872, opened 2023-10-17
build(buck2): create a prettier_check rule for use in lang-js

fnichol:
Prior to this change, no source formatting was being enforced by a build system and thus started to drift in consistency.


fnichol:
bors merge

PR #2880, opened 2023-10-18
feat(web,sdf-server): Remove the need to set handler on functions and assets

stack72:
We now handcode this to be `main` by default. We have already updated all the builtins to use `main` and we have added a typescript lint warning to let people know they need a main entrypoint for a function

Screenshot 2023-10-18 at 22 52 37

### DIFF
--- a/app/web/src/utils/typescriptLinter.ts
+++ b/app/web/src/utils/typescriptLinter.ts
@@ -120,12 +120,25 @@ export const createTypescriptSource = async (
     const doc = view.state.doc;
     // We could be more efficient by updating only the changed spans
     const docString = doc.toString();
+
+    let diagnostics: Diagnostic[] = [];
+
+    // custom lint rule to ensure that we have a `main` function entrypoint
+    if (!docString.includes("function main(")) {
+      diagnostics = diagnostics.concat(
+        diagnosticsForMessage(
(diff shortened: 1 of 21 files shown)


app/web/src/utils/typescriptLinter.ts
```diff
@@ -120,12 +120,25 @@ export const createTypescriptSource = async (
...
+
+    let diagnostics: Diagnostic[] = [];
+
+    // custom lint rule to ensure that we have a `main` function entrypoint
+    if (!docString.includes("function main(")) {
+      diagnostics = diagnostics.concat(
+        diagnosticsForMessage(
+          1,
+          1,
+          "Function should include a `main` function for code execution",
```
paulocsanz, line 132: This may be confusing if the user wrote `const main = async () => {}`, which won't execute properly, but the message suggest that's possible.

paulocsanz: I do love how simple this is!

stack72: Ok, @paulocsanz and I talked in slack and we decided this is fine for now but we can fine tune if it becomes a problem later

stack72:
bors merge

Issue #2659, opened 2023-08-19
[COM-26] [Feature Request] localstack support


Wolfsrudel:
### Is this issue related to an issue?.

I don't like to see an AWS account, in which costs can arise, as a development environment if I can also map the whole thing locally.
Basically, I want to be able to work locally - isolated - independently of AWS.

### Describe the outcome you'd like from us

In order to be able to map a completely local development environment - i.e. to be independent of an AWS account - the integration of [localstack](https://localstack.cloud/)  at API endpoints would be very helpful.

### Describe alternatives you've considered

I believe the alternative to localstack would be the AWS API itself.
Possibly a dry run based on a `terraform plan` without actually generating the resources.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

<sub>[COM-26](https://linear.app/system-initiative/issue/COM-26/[feature-request]-localstack-support)</sub>

keeb:
Hey @Wolfsrudel, thanks for the request! 

 It looks like we have an initial implementation of this support added. See #2882 and let us know if you give it a try.

PR #2882, opened 2023-10-18
feat(si-cli): Add support for custom AWS ENDPOINT URLs

stack72:
Fixes: #2659

This allows us to point our SI instance against a localstack setup in the veritech container
All we need to do is to pass an endpoint url as part of the configure command :) It can even use fake credentials

PLEASE NOTE: Qualifications that use `--dry-run` in AWS will return an empty response from localstack - but you would be able to check the localstack logs for a 200 response:

```
2023-10-18T22:11:31.320  INFO --- [   asgi_gw_0] localstack.request.aws     : AWS ec2.CreateVpc => 200
```

https://share.descript.com/view/Uq3tBNcL3oU


stack72:
bors merge

PR #2883, opened 2023-10-18
right panel improvements

theoephraim:
next steps
- alternate tabs styling (the underline only version)
- make icons throughout open directly to the correct tab in the details panel


theoephraim:
bors merge

PR #2884, opened 2023-10-19
feat: (lang-js) Added zlib to the standard sandbox. 

keeb:
This will allow assets which need gzipped strings to be created.

Tested with a codegen func defined like so

```typescript
async function main(component: Input): Promise < Output > {
    const zippedstring = "This string will be zipped";

    zlib.gzip(zippedstring, (err, compressedData) => {
        if (!err) {
            console.log('Compressed data:', compressedData.toString('base64'));
        } else {
            console.error('Error compressing data:', err);
        }
    });

    return {
        format: "json",
        code: JSON.stringify(component),
    };
}
```

### DIFF
--- a/bin/lang-js/src/sandbox.ts
+++ b/bin/lang-js/src/sandbox.ts
@@ -2,6 +2,7 @@ import os from "os";
 import fs from "fs";
 import path from "path";
 import fetch from "node-fetch";
+import zlib from "zlib";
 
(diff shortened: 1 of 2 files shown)


keeb:
Added some typehints so at least people know zlib.gzip is available. It is not comprehensive.

keeb:
bors merge

PR #2886, opened 2023-10-19
feat(sdf,dal,cyclone): handle js crashes except for attributes

paulocsanz:
Attribute functions will need more work to deal with how to return the failure to the user. As regular attribute functions only return a prop value, like a string or number.

We can implement for code-gen, qualification, etc. Which will be the next steps, but regular propagation functions will be though.


paulocsanz:
bors try

paulocsanz:
bors try

paulocsanz:
bors try

paulocsanz:
bors merge

PR #2887, opened 2023-10-19
feat(web,vue-lib) alternate minimal tab style

theoephraim:
adds new minimal tab style which can be toggled on. For now we'll use wherever we have nested tabs, but may want to consider using in other places as well.

image

also fixes the diff tab when viewing head and other small tab related issues.

### DIFF
--- a/lib/vue-lib/src/design-system/tabs/TabGroup.vue
+++ b/lib/vue-lib/src/design-system/tabs/TabGroup.vue
@@ -41,15 +41,24 @@
             :class="
               clsx(
                 'focus:outline-none whitespace-nowrap',
-                'text-neutral-400 border-b border-neutral-300 dark:border-neutral-600 border-x border-t border-x-neutral-300 border-t-neutral-300 dark:border-x-neutral-600 dark:border-t-neutral-600',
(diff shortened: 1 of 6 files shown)


lib/vue-lib/src/design-system/tabs/TabGroup.vue
```diff
@@ -41,15 +41,24 @@
...
-                'h-11 px-2 text-sm inline-flex items-center rounded-t group-hover:border-shade-100 dark:group-hover:border-shade-0',
-                tab.props.slug === selectedTabSlug
-                  ? 'border-b-white dark:border-b-neutral-800 border-b text-action-700 dark:text-action-300 font-bold'
-                  : themeClasses(
-                      'hover:text-neutral-400 font-medium hover:bg-neutral-100',
-                      'hover:text-neutral-300 font-medium hover:bg-neutral-900',
-                    ),
+                'h-11 px-2 text-sm inline-flex items-center',
                 growTabsToFillWidth && 'flex-grow justify-center',
+                minimal
```
theoephraim, line 46: tailwind is the BEST 🤮 
really missing css variables and overriding with specificity here...

theoephraim:
bors merge

PR #2891, opened 2023-10-20
feat(web): store history state in localstorage

paulocsanz:
- Stores only the last 50 undos and 50 dones
- Deletes buffer if it was written more than a week ago


paulocsanz:
bors merge

PR #2895, opened 2023-10-23
feat(web,dal,sdf): add code editor widget

paulocsanz:
Does not have syntax highlighter, linter, etc because it doesn't know the language, we should improve that somehow (it's not clear to me right now how to do it best and it feels like a immediate benefit even without it)

image

### DIFF
--- a/app/web/src/api/sdf/dal/property_editor.ts
+++ b/app/web/src/api/sdf/dal/property_editor.ts
@@ -15,6 +15,10 @@ export enum PropertyEditorPropKind {
   Map = "map",
 }
 
+export interface PropertyEditorPropWidgetKindCodeEditor {
+  kind: "codeEditor";
+}
+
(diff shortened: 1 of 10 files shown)


paulocsanz:
bors merge

stack72:
The bors failure:

```


Fail: root//app/web:check-type (17.2s)
--
  | ---- STDOUT ----
  | src/components/PropertyEditor/WidgetCodeEditor.vue(10,11): error TS2322: Type 'string \| null' is not assignable to type 'string \| undefined'.
  | Type 'null' is not assignable to type 'string \| undefined'.
  | src/components/PropertyEditor/WidgetCodeEditor.vue(48,9): error TS2322: Type 'string \| null' is not assignable to type 'string \| undefined'.
  |  
  | ---- STDERR ----


```

PR #2898, opened 2023-10-24
feat(web) - secret management panel

wendybujalski:
Screenshot 2023-10-24 at 6 54 25 PM


app/web/src/components/SecretCard.vue
```diff
@@ -3,49 +3,134 @@
...
+            themeClasses(
+              'hover:text-action-500 text-neutral-400',
+              'hover:text-action-400',
+            ),
+          )
+        "
+      >
+        <Icon name="settings" size="lg" />
       </div>
+      <!-- TODO(Wendy) - here's the button we will use when we add deletion -->
```
vbustamante, line 112: Good call annotating the commented out code 😊

app/web/src/store/secrets.store.ts
```diff
@@ -130,6 +135,49 @@ export function useSecretsStore() {
...
+
+            out.push({
+              id: key,
+              latestChange: change,
+              secrets: value.secrets,
+            });
+          }
+
+          out.sort((a, b) => {
+            if (!a.latestChange && !b.latestChange) return 0;
```
vbustamante, line 166: Will we ever have secrets with undefined changes?

app/web/src/store/secrets.store.ts
```diff
@@ -130,6 +135,49 @@ export function useSecretsStore() {
...
+            else if (!a.latestChange && b.latestChange) return 1;
+            else if (a.latestChange && !b.latestChange) return -1;
+            else if (a.latestChange && b.latestChange) {
```
vbustamante, line 169: All of the previous cases returning values mean the code will only get here even without this if. This could be written something like:

```js
out.sort((a, b) => {
  if (!a.latestChange && !b.latestChange) return 0;
  if (!a.latestChange && b.latestChange) return 1;
  if (a.latestChange && !b.latestChange) return -1;

  return new Date(b.latestChange) - new Date(a.latestChange);
});
```

Also, Array.sort expects positive or negative values instead of only 1 and -1, so you could use subtraction of the dates to get a similar result. that's the reason for that last return line. See if you  like it

vbustamante (approved):
Look good! Some suggestions but no blockers

app/web/src/store/secrets.store.ts
wendybujalski: Ah yeah - I figured there was a simpler way to do this, but I kept getting Typescript issues - even the solution you suggested results in a Typescript error, probably because TS is not smart enough to notice that the three if statements filter out any undefined values before that last return statement. 😑

app/web/src/store/secrets.store.ts
wendybujalski: Yes - a secret definition that has been created but no individual secrets have been added yet will have undefined for the latestChange value.

wendybujalski:
bors merge

PR #2900, opened 2023-10-25
feat: deadpool docker implementation

sprutton1:
Super naive implementation of running cyclone in containers. Tons of refactoring work could happen here, but this is enough to get things working. Tests pass! Please crush me with feedback, this is still pretty new to me.


bin/cyclone/BUCK
```diff
@@ -1,4 +1,30 @@
...
+    ],
+)
+
+test_suite(
+    name = "check-format",
+    tests = [
+        ":check-format-rust",
+        ":check-format-shell",
+    ],
+)
```
nickgerace, line 26: I'm surprised these doesn't come along with our `rust_binary` wrapper in `prelude-si`. I'm AFK at the moment, but will look later.

Cargo.toml
```diff
@@ -51,6 +51,7 @@ async-trait = "0.1.68"
 axum = { version = "0.6.18", features = ["macros", "multipart", "ws"] }
 base64 = "0.21.0"
 blake3 = "1.3.3"
+bollard = "0.15.0"
```
nickgerace, line 54: Aren't we using a different library for the `si` binary? Feels like we should pick one and throw it in a theoretical `lib/si-docker` wrapper crate.
cc @stack72 @fnichol

paulocsanz (approved):
This is awesome!

One question, are we rebuilding the docker image with the new cyclone after it or its dependency changes?

This is only for the development process or prod deployment, right? The si binary will run cyclone directly since it's already in veritech's docker image?

lib/deadpool-cyclone/src/instance/cyclone/local_uds.rs
```diff
@@ -90,12 +103,17 @@ impl Instance for LocalUdsInstance {
...
+        match self.runtime.runtime_strategy {
+            LocalUdsRuntimeStrategy::LocalProcess => {
+                if !self.watch_shutdown_tx.is_closed() && self.watch_shutdown_tx.send(()).is_err() {
+                    debug!("sent watch shutdown but receiver was already closed");
+                }
+                return self.runtime.terminate().await;
+            }
+            LocalUdsRuntimeStrategy::LocalDocker => {
+                return self.runtime.terminate().await;
+            }
```
nickgerace, line 115: I'd have to check what `terminate()` does, but you should not have to return them since the function returns `result::Result<(), Self::Error>`. You can likely use the "`try_intro`" operator ("`?`") instead.

lib/deadpool-cyclone/src/instance/cyclone/local_uds.rs
```diff
@@ -311,11 +333,10 @@ impl Spec for LocalUdsInstanceSpec {
...
     async fn spawn(&self) -> result::Result<Self::Instance, Self::Error> {
         let (temp_path, socket) = temp_path_and_socket_from(&self.socket_strategy)?;
-        let mut cmd = self.build_command(&socket);
-
-        debug!("spawning child process; cmd={:?}", &cmd);
-        let child = cmd.spawn().map_err(Self::Error::ChildSpawn)?;
+        let mut runtime =
+            LocalInstanceRuntime::build(&self.runtime_strategy, &socket, self.clone()).await?;
 
+        let _ = runtime.spawn().await;
```
nickgerace, line 339: What are we eating here?

lib/deadpool-cyclone/src/instance/cyclone/local_uds.rs
```diff
@@ -486,6 +476,243 @@ fn temp_path_and_socket_from(
...
+                    .sample_iter(&Alphanumeric)
+                    .take(10)
+                    .map(char::from)
+                    .collect();
+
+                let socket_dir = sock
+                    .parent()
+                    .expect("socket path not available")
+                    .display()
+                    .to_string();
```
nickgerace, line 597: Going through the `Display` trait may not be ideal here. Perhaps [`to_str`](https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.to_str) would be the way to go?

lib/deadpool-cyclone/src/instance/cyclone/local_uds.rs
```diff
@@ -486,6 +476,243 @@ fn temp_path_and_socket_from(
...
+                        source: Some(socket_dir.clone()),
+                        target: Some(socket_dir.clone()),
+                        typ: Some(MountTypeEnum::BIND),
+                        ..Default::default()
+                    },
+                    Mount {
+                        source: Some(spec.cyclone_decryption_key_path),
+                        target: Some(String::from("/tmp/key")),
+                        typ: Some(MountTypeEnum::BIND),
+                        ..Default::default()
```
nickgerace, line 609: Good usage of the `Default` trait! 😄

lib/deadpool-cyclone/src/instance/cyclone/local_uds.rs
```diff
@@ -486,6 +476,243 @@ fn temp_path_and_socket_from(
...
+                    },
+                    Mount {
+                        source: Some(
+                            spec.lang_server_cmd_path
+                                .as_path()
+                                .to_string_lossy()
```
nickgerace, line 615: Similarly to the above, we may want to use `to_str`. This is because [`to_string_lossy`](https://doc.rust-lang.org/std/path/struct.Path.html#method.to_string_lossy) does the following:

> Any non-Unicode sequences are replaced with [U+FFFD REPLACEMENT CHARACTER](https://doc.rust-lang.org/std/char/constant.REPLACEMENT_CHARACTER.html).

I don't think we want to eat that kind of error despite it being impossible to hit.

lib/deadpool-cyclone/src/instance/cyclone/local_uds.rs
```diff
@@ -486,6 +476,243 @@ fn temp_path_and_socket_from(
...
+                        target: Some(String::from("/tmp/langserver")),
+                        typ: Some(MountTypeEnum::BIND),
+                        ..Default::default()
+                    },
+                ];
+
+                let container_id = docker
+                    .create_container(
+                        Some(CreateContainerOptions {
+                            name: format!("cyclone-container-{rand_string}"),
```
nickgerace, line 627: Good use of the `format!` macro!

lib/deadpool-cyclone/src/instance/cyclone/local_uds.rs
```diff
@@ -486,6 +476,243 @@ fn temp_path_and_socket_from(
...
+/// Runtime strategy when spawning [`Instance`]s.
+#[remain::sorted]
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+#[serde(untagged)]
+pub enum LocalUdsRuntimeStrategy {
+    /// Run Docker containers on the local machine
+    LocalDocker,
+    /// Run processes on the local machine
+    LocalProcess,
+}
```
nickgerace, line 488: Nice! We want to keep both, so I am happy to see this.

lib/deadpool-cyclone/src/instance/cyclone/local_uds.rs
```diff
@@ -486,6 +476,243 @@ fn temp_path_and_socket_from(
...
+                    Some(process::Signal::SIGTERM),
+                    None,
+                )
+                .await?;
+                Ok(())
+            }
+            LocalUdsRuntimeStrategy::LocalDocker => {
+                self.docker_api
+                    .as_mut()
+                    .expect("docker socket should be bound")
```
nickgerace, line 698: I would not panic anywhere in the library if you can help it. It should be impossible, but eh??

lib/deadpool-cyclone/src/instance/cyclone/local_uds.rs
```diff
@@ -486,6 +476,243 @@ fn temp_path_and_socket_from(
...
+                    .remove_container(
+                        &self
+                            .container_id
```
nickgerace, line 701: Does this type implement `Copy`? If so, you should not have to clone it.

lib/deadpool-cyclone/src/instance/cyclone/local_uds.rs
```diff
@@ -486,6 +476,243 @@ fn temp_path_and_socket_from(
...
+                            .clone()
+                            .expect("container id should be available"),
```
nickgerace, line 703: Similar note to the above `expect` comment.

lib/deadpool-cyclone/src/lib.rs
```diff
@@ -155,4 +158,55 @@ mod tests {
...
+        let spec = LocalUdsInstance::spec()
+            .try_cyclone_cmd_path(config_file.cyclone.cyclone_cmd_path())
+            .expect("failed to find cyclone program")
+            .cyclone_decryption_key_path(config_file.cyclone.cyclone_decryption_key_path())
+            .try_lang_server_cmd_path(config_file.cyclone.lang_server_cmd_path())
+            .expect("failed to find lang server program")
+            .limit_requests(2)
+            .runtime_strategy(LocalUdsRuntimeStrategy::LocalDocker)
+            .socket_strategy(LocalUdsSocketStrategy::Random)
+            .ping()
```
nickgerace, line 177: What does `ping()` do? Just curious.

lib/deadpool-cyclone/src/lib.rs
```diff
@@ -155,4 +158,55 @@ mod tests {
...
+        instance
+            .execute_ping()
+            .await
+            .expect("failed execute ping")
+            .start()
+            .await
+            .expect("failed to start protocol");
+        instance.ensure_healthy().await.expect("failed healthy");
+
+        instance.terminate().await.expect("failed to terminate");
```
nickgerace, line 210: Great test! One cool thing you could do is write an infallible wrapper around the test and then whenever it returns, you call terminate. Here's an example:

```rust
#[tokio::test]
async fn pow() {
    // Set up the instance here.
    // --snip--

    if let Err(e) = pow_inner().await {
        instance.terminate().await.expect("failed to terminate");
        panic!("{}", e);
    } 
}
```

This way, we pretty much always terminate. It's not foolproof, but it could prove useful.

Cargo.toml
stack72: yes we use [docker-rs](https://github.com/vv9k/docker-api-rs)

lib/deadpool-cyclone/src/lib.rs
sprutton1: It enables the `/ping` route for Cyclone so we can poke it and make sure it's still up.

lib/deadpool-cyclone/src/instance/cyclone/local_uds.rs
sprutton1: What's the right way to handle this? I'm fine propagating the error, I'm just not sure what the right pattern to do so is

Cargo.toml
sprutton1: We talked very briefly about this the other day. The consensus was that `bollard` is a little more feature rich and supported, so if we're going to create a wrapper we should probably use that.

sprutton1:
> One question, are we rebuilding the docker image with the new cyclone after it or its dependency changes? Both for buck2 and cargo run (mostly useful for tests ran through IDEs)?

I've added buck2 targets for building the cyclone container. I could use a hand understanding how to wire it up as you describe.

> This is only for the development process or prod deployment, right? The si binary will run cyclone directly since it's already in veritech's docker image?

Intended for dev use and proof-of-concept that we can add a new runtime type. The exiting method for running still works and is used by default.

sprutton1:
bors try

PR #2902, opened 2023-10-26
feat(web): implement multiple users per workspace

paulocsanz:
- Invite users by email
- We do not validate emails for password logins
- Hard reload page if workspace changes to avoid reactivity issues

### DIFF
--- a/lib/dal/src/user.rs
+++ b/lib/dal/src/user.rs
@@ -11,6 +11,7 @@ use crate::{
 };
 
 const USER_GET_BY_PK: &str = include_str!("queries/user/get_by_pk.sql");
+const USER_GET_BY_EMAIL_RAW: &str = include_str!("queries/user/get_by_email_raw.sql");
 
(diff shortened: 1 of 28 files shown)


lib/dal/src/user.rs
```diff
@@ -154,10 +207,15 @@ impl UserClaim {
...
 
     pub async fn from_bearer_token(
+        workspace_pk: Option<WorkspacePk>,
         public_key: JwtPublicSigningKey,
         token: impl AsRef<str>,
     ) -> UserResult<UserClaim> {
         let claims = crate::jwt_key::validate_bearer_token(public_key, &token).await?;
-        Ok(claims.custom)
+        let mut custom = claims.custom;
+        if let Some(workspace_pk) = workspace_pk {
```
paulocsanz, line 216: I feel this is the biggest risk that could cause a 401

paulocsanz:
bors merge

PR #2903, opened 2023-10-26
fix: bump docker-api crate to git ref

johnrwatson:
Ref: Discord issue regarding issues with SI Launcher.

[Bump models to v1.43](https://github.com/vv9k/docker-api-rs/commit/851f7091e945d04c055b24809a9b375fc3f0c80d) was merged into the docker-api crate on 10th June
The commit/change never made it into the crate itself, being [last published on 5th June](https://crates.io/crates/docker-api/versions) (0.14.0): There isn't a version of this crate with the updated option config in it

To that end, moved to using a git reference off HEAD for the build.

NB: It's noteworthy we have two different crates for docker-api interaction across our monorepo: bollard and docker-api. I suggest at some point we consolidate these/use the most common one.

Added `:si` build target for Buck for convenience

```
~/D/RISCOSPi.5.28 > si update


Message:  Unable to push events to Posthog: ()
Location: bin/si/src/main.rs:166

Error: 
   0: docker api: missing field `VirtualSize` at line 1 column 1626
   1: missing field `VirtualSize` at line 1 column 1626

Location:
   bin/si/src/main.rs:138

```

### DIFF
--- a/BUCK
+++ b/BUCK
@@ -29,6 +29,11 @@ alias(
 )
 
 alias(
+    name = "si",
+    actual = "//bin/si:si",
+)
+
+alias(
     name = "veritech",
     actual = "//bin/veritech:veritech",
 )
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1589,8 +1589,7 @@ checksum = "0688c2a7f92e427f44895cd63841bff7b29f8d7a1648b9e7e07a4a365b2e1257"
 [[package]]
 name = "docker-api"
 version = "0.14.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4576e993f4ef4805931a835b7afe71dec6cbd096e9b08bbd8e8f03e1940b2745"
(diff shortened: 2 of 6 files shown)


jhelwig (approved):

johnrwatson:
Confirmed expected fix is in downstream package:
```
    /// Total size of the image including all layers it is composed of.
    ///
    /// In versions of Docker before v1.10, this field was calculated from
    /// the image itself and all of its parent images. Images are now stored
    /// self-contained, and no longer use a parent-chain, making this field
    /// an equivalent of the Size field.
    ///
    /// Deprecated: this field is kept for backward compatibility, and will be removed in API v1.44.
    pub virtual_size: Option<i64>,
 ```
 
 Confirmed via `buck2 run //bin/si <commands>` that the CLI still functions at a basic level/no major regressions

johnrwatson:
bors merge

nickgerace (commented):
Great PR description too! We need more of these.

PR #2905, opened 2023-10-26
fix(dal): only create payload apa if missing

zacharyhamm:


stack72:
This fixes the issue on main right now! thank you :)

PR #2906, opened 2023-10-26
feat(auth-portal, auth-api): Add the ability to invite a user to a workspace

stack72:
In this PR, we check to see if the invited user is part of our system and if not, we create them a user based account that will be updated when they sign up

We have given the people that are invited to the workspace the “COLLABORATOR” role and we know if they have accepted the invite

There is no remove functionality at this time - this can be added in later as it's just a new route in the Auth API and a new button on the list

This work is using the CREATE_WORKSPACE feature flag so it's only visible to System Initiative users at this time

Screenshot 2023-10-27 at 01 52 33

### DIFF
--- a/app/auth-portal/src/components/WorkspaceLinkWidget.vue
+++ b/app/auth-portal/src/components/WorkspaceLinkWidget.vue
@@ -1,7 +1,11 @@
 <template>
   <div class="flex flex-row items-stretch">
-    <div
+    <RouterLink
       v-if="!compact && featureFlagsStore.EDIT_WORKSPACES"
+      :to="{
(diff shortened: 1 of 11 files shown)


paulocsanz:
I used INVITE_USER feature flag in my PR, we may want to centralize that later. But awesome PR, we can sync tomorrow!

stack72:
> I used INVITE_USER feature flag in my PR, we may want to centralize that later. But awesome PR, we can sync tomorrow! But feel free to merge this

Great catch! I have changed to use the same flag :)

stack72:
bors merge

PR #2916, opened 2023-10-27
feat(web,dal,sdf): stream cursor and show on screen

paulocsanz:
Streams cursors if hovering diagram (but can be easily expanded to other
containers).

### DIFF
--- a/app/web/src/components/Workspace/WorkspaceModelAndView.vue
+++ b/app/web/src/components/Workspace/WorkspaceModelAndView.vue
@@ -21,7 +21,11 @@
     </template>
   </ResizablePanel>
 
-  <div class="grow h-full relative bg-neutral-50 dark:bg-neutral-900">
+  <div
+    class="grow h-full relative bg-neutral-50 dark:bg-neutral-900"
+    @mouseover="mouseOverDiagram"
+    @mouseout="mouseOutDiagram"
+  >
     <!--div
       v-if="!statusStore.globalStatus.isUpdating && isViewMode"
       :class="
@@ -41,6 +45,7 @@
       :customConfig="diagramCustomConfig"
       :nodes="diagramNodes"
       :edges="diagramEdges"
+      :cursors="cursors"
       @insert-element="onDiagramInsertElement"
       @hover-element="onDiagramHoverElement"
       @move-element="onDiagramMoveElement"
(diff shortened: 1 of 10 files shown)


paulocsanz:
bors merge

paulocsanz:
bors cancel

paulocsanz:
bors try

paulocsanz:
bors try

app/web/src/components/Workspace/WorkspaceModelAndView.vue
```diff
@@ -177,8 +187,108 @@ import ComponentCard from "../ComponentCard.vue";
 import EdgeCard from "../EdgeCard.vue";
 import NoSelectionDetailsPanel from "../NoSelectionDetailsPanel.vue";
 
+const realtimeStore = useRealtimeStore();
 const changeSetStore = useChangeSetsStore();
 const fixesStore = useFixesStore();
+const authStore = useAuthStore();
+
+const MOUSE_REFRESH_RATE = 20;
```
theoephraim, line 195: I'd probably move most of this logic to a store

paulocsanz:
bors try

app/web/src/store/realtime/realtime.store.ts
```diff
@@ -183,6 +183,28 @@ export const useRealtimeStore = defineStore("realtime", () => {
     });
   }
 
+  const sendCursor = (req: {
```
theoephraim, line 186: I might abstract this so the realtime store just knows about how to send a message (and handle json stuff) and then something like the "presence store" would deal with handling all the incoming and outgoing stuff about presence

app/web/src/components/Workspace/WorkspaceModelAndView.vue
paulocsanz: I would wait to do that until we want to extend on this. If this logic gets bigger or other places have to replicate it. It will also guide us on the best way to do it.

Doesn't feel required for this first iteration.

app/web/src/store/realtime/realtime.store.ts
paulocsanz: Yeah, we definitely will want to centralize that, but again. Let's wait until other real time events are added. That will show us how to best encapsulate while supporting other cases.

Doesn't feel required for this first iteration.

paulocsanz:
bors merge

PR #2920, opened 2023-10-31
feat(sdf-server, auth-portal, web): adding a page that we can send a post request to that forces auth cleanup

stack72:
This page is basic for now and we redirect both ways but we can make this better. It will compare the users from the auth-api against what’s in our DB - the auth-api is the source of truth!!

There are a number of improvements needed for this but we can start to use it immediately for internal use

In the future, this will be replaced by the use of websockets to do the same job


stack72:
bors merge

PR #2921, opened 2023-10-31
chore(auth-api): Add some tests regarding the new members endpoint in auth-api

stack72:
This will show that it can be accessed, listed, added, deleted


stack72:
bors merge

PR #2922, opened 2023-10-31
bug (app): removed repeat line for testlogs

keeb:
Previously the test log was repeated 1000 times, now it is not.


keeb:
bors merge

PR #2926, opened 2023-11-01
feat(auth-api, auth-portal): Stop allowing unverified users to log into workspaces through the Dashboard

stack72:
Also ensure that we track the invitations in posthog to be able to send email invitations

### DIFF
--- a/app/auth-portal/src/pages/WorkspaceDetailsPage.vue
+++ b/app/auth-portal/src/pages/WorkspaceDetailsPage.vue
@@ -173,6 +173,7 @@ import { useRouter } from "vue-router";
 import { useAuthStore } from "@/store/auth.store";
 import { useWorkspacesStore, WorkspaceId } from "@/store/workspaces.store";
 import { useFeatureFlagsStore } from "@/store/feature_flags.store";
+import { tracker } from "@/lib/posthog";
 
 const authStore = useAuthStore();
 const workspacesStore = useWorkspacesStore();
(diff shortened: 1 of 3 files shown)


app/auth-portal/src/pages/WorkspaceDetailsPage.vue
```diff
@@ -287,6 +288,12 @@ const inviteButtonHandler = async () => {
   const res = await workspacesStore.INVITE_USER(newMember, props.workspaceId);
 
   if (res.result.success) {
+    tracker.trackEvent("workspace_invitation", {
```
theoephraim, line 291: id fire this off from the backend as it will be more reliable

bin/auth-api/src/routes/workspace.routes.ts
```diff
@@ -189,6 +189,10 @@ router.get("/workspaces/:workspaceId/go", async (ctx) => {
   // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
   const authUser = ctx.state.authUser!;
 
+  if (!authUser.emailVerified) {
```
theoephraim, line 192: currently this is going to leave you on an ugly JSON error page - which is something we'll want to deal with eventually anyway... But we should also probably add some more checks in a few places in the auth portal and show a nice error message so they will be unlikely to get here

theoephraim (approved):

bin/auth-api/src/routes/workspace.routes.ts
stack72: We are going to have the ability to be able to trigger a verification check / re-send in the UI at some point soon. I talked ti @britmyerss about this ugly page and it's fine for now ™️

stack72:
bors merge

PR #2927, opened 2023-11-01
feat(app): Add ability to link a user from the workspace to workspace management in auth portal

stack72:
Screenshot 2023-11-01 at 21 27 58


stack72:
bors merge

PR #2929, opened 2023-11-01
feat(sdf-server): Ensure that the user is authorized before allowing any interactions

stack72:
If a user has been removed from a workspace, then we make a best guess at removing them from the workspace by removing them from the user_belongs_to_workspace table

That wasn't fully implemented because if they were still on the workspace then they could still interact with the diagram. If we remove the user from the user_belongs_to_workspace table then this will now check on any SDF interaction and if unathorized, the UI will kick them out

### DIFF
--- a/lib/dal/src/user.rs
+++ b/lib/dal/src/user.rs
@@ -135,9 +135,16 @@ impl User {
         }
     }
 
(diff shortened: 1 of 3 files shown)


stack72:
bors merge

PR #2932, opened 2023-11-03
feat(auth-api, auth-portal): Show the owner of the workspace when you are an editor

stack72:
Also, show when you were invited to a workspace. Any existing invitations will be set to the date we run this migration as we don’t have a backward set of data on this

Screenshot 2023-11-03 at 18 28 23


stack72:
bors merge

PR #2934, opened 2023-11-06
fix(auth-api): Ensure that signing up as an invited user doesn’t create new user

stack72:
When a non-SI user is invited to a workspace, we create a row in the Users table with the email of the user and no other details. Previously, when that user attempted to signup, we didn’t check for a partially created user and instead we allowed the user to create a new account. This meant that the invitation to the workspace was for a “dead” user - i.e. a user that didn’t actually exist.

In this PR, we check for an existing user being either auth0 based OR that has an email and no signup date (email + no signup date means that they are invited but were not previously a member). This means we can now find the correct user and thus update the existing user with the auth0 response rather than creating a new user

### DIFF
--- a/bin/auth-api/src/services/users.service.ts
+++ b/bin/auth-api/src/services/users.service.ts
@@ -60,8 +60,23 @@ export async function createOrUpdateUserFromAuth0Details(auth0UserData: Auth0.Us
   // auth0 docs showing user_id, but looks like "sub" contains the identifier
   // TODO: check data when logging in with other providers
   const auth0Id = auth0UserData.user_id || (auth0UserData as any).sub;
+  const email = auth0UserData.email;
 
-  const existingUser = await getUserByAuth0Id(auth0Id);
+  // We want to ensure we track that we need to create a workspace for a partially
+  // signed up (i.e. invited) user.
+  let requiresWorkspace = false;
+
+  let existingUser = await getUserByAuth0Id(auth0Id);
+  if (!existingUser) {
+    // We should check that we have a user that is potentially partially signed up
+    // a partially signed up user is a user that we have a record for but that record
+    // only has an email and not anything else
+    const potentialUser = await getUserByEmail(email!);
+    if (potentialUser && !potentialUser.signupAt) {
+      existingUser = potentialUser;
+      requiresWorkspace = true;
+    }
+  }
 
   const userData = {
     // TODO: figure out json fields...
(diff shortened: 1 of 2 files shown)


bin/auth-api/src/services/users.service.ts
```diff
@@ -39,7 +39,7 @@ export async function getUserById(id: UserId) {
 }
 export type UserWithTosStatus = NonNullable<Awaited<ReturnType<typeof getUserById>>>;
 
-export async function getUserByAuth0Id(auth0Id: string) {
+export async function getUserByAuth0Details(auth0Id: string) {
```
theoephraim, line 42: why rename? In this case we are passing in the auth0 id, not the full blob of details?

bin/auth-api/src/services/users.service.ts
```diff
@@ -60,8 +60,23 @@ export async function createOrUpdateUserFromAuth0Details(auth0UserData: Auth0.Us
...
+  let requiresWorkspace = false;
+
+  let existingUser = await getUserByAuth0Details(auth0Id);
+  if (!existingUser) {
+    // We should check that we have a user that is potentially partially signed up
+    // a partially signed up user is a user that we have a record for but that record
+    // only has an email and not anything else
+    const potentialUser = await getUserByEmail(email!);
+    if (potentialUser && !potentialUser.signupAt) {
+      existingUser = potentialUser;
```
theoephraim, line 76: we also need to make sure we update the signupAt timestamp!

bin/auth-api/test/auth-routes.test.ts
```diff
@@ -83,7 +85,9 @@ t.test('Auth routes', async () => {
 
     t.test('verify sdf auth token succeeds for whoami', async () => {
       const authData = await decodeAuthToken(validToken);
-      const sdfToken = createSdfAuthToken(authData.userId, 'foo');
+      const user = await getUserById(authData.userId);
+      const workspace = await createWorkspace(user!);
```
theoephraim, line 89: prob need a linting ignore here for the `!` or you can first check its not empty and throw an error

bin/auth-api/src/services/users.service.ts
stack72: sorry, that's a hangup on that I was also going to pass email - will revert that!

bin/auth-api/src/services/users.service.ts
stack72: that's already happening:

```
if (existingUser) {
    if (!existingUser.signupAt) {
      existingUser.signupAt = new Date();
    }
    ```

bin/auth-api/test/auth-routes.test.ts
stack72: Fixed

theoephraim (approved):
🚀

stack72:
bors merge

PR #2935, opened 2023-11-06
fix(auth-api): Ensure we are setting auth0id

stack72:
If a user was invited, then signed up, we would get the auth0Id but never set it to the database. When they logged back in, they would have a signupAt but no auth0Id which would cause a user to be a new user not an existing user

This would have again caused a user to lose their invite as it would be for an account that didn’t exist


stack72:
bors merge

PR #2937, opened 2023-11-07
feat(lang-js, dal, sdf-server): Add ability to set documentation for a prop

stack72:
This means we can do things like:

```
 const propName = new PropBuilder()
  .setName("name")
  .setKind("string")
  .setDocumentation("This is the documentation for the prop")
  .setWidget(new PropWidgetDefinitionBuilder().setKind("text").build())
 .build();
```

And that documentation is plumbed the whole way through the pkg layer, sdf and to the frontend where we can do with it as we wish

### DIFF
--- a/lib/dal/src/migrations/U2404__add_documentation_to_props.sql
+++ b/lib/dal/src/migrations/U2404__add_documentation_to_props.sql
@@ -1,0 +1,64 @@
+-- Add the new columns to the props table.
+ALTER TABLE props ADD COLUMN documentation text;
+
+-- The new prop creation function automatically builds the path using the parent and sets the belongs to relationship.
+CREATE OR REPLACE FUNCTION prop_create_v2(
+    this_tenancy jsonb,
+    this_visibility jsonb,
+    this_name text,
+    this_kind text,
+    this_widget_kind text,
+    this_widget_options jsonb,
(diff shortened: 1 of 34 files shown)


lib/dal/src/migrations/U2404__add_documentation_to_props.sql
```diff
@@ -0,0 +1,64 @@
+-- Add the new columns to the props table.
```
zacharyhamm, line 1: Good on you for going the create_v2 route but we could also just get away with replacing the original v1 at this point

zacharyhamm (approved):
ship it!

zacharyhamm:
looks like some calls to Prop::new were not fixed up

stack72:
> looks like some calls to Prop::new were not fixed up

ah completely forgot our integration tests are different buck2 targets :/

keeb:
Sanity test complete. Verified correct fields are being returned to the UI, and that the UI is not doing anything with it.

Tests run
* Try to set documentation on a socket - both linting and creation failed ✔️ 
* Create asset with no documentation ✔️
* Create asset with all fields having documentation ✔️
* Create asset with some fields having documentation ✔️

LGTM 🚢

stack72:
bors merge

PR #2947, opened 2023-11-11
fix(web): Update terminology when updating an Asset

stack72:
Previous it would say create when it was updating now looks like

Screenshot 2023-11-11 at 17 39 00


stack72:
bors merge

PR #2951, opened 2023-11-13
chore(web): Cleanup old feature flags for test panel and autoattach funcs

stack72:


keeb:
LGTM 👍 

Test func panel works. Every asset has an attribute func. Can edit the asset and attach new funcs

stack72:
bors merge

PR #2953, opened 2023-11-15
Create base Component Attribute Values at Component creation

jhelwig:


lib/dal/src/migrations/U0035__components.sql
```diff
@@ -62,6 +75,309 @@ BEGIN
...
+        -- Create an Attribute Value for the Prop.
+        this_attribute_context := attribute_context_build_from_parts_v1(
+            this_internal_provider.prop_id, -- Prop ID
+            ident_nil_v1(), -- Internal Provider ID
+            ident_nil_v1(), -- External Provider ID
+            this_new_row.id -- Component ID
+        );
+
+        SELECT *
+        INTO this_attribute_prototype
```
paulocsanz, line 256: I changed this from a INTO STRICT because I was observing multiple prototypes for a prop, even on main.

This may also be true for the other branches, for providers, but I don't know enough about this logic to say much. We should take a look into it.

stack72:
bors merge

PR #2954, opened 2023-11-16
new attributes panel!

theoephraim:
very sexy new attributes panel. Much more to do, and we'll keep polishing but this is a huge improvement already :)
image


app/web/src/store/component_attributes.store.ts
```diff
@@ -62,59 +77,90 @@ export const useComponentAttributesStore = (componentId: ComponentId) => {
...
-                    } else {
-                      return true;
-                    }
-                  },
-                );
-              } else {
-                filteredChildValues[parentValue.id] = childValuesIds;
-              }
+          // recombine the schema + values + validations into a single nested tree that can be used by the attributes panel
+          attributesTree: (state): AttributeTreeItem | undefined => {
```
theoephraim, line 81: I hope this should be much easier to follow whats happening... As this stabilizes, we may want to fetch the data from the backend in a shape more like this, but we'll see!

lib/vue-lib/src/design-system/general/JsonTreeExplorer/JsonTreeExplorer.vue
```diff
@@ -0,0 +1,146 @@
+<template>
```
theoephraim, line 1: this is a general purpose JSON tree explorer - super useful for debugging stuff

theoephraim: it also was the basis for the attributes panel since it has a similar recursive setup and shares some collapsing behaviour

app/web/src/components/AttributesPanel/AttributesPanelItem.vue
```diff
@@ -0,0 +1,1009 @@
+<template>
```
theoephraim, line 1: this is where the bulk of magic happens...

I (probably) need to refactor this to pull out each input type into their own components (similar to what the "widget" stuff was doing) but it's much easier to style consistently and figure out how it works when it's all together.

Note that I used regular css here because it was just way too hard to follow what was going on otherwise. Also some visual things really needed CSS rules as it had to do with siblings and parents...

We can refactor this back into tailwind classes at some point if it becomes problematic.

theoephraim:
bors merge

PR #2955, opened 2023-11-16
feat(Web): Change how Apply Changeset button works

stack72:
We now have 3 ways the modal can work:

-> There are no other users in the changeset and no actions - we get a simple confirmation

Screenshot 2023-11-16 at 17 35 08

-> There are no other users in the changeset BUT there are actions - we get a modal with a list of the actions that will be applied

Screenshot 2023-11-16 at 17 35 27

-> There are other users in the workspace - we now show a modal to the user who merged. That modal (in this turn) only gives the ability to `Skip Approval and Merge`

Screenshot 2023-11-16 at 17 35 54

The next follow on will be to show the approvals workflow.

The outstanding question is:

* At what point do we list the actions that will be merged in the changeset? I beliefe that we should show it as part of the voting dialogue…

### DIFF
--- a/app/web/src/components/ApplyChangeSetButton.vue
+++ b/app/web/src/components/ApplyChangeSetButton.vue
@@ -8,44 +8,91 @@
     loadingText="Applying Changes"
     :requestStatus="applyChangeSetReqStatus"
     :disabled="statusStoreUpdating"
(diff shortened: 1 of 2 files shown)


stack72:
bors merge

PR #2960, opened 2023-11-17
codemirror cursor colors

theoephraim:
copying the yjs codemirror plugin into our monorepo for now as activity is low and we will very likely be making more modifications. Eventually we can break it back out and publish, but this should make things easy for now...

also note that we are being dumb here about  leaving their existing build process intact and will just manually call `pnpm dist` and check in the dist folder. We'll very likely convert to proper typescript and fix this setup.


paulocsanz (approved):
Thank you for this!

paulocsanz:
bors merge

theoephraim:
bors retry

theoephraim:
bors merge

paulocsanz:
bors merge

PR #2962, opened 2023-11-17
feat(web, dal, sdf): Create a light approval / notification system

stack72:
There are a number of differnet states going on here:

1. No collaborators on the changeset
 * No Actions so we get a simple confirmation modal (this is new!!)
 * Actions so we get a list of actions to be applied in the modal

2. Collaborators in the changeset
 * We begin a change set approval - changeset status moves to NeedsApproval
 * Each collaborate gets a modal to vote (Accept, Pass, Reject)
 * When a collaborate has voted their changeset usage is locked until the merge has completed or been cancelled
 * If *ANY* of the collaborators reject the merge then the person needs to take an action of override or cancel. * If approved then it's a normal merge process * If cancelled then the changeset is unlocked for everyone and the status changes back to Open
 * If all collaborators choose pass or approve, then the changeset will autoapprove
 * If the person chooses to override the merge, then all collaborators are told so

### DIFF
--- a/app/web/src/api/sdf/dal/change_set.ts
+++ b/app/web/src/api/sdf/dal/change_set.ts
@@ -8,6 +8,7 @@ export enum ChangeSetStatus {
   Failed = "Failed",
   Closed = "Closed",
   Abandoned = "Abandoned",
+  NeedsApproval = "NeedsApproval",
 }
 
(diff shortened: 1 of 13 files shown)


stack72:
bors merge

PR #2956, opened 2023-11-16
feat: add vsock support to cyclone-server

johnrwatson:
Allows cyclone-server to connect via vsock.

The changes for deadpool and cyclone-client (if any) are not ready to be merged. The changes introduced by this should be non-functional/dead until the other half of the change is merged into main.

PR #2964, opened 2023-11-20
Revert "feat: add vsock support to cyclone-server"

stack72:
Reverts systeminit/si#2956

Will re-introduce these changes when we merge the firecracker support for real.


johnrwatson:
bors merge

PR #2965, opened 2023-11-20
Import schema variants and significantly improve graph performance in the switchover branch

nickgerace:


nickgerace:
Cherry picked

PR #2969, opened 2023-11-22
[wip] firecracker spawn variant

johnrwatson:
Adds a new function execution platform variant of (`LocalUdsRuntimeStrategy` named `LocalFirecracker`) which uses Firecracker Micro-VMs to execute the functions within.

This PR also includes conditional compiling of the VSOCK elements which are not supported on Mac `M` series chips and some others. This allows the default compilation execution strategy (LocalProcess) to be used on `all` hosts compatible with SI, not just those which are compatible with the `firecracker specific` elements on the build. It's important to note that due to this, binaries produced on a host which does not support firecracker, will not be able to execute on firecracker even on another host that `does` support it. 

Currently the Firecracker host must be `local` to the context of the rest of the stack. I.e. this PR does not add "remote execution host" functionality to the SI stack although adding this should be possible leveraging a router or similar execution-host manager. Additionally, because of the above and the nature of how firecracker has to interact at a privileged level, to run the stack under the state of this PR, the `buck2 run dev` needs to run as `root` to allow it to take the relevant actions.

<hr />

This change will be dead code, unless the values for  these attributes are set:

Enabling in Cyclone Server:
`lib/deadpool-cyclone/src/instance/cyclone/local_uds.rs`
```
impl Default for LocalUdsRuntimeStrategy {
    fn default() -> Self {
        Self::LocalFirecracker                           // set from Self::LocalProcess
    }
}
```

Enabling in Cyclone Client:
`lib/cyclone-client/src/client.rs`
```
impl Default for ClientConfig {
    fn default() -> Self {
        Self {
            firecracker_connect: true,             // firecracker-setup: set from false
            watch_timeout: Duration::from_secs(10),
        }
    }
}
```

<hr/>

**Kernel and Rootfs Generation:**
- Producing the kernel image at this time is as follows:
  - Clone the kernel repo https://github.com/torvalds/linux on the relevant/chosen git tag
  - Build the kernel image using `make vmlinux` / `make Image` depending on platform, drive the installation with a supported kernel configuration from [here](https://github.com/firecracker-microvm/firecracker/tree/main/resources/guest_configs)
  
 - Producing the Rootfs at this time is as follows:
   - Clone SI repo
   - Run `buck2 run bin/cyclone:image` to generate the cyclone image (This puts it into the docker daemon on the host)
   - Use the outputted cyclone image reference into the `bin/cyclone/create-firecracker-root-fs.sh` script in the `docker run stage` and execute the script. This outputs the new rootfs into the `/firecracker-data/` folder on the host.

We will migrate the production of both of these artifacts into CI at some point in the future.

<hr/>

**Execution Environment**
- Each Firecracker micro-vm is spawned via the jailer binary, a binary which significantly improves the security posture of the host, find more information on exactly what it does here [jailer docs]( https://github.com/firecracker-microvm/firecracker/blob/main/docs/jailer.md)
- Each micro-vm has it's own network namespace and two routes: 
  - One out to the internet via the external host interface
  - One VSOCK interface to allow cyclone-server and cyclone-client (veritech) to talk privately via the virtio-vsock kernel module
- Each has it's own Linux user, group and root filesystem path on the host under which it has limited permissions. The path for these are set to the default for firecracker at the minute, namely: `/srv/jailer/firecracker/<id>/root/`
- Currently we don't pass any cgroups to the jailer, but this can be added for process-level cpu allocation isolation in the future.
- Each execution has it's own process namespace
<hr/>

**Validation on Host**
A test has been added (disabled) within `lib/deadpool-cyclone/src/lib.rs` called `chop` which validates that the firecracker execution environment on the host running the test is valid and works at a high level (can launch firecracker VM and establish a ping/pong websocket connection through to cyclone server).

**Prerequisites on Host**
- The host must have the SI repository and all prerequisites to compile and run the code. 
- The Firecracker VMM is built to be processor agnostic. 64-bit Intel, AMD and Arm CPUs with hardware virtualization support are generally available for production workloads. (i.e. does not support Mac `M` series chips)
- The host must have had the [si-firecracker-config](https://github.com/systeminit/si-firecracker-config/blob/main/machine-userdata.sh) host config executed against it. It's likely at some point in the near future this config will merge into the main `si` monorepo to keep the rust and required system configuration together.

**Further Notes**

- There are various TODO's left in the code which can be dealt with at a later point in time:
  - Hardcoded ProcessRuntime: This should be configurable or otherwise driven by external system/environment settings. 
  - Killall VMs script (currently we only kill one at a time on a host)
  - Observability deep dive 
    - We have no manner of debugging or seeing traces from cyclone-server running within firecracker.
    - How do we detect kernel issues (ooms, etc) - we have no real manner to detect or monitor for these
  - We are using a shared SSH private key on the host to connect to micro-vms. We will likely need to protect this route in more formally using tailscale or similar in the future as customer code will be accessible within that context.
  - Build/test pipeline - As mentioned in the creation of the kernel and rootfs section, we currently manually create both these artifacts and have limited/zero test coverage over them until we manually check their validity.
  - Shareable remote host - As mentioned in the pre-amble description, this PR only supports a single-tenanted SI Stack as we do not talk remotely to a firecracker-capable host to run the execution. 
  - Capacity planning - how many VMs per metal instance, how much memory do the VMs need, etc. These calculations and estimations will come as time passes.
  - No formal performance tests have been executed but the system seems suitably performant at this time.
  - Tying cyclone image build to rootfs generation. Currently this is a manual process whereby 
  - stop.sh is not getting triggered on oom or similar firecracker fault. I.e. the firecracker micro-vm API process is left intact/unterminated. From the UI/cyclone-server perspective you get a connection lost error.
  - The provisioning strategy for micro-vms does not support parallelism at this time. The processes will clash when manipulating the hosts iptables rules and cause the secondary one to fail. When we move to use parallel execution we will need to make mild adjustments here to the pooling/launching strategy from veritech.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1187,7 +1187,7 @@ name = "cyclone-core"
 version = "0.1.0"
 dependencies = [
  "base64 0.21.2",
- "nix",
+ "nix 0.26.2",
  "remain",
  "serde",
  "serde_json",
@@ -1225,6 +1225,7 @@ dependencies = [
  "tokio",
  "tokio-serde",
  "tokio-util",
+ "tokio-vsock",
  "tower",
  "tower-http",
 ]
@@ -1443,7 +1444,7 @@ dependencies = [
  "deadpool",
  "derive_builder",
  "futures",
- "nix",
+ "nix 0.26.2",
  "rand 0.8.5",
  "remain",
  "serde",
@@ -2745,6 +2746,15 @@ dependencies = [
 
 [[package]]
 name = "memoffset"
+version = "0.6.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5aa361d4faea93603064a027415f07bd8e1d5c88c9fbf68bf56a285428fd79ce"
+dependencies = [
+ "autocfg",
+]
+
+[[package]]
+name = "memoffset"
 version = "0.7.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5de893c32cde5f383baa4c04c5d6dbdd735cfd4a794b0debdb2bb1b421da5ff4"
@@ -2920,6 +2930,18 @@ dependencies = [
 
 [[package]]
 name = "nix"
+version = "0.24.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fa52e972a9a719cecb6864fb88568781eb706bac2cd1d4f04a648542dbf78069"
+dependencies = [
+ "bitflags 1.3.2",
+ "cfg-if",
+ "libc",
+ "memoffset 0.6.5",
+]
+
+[[package]]
+name = "nix"
 version = "0.26.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bfdda3d196821d6af13126e40375cdf7da646a96114af134d5f417a9a1dc8e1a"
@@ -2927,7 +2949,7 @@ dependencies = [
  "bitflags 1.3.2",
  "cfg-if",
  "libc",
- "memoffset",
+ "memoffset 0.7.1",
  "pin-utils",
  "static_assertions",
 ]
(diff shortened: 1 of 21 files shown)


fnichol:
bors try

fnichol:
bors merge

PR #2972, opened 2023-11-23
feat(auth) handle email verification and refreshing from auth0

theoephraim:
- add auth api endpoints to refresh profile from auth0 and resend verification email
- refresh auth0 profile (which includes email verification status) while connecting to workspace
- add error banner on auth portal to get users to verify their email
- block workspace login widget when email is not verified
- general cleanup around user creation/login and profile fetching logic
- this uses a new "machine to machine" auth0 app, so we are connecting as our app to fetch user profile data, rather than as the user themselves (using their token)

NOTE - we'll need to add a new env var for the m2m auth0 client secret to deploy

### DIFF
--- a/app/auth-portal/src/App.vue
+++ b/app/auth-portal/src/App.vue
@@ -147,6 +147,31 @@
             <div
               class="m-lg p-lg dark:bg-neutral-800 bg-neutral-200 rounded-md"
             >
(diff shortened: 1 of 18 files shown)


theoephraim:
bors merge

theoephraim:
bors merge

PR #2973, opened 2023-11-23
feat: firecracker-host-config

johnrwatson:
This adds the host firecracker config which has been migrated from https://github.com/systeminit/si-firecracker-config which was used during development. As it turned out we should version these files alongside the codebase as they are directly used and are compatible with the rust of the same version. It's very likely that versions of these files will be compatible on multiple versions of the rust, but keeping these things together makes things easier to track/change in one single PR.

All of this code is "dead" i.e. no functional changes to anything existing in the codebase & can safely be assumed to not cause any regressions or otherwise. Sister PR to https://github.com/systeminit/si/pull/2969 but has no direct dependency on it.

**Explanation of the files below:**
- `component/firecracker-base/machine-userdata.sh`: The userdata wrapper script which clones from git used to provision a new machine to be "firecracker execution" compatible.
- `component/firecracker-base/orchestrate-install.sh`: The script that's called from the userdata above to then pull in the relevant artifacts from S3 and make other dependent changes on the host.
- `component/firecracker-base/default-variables.txt`: A variables file that can be used to drive the install orchestration. Set to defaults for now but can be leveraged in the future for custom installs/OS dependent changes.
- `component/firecracker-base/start.sh`: The "start" file which is called directly in deadpool cylcone to create a new executor
- `component/firecracker-base/start.sh`: The "stop" file which is called directly in deadpool cylcone to stop an existing executor and tidy up the filesystem


johnrwatson:
bors merge

PR #2975, opened 2023-11-23
build(deps): update Nix & Buck2 dependencies (2023-W47)

fnichol:

PR #2978, opened 2023-11-28
chore: update nix dependency in readme after update

johnrwatson:
Updates the version of nix required to run `main` after the platform wide deps upgrade conducted here https://github.com/systeminit/si/pull/2975


johnrwatson:
bors merge

PR #2983, opened 2023-11-28
feat: move jailer configuration to machine setup time

sprutton1:
This:
* Adds a new script that can prepare a jail in advance, including all necessary networking and filesystem bits
* Extends the orchestrate script to run this prepare script _n_ times with the intention that this can be used to prepare jails in advance (like when Veritech starts up)
* Removes all of the startup bits from `start.sh` besides actually running the VM, which should be much faster
* All aforementioned scripts are idempotent, so they can be re-run without explosions happening, except `start.sh` because...

I don't think we are ever running `terminate()` to clean up the VMs. This isn't an issue with docker (or tokio?) processes as they automatically die once complete, but there is a non-trivial amount of cleanup that needs to happen here. We'll need to address this in a follow-up. In addition, cleanup effectively deletes the jail, meaning we will need to recreate it after cleanup. There are race conditions to be wary of here.

I started to implement CoW for the rootfs as it takes a second to copy, but decided to push it out to another PR. That will come next. We are currently limited on VM count by disk size (images are 1GB, we could look into optimizing this). I'm not sure if CoW will help there.

Also, we'll want to add configuration to Veritech and have it run these scripts on startup when using Firecracker.

### DIFF
--- a/component/firecracker-base/orchestrate-install.sh
+++ b/component/firecracker-base/orchestrate-install.sh
@@ -222,6 +222,7 @@ prepare_jailers() {
     wait
   else
     echo "prepare_jailer.sh script not found, skipping jail creation."
+    exit 1
   fi
 }
 
--- a/component/firecracker-base/orchestrate-install.sh
+++ b/component/firecracker-base/orchestrate-install.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# Call this function with: 
+# Call this function with:
 # ./orchestrate-install.sh <filepath to variables>
 
 set -eo pipefail
@@ -27,15 +27,16 @@ check_os_release() {
     echo "Info: /etc/os-release shown below:"
     cat /etc/os-release
     echo "------------------------------------"
-    [[ "$(cat /etc/os-release | grep 'CentOS Linux release 7')" ]] && export OS_VARIANT=centos-7 && return 0 
-    [[ "$(cat /etc/os-release | grep 'CentOS Stream release 8')" ]] && export OS_VARIANT=centos-stream-8 && return 0 
+    [[ "$(cat /etc/os-release | grep 'CentOS Linux release 7')" ]] && export OS_VARIANT=centos-7 && return 0
+    [[ "$(cat /etc/os-release | grep 'CentOS Stream release 8')" ]] && export OS_VARIANT=centos-stream-8 && return 0
     [[ "$(cat /etc/os-release | grep 'Rocky Linux release 8')" ]] && export OS_VARIANT=rocky-8 && return 0
     [[ "$(cat /etc/os-release | grep 'Red Hat Enterprise Linux Server release 7')" ]] && export OS_VARIANT=redhat-7 && return 0
     [[ "$(cat /etc/os-release | grep 'Red Hat Enterprise Linux release 8')" ]] && export OS_VARIANT=redhat-8 && return 0
     [[ "$(cat /etc/os-release | grep 'Amazon Linux release 2')" ]] && export OS_VARIANT=amazon-linux-2 && return 0
     [[ "$(cat /etc/os-release | grep ^NAME | grep Fedora)" ]] && export OS_VARIANT=fedora && return 0
     [[ "$(cat /etc/os-release | grep ^NAME | grep Ubuntu)" ]] && export OS_VARIANT=ubuntu && return 0
-    [[ "$(cat /etc/os-release | grep ^NAME | grep Debian)" ]] && export OS_VARIANT=debian && return 0 
+    [[ "$(cat /etc/os-release | grep ^NAME | grep -i pop!_os )" ]] && export OS_VARIANT=ubuntu && return 0
+    [[ "$(cat /etc/os-release | grep ^NAME | grep Debian)" ]] && export OS_VARIANT=debian && return 0
     [[ "$(cat /etc/os-release | grep ^NAME | grep Mint)" ]] && export OS_VARIANT=mint && return 0
     echo "Error: Operating system could not be determined or is unsupported, could not configure the OS for firecracker node" && exit 1
 
@@ -48,7 +49,7 @@ install_pre_reqs() {
     case $OS_VARIANT in
         centos-7)
             # Insert OS specific setup steps here
-            sudo yum -v update -y 
+            sudo yum -v update -y
         ;;
         redhat-7)
             # Insert OS specific setup steps here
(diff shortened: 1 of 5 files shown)


component/firecracker-base/orchestrate-install.sh
```diff
@@ -94,34 +95,39 @@ execute_configuration_management() {
...
 
         # Remainder of the binaries
-        wget https://si-tools-prod-ec2-firecracker-config.s3.amazonaws.com/firecracker/latest/rootfs.ext4
-        wget https://si-tools-prod-ec2-firecracker-config.s3.amazonaws.com/firecracker/latest/image-kernel.bin
-        wget https://si-tools-prod-ec2-firecracker-config.s3.amazonaws.com/firecracker/latest/firecracker
-        wget https://si-tools-prod-ec2-firecracker-config.s3.amazonaws.com/firecracker/latest/jailer
+        wget -bcq https://si-tools-prod-ec2-firecracker-config.s3.amazonaws.com/firecracker/latest/rootfs.ext4 -O ./rootfs.ext4
+        wget -bcq https://si-tools-prod-ec2-firecracker-config.s3.amazonaws.com/firecracker/latest/image-kernel.bin -O ./image-kernel.bin
+        wget -bcq https://si-tools-prod-ec2-firecracker-config.s3.amazonaws.com/firecracker/latest/firecracker -O ./firecracker
+        wget -bcq https://si-tools-prod-ec2-firecracker-config.s3.amazonaws.com/firecracker/latest/jailer -O ./jailer
```
johnrwatson, line 114: non-blocking: I think this might always return 0 exit code, which could cause us problems, we'd want it to bomb out:
```
root@Threadripper:~/si# wget -bcq http://dead.nothing.com
Continuing in background, pid 142787.
root@Threadripper:~/si# echo $?
0
```

component/firecracker-base/orchestrate-install.sh
```diff
@@ -133,7 +139,7 @@ execute_configuration_management() {
         cp ./jailer /usr/bin/jailer
 
         # Load kernel module
-        modprobe kvm_intel
+        modprobe kvm_intel || echo "loading AMD instead" && modprobe kvm_amd
```
johnrwatson, line 142: nit: I'd probably prefer we remove the log line as we're logging to stdout/stderr anyway & it'll bomb out

component/firecracker-base/orchestrate-install.sh
```diff
@@ -198,18 +205,34 @@ execute_cleanup() {
...
+    esac
 
     rm -Rf /tmp/firecracker-install/*
-    
+
+}
+
+prepare_jailers() {
+  if test -f "./prepare_jailer.sh"; then
+    ITERATIONS="${1:-100}" # Default to 100 jails
```
johnrwatson, line 216: non-blocking: For no other purpose than consistency: I suggest we make the default jail number the same as what veritech is currently hardcoded to (5000) at the minute.

component/firecracker-base/prepare_jailer.sh
```diff
@@ -0,0 +1,164 @@
...
+  echo "Copying kernel image..."
+  cp -v $KERNEL_IMG "$JAIL/$KERNEL"
+fi
+
+if test -f "$JAIL/$ROOTFS"; then
+  echo "Jailed rootfs exists, skipping creation."
+else
+  echo "Copying rootfs..."
+  cp -v $RO_DRIVE "$JAIL/$ROOTFS"
+  # using device mapper for CoW should be faster. Something like this?
```
johnrwatson, line 62: query: If this works I'd totally suggest we include it, the initial implementation of `cp` doesn't scale at all.

component/firecracker-base/prepare_jailer.sh
```diff
@@ -0,0 +1,164 @@
...
+  ip netns exec $JAILER_NS iptables -t nat -A POSTROUTING -o $VETH_DEV -j MASQUERADE
+  ip netns exec $JAILER_NS iptables -A FORWARD -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT
+  ip netns exec $JAILER_NS iptables -A FORWARD -i $TAP_DEV -o $VETH_DEV -j ACCEPT
+fi
+
+########## ############################# #########
+##########        Firecracker Prep       #########
+########## ############################# #########
+
+cat << EOF > $JAIL/firecracker.conf
```
johnrwatson, line 127: Nice one, this is certainly faster than using separate curl's

component/firecracker-base/orchestrate-install.sh
```diff
@@ -198,18 +205,34 @@ execute_cleanup() {
...
+    echo "Creating $ITERATIONS jails..."
+    for (( iter=0; iter<$ITERATIONS; iter++ ))
+    do
+        ./prepare_jailer.sh $iter &
+    done
+    wait
+  else
+    echo "prepare_jailer.sh script not found, skipping jail creation."
```
johnrwatson, line 224: nit: I think it would be preferable to exit non-zero here so we stop

johnrwatson (approved):
Looks good! Some very minor nits.

I'd definitely suggest the CoW implementation that's commented out if it works, it's way better for filesystem usage than my weak `cp` abuse

component/firecracker-base/orchestrate-install.sh
sprutton1: I don't want to do this until we get the CoW bits implemented. 5000 jails and 1gb images will be a bad time unless the disks are big enough

component/firecracker-base/prepare_jailer.sh
sprutton1: I started to mess with it, but it was tricky. I'd like to push it out to another PR so I can focus on it specifically

component/firecracker-base/orchestrate-install.sh
sprutton1: My counterpoint is that we do cleanup after this step and I imagine we still want that to execute. Ignore the fact that our cleanup doesn't do anything on ubuntu...

component/firecracker-base/orchestrate-install.sh
sprutton1: Ah, fair point. I was hoping this would be an easy way to parallelize.  I'll remove it.

component/firecracker-base/orchestrate-install.sh
sprutton1: modprobe only logs if it fails, so you get a log line saying it failed and then nothing else after. The result of the change here is if it fails to install intel, it will tell you and then move on. If it succeeds, it'll just move on.

component/firecracker-base/orchestrate-install.sh
johnrwatson: Alrighty, can we also drop the num in veritech server too in the meantime?
https://github.com/systeminit/si/blob/main/lib/deadpool-cyclone/src/instance/cyclone/local_uds.rs#L700

component/firecracker-base/orchestrate-install.sh
johnrwatson: I'm not sure, I would see this is a "fatal" as the platform wasn't ready for firecracker'ing. Happy to have it either way, we don't really have a "standard" or a cleanup process at all

component/firecracker-base/prepare_jailer.sh
johnrwatson: Sounds good 👍

PR #2984, opened 2023-11-29
Avoid re-propagating siblings when a new children is connected to a parent

paulocsanz:
And removes the unused create attribute value logic from council/dependent values update.


paulocsanz:
bors merge

PR #2988, opened 2023-11-29
feat/add cyclone rootfs target

johnrwatson:
Adds a new buck2 target, toolchain and macro which builds the rootfs required for firecracker to launch micro-vms.

Currently it only supports Cyclone due to some hardcoded values in the build chain but should be easily otherwise extensible to any of the services we which to run on firecracker or build this way.

The new buck2 target can be invoked via:
`buck2 build bin/cyclone:rootfs` 

The target is dependent on the cyclone `binary` not the docker image, so it should be nice and fast compared to if we added it to the tail end of the docker build.

I've added some placeholder content for Promoting the rootfs artifacts through Buck2 too, which should be helpful in the future for on-main merges and similar.

There is some minor work to include the nix flake content within the rootfs, which can come in a separate PR.

Example run:
```
john@Threadripper:~/working-folder/2-repos/si$ buck2 build bin/cyclone:rootfs --show-simple-output
File changed: prelude-si//rootfs/rootfs_build.sh
File changed: prelude-si//rootfs/z_TODO
File changed: prelude-si//rootfs/create-root-fs.sh
6 additional file change events
Build ID: 81af087a-6e28-4408-91cf-575ea3ee87a2
Network: Up: 0B  Down: 0B
Jobs completed: 47. Time elapsed: 3.7s.
Cache hits: 0%. Commands: 1 (cached: 0, remote: 0, local: 1)
BUILD SUCCEEDED
buck-out/v2/gen/root/524f8da68ea2a374/bin/cyclone/__rootfs__/johns-rootfs.tar
```

It's noteworthy this particular buck target `is not` buildable on mac devices, due to the mkfs.ext4 variant that's required. It's possible to do some VMM shenanigans to make this work however.


prelude-si/rootfs/rootfs_build.sh
```diff
@@ -0,0 +1,121 @@
...
+
+# autostart cyclone
+cat << EOF > ${GUESTDISK}/etc/init.d/cyclone
+#!/sbin/openrc-run
+
+name="cyclone"
+description="Cyclone"
+supervisor="supervise-daemon"
+command="cyclone"
+command_args="--bind-vsock 3:52 --decryption-key /dev.decryption.key --lang-server /usr/local/bin/lang-js --enable-watch --limit-requests 1 --watch-timeout 10 --enable-ping --enable-resolver --enable-action-run"
```
fnichol, line 79: I suspect we'll want to instead write out a config file for Cyclone which may be the result of some kind of boot-time "cloud-init" mechanic? We'll need to provide a production Cyclone key to the server at a minimum and that won't be baked in (or if it must be, then the hosting system might have to download a pristine rootfs and create a new one containing a key, or some overlayfs trick?). In the past the Veritech server's config would dictate how the Cyclone server would boot and we may or may not have this ability anymore.
 
Anyway, this is me thinking aloud and keeping some notes for our challenges :)

prelude-si/rootfs/rootfs_build.sh
```diff
@@ -0,0 +1,121 @@
...
+
+auto eth0
+iface eth0 inet static
+        address 10.0.0.1/30
+        gateway 10.0.0.2
+EOZ
+EOL
+
+# run the script, mounting the disk so we can create a rootfs
+docker run \
```
fnichol, line 103: Perhaps in the future this can be an extracted Alpine Linux FS root and we set it up via a `chroot` call? This eliminates the need for a running Docker engine which will be a Docker-in-Docker setup in CI

prelude-si/rootfs/rootfs_build.sh
```diff
@@ -0,0 +1,121 @@
...
+PACKAGEDIR="$BUCKROOT/cyclone-pkg"
+ROOTFS="$PACKAGEDIR/cyclone-rootfs.ext4"
+ROOTFSMOUNT="$PACKAGEDIR/rootfs"
+GUESTDISK="/rootfs"
+INITSCRIPT="$PACKAGEDIR/init.sh"
+
+# create disk and mount to a known location
+mkdir -p $ROOTFSMOUNT
+dd if=/dev/zero of=$ROOTFS bs=1M count=1024
+mkfs.ext4 $ROOTFS
```
fnichol, line 43: We're eventually going to want the `util-linux` Nix package available for builds (although only on Linux). You're right in your PR comment that we'll likely want to restrict the rootfs rule to Linux-only. I can look into seeing if that's possible with Buck2 configuration.

fnichol:
I'm going to double check our existing tasks with a:

bors try

fnichol:
bors merge

PR #2999, opened 2023-12-01
fix(dal): make auth funcs binding  work on import and export

vbustamante:


vbustamante:
bors merge

vbustamante:
bors merge

PR #3002, opened 2023-12-01
feat: add pool setup to veritech start

sprutton1:
This:
* adds some script cleanup to make them a little less chatty
* attempts to parallelize and add some light retrying to the prep scripts. Setting up 5k at once is really flaky otherwise
* adds `setup()` steps for the deadpool manager so each runtime type can do any necessary preparation steps before attempting to run cyclones (like running our setup scripts)
* adds a `pool_size` config value that can be used to configure the... the pool size. What did you think it would do?

The result here is that when using `LocalFirecracker` veritech will attempt to run the setup script from `/firecracker-data/orchestrate-install.sh`. This script should likely be baked into the build somewhere somehow. The default pool size is currently 5000. It takes a loooooong time to set that many up, but it's very fast once they are set up to run through and verify them.


sprutton1:
bors try

PR #3005, opened 2023-12-02
WIP - deep linking to details tabs, sync URL with selection, diagram overhaul

theoephraim:
First I am making diagram less generic, to help simplify some of the logic of syncing state between diagram and stores... In general this should make it easier and the extra separation of concerns was already breaking down and proving to be more of a pain than a help...

This will help make it much easier to sync the URL with the state of "what is selected" within the store and have the diagram update accordingly.

Why is this all necessary? So we can select specific components and tabs via the URL (which is something we'd want anyway so you can send more specific links) which will make it much simpler to move through the UI from anywhere, rather than having to thread those messages all through many layers of components.

I've also introduced a modeling related event bus, so that it should be much easier to communicate with the diagram, right click menus, modals, etc.

Please note a lot of this is still a bit of a mess... but I think this is a first step towards getting a lot of this code seriously simplified.


theoephraim:
bors merge

PR #3010, opened 2023-12-05
feat: small fixes for the startup scripts

sprutton1:
Just some small enhancements to the startup script. I'm pinning the jails to 1k instead of 5k for now. Creating 5k jails takes way too long, get exponentially slower and flakier as more jails are created. I think this is the best this script can do and that the more correct answer here is to have a discrete service or a background thread in Veritech that recycles jails as needed.


sprutton1:
bors try

PR #3012, opened 2023-12-06
feat(buck2): create `artifact_promote` rule & update `artifact_publish`

fnichol:
`artifact_promote` Buck2 Rule
-----------------------------

This change adds a new Buck2 rule, `artifact_promote` which will create
or update AWS S3 [website redirect location] metadata in empty S3 objects
to serve as HTTP/301 redirects. Additionally, if the
`--cloudfront-distribution-id` option is provided or if the
`AWS_CLOUDFRONT_DISTRIBUTION_ID` environment variable is set, the
promotion will invalidate the appropriate paths in AWS CloudFront to
ensure that the redirects are up to date.

[website redirect location]: https://docs.aws.amazon.com/AmazonS3/latest/userguide/how-to-page-redirect.html#redirect-requests-object-metadata

Updated S3 Path Strategy
------------------------

This change also updates the S3 object locations to importantly include
a version path entry. This allows us to group many different artifact
variants by a release version which may be helpful when
yanking/purging/aging out older releases. Additionally the version path
slot is used during promotion with the channel value.

Here's a more concrete example. A 64-bit ARM release of the Cylone
omnibus package for Linux would be found at:

```
/cyclone/20231206.001546.0-sha.e6ba9be23/omnibus/linux/aarch64/cyclone-20231206.001546.0-sha.e6ba9be23-omnibus-linux-aarch64.tar.gz
```

This scheme generalizes to:

```
/$family/$version/$variant/$os/$arch/$family-$version-$variant-$os-$arch.$extension
```

When being promoted to a "release channel" (in our case we use the
"stable" channel name by default), the above artifact would be found as
the current version in the "stable" channel at:

```
/cyclone/stable/omnibus/linux/aarch64/cyclone-stable-omnibus-linux-aarch64.tar.gz
```

Similar to above, this scheme generalizes to:

```
/$family/$channel/$variant/$os/$arch/$family-$version-$variant-$os-$arch.$extension
```

CloudFront Invalidation Pathing
-------------------------------

With the updated S3 path strategy, we can compute simplified [CloudFront
invalidations] that are wildcarded after the variant path segment. In
the case of the example above, and even assuming that multiple
`$os-$arch` versions are included in the promotion, only 1 CloudFront
path entry will need invalidation:

```
/cyclone/stable/omnibus/*
```

[CloudFront invalidations]: https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/Invalidation.html

### DIFF
--- a/bin/cyclone/BUCK
+++ b/bin/cyclone/BUCK
@@ -78,24 +78,3 @@ rootfs(
         "//bin/lang-js:omnibus",
     ],
 )
-
-load(
-    "@prelude-si//:artifact.bzl",
-    "artifact_promote",
-    "artifact_publish",
-)
-
-artifact_publish(
-    name = "publish-omnibus",
-    artifact = ":omnibus",
-    destination = "s3://si-artifacts-prod",
-    cname = "artifacts.systeminit.com",
-)
-
-artifact_promote(
-    name = "promote-omnibus",
-    family = "cyclone",
-    variant = "omnibus",
-    destination = "s3://si-artifacts-prod",
-    cname = "artifacts.systeminit.com",
-)
--- a/bin/cyclone/BUCK
+++ b/bin/cyclone/BUCK
@@ -1,9 +1,9 @@
 load(
     "@prelude-si//:macros.bzl",
     "docker_image",
-    "rootfs",
     "export_file",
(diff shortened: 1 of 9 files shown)


fnichol:
bors merge

PR #3013, opened 2023-12-06
feat: terminate on recycle, startup fixes

sprutton1:
This adds some robustness to our startup script so that we can dynamically choose whether to do things like download new rootfs or kernel images. It also adds force recreating jails if necessary (such as if you download a new rootfs). I added some flags to make this easier to use on the command line and a help block.

I also made it so we actually call `terminate()` when when recycling the pool. In the case of Firecracker, we will clean the jail and set it up again. I think a better way to do this would be just clean replace the rootfs and remove the jailer bits, but this is fine for now.


sprutton1:
bors try

johnrwatson (approved):
Looks good to me, variable parsing is nice! Just needs a lint I think (CI checks)

PR #3014, opened 2023-12-06
feat(dal,web) Make secrets work inside the dependency reactivity system

vbustamante:
image

By setting the toggle above, `root/secrets` will be set as input to the function. At runtime, the system will detect the dependency on secrets, wipe that from the args of the main function to be executed and trigger fetching the before functions.

As a corollary, auth functions now  will only run when explicitly set by the user

### DIFF
--- a/lib/dal/src/schema/variant/leaves.rs
+++ b/lib/dal/src/schema/variant/leaves.rs
@@ -68,6 +68,8 @@ pub enum LeafInputLocation {
     Domain,
     /// The input location corresponding to "/root/resource".
     Resource,
+    /// The input location corresponding to "/root/secrets".
+    Secrets,
 }
 
 // We only want to allow converting an input location into a root prop child and root the other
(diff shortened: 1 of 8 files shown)


vbustamante:
bros try

vbustamante:
bors try

lib/dal/src/schema/variant/leaves.rs
```diff
@@ -80,6 +82,7 @@ impl Into<RootPropChild> for LeafInputLocation {
             LeafInputLocation::Domain => RootPropChild::Domain,
             LeafInputLocation::Resource => RootPropChild::Resource,
             LeafInputLocation::DeletedAt => RootPropChild::DeletedAt,
+            LeafInputLocation::Secrets => RootPropChild::Secrets,
```
fnichol, line 85: Ah, this is how, nice! I was so wrong that the solution would likely need SQL query changes.

fnichol (approved):
Nicely done, and I know how *not* easy this was to figure out!

vbustamante:
bors merge

vbustamante:
bors merge

PR #3015, opened 2023-12-06
feat: add artifact-status webpage

johnrwatson:
Adds a simple webpage which is currently hosted at the root of the bucket behind https://artifacts.systeminit.com which gives a traversable view of the content of our production CDN.

We can extend this pretty easily in the future to also cover our docker registries, or other repositories as our footprint grows.

NB: Simple client-side implementation that leaves `a lot` of room for improvement.


johnrwatson:
Merging as is, as it's the current representation of live, not because it's totally correct.

Likely need some work here to make it closer to what we want on brand.

johnrwatson:
bors merge

PR #3019, opened 2023-12-06
Create build and distribution uniformity across binaries/packages

johnrwatson:
Achieves parity for build and distribution on our consolidated Artifacts CDN across our set of services by adding these variants:
- council: `[omnibus]`
- cyclone: `[rootfs]`
- lang-js: `[omnibus]`
- module-index: `[omnibus]`
- pinga: `[omnibus]`
- sdf: `[omnibus]`
- si (cli): `[omnibus]`

There are three main reasons why we want to do this:
- This grants us a lot of options in terms of how we run SI in Production, including how we manage updates across the stack.
- It gives open-source/external users plenty of options for how they wish to consume SI, either in parts/full stack and using the associated nix dependencies or not.
- It allows us to upgrade, rollback or patch a specific part of the stack from any git branch from a centralised place.


johnrwatson:
bors merge

PR #3020, opened 2023-12-06
WIP: do not process duplicates for category nodes

nickgerace:
- Do not process duplicate "to rebase indices" for category nodes
  - We use a HashSet instead of a Vec to accomplish this
- Move all tests to a submodule for graph
- Add test to simulate rebase
- Don't take in a mutable "onto" graph (leftover from when "perform_updates" was in the rebaser and it needed to be mutable for the snapshot to have a working copy)


lib/dal/src/workspace_snapshot/graph.rs
```diff
@@ -546,16 +547,16 @@ impl WorkspaceSnapshotGraph {
                     event
                 })?;
 
-                let mut to_rebase_node_indexes = Vec::new();
-                if onto_node_index == onto.root() {
+                let mut to_rebase_node_indexes = HashSet::new();
```
nickgerace, line 550: This is the primary change

lib/dal/src/workspace_snapshot/graph.rs
```diff
@@ -1744,7 +1746,7 @@ impl WorkspaceSnapshotGraph {
     pub fn perform_updates(
         &mut self,
         to_rebase_change_set: &ChangeSetPointer,
-        onto: &mut WorkspaceSnapshotGraph,
+        onto: &WorkspaceSnapshotGraph,
```
nickgerace, line 1749: No longer needs to be mutable. This was leftover from when it was in the rebaser. In there, the snapshot needed to be mutable to access a working copy.

PR #3026, opened 2023-12-07
feat(buck2): add `support/buck2:sync-cargo-deps` target

fnichol:
This change introduces 2 new targets to our system to help with synchronizing Rust crate depdendencies from the root Cargo.toml into Buck2's `third-party/rust/Cargo.toml` which is used by the Reindeer tool.

Now when adding or altering the root `Cargo.toml`, you can run:

```sh
buck2 run support/buck2:sync-cargo-deps
```

This will copy over all the `[workspace.depdencies]` from the Cargo workspace over to the `[workspaces]` section in
`third-party/rust/Cargo.toml` as well as the `[path.crates-io]` sections for temporary forks. Next, it will run `reindeer --third-party-dir
third-party/rust buckify` to update both the `third-party/rust/Cargo.lock` and `third-party/rust/BUCK` files. The
program will report if any files have changed and advise the user to commit these changes to a Git commit.

Additionally, there is a check version of this target which will passively check whether or not the sync task *should* be run. You can invoke this with:

```sh
buck2 run support/buck2:check-cargo-deps
```

Future work may tie the `support/buck2:check-cargo-deps` target into CI as a check when reviewing pull requests, making it easier to detect if the Cargo and Buck2 worlds are out of sync.

To recap, a reasonable way to approach adding a new Rust into the system if it isn't already present in the root `Cargo.toml` is to add it (please keep the dependencies in sorted lexical order to help with tooling and diffs) and run `buck2 run support/buck2:sync-cargo-deps` to create the Buck2 crate dependency targets.


fnichol:
bors try

fnichol:
bors merge

fnichol:
bors merge

PR #3028, opened 2023-12-07
WIP: obliterate compiler and clippy warnings

nickgerace:
- Obliterate compiler and clippy warnings
- Allow dead code where it will likely be reused in the near future
  - This could backfire a bit, but it should be fine


nickgerace:
Cherry-picked

PR #3030, opened 2023-12-07
feat: veritech flags to configure runtime

sprutton1:
This adds flags to veritech to configure the runtime type and the size of the pool. These flags are passed down into cyclone-server/client to configure the respective settings.  As an example, to tell Veritech to use Firecracker:
```bash
$ buck2 run bin/veritech -- --cyclone-local-firecracker --cyclone-pool-size 500
```


sprutton1:
bors try

PR #3033, opened 2023-12-07
feat: add veritech service file

sprutton1:
Adds the systemd service file for veritech. Ideally we ship this as an artifact along with veritech itself so we can version it with the binary. We can then just grab both in user data and `systemctl daemon-reload && systemctl enable veritech --now`


sprutton1:
bors try

johnrwatson (approved):

PR #3040, opened 2023-12-09
more modeling fixes

theoephraim:
- show loader instead of empty model 
- middle click and drag to pan
- pan to element on selection via outline
- center on element for outline double click
- clean up navbar!
- fix popup dismissal logic for apply change set modal
- fix right click behaviour in outline
- clean up style on confirm delete change set modal
- add "+ New" back into change set selection dropdown


theoephraim:
git merge

theoephraim:
bors merge

PR #3047, opened 2023-12-11
fix: startup script error handling

sprutton1:
This removes startup parallelization to avoid some locking issues we found. It also appropriately handles file copy errors.


sprutton1:
bors try

PR #3048, opened 2023-12-12
Implement nested frames

paulocsanz:
Notes:

- arity many sockets will not cause nested frames to disconnect the parents, they will accumulate the data
- node creation/dropping into a frame may be a bit slow, I need us to test it more to see how slow it is, it's easy to optimize but I prefer merging this first to then optimize based on a working state (and only if/when needed)
- sometimes you want to drop a component in a frame but the component is rendered behind the frame so you are unable

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4406,6 +4406,7 @@ dependencies = [
 name = "sdf-server"
 version = "0.1.0"
 dependencies = [
+ "async-recursion",
  "async-trait",
  "axum",
  "base64 0.21.2",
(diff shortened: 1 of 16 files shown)


keeb:
This looks really good and is WORKING!! with my custom assets 

image

Some issues:
- Deleting nested assets is not working (cannot remove accidentally created nested host asset)
- Moving things out of nested frame does not work (detach) vs. attach

Feedback:

 - Sometimes I am going to want my asset to not be a frame, I cannot switch types but instead I'd need to duplicate types. This isn't a show stopper but it is something to notice.

stack72:
bors merge

PR #3056, opened 2023-12-13
feat: add cred string and file options for NATs auth.

sprutton1:
This adds the ability to authenticate with nats using either a credential file or a credential string. These credentials must follow the format defined [here](https://docs.nats.io/using-nats/developer/connecting/creds). 

This is only required if the NATs server is using authentication, such as if you are using [NGS](https://app.ngs.global/).


sprutton1:
bors try

sprutton1:
bors try

PR #3058, opened 2023-12-14
feat(nats): add connection name to nats connection

sprutton1:
This allows giving a connection name to nats. This makes it easier to identify who is connecting to nats. We're considering using a managed nats solution and without this the UI shows all of the connections as `unknown` . Here's a before and after:
image
image


sprutton1:
bors try

PR #3062, opened 2023-12-14
Fix/firecracker rootfs

johnrwatson:
Minor fixes to the rootfs, including pulling the rootfs from our artifact store.

The issue with the rootfs is the way we were mounting/extracting the omnibus package into the disk, it was `removing` the actual filesystem from the contents so that it was totally broken. I think this happened as some kind of regression during development/testing but it's resolved via this PR.

Also changes the default jail number and jail-creation-parallelisation to 100, as we were hitting throttling/race conditions. This stabilises it until we come up with something better.


johnrwatson:
bors merge

PR #3065, opened 2023-12-15
fix: rootfs script fixes to ensure it is bootable

sprutton1:
This script was not creating a bootable rootfs. To fix this:
* At some point we moved some of the cyclone artifacts to `/run`. A tempfs gets mounted at `/run` at boot, so all files there were being masked. I moved all of these to `/cyclone`.
```bash
c8b21330e556:~# df
Filesystem           1K-blocks      Used Available Use% Mounted on
/dev/root              1992552    867572   1003740  46% /
devtmpfs                 10240         0     10240   0% /dev
**tmpfs                    98748        44     98704   0% /run**
shm                     246864         0    246864   0% /dev/shm
```
* I moved back the gathering of the binary inputs. This didn't need to happen later in the script like we thought before.
* Added back mingetty and the root shell so we could actually drop into the VM to figure out what is going wrong when it breaks. This may need to be thought out a bit more, but should be fine for now. The only way to abuse this would require root access on the Veritech box, and we're basically sunk at that point anyhow.
* Changed the copying of the rootfs we create. I now unmount the drive and then move it where we expect it. This make no sense to me whatsoever, but I was able to see several occasions where the copied rootfs was missing bytes and completely unusable. Unable to recreate this after making this change.

### DIFF
--- a/prelude-si/rootfs/rootfs_build.sh
+++ b/prelude-si/rootfs/rootfs_build.sh
@@ -84,7 +84,7 @@ setup_traps cleanup
 
 cyclone_args=(
   --bind-vsock 3:52
-  --decryption-key /run/cyclone/decryption.key
+  --decryption-key /cyclone/decryption.key
   --lang-server /usr/local/bin/lang-js
   --enable-watch
   --limit-requests 1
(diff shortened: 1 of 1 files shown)


sprutton1:
bors try

sprutton1:
bors try

PR #3067, opened 2023-12-15
feat: refactor rootfs to not use docker

sprutton1:
This changes the rootfs build from using the alpine docker image to downloading the rootfs directly and `chroot`ing it into the right shape. The docker path works fine locally, but fails in CI due running this as docker in docker, which makes bind mounting the scripts and directories fraught with peril as it tries to mount from the **host**, not the local container. This should make this process more reliable in general and a little faster. @fnichol I think you made changes to make docker in docker work correctly. I'm not sure if we should go back and remove those, I'm not sure exactly where they live.

### DIFF
--- a/prelude-si/rootfs/rootfs_build.sh
+++ b/prelude-si/rootfs/rootfs_build.sh
@@ -4,7 +4,7 @@
 # components we need rootfs' for but there are some cyclone-specifics bits that
 # will cause us problems
 
-set -euo pipefail
+set -euxo pipefail
 
 # TODO(johnrwatson): We need to port this to python or similar, and check for
 # OS-dependencies that are required. i.e. docker and basic priviledged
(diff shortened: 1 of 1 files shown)


sprutton1:
bors try

sprutton1:
bors try

sprutton1:
bors try

sprutton1:
bors try

sprutton1:
bors try

sprutton1:
bors try

sprutton1:
bors try

sprutton1:
bors try

sprutton1:
bors try

sprutton1:
bors try

sprutton1:
bors try

sprutton1:
bors try

sprutton1:
bors try

sprutton1:
bors try

sprutton1:
bors try

PR #3070, opened 2023-12-15
feat(sdf): use schema names for model name defaults

adamhjk:
Currently, every asset is named in a `si-1234` style. With this change, they now default to `Schema Name 1234` style, which is much nicer on the old eyeballs.

image


paulocsanz (approved):

adamhjk:
bors merge

PR #3079, opened 2023-12-18
feat(web) - various UI improvements

wendybujalski:
- disable the change set abandon button while on HEAD
- prevent users from naming a change set "head" (on the front end only)
- add map item error message if you try to add an item without a key
- AssetActionDetails UI more closely matches the rest of the UI
- hover states for clickable diff icons on the diagram and ComponentCard
- "No Code" message in AssetDiffDetails shows outside of a CodeViewer


wendybujalski:
bors merge

PR #3080, opened 2023-12-19
updating vue/vite deps

theoephraim:
encountering a vite issue lately that requires clearing the cache to unstick your local env...

No idea if this will fix it as we dont have clear reproduction steps, but this should hopefully help.


app/auth-portal/src/App.vue
```diff
@@ -103,7 +103,7 @@
               variant="transparent"
               tone="shade"
               @mousedown.prevent
-              @click.prevent="profileMenuRef?.open"
+              @click.prevent="profileMenuRef?.open || _.noop"
```
theoephraim, line 106: shouldnt really need this but it was causing a vue internals bug... I expect it to be fixed pretty quickly.

https://github.com/vuejs/core/issues/9865

theoephraim:
bors merge

theoephraim:
bors merge

app/auth-portal/src/App.vue
theoephraim: looks like the issue itself is resolved but we'll need to wait for a few dependencies to update to the latest vue (not just us)

theoephraim:
bors merge

PR #3084, opened 2023-12-19
Optimize components paste

paulocsanz:
- Parallelize component pasting (without edges)
- Filter existing attribute values/prototypes in rust (check context in rust)
- Commit more frequently to give the user the feeling of progress


paulocsanz:
bors merge

PR #3087, opened 2023-12-20
fix: using alternative delimiter permits http-type forms

johnrwatson:
Using a `/` delimiter in the replace was a poor choice, as something like `https://app.systeminit.com/api` as the SI__SDF__HOST variable escapes the sed and throws a non-zero exit code, failing the replace.

Using `^` should be better as we'll need want to pass various strings in there over time


johnrwatson:
bors merge

PR #3088, opened 2023-12-20
fix: pass the encryption key path for the launcher

sprutton1:
I changed how the defaults work so I could pass this as a base64 encrypted string. This adds the path to the containers directly.

Verified on both linux and macos.


sprutton1:
bors try

PR #3090, opened 2023-12-20
feat: allow passing the secret as a base64 string

sprutton1:
This lets us pass the cbor encoded key that we use to encrypt secrets as a base64 string so we can host in ECS. This was a bit more surgery than the last one, but we got there in the end.


sprutton1:
bors try

PR #3091, opened 2023-12-20
debounce ChangeSetWritten

theoephraim:
not a good longterm solution, as this actually introduces delay between receiving events and things happening... BUT this is a good fix for now until we fire off smarter events instead of refreshing everything constantly


theoephraim:
bors merge

PR #3092, opened 2023-12-20
feat: allow passing the jwt as a base64 string

sprutton1:
This adds the ability to pass in the jwt to SDF as a base64 string. You know the drill!


sprutton1:
bors try

sprutton1:
bors try

PR #3104, opened 2023-12-22
Theo/bye bye 2023

theoephraim:
- dont fetch proposed actions on head
- auto open new changeset if adding action on head
- fix logic around change set auto selection


theoephraim:
bors merge

PR #3107, opened 2023-12-23
feat: add ssl to postgres

sprutton1:
Shhhhhhhhh, this PR didn't happen today. I definitely posted this yesterday. 

Anyways, this adds SSL to postgres connections. I also added a root cert to the docker image so we can use SSL locally and in the launcher. I'm really not sure how to test this for the module-index and I have a feeling all of the tests are going to fail unless we stage this in multiple parts. Merging this should wait until next year.

P.S. I need to do some more work ensuring the launcher still works.


paulocsanz:
bors try

sprutton1:
bors try

sprutton1:
bors try

sprutton1:
bors try

PR #3111, opened 2024-01-03
Create and display edges in the UI with the new engine (ENG-2203)

nickgerace:


README_NEW_ENGINE.md
nickgerace, line 1: May not be useful in the long run, but wanted to write down my CI-less routine somewhere

app/web/src/components/ModelingDiagram/DiagramGroup.vue
```diff
@@ -427,8 +427,8 @@ const isAdded = computed(() => props.group.def.changeStatus === "added");
 
 const childCount = computed(() => {
   const mappedChildren = _.map(
-    props.group.def.childNodeIds,
-    (child) => useComponentsStore().componentsByNodeId[child],
+    props.group.def.childComponentIds,
+    (child) => useComponentsStore().componentsById[child],
   );
```
nickgerace, line 432: I'll refrain from leaving many comments like this, but the `mappedChildren` abstraction is an example of many that are no longer needed because `Node <--> Component` no longer exists. We will fix abstractions like these as we begin to use them.

jhelwig (approved):

lib/dal/src/diagram/edge.rs
```diff
@@ -0,0 +1,138 @@
...
+    pub from_external_provider_id: String,
+    pub to_component_id: String,
+    pub to_explicit_internal_provider_id: String,
+    // pub change_status: ChangeStatus,
+    // pub created_info: Option<HistoryEventMetadata>,
+    // pub deleted_info: Option<HistoryEventMetadata>,
+}
+
+impl DiagramEdgeView {
+    pub async fn list(ctx: &DalContext) -> DiagramResult<Vec<Self>> {
```
nickgerace, line 32: First pass... can likely be improved

PR #3112, opened 2024-01-03
Avoid blocking frontend when actions are running

paulocsanz:
This may trigger unexpected behavior, but it's not something we want in the long run and it's probably not needed anymore.

The only missing piece to decide is, should we block running the same action on the same component if one is running? What if the action source code changed, or if the user changed a property in between?

But if 2 create actions for the same component run in parallel they may lose the resource from the first one to finish, since it will be overwritten.


paulocsanz:
bors merge

PR #3118, opened 2024-01-04
window/keyboard event handling priority

theoephraim:
implement event priority-aware event manager so we can still use window-bound key events, but handle them in the right order. Previously the modeling diagram component was catching all key events, so hitting escape was actually deselecting the current selection - which triggered modals/popovers to close. Now the popover/modal key handlers have a higher priority, so we can close the modal, and stop the event from triggering deselection.

Bonus - ignore mouse events on blue drag-to-select box, so that quick mouse movements dont mess with hover states

### DIFF
--- a/app/web/src/components/AssetPalette.vue
+++ b/app/web/src/components/AssetPalette.vue
@@ -57,7 +57,7 @@
                       : '',
                   )
                 "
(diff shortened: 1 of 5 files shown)


theoephraim:
bors merge

PR #3119, opened 2024-01-04
feat: serialize jail usage

sprutton1:
This does two things:

1. We now manage a small file that is used to keep track of jail ids. We want to never reuse jails, the overhead of repreparing them is too much and the risk of failure is too high. We cannot rely on random id generation as the jails must exist on disk at a known path. This will (mostly) ensure that we use the next available jail and that jails are not reused. There are really big contention problems with this approach, but it should be enough to get us by until we can build a proper pool management suite.
2. This change means that our stop/start processes for jails is down to only interacting with the jailer process itself. Therefore, we can remove the `{start,stop}.sh` scripts from the path and manage the jailer process directly in rust. Hurray!

### DIFF
--- a/bin/veritech/scripts/stop.sh
+++ b/bin/veritech/scripts/stop.sh
@@ -4,40 +4,5 @@ set -euo pipefail
 
 SB_ID="${1:-null}"
 
-JAILER_BINARY="/usr/bin/jailer"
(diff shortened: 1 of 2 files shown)


sprutton1:
bors try

PR #3120, opened 2024-01-04
WorkspaceSnapshoGraph cleanup speedup

jhelwig:
* Only prune & write out new workspace snapshot graph if there have been changes in rebaser
* Speed up garbage collection of un-referenced nodes in WorkspaceSnapshotGraph

    We want to remove all of the "garbage" we've accumulated while operating on the graph.
    Anything that is no longer reachable from the current `self.root_index` should be
    removed as it is no longer referenced by anything in the current version of the graph.

    Fortunately, we don't need to walk the graph to find out if something is reachable from
    the root, since `has_path_connecting` is slow (depth-first search). Any node that does
    *NOT* have any incoming edges (aside from the `self.root_index` node) is not reachable,
    by definition. Finding the list of nodes with no incoming edges is very fast.

    If we remove all nodes (that are not the `self.root_index` node) that do
    not have any incoming edges, and we keep doing this until the only one
    left is the `self.root_index` node, then all remaining nodes are
    reachable from `self.root_index`.

### DIFF
--- a/lib/dal/src/workspace_snapshot/graph.rs
+++ b/lib/dal/src/workspace_snapshot/graph.rs
@@ -490,18 +490,53 @@ impl WorkspaceSnapshotGraph {
     }
 
     pub fn cleanup(&mut self) {
-        self.graph.retain_nodes(|frozen_graph, current_node| {
-            // We cannot use "has_path_to_root" because we need to use the Frozen<StableGraph<...>>.
-            algo::has_path_connecting(&*frozen_graph, self.root_index, current_node, None)
-        });
+        let start = tokio::time::Instant::now();
+
+        // We want to remove all of the "garbage" we've accumulated while operating on the graph.
+        // Anything that is no longer reachable from the current `self.root_index` should be
+        // removed as it is no longer referenced by anything in the current version of the graph.
+        // Fortunately, we don't need to walk the graph to find out if something is reachable from
+        // the root, since `has_path_connecting` is slow (depth-first search). Any node that does
+        // *NOT* have any incoming edges (aside from the `self.root_index` node) is not reachable,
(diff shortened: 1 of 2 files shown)


lib/dal/src/workspace_snapshot/graph.rs
```diff
@@ -490,18 +490,53 @@ impl WorkspaceSnapshotGraph {
...
+        // *NOT* have any incoming edges (aside from the `self.root_index` node) is not reachable,
+        // by definition. Finding the list of nodes with no incoming edges is very fast. If we
+        // remove all nodes (that are not the `self.root_index` node) that do not have any
+        // incoming edges, and we keep doing this until the only one left is the `self.root_index`
+        // node, then all remaining nodes are reachable from `self.root_index`.
+        let mut old_root_ids: HashSet<NodeIndex>;
+        loop {
+            old_root_ids = self
+                .graph
+                .externals(Incoming)
```
nickgerace, line 509: Ooooo [this](https://docs.rs/petgraph/latest/petgraph/graph/struct.Externals.html) is nice!

lib/rebaser-server/src/server/change_set_loop.rs
```diff
@@ -170,15 +170,20 @@ async fn process_delivery(
...
+            // and update the pointer.
+            to_rebase_workspace_snapshot
+                .write(ctx, to_rebase_change_set.vector_clock_id())
+                .await?;
+            info!("Snapshot written: {:?}", start.elapsed());
+            to_rebase_change_set
+                .update_pointer(ctx, to_rebase_workspace_snapshot.id())
+                .await?;
+            info!("Pointer updated: {:?}", start.elapsed());
+        }
```
nickgerace, line 186: Smart

nickgerace (approved):

PR #3122, opened 2024-01-04
fix: revert changes to stop script

sprutton1:
We no longer use this script when stopping VMs, but we do use it when cleaning up all of the jails on veritech start. Just reverting this back to the previous incarnation.

### DIFF
--- a/bin/veritech/scripts/stop.sh
+++ b/bin/veritech/scripts/stop.sh
@@ -4,5 +4,40 @@ set -euo pipefail
 
 SB_ID="${1:-null}"
 
+JAILER_BINARY="/usr/bin/jailer"
+TAP_DEV="fc-${SB_ID}-tap0"
+
 # Kill the firecracker process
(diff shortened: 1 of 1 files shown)


bin/veritech/scripts/stop.sh
```diff
@@ -4,5 +4,40 @@ set -euo pipefail
...
+
+# We are not currently creating these.
+# umount /srv/jailer/firecracker/$SB_ID/root/image-kernel.bin || true
+# dmsetup remove kernel-overlay-$SB_ID || true
+
+# TODO(scott): figure out a better way to do this.
+# this will only detach devices removed from device-mapper
+# but it still feels bad
+losetup --detach-all
+rm -rf /srv/jailer/firecracker/$SB_ID
```
nickgerace, line 43: Be one with the jank

nickgerace (approved):

bin/veritech/scripts/stop.sh
nickgerace: In other words, it looks great to me 😄. I don't think we have other devices on these systems either, right?

PR #3123, opened 2024-01-04
Fix RabbitMQ infinite hangs when it isn't ready (ENG-2214)

nickgerace:
Fix RabbitMQ infinite hangs by using a timeout-and-retry technique. When RabbitMQ is booting up and a user tries to create an "Environment" using the streams port, there can be an infinite hang. That is an issue in its own right, but this change seeks to avoid it entirely.

On macOS aarch64 with an M1 Max, it appears that RabbitMQ needs about ~3-4 seconds for the streams feature to become available. It is unclear if other RabbitMQ services need the same amount of time.

### DIFF
--- a/lib/si-rabbitmq/src/environment.rs
+++ b/lib/si-rabbitmq/src/environment.rs
@@ -1,10 +1,14 @@
 use rabbitmq_stream_client::error::{StreamCreateError, StreamDeleteError};
 use rabbitmq_stream_client::types::{ByteCapacity, ResponseCode};
 use rabbitmq_stream_client::Environment as UpstreamEnvironment;
+use std::time::Duration;
+use telemetry::prelude::{info, trace, warn};
 
-use crate::{config::Config, error::RabbitResult};
+use crate::{config::Config, error::RabbitResult, RabbitError};
 
 const STREAM_LENGTH_CAPACTIY_IN_MEGABYTES: u64 = 10;
+const ENVIRONMENT_CREATION_TIMEOUT_DURATION_IN_SECONDS: u64 = 4;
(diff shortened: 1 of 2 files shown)


lib/si-rabbitmq/src/environment.rs
```diff
@@ -1,10 +1,14 @@
...
 use rabbitmq_stream_client::types::{ByteCapacity, ResponseCode};
 use rabbitmq_stream_client::Environment as UpstreamEnvironment;
+use std::time::Duration;
+use telemetry::prelude::{info, trace, warn};
 
-use crate::{config::Config, error::RabbitResult};
+use crate::{config::Config, error::RabbitResult, RabbitError};
 
 const STREAM_LENGTH_CAPACTIY_IN_MEGABYTES: u64 = 10;
+const ENVIRONMENT_CREATION_TIMEOUT_DURATION_IN_SECONDS: u64 = 4;
```
nickgerace, line 10: It takes ~3-4 for the "Streams" service/feature to be ready. It's unclear if this is platform specific (i.e. macOS aarch64, resources specific (i.e. too few cores and RAM allocated), and/or feature specific (i.e. solely "Streams" or other RabbitMQ features too).

lib/si-rabbitmq/src/environment.rs
```diff
@@ -1,10 +1,14 @@
...
+const ENVIRONMENT_CREATION_RETRIES: u64 = 3;
```
nickgerace, line 11: If it's not ready in 12 seconds (3 retries x 4 second timeouts), then it is unlikely to be ready at all.

PR #3126, opened 2024-01-04
show loader until new components appear when pasting

theoephraim:
uses same strategy that insert currently uses, which is very imperfect, but works ok until we rearchitect the realtime stuff to just push new components right away. This means

- on paste, we optimistically add show a loader at a certain position tracked under a temporary id
- when paste api request returns, we attach a new component id to the temporary loader (in this case we just use the first one if there are multiple components pasted)
- whenever we fetch the diagram, we look through any temporary loaders that are showing and if we find the new component id in the data fetched, we hide the loader


theoephraim:
bors merge

PR #3127, opened 2024-01-05
option + click/drag to select within frames

theoephraim:
if holding option when clicking and dragging, it will force drag-to-select (ie the blue box) mode. This will also have special behaviour which excludes the item that was clicked on and its parents. This allows you to drag to select within frames.


theoephraim:
bors merge

PR #3128, opened 2024-01-05
fix(web): Deal with unset Joi validations correctly

vbustamante:
Also display array and map children better on the error log 
image


vbustamante:
bors merge

PR #3134, opened 2024-01-09
build(deps): patch `tokio-postgres-rustls` to use ring 0.17

fnichol:
This change bumps all Rust dependencies to use the ring 0.17.x release stream, leaving one version of ring to build, and benefits from the pin to ring 0.17.5 which the Buck2/Reindeer fixup supports.

References: jbg/tokio-postgres-rustls#17

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -30,9 +30,9 @@ dependencies = [
 
 [[package]]
 name = "ahash"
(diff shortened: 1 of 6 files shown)


fnichol:
bors merge

fnichol:
bors retry

PR #3135, opened 2024-01-09
feat: close the firecracker VM when cyclone exits

sprutton1:
Ok, here me out. This reverts us back to the old behavior of never calling the `terminate()` methods and instead relying on cyclone exiting by itself. I figured out I can replicate this in Firecracker by simply following the cyclone call with `&& reboot`, which will kill the firecracker VM. This whole thing feels weird to me as we define terminate methods that are never used, but I think this method may be better as it's currently possible that we'll call terminate on a running VM because the healthcheck will return unhealthy if it's already (or currently) running one. I think this may be leading to some strange performance issues that are difficult to diagnose, but I'm not entirely sure.

@fnichol at some point I'd love to talk to you about this and how deadpool works. I'm not sure I entirely understand the recycle semantics and what you were planning here.

### DIFF
--- a/lib/deadpool-cyclone/src/lib.rs
+++ b/lib/deadpool-cyclone/src/lib.rs
@@ -106,13 +106,7 @@ where
         obj: &mut Self::Type,
         _: &Metrics,
     ) -> managed::RecycleResult<Self::Error> {
-        match obj.ensure_healthy().await {
-            Ok(_) => Ok(()),
-            Err(err) => {
-                obj.terminate().await?;
-                Result::map_err(Err(err), Into::into)
-            }
-        }
+        obj.ensure_healthy().await.map_err(Into::into)
     }
 }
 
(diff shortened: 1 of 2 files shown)


fnichol:
Yeah, I forgot there was a terminate behavior. Looking back this is mostly on me: I was working on a trait to manage an instance (of Cyclone server) but hadn't come upon this idea that the Cyclone server would self terminate. I guess I thought we'd need a way to explicitly kill/terminate/shutdown instances but instead we're using deadpool's idea of "health" to determine if we spawn a new instance.

In essence, after one request a Cyclone server is not longer healthy (that is, the client loses it's watch on the server because it's self-terminated).

I'll have to re-read this code again, but we may be able to eliminate the `terminate` behavior if it isn't being invoked. It's also possible that what we need is something like an instance "cleanup" which could terminate any still-running processes/containers, clean up left-over state, PID files, etc.

sprutton1:
> It's also possible that what we need is something like an instance "cleanup" which could terminate any still-running processes/containers, clean up left-over state, PID files, etc.

I'm not sure if this fits in with the `pool-noodle` concept, but more and more it seems like we need something a little more robust than deadpool-managed or that we need something ensuring our runtime environment is always in a good state.

fnichol:
bors retry

PR #3146, opened 2024-01-10
feat: added otelcol to cyclone within firecracker

johnrwatson:
Adds a static route from cyclone into 10.0.0.3 which is then picked up via a NAT iptables route to push it into the otelcollector running on the veritech host on 4317.

This includes:
- Amendments to the rootfs for cyclone
- Adds docker to the veritech node
- Pulls & runs our otelcol container with the relevant honeycomb configuration


sprutton1 (approved):

johnrwatson:
bors merge

PR #3147, opened 2024-01-10
Remove debouncer from code editor to avoid data loss

paulocsanz:
Also saves function before running test to ensure its never stale

This is a proper fix for the bug Salmon experienced and reported on discord.


paulocsanz:
bors merge

paulocsanz:
bors merge

PR #3149, opened 2024-01-10
fix: amend ip for otel within cylcone so it doesn't clash with any internal ips

johnrwatson:
Amends the ip allocated from the context of the cyclone jail to permit communication into otel on the host network. `1.0.0.1` should never clash with any internal IPs.


johnrwatson:
bors merge

PR #3152, opened 2024-01-11
fix: added missing scheme to otel endpoint for cyclone in firecracker

johnrwatson:
The scheme was missing for the otel ingestion for cyclone in firecracker, the remainder of the process/routes have been proven out.

I've tested this change locally using firecracker and it works! Horray!

It just needs merged into main and then revalidated on a new veritech node.


sprutton1 (approved):

johnrwatson:
bors merge

PR #3153, opened 2024-01-11
feat: summary table for qualifications

adamhjk:
The get_summary endpoint was taking 30+ seconds to respond in production. This commit introduces the idea of 'summary tables', which get updated when the internal provider for the root prop is updated. We take the raw data, and then compute custom tables that cache the data, getting rid of the need to do the complex queries to get the information.

This is in a pretty raw state - if we decide that summary tables like this are a pattern we want more broadly, it'll certainly need some refactoring (like moving the summary table code into a single module, etc.)


stack72:
bors merge

PR #3155, opened 2024-01-11
fix: docker + iptables conflicting rules amendment

johnrwatson:
TLDR: Docker was being a pain

These were the existing rules in the iptables nat table regarding otel:
```
-A PREROUTING -d 1.0.0.1/32 -p tcp -m tcp --dport 4317 -j DNAT --to-destination 10.1.151.8:4317
-A DOCKER ! -i docker0 -p tcp -m tcp --dport 4317 -j DNAT --to-destination 172.17.0.2:4317
```

The two iptables rules used are for different scenarios, but there is a potential conflict depending on the specific use case. In our case, they were conflicting causing irrational packet loss.

If there is a possibility that packets can match both rules, there might be a conflict. The conflict arises because both rules are trying to DNAT packets with the same destination port (4317) but to different destination addresses (10.1.151.8:4317 and 172.17.0.2:4317).

To avoid conflicts, make sure that the conditions for each rule are mutually exclusive. If you have specific criteria to distinguish between the scenarios these rules are meant for, you should adjust the rules accordingly. Additionally, you may want to consider the order of the rules and how they interact with other rules in your iptables configuration.


johnrwatson:
bors merge

PR #3156, opened 2024-01-11
feat(sdf,telemetry): update HTTP status on all request spans

fnichol:
This change sets the `http.response.status_code` span field on the parent span of each incoming HTTP request.

Future and different work will be needed to augment the span with error-specific information as it will involve implementating another Tower middleware to intercept the error types.


fnichol:
bors merge

PR #3157, opened 2024-01-11
fix: add missing iptables rule for default network namespace forwarding

johnrwatson:
For some reason on some launches this rule isn't required and is already set.

This ensures it is set for all veritech launches (it's required for the NAT traffic to leave for the firecracker VM's to reach the internet)


johnrwatson:
bors merge

PR #3160, opened 2024-01-12
frame attach/detach improvements

theoephraim:
still wip but coming together

- show visual indicator of whether components are attached to their parent
- hovering that indicator highlights the parent frame(s)
- add "detach from parent(s)" to right click context menu and wire up (backend bug)

dragging components into frame behaviour still needs to be adjusted
- show indicator that an attachment is about to happen depending on mouse position
- determine attachment based on _mouse position_ rather than component bounds
- on attaching, we should bump components inside of frame bounds (or expand frame?) so we dont end up with components that are attached but outside the frame

BONUS
- small konva perf tweaks
- hover meta info moved to the store
- refactor hover handling - components now talk directly to the store rather than emitting events to ModelingDiagram

### DIFF
--- a/app/web/src/components/ModelingDiagram/DiagramEdge.vue
+++ b/app/web/src/components/ModelingDiagram/DiagramEdge.vue
@@ -85,6 +85,7 @@ import {
   useTheme,
   getToneColorHex,
 } from "@si/vue-lib/design-system";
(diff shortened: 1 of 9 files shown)


theoephraim:
bors merge

theoephraim:
bors merge

PR #3171, opened 2024-01-17
Act on http request asynchronously, returning immediately

paulocsanz:


app/web/src/components/ModelingDiagram/ModelingDiagram.vue
```diff
@@ -2210,6 +2212,11 @@ const pasteElementsActive = computed(() => {
     componentsStore.selectedComponentIds.length > 0
   );
 });
+
+const installId = ref<string | null>(null);
```
theoephraim, line 2216: I think we'll have an easier time if you keep this stuff in the store - and keep the realtime communication there as well

paulocsanz: I see, to me it doesn't make sense to have information that is exclusive for a module in the store, specially if it's temporary. Why do you prefer to move that stuff to the store, instead of encapsulating and keeping context where it makes sense?

But ok, I can move that tomorrow.

theoephraim: Keeping the communication with the API more encapsulated within the stores, and letting the components just visualize the current state tends to be easier in the long run. As soon as this condition might be triggered another way (like from multiplayer, or undo/redo/etc) then you'll end up wanting to move it there anyway. Also you'll  notice that nowhere in the codebase are we editing store data from components directly and instead always do so via actions. Again it's not strictly necessary, but keeping that stuff contained as much as possible within the stores after the user shows some intent to do something tends to keep things cleaner.

Of course there's a million ways to do everything, but these things come from lessons learned over the years... and its also nice to keep consistency with the rest of the codebase.

theoephraim: oh also, if you do leave it this way, you need to clean up the subscriptions before unmount in case they never finish - or else you'll create a memory leak

paulocsanz: Great point about cleaning on unmount, thank you! I will migrate it to the store

paulocsanz: Done!

paulocsanz:
bors merge

paulocsanz:
bors merge

PR #3173, opened 2024-01-18
feat: augment NATS tracing & add INFO span support for subscriptions

fnichol:
This change is doing...a lot, but is primarily focused on enabling a
tracing span that can sit over an invoming NATS message that ultimately
send one or more reply messages at the end (in this way, a very RPC-like
interaction).

The low level span-creating machinery and OpenTelemetry header
injection/extraction is contained in the `lib/telemetry-nats-rs` Rust
crate. This machinery is wired in to our somewhat generic NATS
subscription streaming crate `lib/nats-subscriber` as Pinga, Veritech,
and parts of SDF already use this as their interface to a NATS
subscription.

As part of creating this "request span", a fair amount of span metadata
has been updated in the `lib/si-data-nats` crate to help with more
meaningful/contextual information.

At present in `lib/nats-subscriber`, the `Subscriber` stream yields a
generic `Request` type (as before) but now contains this "request span"
and as long as this field and/or the request type is kept in some scope
(that is, it isn't dropped), this will keep the "request span" open.
Some effort in this change (and more in future work) is being made to
make all associated work be captured in child spans or "follows from"
spans if it is related work but not bound to the lifetime of the
request/reponse lifetime.

At the moment, Pinga is most mature here, creating a `pinga-jobs
receive` span for each incoming job request. While not 100% of the
related work is fully parented/associated, almost all of it is ;)


fnichol:
bors try

fnichol:
bors try

fnichol:
bors merge

PR #3175, opened 2024-01-18
chore: revert "build(deps): upgrade to NATS 2.10.9"

fnichol:
This reverts commit b9f26c80506f7883297832888132493bd9f3b951.

I would like to better understand if there is a change in the connection details between these versions of NATS. Granted I made a massive jump and over 2 years of upstream releases :)


fnichol:
bors merge

PR #3181, opened 2024-01-19
feat(web) snap frame attach/detach with snapping

theoephraim:
probably needs some cleanup, but this is feeling pretty good!

- you can now drag items in/out of frames to attach/detach
- New parent is based on cursor position rather than what each component is hovering over
- we highlight the new parent so you can see what it will be attached to
- we optimistically update parent id so it feels snappy
- there is logic that bumps children into the new parent if they are outside, which means things sort of snap into the parent...

It's not perfect, but it feels pretty good!

Main issue is that when snapping things into place, you can still end up with items hovering over each other that are not actually attached.

### DIFF
--- a/app/web/src/components/ModelingDiagram/DiagramGroup.vue
+++ b/app/web/src/components/ModelingDiagram/DiagramGroup.vue
@@ -12,7 +12,7 @@
   >
     <!-- selection box outline -->
     <v-rect
-      v-if="isHovered || isSelected || highlightParent"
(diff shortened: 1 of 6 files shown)


stack72:
@theo I just fixed up the conflict here - this PR isn't quite ready - *any* movement of a component inside the frame will actually detach it - it's not calculating that the component is outside the frame to detach

theoephraim:
Will take a look! Likely a few other issues as well… but hopefully it feels
pretty good?

stack72:
bors merge

PR #3184, opened 2024-01-19
chore(web) - Upgrade cypress to latest, refactor structure/tests, support auth0

britmyerss:
This change upgrades Cypress to latest - we were on quite an old version so I had to restructure the config to adopt Cypress' new ways. I added a command to login to auth0 and a very simple (and probably not very reliable) test to create an AWS Credential on the diagram. We'll need to add the right data to the components we want to drive but that can come later.


britmyerss:
bors try

britmyerss:
bors merge

PR #3185, opened 2024-01-19
feat: get_diagram is now a summary table

adamhjk:
Creating a summary table for get_diagram, in two parts - one for components, and one for edges. This one is a bit more complicated than the qualification summary table, as it has more spots where the data needs to be updated.

Add summary edges and collect parentNodeIds for components

Summary edges are now deleted and created automatically. The migration for summary diagram tables and functions has been incremented due to a conflict when rebasing with main. "ChildNodeIds" are now collected within the frontend and only "parentNodeId" is collected for summary components.

Correctly determine if the attribute value being processed in dependent values update corresponds to a root prop and a component. Fix integration tests broken by changing the diagram edges and nodes types.


adamhjk:
bors merge

adamhjk:
bors merge

PR #3190, opened 2024-01-22
Add instance timeout to poolnoodle

sprutton1:
* moves the PoolNoodle invocation to inside of local_uds instead of at the manager level to avoid repeating decision logic  specific to firecracker
* adds an `active` pool when we start an instance that keeps track of the start_time
* we now check if the pools are timed out and move them to the `to_be_cleaned` pool
* this check is pretty lazy and only happens if we don't have new jails to make or old jails to clean. Given that the timeout is 60 minutes, another few seconds should be fine


sprutton1:
bors try

sprutton1:
bors try

sprutton1:
bors try

PR #3192, opened 2024-01-22
feat: only reload properties and schema for impacted components

adamhjk:
This PR makes it so that we only re-load the properties and schema for impacted components. It adds a new WsEvent called "ComponentUpdated", which provides the component id and change set id. The component_attributes store then uses that event to trigger its update, rather than the raw ChangeSetUpdated event.


adamhjk:
bors merge

PR #3196, opened 2024-01-22
chore: amend wording in new change set dropdown

johnrwatson:
Amends the wording to reflect "Change Set <number>" rather than "Demo <number>" as it's more appropriate as we move towards Production.


johnrwatson:
bors merge

PR #3197, opened 2024-01-22
fix(web): calculate ancestry in frontend to ignore parentId stored for get_diagram

vbustamante:
Fixes bug in which we were showing implicit frame edges

image


vbustamante:
bors merge

nickgerace (approved):

PR #3201, opened 2024-01-23
fix(sdf,dal): Update frame children list on delete_component

vbustamante:


vbustamante:
bors merge

vbustamante:
bors merge

PR #3203, opened 2024-01-23
Allow updating local asset

paulocsanz:
This PR should be ready now. It was a wild ride to bring it here, but it seems done. The updates are behind a feature flag and it has been thoroughly tested, including how it affects workspace imports.

I need to merge this, so I don't want to make more changes in this PR. So the endpoint is not async yet.

I can't promise there aren't problems with updates, or even regressions on imports, so we should be on the lookout for bugs, and if some is found please ping me and I will take a look. But I did what I could to diminish the risk.


paulocsanz:
bors try

paulocsanz:
bors cancel

paulocsanz:
bors try

app/web/src/components/AssetDetailsPanel.vue
```diff
@@ -248,18 +255,8 @@ const updateAsset = async () => {
   }
 };
 
-const disabled = computed(() => !!(editingAsset.value?.hasComponents ?? false));
-
-const disabledWarning = computed(() => {
-  if (editingAsset.value?.hasComponents) {
```
paulocsanz, line 254: Undo this and put behind a feature flag

paulocsanz:
bors try

paulocsanz:
bors try

paulocsanz:
bors try

paulocsanz:
bors try

lib/dal/src/pkg/import.rs
```diff
@@ -211,7 +215,7 @@ async fn import_change_set(
     // Code level feature flag to allow updating any schema by simpling installing a package with a schema of the same name
     // This existis mostly to help debugging
     // Should always be `false` in production
-    let update_even_if_not_builtin = false;
+    let update_even_if_not_builtin = true;
```
zacharyhamm, line 218: Curious about this flag and the comment (which says it should always be false)

lib/dal/src/pkg/export.rs
```diff
@@ -553,7 +518,7 @@ impl PkgExporter {
                 .arity(socket.arity())
                 .ui_hidden(socket.ui_hidden());
 
-            let mut has_custom_func = false;
+            // let mut has_custom_func = false;
```
zacharyhamm, line 521: What's going on with this flag?

lib/dal/src/pkg/export.rs
```diff
@@ -178,27 +165,25 @@ impl PkgExporter {
...
                 }
-                schema_spec_builder.variant(variant_spec);
             }
+            schema_spec_builder.variant(variant_spec);
         }
 
-        if in_change_set && is_deleted {
+        if is_deleted {
             schema_spec_builder.deleted(true);
-        } else if in_change_set {
```
zacharyhamm, line 201: This means we will export all schemas into this change set in a workspace export whether they're only present on head or in the change set. I have to check the import logic but at the very least it's stuffing a ton of data into the export that doesn't need to be there.

lib/dal/tests/integration_test/internal/pkg.rs
```diff
@@ -627,8 +627,8 @@ async fn test_install_pkg(ctx: &DalContext) {
         .data(
             FuncSpecData::builder()
                 .name("si:truthy")
-                .display_name("truth")
-                .description("it returns true")
+                .display_name("truth".to_owned())
```
zacharyhamm, line 630: What happened here to make this necessary?

lib/sdf-server/src/server/service/variant_definition/exec_variant_def.rs
```diff
@@ -165,106 +164,430 @@ pub async fn exec_variant_def(
...
+            }
+        }
 
-            let schema_variant = SchemaVariant::get_by_id(&ctx, &schema_variant_id)
-                .await?
-                .ok_or(SchemaVariantError::NotFound(schema_variant_id))?;
+        let mut func_ids: Vec<String> = Vec::new();
 
-            let attribute_prototypes = migrate_attribute_functions_to_new_schema_variant(
+        exporter
```
zacharyhamm, line 279: What does this do if we throw away the return value?

zacharyhamm (commented):
The changes to the export logic's handling of visibility mean that schema variants in head get exported into every change set in the package. It also means that those schema variants get imported into every change set in the package, even if they're only present on head. All attribute prototype for them get written into both HEAD and the change set, which means that changes to HEAD will not be reflected in the changeset. So for example if you had only modified a single function's prototype in the change set, if you export your workspace now, then import it, and go to HEAD, the changes you make in head to the schema variant will not propagate into the changeset version, since every prototype has been created in both changesets now and the changeset visibility version will be chosen.

zacharyhamm:
I exported my workspace with a single custom variant, then re-imported it. I tried to add a prop to the variant's prop tree. Got this error:

lib/dal/src/pkg/import.rs
paulocsanz: You are right about the comment, I may change it.

Basically the original spec defined that only builtins would behave like that, but we realized that this was not ideal. I didn't want to remove that flag right now to allow us to quickly toggle the feature.

lib/dal/src/pkg/export.rs
paulocsanz: I have no idea! We used to check it before setting the data, but it doesn't make sense to me to only set data if has custom func.

What do you think?

But I didn't want to remove that from the code, since we may want to restore it quickly and without missing information (I'm focusing on making it work to merge this, as the PR has lived for too long and it gets harder and harder to make all parts move together while rebasing on main)

lib/dal/src/pkg/export.rs
paulocsanz: Hmm, I need to test this, you are right on this. The problem is that sometimes we need to export the schema in the change-set and sometimes we don't.

lib/dal/tests/integration_test/internal/pkg.rs
paulocsanz: We were removing the option in the setter, which made it annoying to add an option to it. Since we were building the spec from the pkg, and the field in the package also is an option, it doesn't make sense to have to if let in the field for the package to set an optional to the spec.

lib/sdf-server/src/server/service/variant_definition/exec_variant_def.rs
paulocsanz: What do you mean by that?

lib/sdf-server/src/server/service/variant_definition/exec_variant_def.rs
```diff
@@ -165,106 +164,430 @@ pub async fn exec_variant_def(
...
+
+        'outer: for old_socket in exporter
+            .export_sockets(
+                &ctx,
+                Some(ctx.visibility().change_set_pk),
+                *previous_variant.id(),
+            )
+            .await?
+        {
+            for socket in &mut variant_spec.sockets {
```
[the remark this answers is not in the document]
paulocsanz: Sure, syntax differences.

paulocsanz:
I was not able to reproduce that last bug

paulocsanz:
bors merge

PR #3205, opened 2024-01-23
Ensure pinga fails on startup if it cannot connect to PG (BUG-68)

nickgerace:
Ensure pinga fails on startup if it cannot connect to PG. SDF works similarly by failing upon migrating. This does result in a duplicate test of the database connection. However, we want to retain both tests because the first one is used by "si-data-pg" to warm up the pool. This second one is used to block the listener loop from occuring.


nickgerace:
bors merge

PR #3208, opened 2024-01-24
Fix components from HEAD not having new positions persist (BUG-71)

nickgerace:
Fix components that existed on HEAD not having new positions (from a new change set) persist. This fix inserts the row into the summary table, like the "component_update" function, when setting geometry for a summary component.

### DIFF
--- a/lib/dal/src/migrations/U2412__summary_diagram_update_geometry_insert_row.sql
+++ b/lib/dal/src/migrations/U2412__summary_diagram_update_geometry_insert_row.sql
@@ -1,0 +1,77 @@
+DROP FUNCTION IF EXISTS summary_diagram_component_update_geometry_v1(
+    this_tenancy jsonb,
+    this_visibility jsonb,
+    this_node_id ident,
+    this_position jsonb,
+    this_size jsonb,
+    OUT object json
+);
+
+CREATE OR REPLACE FUNCTION summary_diagram_component_update_geometry_v2(
+    this_tenancy jsonb,
+    this_visibility jsonb,
+    this_node_id ident,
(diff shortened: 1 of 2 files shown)


lib/dal/src/migrations/U2412__summary_diagram_update_geometry_insert_row.sql
```diff
@@ -0,0 +1,77 @@
...
+    this_tenancy_record    tenancy_record_v1;
+    this_visibility_record visibility_record_v1;
+    this_new_row           summary_diagram_components%ROWTYPE;
+BEGIN
+    this_tenancy_record := tenancy_json_to_columns_v1(this_tenancy);
+    this_visibility_record := visibility_json_to_columns_v1(this_visibility);
+
+    IF NOT EXISTS (SELECT
+                   FROM summary_diagram_components
+                   WHERE node_id = this_node_id
```
nickgerace, line 28: Subtly different from the `component_update` function.

lib/dal/src/migrations/U2412__summary_diagram_update_geometry_insert_row.sql
```diff
@@ -0,0 +1,77 @@
+DROP FUNCTION IF EXISTS summary_diagram_component_update_geometry_v1(
```
paulocsanz, line 1: Isn't it better to use v2 instead of replacing v1?

nickgerace: Yes

nickgerace:
bors merge

PR #3209, opened 2024-01-24
bug(geometry): this fixes the geometry upsert

adamhjk:
Fixes the issue where 'this_change_status' is seen as not existing as a column, because it's not set as a variable. We actually don't want to update the change status when you move the geometry anyway, so easy fix.


stack72:
bors merge

PR #3212, opened 2024-01-24
fix(web) url issues when switching change sets

theoephraim:
we were running into issues sometimes when swapping change sets

This stuff gets pretty gnarly since we are keeping some state in sync between the URL, the store, and the tabs.
Unfortunately, there's not a good way out to make it much cleaner...


stack72:
bors merge

PR #3220, opened 2024-01-26
bug: stop trying to find the span in our transaction libs

adamhjk:
This is a brutal hammer that disables some logic that tried to find the correct span for a transaction. It should make the spans a little weirder, but we think it will fix our performance problem.


adamhjk:
bors merge

PR #3221, opened 2024-01-26
Fail fast on failed fixes (BUG-72)

nickgerace:
Rather than processing fixes after the current batch is complete, let's process failures as they happen. When processing failures, we should log errors rather than bubbling them up.

For context, at the time of writing, "fixes" are the backend mechanism for "actions" performed in the frontend.

### DIFF
--- a/lib/dal/src/job/definition/fix.rs
+++ b/lib/dal/src/job/definition/fix.rs
@@ -169,40 +169,62 @@ impl JobConsumer for FixesJob {
             });
         }
 
-        let mut failed_fixes = VecDeque::new();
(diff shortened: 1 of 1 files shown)


nickgerace:
bors merge

nickgerace:
bors cancel

nickgerace:
bors merge

nickgerace:
Okay Bors, they are all set, so let's try this again.

nickgerace:
bors merge

nickgerace:
Yeah, I gotcha.

PR #3222, opened 2024-01-26
chore(si-data-pg,si-data-nats): compute appropriate Span for txns/subs

fnichol:
This change incorporates an upstream fix in `tracing-opentelemetry` to prevent parent and "follows-from" span association from panicking under certain scenarios.

Additionally, the strategy to determine the appropriate span for PostgreSQL and NATS transactions/subscriptions has been updated slightly. The macros present prior to this change attempted to prevent another "Mutex poisoned" issue encountered a year ago. The updated `tx_span()` and `sub_span()` functions check to see if `Span::current()` is disabled, in which case a `Span::none()` is returned, otherwise the current Span is returned. The former macros were slightly less flexible in that they checked `target` and `level` span metadata--which happened to be correct in these two cases.

References: tokio-rs/tracing-opentelemetry#14
References: tokio-rs/tracing-opentelemetry#86
References: tokio-rs/tracing#2399
References: tokio-rs/tracing#1766

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -578,7 +578,7 @@ checksum = "b58071e8fd9ec1e930efd28e3a90c1251015872a2ce49f81f36421b86466932e"
 dependencies = [
  "serde",
  "serde_repr",
- "serde_with 3.4.0",
+ "serde_with 3.5.1",
 ]
 
 [[package]]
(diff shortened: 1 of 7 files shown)


fnichol:
bors merge

PR #3223, opened 2024-01-26
Fix doc and clippy lints (2024-W04) (ENG-2278)

nickgerace:
## Description

Ran the following two commands to fix warnings found:

- `cargo clippy --all-targets --all-features`
- `cargo doc --all --all-features --document-private-items`

## Additional Note

I wonder why our `prelude-si` checks aren't including the `cargo doc` equivalent checks, in particular. Something to investigate...


nickgerace:
bors merge

lib/cyclone-server/src/tower.rs
```diff
@@ -1,5 +1,3 @@
-pub use limit_requests::{LimitRequest, LimitRequestLayer};
```
fnichol, line 1: Why are these type re-exports removed?

lib/pinga-server/src/config.rs
```diff
@@ -1,18 +1,16 @@
...
 use si_crypto::{CryptoConfig, SymmetricCryptoServiceConfig, SymmetricCryptoServiceConfigFile};
 use si_data_nats::NatsConfig;
 use si_data_pg::PgPoolConfig;
 use si_std::CanonicalFileError;
+use std::{env, path::Path};
 use telemetry::prelude::*;
 use thiserror::Error;
+use ulid::Ulid;
 
-pub use si_crypto::CycloneKeyPair;
```
fnichol, line 13: Why is this export removed?

fnichol (commented):
I'm wondering about a few of the type exports that were removed? Was this an oversight or on purpose?

fnichol:
@nickgerace The reason these checks aren't yet in CI was for 2 reasons:

1. There were a *lot* of failures that needed to be addressed
2. The current Buck2 rule relies on running `cargo doc` which all take the global Cargo lock at made running these pretty untenable. Theses days, from what I recall, I think I'd be able to make the Cargo-free version of these rules but until then it was going to be far more painful on the runtime of `si/review-pr` to be worth turning them on.

nickgerace:
bors cancel

nickgerace:
Need to review your comments @fnichol! Cancelled the merge.

lib/pinga-server/src/config.rs
nickgerace: Clippy seems to believe the import is unused. Does not mean that we don't want to keep though, but I see I've seen it on `main` for a couple weeks now.
```
warning: unused import: `si_crypto::CycloneKeyPair
```

lib/cyclone-server/src/tower.rs
nickgerace: Clippy seems to believe the import is unused. Does not mean that we don't want to keep though, but I see I've seen it on `main` for a couple weeks now.
```
warning: unused import: limit_requests::{LimitRequest, LimitRequestLayer}
```

nickgerace:
@fnichol good to know both of those reasons! I run into these lints pretty often, so found some down time and decided to hand comb through them. As far as the re-exports go, `clippy` might be being overly enthusiastic. I replied to your comments accordingly.

nickgerace:
Now's not the time and there's increasing drift with main.

PR #3231, opened 2024-01-29
bug(dal): fix some deadlocks

adamhjk:
This fixes deadlocks relating to the calculation of child_node_ids by the backend. We remove the trigger that calcualtes them, and now the summary tables always return an empty array. The frontend then calculates the correct child node ids itself by manipulating the raw response.

CC: Victor!


adamhjk:
bors merge

adamhjk:
bors cancel

stack72:
bors cancel

adamhjk:
bors merge

PR #3232, opened 2024-01-30
feat: refactor application telemetry

fnichol:
This change aggresively refactors the telemetry setup for all services.
The enabling/disabling of OpenTelemetry has been removed and is now
*always* active and set up.

The dynamic tracing level functionality remains but its approach is
different. In this current iteration, both the console logger and the
OpenTelemetry layer have their own reloadable `EnvFilter` layer on top,
whereas before the `EnvFilter` layer was its own third layer in the
`Registry` stack. That is:

```
- Registry
  - console_log
    - env_filter
  - otel
    - env_filter
```

At the moment both of the `EnvFilter`s are updated identically and at
the same time, but these may become more independent in future work.

Additionally, long-running the signal handling and level-updating tasks
have been re-written to be their own structs.

Finally, a new pattern is introduced into each of the service binary
crates which make use of 2 types from the `tokio-util` crate:
`TaskTracker` and `CancellationToken`. These 2 types work extremely well
together and make cancellation/shutown signalling and waiting (i.e. a
process-wide graceful shutdown) much more straight forward. Future work
will embed this idea further down into each service, allowing much of
the existing shutdown channel infrastructure to be pulled out.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -108,9 +108,9 @@ dependencies = [
 
 [[package]]
 name = "anstyle"
-version = "1.0.4"
+version = "1.0.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7079075b41f533b8c61d2a4d073c4676e1f8b249ff94a393b0595db304e0dd87"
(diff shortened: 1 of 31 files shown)


fnichol:
bors merge

PR #3233, opened 2024-01-30
bug(web,sdf,dal): remove change_set_written

adamhjk:
Pour one out for our fallen homie, Change Set Written. It updated the UI when nothing else would, and kept us all safe from bad data. But it had long since served its purpose, and is no longer the loving friend it once was. Goodbye, old friend.

* change set store no longer reacts to change set written
* fixes store no longer reacts to change set written
* qualifications store no longer reacts to change set written
* qualifications now correctly update their status when merged to head, along with names in the outliner
* fixes the components and qualficiations stores to not use change set written
* action store no long uses changeSetWritten
* secret store no longer uses change set written
* asset store no longer responds to change set written
* func store no longer uses change set write
* remove change_set_written events that are now needless

Less spurious reloading, here we come.


adamhjk:
bors check

fnichol:
bors try

nickgerace (approved):

stack72:
bors merge

PR #3234, opened 2024-01-30
chore: inject otel ctx when dal submits jobs to pinga

fnichol:
This change marks the first transmission of OpenTelemetry trace metadata between 2 service components. In this case, from a `NatsProcessor` (in the DAL) to a Pinga job. Future work will connect more. Yay!


fnichol:
bors merge

fnichol:
bors retry

PR #3236, opened 2024-01-30
feat: allow mounting cyclone decryption key into jails

sprutton1:
This adds a volume to the firecracker jails that is expected to include the cyclone decryption key. This volume must be present in the `/firecracker-data` directory when preparing the jail. The volume is teeny, one whole megabyte, so I thought it would be fine to just copy it around.

I also updated the userdata script to match what we have in the tools env plus having it pull the key from an aws secret, make the volume, and place it where it should go. 

I am removing the dev key from the rootfs entirely so this will break local development until I figure out how to add this to buck2. It's an easy fix, just create the volume with the dev key in it and put it where it goes.

### DIFF
--- a/bin/veritech/scripts/userdata.sh
+++ b/bin/veritech/scripts/userdata.sh
@@ -1,7 +1,71 @@
 #!/bin/bash
 
-POOL_SIZE=${1:-1000}
+# Detects all ssd devices present on the machine,
+function detect_devices {
+  local block_devices=$(ls /dev | grep nvme[1-9]*n[0-9]*$)
+  local result=""
+  for device_name in $block_devices
+  do
+    device="/dev/$device_name"
+    # if it is a block device and not a root volume
+    if [[ -b $device ]] && [[ ! "$(file -b -s $device)" == *"boot sector"* ]]
+    then
+      result+=" $device"
+    fi
+  done
+  echo $result
+}
+
+# create RAID array
+function create_md_array {
+  local devices="$(detect_devices)"
+  local dev_count=$(echo $devices | wc -w)
+
+  echo "Total devices: $dev_count"
+  echo "Devices found: $devices"
+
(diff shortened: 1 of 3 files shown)


sprutton1:
bors try

bin/veritech/scripts/userdata.sh
```diff
@@ -12,17 +76,17 @@ curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip
...
 curl -fsSL get.docker.com | bash
 
 docker run \
  --restart always \
  --env SI_OTEL_COL__CONFIG_PATH=/etc/otelcol/honeycomb-config.yaml \
- --env SI_OTEL_COL__HONEYCOMB_API_KEY=$(aws secretsmanager get-secret-value --region us-east-1 --secret-id si-honeycomb-api-key | jq -r '.SecretString') \
+ --env SI_OTEL_COL__HONEYCOMB_API_KEY=$(aws secretsmanager get-secret-value --region us-east-1 --secret-id $HONEYCOMB_API_SECRET | jq -r '.SecretString') \
  -p 4317:4317 \
- systeminit/otelcol:stable
+ -d systeminit/otelcol:stable
```
johnrwatson, line 89: Nice, apologies I totally forgot to backmerge this after I got it working

prelude-si/rootfs/rootfs_build.sh
```diff
@@ -169,17 +173,6 @@ EOL
 # we created above. This will cumulatively stack the content of each.
 for binary_input in "${binary_inputs[@]}"; do
   sudo tar -xpf "$binary_input" -C "$ROOTFSMOUNT"
-
-  # TODO(johnrwatson): This can never make it into Production We need to figure
```
johnrwatson, line 173: swish

PR #3237, opened 2024-01-30
feat(sdf): add `--generate-symmetric-key-path` to SDF binary

fnichol:
This new flag operates like the former Cyclone key generation flags (i.e. `--generate-cyclone-secret-key-path` and
`--generate-cyclone-public-key-path`) in that it will generate a new key, write it to the file as requested and then terminate the program.


fnichol:
bors merge

PR #3243, opened 2024-01-31
Add nats multiplexer for crdt, ws and general use (ENG-2281)

nickgerace:
## Summary

Alright! Here we go. This PR adds three crates: `nats-multiplexer`, `nats-multiplexer-client` and `nats-multiplexer-core`. Let's talk about them.

## Multiplexer Crates

Why use ["multiplexing"](https://en.wikipedia.org/wiki/Multiplexing)? Well, we were creating too many damn [NATS](https://nats.io) subscriptions... that's it.
Now, we have three crates to help alleviate the issue.

- `nats-multiplexer`: contains `Multiplexer`, which subscribes to one NATS subject and communicates over channels to receivers requested by a single client
- `nats-multiplexer-client`: contains `MultiplexerClient`, which communicates with a single, running `Multiplexer` to get receivers for a given subject (the subject can include a _terminating_ wildcard)
- `nats-multiplexer-core`: contains common dependencies for the two aforementioned crates (ensures that the client crate does not depend on the main crate for portability)

## SDF Changes

SDF now uses two multiplexers to subscribe to `"crdt.>"` for crdt logic and to `"si.>"` for workspace updates logic. The multiplexers prevent the ability to create too many NATS subscriptions driven by SDF users (e.g. `app/web`).

By design, there can only be one client per multiplexer. Clients can spawn many `broadcast::Receivers` for a given `MultiplexerKey` (i.e. a subject that can contain a _terminating_ wildcard), but there can only be one client. SDF's `AppState` wraps both clients in an `Arc<Mutex<T>>` as a result. We may want more clients in the future, but one client and multiplexer per SDF may work for now and forever... yeah, uh, famous last words.

We should eventually change how `app/web` interacts with SDF to avoid NATS misuse, but now that all communication must run through one of the two multiplexers, we ultimately avoid the issue.

## Closing Thoughts

There are remaining questions about `Tenancy` when request receivers, garbage collection within the multiplexer,  etc. However, all are minor for the near term future. Overall, these changes should not only resolve runaway NATS subscription issues in SDF, but should also provide libraries for creating NATS multiplexers moving forward.


nickgerace:
This is ready for review as far as the core bits go, but I have moved it back to drafts since we have to the `client_failure_doesnt_affect_others` crdt test failing and we have to finish plumbing the multiplexers through the `sdf_test` macro.

lib/sdf-server/tests/service_tests/mod.rs
```diff
@@ -6,14 +6,14 @@ use axum::{
...
-mod functions;
-mod scenario;
-mod schema;
-mod secret;
-mod session;
+// mod functions;
+// mod scenario;
+// mod schema;
+// mod secret;
+// mod session;
```
nickgerace, line 16: Temporary until the multiplexers are plumbed through the `sdf_test` macro. This PR will move out of drafts once that is done and working.

lib/sdf-server/tests/service_tests/crdt.rs
nickgerace, line 1: We need to solve an aforementioned test failure here before moving this PR out of drafts.

lib/si-test-macros/src/sdf_test.rs
nickgerace, line 1: Work in progress with plumbing the multiplexers through the `sdf_test` macro. This PR will move out of drafts once that is done and working.

Cargo.toml
```diff
@@ -136,11 +139,11 @@ tokio = { version = "1.28.0", features = ["full"] }
 tokio-postgres = { version = "0.7.8", features = ["runtime", "with-chrono-0_4", "with-serde_json-1"] }
 tokio-postgres-rustls = { version = "0.11.0" }
 tokio-serde = { version = "0.8.0", features = ["json"] }
-tokio-stream = "0.1.14"
+tokio-stream = { version = "0.1.14", features = ["sync"] }
```
nickgerace, line 142: You get `BroadcastStream` from this, which is nice a drop in replacement in crdt.

nickgerace:
bors try

nickgerace:
bors try

nickgerace:
bors merge

PR #3247, opened 2024-01-31
fix(dal,web,sdf) - fix up component created event firing and handling

britmyerss:
This PR modifies the ComponentCreated event to pass along the changeset pk and component id, which is used by the front end to re-fetch the diagram when components are created


nickgerace (approved):

britmyerss:
bors merge

PR #3249, opened 2024-02-01
Upgrade package protocol for components

paulocsanz:
I have no idea how to properly upgrade the package structure. Changing locally does not work because it requires the module service to be upgraded.

Please help

This gets merged to my asset upgrade PR, since it's related, but also to help inspecting the individual changes. But it's not urgent, nor blocking. It just hides upgraded components in the change-set they got upgraded.


paulocsanz:
bors try

PR #3250, opened 2024-02-01
bug(pinga): don't use nats for recursion

adamhjk:
Pinga was using Nats to recurse the fix jobs for a given batch. This makes it use a loop instead, with a limit on the total number of times we will go over the list of fixes.

We don't think this fixes the underlying issue, but we do think it adds a bunch of useful telemetry and significantly reduces the blast radius.


adamhjk:
bors merge

PR #3251, opened 2024-02-01
fix(dal,sdf): Fix how change status is computed for summary

vbustamante:
Make change status for summaries be computed in the same way as it is for components:
- If component has deleted timestamp, it's `deleted`
- If component on changeset does no exist on head, it's `added`
- If component gets modified and exists on head, it's `modified`

This PR also simplifies  our parentId setting code for summaries


vbustamante:
bors merge

nickgerace (approved):

vbustamante:
bors merge

PR #3252, opened 2024-02-01
chore(telemetry): aggressively dial back default tracing spans

fnichol:
This change targets most of our `#[instrument]` derive macro callsites in the Rust codebase.

A default log level setting for an `#[instrument]` call is `INFO` which cause these spans to be includes in all traces at the default logging level. What we're probably wanting (in the default case) is less/higher level spans in each service, espeically spans that happen at the service boundaries (i.e. make a job call over nats, etc.).

This change gives a much higher preference to service-specific spans to be at `INFO` level, and our other supporting crates tend to be at `DEBUG` and `TRACE` levels.

Our hope is that this change may lead to a clearer high level story without having to dial up the telemetry verbosity.

### DIFF
--- a/lib/council-server/src/server.rs
+++ b/lib/council-server/src/server.rs
@@ -315,7 +315,6 @@ pub enum Error {
     UnknownNodeId,
 }
 
-#[instrument(level = "info")]
 pub async fn register_graph_from_job(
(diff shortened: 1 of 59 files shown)


nickgerace (approved):

fnichol:
bors merge

PR #3255, opened 2024-02-02
fix(sdf): make connect_component_sockets_to_frame_inner iterative

fnichol:
This change converts the `connect_component_sockets_to_frame_inner` function from an async/recursive strategy to an iterative strategy. Prior to this change, when working with a large workspace containing multiple levels of nested frames (like 5 and more), calls to this function would recurse enough times that it would exhaust the stack space and cause the program to panic and abort. The iterative approach guards against an *arbitrarily* deep set of nested frames.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4488,7 +4488,6 @@ dependencies = [
 name = "sdf-server"
 version = "0.1.0"
 dependencies = [
- "async-recursion",
  "async-trait",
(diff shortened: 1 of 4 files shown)


paulocsanz (approved):
This is genius, such a simple change with such a big impact. Great way to avoid recursion!

fnichol:
bors merge

PR #3259, opened 2024-02-02
feat(veritech): add service-spanning telemetry to/from Veritech

fnichol:
This change adds tracing context propagation between the Veritech client and server code (in `lib/veritech-client` and `lib/veritech-server` respectively). With this change, you can now track each incoming function request into the Veritech server and see the corresponding output and result messaging between client and server. Very cool!


fnichol:
bors merge

PR #3260, opened 2024-02-02
fix(module-index-server): increase upload limit to 512Mb

zacharyhamm:


lib/module-index-server/src/routes/upsert_module_route.rs
```diff
@@ -73,8 +75,10 @@ pub async fn upsert_module_route(
     info!("Got part data");
 
     // SiPkg using old term "package" but we are dealing with a "module"
-    let loaded_module = dbg!(SiPkg::load_from_bytes(data.to_vec()))?;
-    let module_metadata = dbg!(loaded_module.metadata())?;
+    let loaded_module = SiPkg::load_from_bytes(data.to_vec())?;
+    let module_metadata = loaded_module.metadata()?;
+
+    info!("upserting module: {:?}", &module_metadata);
```
stack72, line 81: <3

PR #3264, opened 2024-02-02
fix(dal): fire ws_event when component position is set

britmyerss:
Small change - doesn't fix the frames issue we're seeing but it at least makes it a little easier to collaborate within a changeset


stack72:
bors merge

PR #3265, opened 2024-02-03
Calculate AttributeValue dependency graph ahead of time for DependentValuesUpdate

jhelwig:
Calculating the dependency graph whenever we need to figure out where all the data should flow to during a DependentValuesUpdate takes a significant amount of time when we do the entire calculation from scratch.

In order to speed up how longs the DependentValuesUpdate takes, we're now spreading out the cost by calculating the 1-hop "sources" of data for every AttributeValue, whenever a the AttributeValue, its AttributePrototype, or its AttributePrototypeArguments are updated. We then use these pre-calculated 1-hop sources to build the full graph.


britmyerss:
bors try

lib/dal/src/migrations/U2600__attribute_value_dependency_graph_summary.sql
```diff
@@ -0,0 +1,1732 @@
...
+CREATE INDEX ON prop_belongs_to_prop(tenancy_workspace_pk);
+CREATE INDEX ON internal_providers(visibility_deleted_at) WHERE (visibility_deleted_at IS NOT NULL);
+CREATE INDEX ON internal_providers(tenancy_workspace_pk);
+CREATE INDEX ON external_providers(visibility_deleted_at) WHERE (visibility_deleted_at IS NOT NULL);
+CREATE INDEX ON external_providers(tenancy_workspace_pk);
+
+DROP INDEX IF EXISTS attribute_value_belongs_to_attribute_prototype_single_associati;
+CREATE UNIQUE INDEX attribute_value_belongs_to_attribute_prototype_single_associati
+    ON attribute_value_belongs_to_attribute_prototype (object_id, tenancy_workspace_pk, visibility_change_set_pk)
+    WHERE visibility_deleted_at IS NULL;
```
jhelwig, line 41: The old constraint didn't have the `WHERE visibility_deleted_at IS NULL`, which meant that you couldn't `unset_belongs_to` and `set_belongs_to` for the same `object_id` within the same change set, as there would both be a soft-deleted record and a (conflicting) not-deleted record.

lib/dal/src/migrations/U2600__attribute_value_dependency_graph_summary.sql
```diff
@@ -0,0 +1,1489 @@
...
+        FROM attribute_values_v1(this_tenancy, this_visibility) AS av
+        WHERE av.attribute_context_component_id = this_new_row.id
+    LOOP
+        PERFORM attribute_value_dependencies_update_v1(
+            this_tenancy_record.tenancy_workspace_pk,
+            this_visibility_record.visibility_change_set_pk,
+            this_visibility_record.visibility_deleted_at,
+            this_new_attribute_value_id
+        );
+    END LOOP;
```
jhelwig, line 783: The new bit in `component_create_v1`.

lib/dal/src/migrations/U2600__attribute_value_dependency_graph_summary.sql
```diff
@@ -0,0 +1,1489 @@
...
+                           'ALTER TABLE %1$I '
+                           '    ADD CONSTRAINT %1$s_object_id_is_valid '
+                           '        CHECK (check_id_in_table_v1(%2$L, object_id)); '
+                           'ALTER TABLE %1$I '
+                           '    ADD CONSTRAINT %1$s_belongs_to_id_is_valid '
+                           '        CHECK (check_id_in_table_v1(%3$L, belongs_to_id)); '
+                           'CREATE UNIQUE INDEX %1$s_single_association ON %1$I (object_id, '
+                           '                                        tenancy_workspace_pk, '
+                           '                                        visibility_change_set_pk) '
+                           '    WHERE visibility_deleted_at IS NULL; '
```
jhelwig, line 1344: The old one wasn't taking into account that we only want **undeleted** records to count for the uniqueness.

jhelwig:
bors try

lib/dal/src/migrations/U2600__attribute_value_dependency_graph_summary.sql
vbustamante: Haven't read all of it yet, but does this mean we're not getting attribute values for deleted components at all?

lib/dal/src/attribute/value.rs
```diff
@@ -1151,6 +1240,20 @@ impl AttributeValue {
...
+        // special, and can't actually be re-run. Their values are static anyway, so re-running it
+        // wouldn't change anything. The "si:setObject", "si:setArray", and "si:setMap" functions
+        // are a bit special, however, as the "local" value will always be an empty object, array,
+        // or map.
+        let func = Func::get_by_id(ctx, &attribute_prototype.func_id())
+            .await?
+            .ok_or_else(|| {
+                AttributeValueError::MissingFunc(format!("Unable to get func for {:?}", self.id()))
+            })?;
+        if func.name().starts_with("si:set") || func.name() == "si:unset" {
```
zacharyhamm, line 1253: it might be safer to use the `InstrincFunc::maybe_from_str(func.name())` and just see that it's `Some` (but not `Some(IntrinsicFunc::Identity)`, just in case someone adds a custom function that starts with si:set

lib/dal/src/migrations/U2600__attribute_value_dependency_graph_summary.sql
vbustamante: we still show soft deleted components that  exist on head inside of the changeset as the ghosted components, so you can still look at their attribute editor. So we still need those values in the summary

vbustamante: nevermind, we don't show the attribute editro for the ghosted components

lib/dal/tests/integration_test/internal/provider/inter_component.rs
```diff
@@ -785,6 +809,17 @@ async fn with_deep_data_structure(ctx: &DalContext) {
...
+    ctx.txns()
+        .await
+        .unwrap()
+        .pg()
+        .execute(
+            "SELECT attribute_value_dependencies_migration_populate_v1()",
+            &[],
+        )
+        .await
+        .unwrap();
```
nickgerace, line 821: This looks like leftover debugging code.

lib/dal/src/queries/attribute_value/ids_with_dynamic_functions.sql
```diff
@@ -0,0 +1,11 @@
+SELECT avbtap.id AS attribute_value_id
+FROM attribute_value_belongs_to_attribute_prototype_v1($1, $2) AS avbtap
+    INNER JOIN attribute_prototypes_v1($1, $2) AS ap
+        ON ap.id = avbtap.belongs_to_id
+    INNER JOIN funcs_v1($1, $2) AS funcs
+        ON funcs.id = ap.func_id
+WHERE avbtap.object_id = ANY($3)
+    AND NOT (
+        funcs.name LIKE 'si:set%'
+        OR funcs.name = 'si:unset'
```
nickgerace, line 10: We'll need to keep an eye on `si:` funcs in the future.

lib/dal/src/migrations/U2600__attribute_value_dependency_graph_summary.sql
```diff
@@ -0,0 +1,1489 @@
...
+        (
+            attribute_value_id           ident,
+            dependent_attribute_value_id ident
+        )
+AS
+$$
+    WITH RECURSIVE dependency_graph AS (
+      SELECT
+          destination_attribute_value_id AS attribute_value_id,
+          source_attribute_value_id AS dependent_attribute_value_id
```
nickgerace, line 389: Potential naming confusion, but looks good to me!

nickgerace (approved):

zacharyhamm (commented):
There are three functions already in the buitlins that will break with this,

 si:setAwsDatabasePassword, si:setAwsListenerName and si:setAwsSecretString

It might be worth making the check more explicit about matching the "static" si:set functions. The other solution is to update the packages by changing the names of those functions and re-exporting them.

stack72:
> There are three functions already in the buitlins that will break with this,
> 
> si:setAwsDatabasePassword, si:setAwsListenerName and si:setAwsSecretString
> 
> It might be worth making the check more explicit about matching the "static" si:set functions. The other solution is to update the packages by changing the names of those functions and re-exporting them.

Those functions are already in use in production so we would need to coordinate here

lib/dal/src/migrations/U2600__attribute_value_dependency_graph_summary.sql
```diff
@@ -156,7 +156,14 @@ BEGIN
         -- If the prototype's func isn't one of the 'si:set*', or 'si:unset', then it's a
         -- "dynamic" function. We also need to update the InternalProvider for any child props
         -- to have this AttributeValue as a dependency.
-        SELECT funcs.name NOT LIKE 'si:set%' AND funcs.name != 'si:unset'
+        SELECT NOT (funcs.name = 'si:setObject'
```
britmyerss, line 159: you've got double `si:setObject` checks - reallly don't like objects, eh? :)

jhelwig:
bors merge

PR #3266, opened 2024-02-03
feat(cyclone): add initial service-spanning telemetry to/from Cyclone

fnichol:
This is a first-pass addition of propagating telemetry from Cyclone
clients (that is mainly Veritech by way of the deadpool-cyclone library)
and Cyclone servers. The server responses don't yet have trace
propagation headers but this will appear in future work. Additionally,
only the pre-WebSocket HTTP round trip is being accounted for so far.
Once the HTTP/101 (upgrade connection) response is issued then both the
client and server are done with HTTP and so axum considers that work as
complete. Future work will attempt to thread the lifetime of a "request
span" over the duration of the WebSocket session.


fnichol:
bors merge

PR #3269, opened 2024-02-06
build(deps): upgrade to NATS 2.10.10

fnichol:
Confirmed we're good to upgrade--this also brings our development environments to closer parity with our production environment.


fnichol:
bors merge

PR #3270, opened 2024-02-06
chore: support opt-in ANSI coloring of telemetry logging output

fnichol:
This change updates the telemetry setup and adds a couple of extra CLI arguments and environment variables to help control when it's appropriate for logging output to contain ANSI escape characters.

Each service has a new CLI argument of `--no-color` which will disable ANSI coloring of telemetry output when set. An associated environment variable of `SI_NO_COLOR` is also supported by setting the value to either `true` or `false`. Finally as a fallback of last resort, and to better support other operational ecosystems the `NO_COLOR` environment variable is also inspected and will disable coloring if this value is set to something non-empty.

In an attempt to make our services better behaved as Unix programs, they will now automatically disable ANSI coloring if the standard output of the process doesn't refer to a terminal or TTY (that is, the session is "non-interactive").

Our Tilt development environment can support output from processes with ANSI coloring, however these processes are considered "non-interactive" and thus won't be colored by default. As a consequence of this, there is one more new CLI argument of `--force-color` which will override the terminal/TTY detection logic and will unconditionally add ANSI coloring. An associated environment variable of `SI_FORCE_COLOR` will also work in place of using the `--force-color` CLI flag.

### DIFF
--- a/bin/council/src/args.rs
+++ b/bin/council/src/args.rs
@@ -19,6 +19,28 @@ pub(crate) struct Args {
     #[arg(short = 'v', long = "verbose", action = ArgAction::Count)]
     pub(crate) verbose: u8,
 
+    /// Disables ANSI coloring in log output, even if standard output refers to a terminal/TTY.
+    ///
+    /// For more details, visit: <http://no-color.org/>.
+    #[arg(
+        long,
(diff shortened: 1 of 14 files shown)


fnichol:
bors merge

PR #3271, opened 2024-02-06
fix: treat `--no-color`/`--force-color` as CLI flags

fnichol:
This change ensures that the new color CLI flags are flags and *not* options. That means that passing `--no-color` sets the value to `true` rather than an option where `--no-color false` is valid (this was not intended behavior).

In an effort to determine whether a CLI flag was set the telemetry config was also updated so that the right config precedence is in effect.


fnichol:
bors merge

PR #3272, opened 2024-02-06
feat: support telemetry logging option as JSON lines

fnichol:
This change adds a new `--log-json` flag to all services which will configure the telemetry to output logging as JSON objects, one per line. An associated environment variable of `SI_LOG_JSON` is also supported to enable this output mode.


fnichol:
bors merge

PR #3276, opened 2024-02-07
chore: revisit telemetry tracing levels to make debugging easier

fnichol:
This change updates the tracing from 5 prescribed levels to 7. This means that before there were a max of 4 `-v` levels and now there are 6.

Additionally, the `TelemetryConfig` has 2 new values that influence the tracing levels in addition to `app_modules`:

- `interesting_modules`: a `Vec` of Rust crate/module names that are not strictly the server's library code but rather "crates of interest" the application. For example, we care about `si-data-nats` when looking at the Veritech server as it's driven primarily via NATS. The default is an empty collection of modules.
- `never_modules`: a `Vec` of Rust crate/module names that will never be logged/transmitted/considered when setting levels. In other words, these modules are always set to `NAME=off` at any verbosity level. After operating our services several Rust crates demonstrated themselves to be extremely chatty at `DEBUG` and `TRACE` levels and the value we derive from this logging is low to nothing. By default a couple of initial crates are supplied here, namely `h2` and `hyper`.

So now there are 4 groups of modules to consider:

1. app modules
2. interesting modules
3. never modules
4. all other modules

The updated tracing level increases work as follows:

1. `INFO` for all modules (the default for all services)
2. `DEBUG` for app modules and `INFO` for all others
3. `DEBUG` for app and interesting modules, and `INFO` for all others
4. `TRACE` for app modules, `DEBUG` for interesting modules, and `INFO` for all others
5. `TRACE` for app modules, `TRACE` for interesting modules, and `INFO` for all others
6. `TRACE` for app and interesting modules, and `DEBUG` for all others
7. `TRACE` for all modules

### DIFF
--- a/bin/council/src/args.rs
+++ b/bin/council/src/args.rs
@@ -15,7 +15,7 @@ pub(crate) fn parse() -> Args {
 pub(crate) struct Args {
     /// Sets the verbosity mode.
     ///
-    /// Multiple -v options increase verbosity. The maximum is 4.
+    /// Multiple -v options increase verbosity. The maximum is 6.
     #[arg(short = 'v', long = "verbose", action = ArgAction::Count)]
     pub(crate) verbose: u8,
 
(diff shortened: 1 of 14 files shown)


fnichol:
bors merge

PR #3280, opened 2024-02-07
chore: add optional API to make tracing level changes "wait-able"

fnichol:
This change adds several `*_and_wait()` method options to the `ApplicationTelemetryClient` type. When our services boot up we check the parsed CLI argument for verbosity flags. If there are more than one set then we issue a command to raise the tracing level but prior to this change, this operation was non-blocking/async. That means that the level may not have raised by the time the next line of code executes resulting in potential missing log messages.


fnichol:
bors merge

PR #3281, opened 2024-02-07
chore: prevent credentials & secrets from display in CLI parsing

fnichol:
This change is a tweak to #3257 which has the same goal of not accidentally logging sensitive values such as keys, certificates, etc. In this version, we're using our `si_std::SensitiveString` type which has overridden `Display` and `Debug` trait implementations. So when a `debug!()` or `trace!()` call prints the contents of our `Args` CLI type, we won't see the contents of certain "sensitive" fields.


fnichol:
bors merge

PR #3283, opened 2024-02-07
Ensure content store works with TLS config (ENG-2315)

nickgerace:
Now that the content store (Pg) has been moved to the ServicesContext, it must also use the TLS config that the regular "PgPoolConfig" uses.


nickgerace:
Informed @sprutton1 that we can optimize later. This PR unblocks getting frames working on the new engine... well, any new engine work really.

PR #3286, opened 2024-02-08
feat(telemetry): add path-filterable support into HTTP request spans

fnichol:
This change adds a general capability to "filter" HTTP requests by their URI path and change the tracing level of the request span away from its default.

In our SDF and Cyclone server setups, we're creating an HTTP request span at the `INFO` level by default, and this change alters any request in SDF at `/api/` to create an HTTP request span at the `TRACE` level, effectively disabling these spans in all but the most intense debugging scenarios.


fnichol:
bors merge

PR #3287, opened 2024-02-08
chore(cyclone): revert cyclone's default tracing level to `INFO`

fnichol:
This change helps to support Cyclone server spans being emitted by default in development and production. Long prior to this change, the output from a Cyclone instance was so chatty that we dialed down the default log level but this has since been remedied. To further muddy the waters, the Tilt configuration has been running with `SI_LOG=debug` for the Veritech service meaning that all Cyclone server instances were also being hard-coded into running a `DEBUG` level. This development setting should eventually be removed as it widens the gap between development and production and should be increasingly less desirable under normal circumstances.

### DIFF
--- a/bin/cyclone/src/main.rs
+++ b/bin/cyclone/src/main.rs
@@ -5,14 +5,6 @@ use tokio_util::{sync::CancellationToken, task::TaskTracker};
 
 mod args;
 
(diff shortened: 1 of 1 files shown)


fnichol:
bors merge

PR #3288, opened 2024-02-08
feat(telemetry-http): add "matched" path to HTTP span names

fnichol:
Yay! There is an extension on the Axum request type that returns the router matching string. This means that if the request path in the Axum router was `/api/summary/:workspace_id`, then a related request span name would now be `GET /api/summary/:workspace_id`.

With this implementation detail sorted we can now include the request path without matched ids and other path variables in the span name. This keeps the number of span names lower (in OpenTelemetry jargon "low cardinality") and thus helps to increase the power of querying later on.


fnichol:
bors merge

PR #3293, opened 2024-02-09
Remove unused frame sockets and add multiple frames test (ENG-2325,ENG-2327)

nickgerace:
## Description

Remove unused frame sockets (external provider and explicit internal provider) when creating schema variants. Add an extended test with not only multiple frames, but also a connection from a frame to an unattached component.

Misc change: now we import AWS EC2 PKG and AWS PKG for integration tests (again).

## Test Contents

The test runs through a few non-trivial scenarios when working with multiple, non-nested frames. They are include, but are not limited to, the following:

- Ensure multiple frames with multiple connections work
- Ensure connections within frames work
  - Example: two components are placed within a frame. They have an edge between one another for a given set of sockets.
- Ensure edges from frame sockets to unattached components work
  - Example: create an AWS Region and create a AWS AMI outside of the frame. Draw an edge between the AWS Region's "region" socket to the AWS AMI's "region" socket. The AMI should remain unattached.

The test does not check for values propagation, but it does check for data setup and potential drift.

### DIFF
--- a/lib/dal/src/schema/variant.rs
+++ b/lib/dal/src/schema/variant.rs
@@ -194,26 +194,6 @@ impl SchemaVariant {
         let root_prop = RootProp::new(ctx, schema_variant_id).await?;
         let func_id = Func::find_intrinsic(ctx, IntrinsicFunc::Identity).await?;
 
-        InternalProvider::new_explicit(
-            ctx,
-            schema_variant_id,
-            "Frame",
-            func_id,
-            ProviderArity::Many,
-            ProviderKind::Frame,
-        )
-        .await?;
-        ExternalProvider::new(
-            ctx,
-            schema_variant_id,
-            "Frame",
-            None,
-            func_id,
-            ProviderArity::Many,
-            ProviderKind::Frame,
-        )
-        .await?;
-
         let schema_variant = Self::assemble(id.into(), content);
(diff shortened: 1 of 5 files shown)


lib/dal/src/schema/variant.rs
```diff
@@ -194,26 +194,6 @@ impl SchemaVariant {
...
-        ExternalProvider::new(
-            ctx,
-            schema_variant_id,
-            "Frame",
-            None,
-            func_id,
-            ProviderArity::Many,
-            ProviderKind::Frame,
-        )
-        .await?;
```
nickgerace, line 215: This is the only change that affects running the entire stack

lib/dal/tests/integration_test/internal/new_engine/frame.rs
```diff
@@ -106,3 +109,769 @@ async fn convert_component_to_frame_and_attach(ctx: &mut DalContext) {
         fallout_parent_node_id.expect("no parent node id for fallout component")
     );
 }
+
+#[test]
+async fn multiple_frames_with_complex_connections_no_nesting(ctx: &mut DalContext) {
```
nickgerace, line 114: This test runs through multiple scenarios with frames (and no nesting) in the new engine.

PR #3294, opened 2024-02-10
fix(sdf): don't hard delete stuff on HEAD

vbustamante:
This makes restore component safer.

Also updates component restore to hard delete the changeset version of summary components, instead of updating them. This was probably the cause of some ghost components too.


stack72:
bors try

stack72:
bors try

stack72:
bors merge

PR #3296, opened 2024-02-10
Run validations in the backend

paulocsanz:
New engine has been plugged to the old validations design, the new design implementation will come next

Display is behind a feature-flag, but execution is not, as dependent values update is tricky, since the user triggering it may not be the same as the one observing the value, and there is low risk of fallibility.

Validations functions were implemented on top of attribute functions, bypassing func binding, since it's not clear how we want to further integrate them in the system (there have been many failed attemps requiring a lot of rework), so a simple quick implementation that can be easily expanded on later feels the way to go. We will have to decide how to store the Func in the backend, if a new function type should exist (in my ideal world lang-js wouldn't have any metadata on the functions and the dal should handle that, but it's not how we have been doing it), if we want to make a summary of them, if they should be part of the attribute values graph, etc.

image

image


paulocsanz:
bors merge

PR #3298, opened 2024-02-12
Fix Module Index Server Startup

stack72:
We can't load a path from a SensitiveString so needed to change the path back to a string


stack72:
bors merge

PR #3300, opened 2024-02-13
Use NATS Jetstream for the new engine (ENG-2331)

nickgerace:
## Description

Welcome to [NATS Jetstream](https://docs.nats.io/nats-concepts/jetstream) for the new engine. Jetstream enables us to continue using our proven NATS stack while fulfilling the requirement to use a fast, durable queue for the new engine.

The speed increases are immediately apparent and we do not encounter platform-specific errors when using Jetstream. Moreover, there is concurrent work on `main` (coordinating with @fnichol) that is leveraging Jetstream, so the new engine will not be the sole user.

As a result of this change, the rebaser architercture has been significantly improved. Instead of managing both management and change set loops, the rebaser will now consume on one Jetstream stream. Replies to requesters are handled via NATS-generated reply mailboxes outside of the stream.

This PR also more easily enables use to deploy more than one rebaser due to multiple NATS Jetstream features, such as durable, shared consumers and a work queue retention policy.

## Other Changes

Council is no longer deployed via `dev:up`. It is not used in the new engine, so we will no longer include it in the running stack.

We have also dropped down to one net new dependency (`postcard` for de/ser) from two.

## Disclaimer

There are examples below, but this PR is a "first pass" style PR and does not do _everything_ we want with the rebaser architecture right away.

## Known Issues

While this PR takes us a step closer to deploying more than one rebaser at a time, additional work is needed to ensure that it works reliably. We likely need to use [work queue retention policy](https://docs.nats.io/using-nats/developer/develop_jetstream/model_deep_dive) on the stream as well as decide how to handle individual rebases.

For now, we assume that exactly one rebaser will be deployed (and restarted as needed). It blocks between rebase requests and does not spawn off tasks.

Another issue is that we are leaking `Tenancy` into connection state. I believe this is fine for this "first pass" style of PR as it is very specifically for rebasing and that's it. This may be fine long term and a non-issue, but I wanted to call it out.

## Rough Architecture Diagram

Attached is a rough architecture diagram for how we want the rebaser to work with Jetstream both now and in the near future. Not everything pictured is included in this PR (e.g. `pinga` and `sdf` do not try to create the rebaser stream). It is subject to change and may contain mistakes/misunderstandings as it was solely used for discussion and planning.

image


component/rabbitmq/Dockerfile
nickgerace, line 1: Will need to delete this on `main` as well.

Cargo.toml
```diff
@@ -118,7 +117,6 @@ postgres-types = { version = "0.2.5", features = ["derive"] }
 pretty_assertions_sorted = "1.2.1"
 proc-macro2 = "1.0.56"
 quote = "1.0.27"
-rabbitmq-stream-client = "0.3.0"
```
nickgerace, line 121: Now, the new engine only has one new dependency vs. `main`: `postcard`.

dev/Tiltfile
```diff
@@ -113,23 +111,6 @@ local_resource(
...
-    serve_cmd = "buck2 run {}".format(council_target),
-    serve_env = {"SI_FORCE_COLOR": "true"},
-    allow_parallel = True,
-    resource_deps = [
-        "nats",
-        "otelcol",
-    ],
-    deps = _buck2_dep_inputs(council_target),
-    trigger_mode = trigger_mode,
-)
```
nickgerace, line 131: Drop council from `dev:up` since we are not using it with the new engine

lib/rebaser-core/src/messaging_config.rs
nickgerace, line 1: This replaces the `SiRabbitqMqConfig` that was being used for both client and server. We only need the `subject_prefix` field from it since we rely on an existing NATS config for Jetstream.

lib/rebaser-core/src/subject.rs
```diff
@@ -0,0 +1,44 @@
...
+impl SubjectGenerator {
+    /// Returns the root subject for all rebaser-related-messages.
+    pub fn root(subject_prefix: Option<impl AsRef<str>>) -> String {
+        Self::assemble_with_prefix("rebaser", subject_prefix)
+    }
+
+    /// Returns the subject covering all rebaser-related messages.
+    pub fn all(subject_prefix: Option<impl AsRef<str>>) -> String {
+        Self::assemble_with_prefix("rebaser.>", subject_prefix)
+    }
```
nickgerace, line 19: All rebaser servers share a durable consumer that consumes on this subject.

lib/rebaser-core/src/subject.rs
```diff
@@ -0,0 +1,44 @@
...
+/// A generator that provides subject names in a centralized location.
+#[allow(missing_debug_implementations)]
+pub struct SubjectGenerator;
+
+impl SubjectGenerator {
+    /// Returns the root subject for all rebaser-related-messages.
+    pub fn root(subject_prefix: Option<impl AsRef<str>>) -> String {
+        Self::assemble_with_prefix("rebaser", subject_prefix)
+    }
+
```
nickgerace, line 15: This is the subject for the Jetstream stream. It is the root subject.

lib/rebaser-core/src/subject.rs
```diff
@@ -0,0 +1,44 @@
...
+    pub fn request(
+        workspace_id: Ulid,
+        change_set_id: Ulid,
+        subject_prefix: Option<impl AsRef<str>>,
+    ) -> String {
+        Self::assemble_with_prefix(
+            format!("rebaser.ws.{workspace_id}.cs.{change_set_id}"),
+            subject_prefix,
+        )
+    }
```
nickgerace, line 31: When requesting rebases, the clients will publish using this subject.

lib/rebaser-server/src/config.rs
```diff
@@ -168,10 +156,6 @@ fn default_cyclone_encryption_key_path() -> String {
     "/run/rebaser/cyclone_encryption.key".to_string()
 }
 
-fn default_recreate_management_stream() -> bool {
-    false
-}
```
nickgerace, line 173: This was never used and we'll want to think about configuration differently in the long term anyway with N rebaser servers.

lib/rebaser-server/src/server.rs
```diff
@@ -179,27 +138,26 @@ impl Server {
...
             self.job_processor,
             self.symmetric_crypto_service,
             self.encryption_key,
             self.shutdown_watch_rx,
-            self.rabbitmq_config,
+            self.messaging_config,
             self.content_store_pg_pool,
         )
-        .await;
+        .await?;
```
nickgerace, line 160: It's only fallible in the setup phase. The loop itself is infallible within our control.

lib/rebaser-server/src/server/core_loop.rs
nickgerace, line 1: This replaces the old loop files.

lib/rebaser-server/src/server/core_loop.rs
```diff
@@ -0,0 +1,190 @@
...
+    // Setup the subjects.
+    let subject_all = SubjectGenerator::all(messaging_config.subject_prefix());
+    let subject_root = SubjectGenerator::root(messaging_config.subject_prefix());
+    info!(%subject_all, %subject_root, "created services context and prepared subjects");
+
+    // Setup the stream and the consumer.
+    let jetstream_ctx = nats.clone().to_jetstream_ctx();
+    let rebaser_jetstream_stream = jetstream_ctx
+        .get_or_create_stream(&subject_root, vec![subject_all])
+        .await?;
```
nickgerace, line 77: Like before, the rebaser server(s) will try to find or create the stream they consume on. In the future, we should consider other services trying to do this too (like pinga and sdf) for rolling deployments.

lib/rebaser-server/src/server/core_loop.rs
```diff
@@ -0,0 +1,190 @@
...
+    let ctx_builder = DalContext::builder(services_context.clone(), false);
+
+    info!("setup complete, entering core loop");
+    core_loop_infallible(ctx_builder, stream, shutdown_watch_rx).await;
+    info!("exited core loop");
+
+    Ok(())
+}
+
+async fn core_loop_infallible(
```
nickgerace, line 92: This is the main loop for the rebaser servers.

lib/rebaser-server/src/server/core_loop.rs
```diff
@@ -0,0 +1,190 @@
...
+    let reply_subject = reply_subject.to_subject();
+
+    let reply_message = perform_rebase(ctx, message).await.unwrap_or_else(|err| {
+        error!(error = ?err, ?message, ?reply_subject, "performing rebase failed, attempting to reply");
+        ReplyRebaseMessage::Error {
+            message: err.to_string(),
+        }
+    });
+
+    match serde_json::to_vec(&reply_message) {
```
nickgerace, line 177: Reply back to the requester using regular NATS via a reply mailbox. We will do this for errors too.

lib/rebaser-server/src/server/rebase.rs
nickgerace, line 1: Moved the rebasing logic to its own file and function since it was part of the old `change_set_loop`, which is no longer needed.

lib/rebaser-server/src/server/core_loop.rs
```diff
@@ -0,0 +1,191 @@
...
+                continue;
+            }
+        };
+        ctx.update_visibility(Visibility::new_head(false));
+        ctx.update_tenancy(Tenancy::new(WorkspacePk::NONE));
+
+        // TODO(nick): decide if we want this to remain blocking.
+        let start = Instant::now();
+        perform_rebase_and_reply_infallible(&mut ctx, request_message, reply_subject).await;
+        info!("perform rebase and reply total: {:?}", start.elapsed());
```
nickgerace, line 159: We may want to spawn off tasks. This is where the new code is likely most subject to change.

We need a [work queue retention policy](https://docs.nats.io/using-nats/developer/develop_jetstream/model_deep_dive) and decide how we want to handle "acking" in progress.

For now, we assume we have exactly one rebaser server and that it will block on rebases and perform them in sequence.

lib/si-data-nats/src/jetstream.rs
nickgerace, line 1: This may conflict with what @fnichol is doing on `main` and is a work in progress. This is the first pass at extending the `si-data-nats` crate with Jetstream functionality.

lib/si-data-nats/src/jetstream/context.rs
```diff
@@ -0,0 +1,109 @@
...
+use bytes::Bytes;
+use telemetry::prelude::*;
+
+use crate::jetstream::{Consumer, JetstreamResult};
+use crate::Client;
+use crate::HeaderName;
+
+/// Used in a [`HeaderMap`] for consumers of stream to reply back to requesters (via a reply mailbox and outside of the
+/// stream).
+pub static REPLY_SUBJECT_HEADER_NAME: HeaderName = HeaderName::from_static("X-Reply-Subject");
```
nickgerace, line 12: When consumers of a stream need to reply, they will likely do so outside of a Jetstream and do it via regular NATS with a generated reply mailbox (which NATS knows how to garbage collect).

lib/si-data-nats/src/jetstream/context.rs
```diff
@@ -0,0 +1,109 @@
...
+        let stream = self
+            .inner
+            .get_or_create_stream(Config {
+                name: name.into(),
+                subjects,
+                max_messages: DEFAULT_MAX_MESSAGES,
+                // TODO(nick): for the rebaser, we must have a work queue retention policy. This is temporary, just for
+                // the first pass. I hope this comment doesn't come back to haunt me.
+                // retention: RetentionPolicy::WorkQueue,
+                ..Default::default()
```
nickgerace, line 45: Referencing the message earlier about work queues, this is the other half: when we get or create the stream, we need to decide it's a work queue. For this PR, we just use the defaults.

lib/si-rabbitmq/Cargo.toml
nickgerace, line 1: The wrapper library is no longer needed.

lib/si-data-nats/src/jetstream/context.rs
```diff
@@ -0,0 +1,136 @@
...
+            ' ' | '.' | '>' | '*' | '/' | '\\' => false,
+            char if char != '-' && !char.is_alphanumeric() => false,
+            _ => true,
+        };
+        if !valid {
+            return Err(JetstreamError::InvalidSubjectName(subject.to_string()));
+        }
+    }
+    Ok(())
+}
```
nickgerace, line 122: Without this, we would run into infinite hangs for tests. Perhaps, that's an upstream issue with the client. The prefix cannot be joined by a '.' character (causes the hang), so we use a '-' character.

However, stream name sanitation isn't the worst idea either. We should likely have a longer-term vision for how we want to centralize and construct NATS subjects across SI.

nickgerace:
New engine tests pass as well as other relevant unit and integration tests. Frontend basic modeling and values updating works too.
```
running 15 tests
test integration_test::internal::new_engine::builtins::builtin_funcs_and_schemas_are_not_empty ... ok
test integration_test::internal::new_engine::prop::prop_path ... ok
test integration_test::internal::new_engine::sdf_mock::list_schema_variant_views ... ok
test integration_test::internal::new_engine::builtins::docker_image_has_one_qualfiication_map_prop ... ok
test integration_test::internal::new_engine::component::create_and_determine_lineage ... ok
test integration_test::internal::new_engine::property_editor::assemble ... ok
test integration_test::internal::new_engine::rebaser::delete_func_node ... ok
test integration_test::internal::new_engine::rebaser::modify_func_node ... ok
test integration_test::internal::new_engine::rebaser::func_node_with_arguments ... ok
test integration_test::internal::new_engine::frame::multiple_frames_with_complex_connections_no_nesting ... ok
test integration_test::internal::new_engine::component::update_and_insert_and_update ... ok
test integration_test::internal::new_engine::frame::convert_component_to_frame_and_attach_no_nesting ... ok
test integration_test::internal::new_engine::component::set_the_universe ... ok
test integration_test::internal::new_engine::component::through_the_wormholes ... ok
test integration_test::internal::new_engine::connection::connect_components has been running for over 60 seconds
test integration_test::internal::new_engine::connection::connect_components ... ok

test result: ok. 15 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 66.22s
```

lib/si-data-nats/src/jetstream/context.rs
nickgerace: Update: filed an issue https://github.com/nats-io/nats.rs/issues/1213

PR #3306, opened 2024-02-16
fix: hardcode the baseurl as it isn't working in CI

johnrwatson:
Compatibility with local runs `must` be retained but without this it isn't passing in CI (even if without it it passes locally against remote) 

Temporarily hardcoding this to give us stability coverage over the weekend + will amend properly config-wise next week when we are up and running.


johnrwatson:
bors merge

PR #2867, opened 2023-10-16
Add systeminit/rabbitmq DockerHub image (ENG-1990)

nickgerace:
## Description

Add si-rabbitmq DockerHub image for ENG-1703 ("Mostly Everything is a Node or an Edge") graph work. This directory is being added to main to get the image building and published in advance to the long running dev branch getting merged.

The new image is not used for anything other than image building, promoting and publishing.

## Reference

The code in this PR was taken from https://github.com/systeminit/si/pull/2424.

PR #3309, opened 2024-02-16
Remove RabbitMQ component (ENG-2333)

nickgerace:
This PR undoes the work from #2867 and is a follow-up to #3300 since the new engine is now using NATS Jetstream.


nickgerace:
bors merge

PR #3311, opened 2024-02-16
feat(si-generator): add codegen and create funcs

adamhjk:
The generator can now produce codeGen and createFuncs.

In theory, there should only need to be 1 codeGen func created, and it should just be re-used across every asset. It should be named 'si:genericAwsCreate', which will then be re-used in the create functions.


adamhjk:
bors merge

PR #3313, opened 2024-02-16
chore(ci): set OpenTelemetry environment varible in integration tests

fnichol:
This change configures the OpenTelemetry collector endpoint for any Cyclone servers running as part of an integration test. Prior to this change (and given past work), Cyclone servers try in vain to connect to a localhost collector only to fail. This causes a *lot* of stderr output in the CI logs, even though this is not a failure at all.


paulocsanz (approved):

stack72:
bors merge

PR #3315, opened 2024-02-19
Add delete and action generators

adamhjk:
Adds a 'delete' and 'action' generator.

Like the refresh generator, they essentially construct the arugments for the cli skeleton from 'input' arguments that specify the AWS Input path on the left, and the property to set it from on the right. Then it calls the API and checks the exit code. 

Delete should work for any AWS API - actions should work for anything except tagging, which is just fraught with peril.


stack72:
bors merge

PR #3316, opened 2024-02-20
feat: replace deadpool with poolnoodle

sprutton1:
This rips out deadpool for spinning up Cyclone instances and replaces the pool logic with PoolNoodle.

## The Problem

1. Deadpool only creates new instances when the pool is empty, so we have no way to warm up instances in advance
2. Deadpool pushes off the recycle task until an instance is requested, so dead instances would linger in the mean time
3. We never want to reuse an instance, so the pooling mechanism of Deadpool is completely redundant

## The Solution

1. Remove Deadpool as the interface for getting Cyclone instances
2. Replace it with a similar implementation using PoolNoodle
  a. The core loop of PoolNoodle is effectively unchanged, except it now looks out for Drop messages from dead instances
  b. Instances requested from PoolNoodle are given back in a wrapper (LifeGuard) that implements Drop so we can handle them going out of scope. On drop, we send a message on a channel that is handled in the core loop, as mentioned above. Deref is also implemented to allow referencing the inner Instance
  c. We now call the native Instance methods, so we are no longer hardcoded to only working with Firecracker
  d. We now always try to fill the entire pool (so pool sizes should be kept lower). When doing so, we start the instance, effectively making our pool warm! Not that this works with both the Tokio and Firecracker runtimes

I also modified the existing tests and added a few more to test the pool mechanics.

I do think that there is an issue with the core loop. It feels weird to be both checking vecs and reading messages from a channel. I definitely think there is a more event-driven approach that can be taken here, but I'm too close to this solution right now to walk it out.


sprutton1:
bors try

sprutton1:
bors try

sprutton1:
bors try-

sprutton1:
bors try

PR #3318, opened 2024-02-20
Fix validations for maps/arrays and remove default validation

paulocsanz:


lib/dal/src/prop.rs
```diff
@@ -724,7 +728,7 @@ impl Prop {
                     handler: "main".to_owned(),
                     component: veritech_client::ResolverFunctionComponent {
                         data: veritech_client::ComponentView {
-                            properties: serde_json::json!({ "format": serde_json::to_value(validation_format)?, "value": value }),
+                            properties: serde_json::json!({ "format": serde_json::to_value(dbg!(validation_format))?, "value": dbg!(value) }),
```
stack72, line 731: did you mean to leave the dbg here?

paulocsanz: Oh I totally forgot it, thank you!

paulocsanz:
bors merge

PR #3319, opened 2024-02-20
 Stop migrating most builtins for tests 

paulocsanz:
6 external tests disabled that basically tested the assets

We should test those assets, but the cypress baked in the package feels a better way.

Now the tests take 5 minutes instead of 20.


paulocsanz:
bors try

paulocsanz:
bors try

paulocsanz:
bors try

paulocsanz:
bors merge

PR #3320, opened 2024-02-20
feat(dal): restore edges on component upgrade

zacharyhamm:
When a component is upgraded to a different schema variant, we need to recreate all the edges (if possible). This attempts to do that, and also updates the component diagram summary row to have the new sockets.


vbustamante (approved):
LGTM

PR #3324, opened 2024-02-21
feat: one size fits all docker compose deployment tool

sprutton1:
This creates a docker compose deployment template that we can use to deploy any of our services. I created an init container that talks to Secrets Manager and Parameter Store to populate our config files with the appropriate values and put all the things in all the right places.

* Move all supporting services into docker-compose
  * Otelcol (traces)
  * NodeExporter (vm metrics)
  * pgbouncer (when applicable)
  * Vector (logs)
* Create an init container that runs first, interrogates AWS for config values, and populates the supporting service + SI service configs
  * This will look for all Secrets/Parameters that start with `$SI_HOSTENV` and set local env vars to `${Secret/Param Description}=${Secret/Param value}`. Therefore, all secrets/params we want to use in our services MUST in clude a hostenv prefix (tools, prod) and the name of the env var to transform in the description (SI_PG_HOST, SI_NATS_CREDS, etc.)
  * Depending on the `$SI_SERVICE` value, it will create a config toml at `/etc/${SI_SERVICE}/${SI_SERVICE}.toml`
* Pull the SI service from the artifact store and run it as a systemd service. The service will read its config from the aforementioned toml file. **NOTE** we may move these into the docker compose assuming we can get the networking bits to play nice with Veritech


component/deploy/docker-compose.yaml
```diff
@@ -0,0 +1,68 @@
+---
+version: "3.8"
+
+services:
+
+  init:
+    container_name: init
+    image: systeminit/init:${SI_VERSION}
+    profiles: [council, pinga, sdf, veritech]
```
nickgerace, line 9: Do we need to handle `web`?

nickgerace (approved):

sprutton1:
bors try

sprutton1:
bors try

sprutton1:
bors merge

sprutton1:
bors merge

PR #3329, opened 2024-02-22
fix: increase timeout permitted to create_node as we know it's slow atm

johnrwatson:
Cypress tests have been failing in Production as overnight the create_node api calls degraded to over 4000ms. This allows our cypress tests to go green again to then allow us to debug further whilst maintaining coverage.

Validaed the timeout fix locally against Production


johnrwatson:
bors merge

PR #3330, opened 2024-02-22
fix: amend other reference timeout

johnrwatson:
Missed a timeout in the delete test which happened to pass when I checked locally


johnrwatson:
bors merge

PR #3331, opened 2024-02-22
feat: save recordings to github artifacts

johnrwatson:
Save the cypress recordings for easy access/allowing you to see what didn't work during any run without having to re-run it to try and understand what went wrong.

**Example here**
https://github.com/systeminit/si/actions/runs/8005441473

See Artifacts: Produced during runtime


johnrwatson:
bors merge

PR #3333, opened 2024-02-22
Restore "force new" for change sets in the new engine

nickgerace:
## Description

This PR restores the "force new" functionality for change sets in the new engine. It uses the same interface through the existing `ChangeSet` struct, but goes through the new `ChangeSetPointer` interface under the hood.

As a result of this change, existing logic attempting to coerce change set pointer data into visibility has been unified to one method.

## Reminder

The `ChangeSetPointer` struct will eventually replace the existing `ChangeSet` struct. After or during replacement, it will likely be renamed to `ChangeSet`. Both will likely happen once the existing `ChangeSet` struct is no longer used and the new engine reaches near feature parity with `main`.


zacharyhamm (approved):
looks like the right way forward to me

PR #3337, opened 2024-02-22
Restore ability to update and list secrets in the new engine

nickgerace:
## Description

This PR restores the ability to update and list secrets. Updating secrets remains the same, but now the referential secrets on the snapshot are updated. Listing secrets uses the same interfaces, but now the graph is purely used and mimics the original logic (this first pass is likely unoptimized as a result).

This is also the first time integration tests have been restored that aren't part of the `new_engine` module! The secrets tests return. We needed to expose `WorkspaceSignup` from the `dal-test` crate as a result, since `WorkspaceSignup` is still in the `dal` on main. The `test_harness` module has been partially restored as well.

## Disclaimer

We need to figure out the long term shape and relationship between referential secrets and encrypted secrets. The `update_secrets` route contains side effect logic for updating the referential secret. This is brittle, but will work for the first pass. However, it is unlikely to work in the long term.

### DIFF
--- a/README_NEW_ENGINE.md
+++ b/README_NEW_ENGINE.md
@@ -27,8 +27,8 @@ This guide provides the minimum steps for preparing changes for the new engine.
 1. `cargo check --all-targets --all-features`
 1. `cargo fmt`
 1. `cargo clippy --all-targets --all-features`
-1. `buck2 run //lib/dal:test-unit -- workspace_snapshot::graph`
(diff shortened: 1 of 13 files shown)


nickgerace:
There are still some UI quirks with this PR, including the `secrets` route not always working. However, there are no regressions and this is a first pass. Merging as a result.

PR #3339, opened 2024-02-22
Optimize internal provider values creation when creating components

paulocsanz:


lib/dal/src/migrations/U2424__optimize_component_creation.sql
```diff
@@ -0,0 +1,365 @@
...
+        PERFORM set_belongs_to_v1(
+            'attribute_value_belongs_to_attribute_prototype',
+            this_tenancy,
+            this_visibility,
+            this_new_attribute_value ->> 'id',
+            this_attribute_prototype.id
+        );
+    END LOOP;
+
+    SELECT array_agg(ip.*)
```
paulocsanz, line 168: The significant changes to this function happened from this line up to the foreach (the other changes were to use the values we stored in the arrays instead of selecting from the database)

lib/dal/src/migrations/U2424__optimize_component_creation.sql
```diff
@@ -0,0 +1,365 @@
...
+    SELECT array_agg(ap.*)
+    INTO STRICT this_attribute_prototypes
+    FROM attribute_prototypes_v1(this_tenancy, this_visibility) AS ap
+    WHERE ap.attribute_context_internal_provider_id = ANY (SELECT id FROM UNNEST(this_internal_providers))
+          OR ap.attribute_context_prop_id = ANY (SELECT prop_id FROM UNNEST(this_internal_providers));
+
+    SELECT array_agg(av.*)
+    INTO STRICT this_attribute_values
+    FROM attribute_values_v1(this_tenancy, this_visibility) AS av
+    WHERE av.attribute_context_prop_id = ANY (SELECT prop_id FROM UNNEST(this_internal_providers));
```
jhelwig, line 184: Was there a specific reason to not use the `belongs_to` relationship between `AttributeValue`s and `AttributePrototype`s? There _shouldn't_ be stray `AttributeValue`s sitting around that might match based on the `attribute_context_prop_id`, but I'm pretty sure there are places that aren't cleaning up after themselves as well as they should be (which means there are probably more that I'm not thinking of).
```suggestion
    SELECT array_agg(av.*)
    INTO STRICT this_attribute_values
    FROM attribute_values_v1(this_tenancy, this_visibility) AS av
        INNER JOIN attribute_value_belongs_to_attribute_prototype_v1(this_tenancy, this_visibility) AS avbtap
            ON avbtap.object_id = av.id
    WHERE avbtap.belongs_to_id = ANY (SELECT id FROM UNNEST(this_attribute_prototypes));
```

paulocsanz: I just copied the query we used inside of the loop to outside of it.

Inside of the loop we used to do this:

```
        SELECT *
        INTO STRICT this_prop_attribute_value
        FROM attribute_values_v1(this_tenancy, this_visibility) AS av
        WHERE in_attribute_context_v1(
            attribute_context_build_from_parts_v1(
                this_internal_provider.prop_id,
                ident_nil_v1(),
                ident_nil_v1(),
                ident_nil_v1()
            ),
            av
        );
```

Since I had the prop ids beforehand I fetched all avs with that prop_id, is my thinking incorrect?

paulocsanz:
bors try

paulocsanz:
bors merge

PR #3340, opened 2024-02-22
feat: run workflow with option for multiple executions, improve cypress execution logic

johnrwatson:
Adds the possibility of driving cypress for test magnitude using `VITE_SI_CYPRESS_MULTIPLIER` to multiply the amount of times the tests are executed on each run to simulate user load. Additionally modified the CI build to be a matrix (of 1) so we can easily multiply on that vector too to give larger scale if needed. i.e.

CI Executor Multiples x VITE_SI_CYPRESS_MULTIPLIER x Test Number = Load

I also added a Posthog command to push a custom UUID into Posthog for each test to allow us to easily look at the user session for each event. This should be super helpful when we get more complicated tests or fragile tests to figure out what is wrong.

### DIFF
--- a/.github/workflows/e2e-validation.yml
+++ b/.github/workflows/e2e-validation.yml
@@ -26,8 +26,8 @@ jobs:
       - name: Setup authentication for vite & other synthetic test variables
         working-directory: app/web
         run: |
-          apt update
-          apt install uuid -y
+          sudo apt update
+          sudo apt install uuid -y
           cp .env .env.local
           echo "VITE_AUTH0_USERNAME=${{ secrets.VITE_AUTH0_USERNAME }}" >> .env.local                           # Production Synthetic User Email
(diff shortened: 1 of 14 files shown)


.github/workflows/e2e-validation.yml
```diff
@@ -4,14 +4,19 @@ on:
   workflow_dispatch:
   push:
     branches:
-      - main
+      - feat/add-more-synthetics
```
stack72, line 7: I don't think we want this right @johnrwatson ?

stack72: Also, ANY push to main will trigger these tests - that doesn't feel right as that code won't have made production yet - should this be limited to any changes to the cypress tests?

johnrwatson:
Looks good:
image

johnrwatson:
bors merge

PR #3342, opened 2024-02-23
Disable PG full page writes in test DB instances

jhelwig:
The PG [docs for the fsync option][1] mention that you should consider turning off `full_page_writes` if you're disabling `fsync`. This doesn't make anywhere near as much of a difference as disabling `fsync` does, but should help some in more I/O bound environments when running tests.

[1]: https://www.postgresql.org/docs/16/runtime-config-wal.html#GUC-FSYNC

### DIFF
--- a/.ci/docker-compose.test-integration.yml
+++ b/.ci/docker-compose.test-integration.yml
@@ -33,6 +33,8 @@ services:
     command:
       - "-c"
       - "fsync=off"
+      - "-c"
+      - "full_page_writes=off"
 
   nats:
     image: systeminit/nats:stable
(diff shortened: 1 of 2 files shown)


jhelwig:
@fnichol Hm. We may want changes to the compose file(s) to trigger all the tests? Doesn't look like the "what should I run" logic things much of anything should run.

stack72:
Closed this in favour of #3407 which has the *exact* same commit - I just rebased it so that I could test the functionality to trigger the entire build suite when the compose files have changed

stack72:
[PS I opened a new PR as I didn't feel right force-pushing to your branch]

PR #3343, opened 2024-02-23
feat(web): Allow navigating to collaborators changeset

stack72:
By clicking on a user in the collaborators window, you will go to their changeset


stack72:
bors merge

PR #3348, opened 2024-02-23
Fix graph unit tests lacking ordinal edges

nickgerace:
## Description

When ordinal edges were added to the system as part of the dependent values update work, three unit tests regressed. This PR returns optional ordinal edge information from `add_ordered_edge` without performing any additional fetch logic. The three tests then use that information to assemble the missing edge.

### DIFF
--- a/lib/dal/src/workspace_snapshot/graph.rs
+++ b/lib/dal/src/workspace_snapshot/graph.rs
@@ -289,7 +289,7 @@ impl WorkspaceSnapshotGraph {
         from_node_index: NodeIndex,
         edge_weight: EdgeWeight,
         to_node_index: NodeIndex,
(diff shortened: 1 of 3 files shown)


lib/dal/src/workspace_snapshot/graph.rs
```diff
@@ -327,9 +327,17 @@ impl WorkspaceSnapshotGraph {
...
                 self.replace_references(container_ordering_node_index)?;
             }
-        }
 
-        Ok(new_edge_index)
+            Some((
+                ordinal_edge_index,
+                container_ordering_node_index,
+                to_node_index,
+            ))
```
nickgerace, line 335: Don't perform additional fetches. Just return what's already there.

PR #3350, opened 2024-02-24
fix: user session must be active to push event into posthog

johnrwatson:
Page must be visited before posthog event being sent. Validated locally against Production to be working, see https://us.posthog.com/project/20369/data-management/events/018dd875-9f8b-0000-3015-bb671a4893c6


johnrwatson:
bors merge

PR #3353, opened 2024-02-26
feat(dal): support workpace imports/exports with multi variants

zacharyhamm:
Exports non-builtin variants for builtin schemas and imports them. Imports the components against the imported new schema variant

### DIFF
--- a/lib/dal/src/migrations/U2610__clear_dependencies_for_deleted_values.sql
+++ b/lib/dal/src/migrations/U2610__clear_dependencies_for_deleted_values.sql
@@ -1,0 +1,24 @@
+CREATE OR REPLACE FUNCTION clear_dependencies_for_deleted_values_v1(
+    this_tenancy jsonb,
+    this_visibility jsonb
+)
+RETURNS VOID
+AS
+$$
+DECLARE
+    dependency_row_id ident;
+BEGIN
+    FOR dependency_row_id IN
+        SELECT avd.id
+        FROM attribute_value_dependencies_v1(this_tenancy, this_visibility) AS avd
+        LEFT JOIN attribute_values_v1(this_tenancy, this_visibility) AS attribute_values_source_join 
+            ON avd.source_attribute_value_id = attribute_values_source_join.id
+        LEFT JOIN attribute_values_v1(this_tenancy, this_visibility) AS attribute_values_dest_join 
+            ON avd.destination_attribute_value_id = attribute_values_dest_join.id
+        WHERE attribute_values_source_join.id IS NULL
+            OR attribute_values_dest_join.id IS NULL
+    LOOP
+        PERFORM delete_by_id_v1('attribute_value_dependencies', this_tenancy, this_visibility, dependency_row_id);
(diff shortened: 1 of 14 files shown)


lib/dal/src/migrations/U2610__clear_dependencies_for_deleted_values.sql
```diff
@@ -0,0 +1,24 @@
+CREATE OR REPLACE FUNCTION clear_dependencies_for_deleted_values_v1(
```
zacharyhamm, line 1: @jhelwig I had to add this to ensure that the dependent values update jobs run (otherwise pinga reports that it can't find some of the attribute values). I believe something like this needs to be added to the dependency update function for components, which I think may not prune deleted attribute values?

jhelwig: It's intended to remove dependencies for both individual attribute values that have been deleted, and for attribute values of entire components that have been deleted, but it's entirely possible that it's not quite right yet.

Whenever anything that would affect the dependencies within (or related to) a component, the dependencies related to that component are recalculated, which should handle removing deleted attribute values from the summary table.

jhelwig: It might be worth modifying the "get the graph" SQL method to have a little more of a safeguard, but I haven't tested to see exactly how it affects the performance.

```
CREATE OR REPLACE FUNCTION attribute_value_dependency_graph_v1(
    this_tenancy                  jsonb,
    this_visibility               jsonb,
    this_attribute_value_id_list  ident[]
)
    RETURNS TABLE
        (
            attribute_value_id           ident,
            dependent_attribute_value_id ident
        )
AS
$$
    WITH RECURSIVE dependency_graph AS (
      SELECT
          destination_attribute_value_id AS attribute_value_id,
          source_attribute_value_id AS dependent_attribute_value_id
      FROM attribute_value_dependencies_v1(this_tenancy, this_visibility) AS avd
      WHERE source_attribute_value_id = ANY(this_attribute_value_id_list)
      UNION
      SELECT
          destination_attribute_value_id AS attribute_value_id,
          source_attribute_value_id AS dependent_attribute_value_id
      FROM dependency_graph
          INNER JOIN attribute_value_dependencies_v1(this_tenancy, this_visibility) AS avd
              ON dependency_graph.attribute_value_id = avd.source_attribute_value_id
    )
    SELECT dependency_graph.* FROM dependency_graph
        INNER JOIN attribute_values_v1(this_tenancy, this_visibility) AS av
            ON dependency_graph.attribute_value_id = av.id
        INNER JOIN attribute_values_v1(this_tenancy, this_visibility) AS av2
            ON dependency_graph.dependent_attribute_value_id = av2.id
$$ LANGUAGE SQL;
```

It _might_ be possible for the additional `INNER JOIN`s at the end to result in multiple, disjoint graphs (and possible overwork), **but** it should guarantee that there are none of the pinga errors about being unable to find attribute values.

PR #3360, opened 2024-02-27
fix/politely wait after attribute change

johnrwatson:
Amend the cypress synthetic to wait until the call to /api/component/update_property_editor_value finishes before navigating away to check the value on the child component.

In the synthetic UI validation workspace something is definitely awry as the attribute updates are taking ~ 30s to finish In another workspace with zero other content/change sets in it, the update_property_editor_value calls take ~5s. We will investigate this further.

NB: What's pretty gnarly about this is that if the user navigates away after making the attribute change (such as refreshing the page) the attribute change is lost in the ether. 

This modifies the synthetic test to politely wait for this API call to finish before navigating away which should put us back to green, but we should definitely investigate what's up in the Synthetic workspace.

</hr>

This also adds a log out in the cypress log for the uuid of the run, so we can more easily identify the runs in posthog.

### DIFF
--- a/app/web/cypress.config.ts
+++ b/app/web/cypress.config.ts
@@ -5,9 +5,19 @@ import vitePreprocessor from 'cypress-vite'
 export default defineConfig({
   e2e: {
     setupNodeEvents(on, config) {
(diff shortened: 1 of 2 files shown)


johnrwatson:
bors merge

PR #3365, opened 2024-02-28
Fix/amend cypress data selector for component card

johnrwatson:
Modifies the selector for the new type of asset cards

Validates that it passes locally.

I've re-seeded the cypress synthetic workspace with the lego components again.


johnrwatson:
bors merge

PR #3366, opened 2024-02-28
feat(cypress): increase frequency and add retry

johnrwatson:
Moves it to every 5 minutes + retries on failure 3 times before accepting failure + finishing.

Also adds a basic Slack webhook back into #_alerts when failure occurs


johnrwatson:
bors merge

PR #3367, opened 2024-02-28
feat(dal,sdf): track whether a component is using the default variant

zacharyhamm:
Changes the component summary table to have a boolean flag indicating whether a component is using the default variant for a schema. (This lets us determine whether we should upgrade the component to the default).


zacharyhamm:
bors try

zacharyhamm:
bors cancel

PR #3371, opened 2024-02-29
Fix/userid

sprutton1:
We switched to Amazon Linux 2023 as our base ami. In some cases, the MAX_UID for the base of that AMI is set to a number lower than the UID we set for the firecracker users. This is reported as a warning, but propagated as an error. This can cause jail preparation to fail. This reduces our UIDs to being under that limit. I tested this locally and an AML 2023 vm.


sprutton1:
bors try

PR #3372, opened 2024-02-29
Avoid redundant work in component creation

paulocsanz:


lib/dal/src/migrations/U2612__avoid_redundant_work_during_component_creation.sql
```diff
@@ -0,0 +1,382 @@
...
+        PERFORM set_belongs_to_v1(
+            'attribute_value_belongs_to_attribute_value',
+            this_tenancy,
+            this_visibility,
+            this_attribute_value_id,
+            this_parent_attribute_value_id
+        );
+    END LOOP;
+
+    -- -- Make sure we've populated the dependency graph for this (new) component.
```
paulocsanz, line 356: This is the biggest change, we remove that block because this does exactly what `attribute_value_dependencies_update_component_v1` does, but without deduplicating.

So now we schedule that work during create so it can be deduplicated. I'm observing Component::new go from 7s locally to 1s, those 6s are exactly the same time `attribute_value_dependencies_update_component_v1` takes. So we can be sure that we were duplicating that processing and now arent.

paulocsanz:
bors merge

PR #3374, opened 2024-02-29
fix: need both uids to match

sprutton1:
UID was still too big and the jailer was not selecting the correct UID, so runs were failing.


sprutton1:
bors try

PR #3378, opened 2024-03-01
Import authentication prototypes and associations

vbustamante:

### DIFF
--- a/lib/dal/src/func/before.rs
+++ b/lib/dal/src/func/before.rs
@@ -50,57 +50,21 @@ pub async fn before_funcs_for_component(
         let secrets_prop =
             SchemaVariant::find_root_child_prop_id(ctx, schema_variant, RootPropChild::Secrets)
                 .await?;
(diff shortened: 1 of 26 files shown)


lib/dal/src/func/before.rs
```diff
@@ -1,55 +1,165 @@
...
+    SchemaVariant(#[from] SchemaVariantError),
+    #[error("secret error: {0}")]
+    Secret(#[from] SecretError),
+    #[error("standard model error: {0}")]
+    StandardModel(#[from] StandardModelError),
 }
 
+pub type BeforeFuncResult<T> = Result<T, BeforeFuncError>;
+
 pub async fn before_funcs_for_component(
```
nickgerace, line 44: For reviewers: we acknowledge there are some algorithmic-ally "bleh" traversal methods in here. However, this was more-or-less a direct port of how `before_funcs` assembly works on `main`, so we can optimize in the future as `new-engine` becomes the new `main`.

PR #3380, opened 2024-03-01
Wendy/eng 2355 hide the add array item button when array is set from an

wendybujalski:
- Add Map/Array Item button only shows when the parent Map/Array is set to be edited manually
- Dropdown menu on any parent prop which has children (maps, arrays, objects) that is set via a socket or attribute function
- The dropdown menu currently only allows you to switch to manually editing that prop and its children, but will have the option to connect a socket or attribute function in the future
- Props which have children and have a parent that is not set manually do not show this dropdown since their parent must be set to manual before they can be set to manual
- Ability to view textbox and codeviewer prop data if the prop is not editable


wendybujalski:
bors merge

PR #3382, opened 2024-03-04
feat(si-layer-cache): adds an experimental layer cache

adamhjk:
This adds an experimental layer cache to the new engine, which will allow us to start rebuilding the underlying write/read path for the new engine.

### DIFF
--- a/lib/si-layer-cache/src/disk_cache.rs
+++ b/lib/si-layer-cache/src/disk_cache.rs
@@ -1,0 +1,65 @@
+use std::path::Path;
+
+use sled::{self, IVec};
+
+use crate::{error::LayerCacheResult, CacheType};
+
+#[derive(Debug)]
+pub struct DiskCache {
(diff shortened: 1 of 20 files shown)


lib/si-layer-cache/src/disk_cache.rs
```diff
@@ -0,0 +1,65 @@
+use std::path::Path;
+
+use sled::{self, IVec};
```
sprutton1, line 3: Do we care that this is beta? Seems like the maintainer is low-key rewriting it prior to a 1.0 release

adamhjk: for now I think ti's fine, only because we don't care about the data. we should move it to rocksdb eventually, but it's going to take some time to get a stable build of the rocksdb rust bindings

adamhjk:
bors merge

PR #3386, opened 2024-03-05
feat(dal): remove direct node/edge weight accesses

zacharyhamm:
In order to support the small graph and access to node and edge weights
through the caching layer, we cannot use the petgraph `node_weight` or
`edge_weight` methods directly in most cases. Instead we need to access
the weights through a method that will look them up in the caching layer
(or fetch it from the content store). This removes all direct calls to
the petgraph `node_weight` and `edge_weight` methods, replacing them
with calls to `get_node_weight` and `get_node_weight_opt`. These methods
are sync right now but will become async. Some other accesses remain
(calls to `edges_directed`, `node_weights()` and calls to
`node_weight_mut`, etc.) that also have to be made indirect.


nickgerace:
Good PR! 🎉

PR #3388, opened 2024-03-05
Rename providers to sockets in the dal (and clean it up)

nickgerace:
## Introduction

REST IN PEACE PROVIDERS! You've served us well back when we first added you in April 2022 (https://github.com/systeminit/si/pull/930). You provided values well to the rest of the system and now you can go the way of the dinosaur.

## Description

This PR renames providers to sockets in the dal and in immediately affected crates. This is not a holistic naming conversion, but a first pass to prevent future conflicts and create a massive PR that is untenable to rebase and review. 

Essentially, only in **active** modules of `lib/dal` were providers renamed to sockets. What does that mean in practice?

- Wherever providers were renamed to sockets, doc comments, code comments and string literals were changed too
- All queries using old sockets or old providers were deleted (though all migrations remain in place since we will handle migrations as `main` stabilizes)
- Inactive modules in `lib/dal/tests` were also excluded from the changes since those tests may need to re-evaluated entirely
- Some exceptions to renaming were made in the `resource` and `pkg/import` modules since contextual comments are currently being used to piece those domains back together (i.e. old code from `old-engine`) that's been commented out means little if we rename providers for the sake of renaming them yet)
- Delete unused unused attribute code (values, prototypes, etc.) since they touched the old provider system and now `DependentValuesUpdate` works in the new engine

## Context

The original "provider" naming made the most sense back in ~Spring 2022 when they signaled that they provided values for others to consume from. Now that implicit `InternalProviders` do not exist in the new engine and that sockets and edges are dead, we can rename providers to "sockets". They are "just sockets" in the new engine and our ubiquitous language between all domains using them should reflect this.

## Disclaimer

This is part one of a two part attempt to perform the refactor. We want to avoid having one massive unreadable PR that creates a ton of conflicts and is impossible to rebase. The follow up PR covering `app/web` and all other places will come soon.


nickgerace:
bors merge

nickgerace:
bors merge

PR #3391, opened 2024-03-06
chore(dal): Reformat the dal tests to account for what is old vs new

stack72:
```
test result: ok. 20 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 38.57s
```


stack72:
bors merge

PR #3392, opened 2024-03-06
chore(build): duplicate Tokio unstable flags in target block

fnichol:
The Tokio unstable flags need to be duplicated in this target-specific section.

> There are four mutually exclusive sources of extra flags. They
> are checked in order, with the first one being used

See: https://doc.rust-lang.org/cargo/reference/config.html#buildrustflags


fnichol:
I have an alternate idea...closing

PR #3393, opened 2024-03-06
chore(dal-test): Reorganise the test schemas to install only in dal-test

stack72:
This stops the dal crate knowing anything about the test specific schemas


stack72:
bors merge

lib/dal-test/src/schemas/mod.rs
```diff
@@ -0,0 +1,7 @@
+mod test_exclusive_schema_bethesda_secret;
+mod test_exclusive_schema_fallout;
+mod test_exclusive_schema_starfield;
+
+pub use test_exclusive_schema_bethesda_secret::migrate_test_exclusive_schema_bethesda_secret;
+pub use test_exclusive_schema_fallout::migrate_test_exclusive_schema_fallout;
+pub use test_exclusive_schema_starfield::migrate_test_exclusive_schema_starfield;
```
nickgerace, line 7: Love the use of private modules with select exports.

PR #3394, opened 2024-03-06
chore(dal): use alternative to JoinSet Id to drop unstable requirement

fnichol:
This change tracks an alternate "task ID" (in this case a random `Ulid`) which is used in the same `task_id_to_av_id` hash map logic, thus removing the need for the unstable task `Id` type. In this way it's a little easier to setup a development environment with editors that use `cargo clippy` and `rust-analyzer` underneath.

### DIFF
--- a/lib/dal/src/job/definition/dependent_values_update.rs
+++ b/lib/dal/src/job/definition/dependent_values_update.rs
@@ -8,6 +8,7 @@ use serde::{Deserialize, Serialize};
 use telemetry::prelude::*;
 use thiserror::Error;
 use tokio::task::{JoinError, JoinSet};
+use ulid::Ulid;
 
 //use crate::tasks::StatusReceiverClient;
 //use crate::tasks::StatusReceiverRequest;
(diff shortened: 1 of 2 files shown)


fnichol:
bors try

lib/dal/src/job/definition/dependent_values_update.rs
```diff
@@ -143,16 +144,19 @@ impl DependentValuesUpdate {
                 let attribute_value_id = attribute_value_id.to_owned(); // release our borrow
 
                 if !seen_ids.contains(&attribute_value_id) {
-                    let join_handle = update_join_set.spawn(
-                        values_from_prototype_function_execution(ctx.clone(), attribute_value_id),
-                    );
-                    task_id_to_av_id.insert(join_handle.id(), attribute_value_id);
+                    let id = Ulid::new();
```
nickgerace, line 147: @fnichol this is a nit, but I'm wondering if it would be valuable to use the same `Ulid` generator for each loop iteration?

fnichol: A generator to make a new `Ulid`? The only real thing I was after was making sure that each task had a unique identifier across all the tasks so we could return it as part of the task response to make it similar to a unique task id. I could have even been a regular old number (like a `u8`) but I didn't want to do any incrementing bookkeeping 😄 I might be misunderstanding though

fnichol:
bors merge

zacharyhamm:
@fnichol what happens here if the task panics?

PR #3397, opened 2024-03-06
feat(si-layer-cache): Make memory cache generic, store deserialized data

zacharyhamm:
Moka has a strange requirement: it's Key and Value types have a `'static` lifetime bound. This makes it tricky to capture it in a generic type, since it means that type's generics also have to have a static lifetime bound. We work around this by using a trait and implementing that trait for all Cache<K, V>. A consequence is that the Layer cache now has to take the concrete Moka cache type as a dependency.


lib/si-layer-cache/benches/insert_speed.rs
nickgerace, line 1: I didn't know there was a bench! Neat.
_(cc @johnrwatson @britmyerss @fnichol @stack72 since we have been talking about `dal` testing)_

lib/si-layer-cache/src/disk_cache.rs
```diff
@@ -1,65 +1,41 @@
...
-    pub object_tree: sled::Tree,
-    pub graph_tree: sled::Tree,
+pub struct DiskCache<K>
+where
+    K: AsRef<[u8]>,
+{
+    tree: sled::Tree,
+    // We have to make it appear that we hold on to a K and a V when we don't
+    // actually do so. This allows us to use static dispatch, etc.
+    _phantom_of_the_opera: PhantomData<K>,
```
nickgerace, line 14: Got it. Name checks out.

lib/si-layer-cache/benches/insert_speed.rs
```diff
@@ -10,12 +11,21 @@ const ASCII_LOWER: [u8; 26] = [
 
 const ONE_MB: usize = 1_000_000;
 
+type BenchmarkCache = Cache<[u8; 1], Vec<u8>>;
+
 pub async fn fresh_cache_count(objects: &[Vec<u8>], count: usize) {
-    let tempdir = tempfile::TempDir::new_in("/home/adam/benches").expect("cannotc reate tempdir");
```
nickgerace, line 14: Oops

lib/si-layer-cache/benches/insert_speed.rs
```diff
@@ -10,12 +11,21 @@ const ASCII_LOWER: [u8; 26] = [
...
-    let layer_cache = LayerCache::new(tempdir).expect("cannot create layer cache");
+    let target_dir = "/tmp";
```
nickgerace, line 17: Feels like we should instead allow `TempDir` to handle the parent since it has OS-specific logic to do so?

zacharyhamm: yes, you're right!

PR #3398, opened 2024-03-06
Add color to rebaser telemetry and close tracker (ENG-2381)

nickgerace:
## Description

Mimic sdf and pinga behavior by adding color to rebaser telemetry config and closing the tracker. We also add args the mimic the pinga args of the same name.


nickgerace:
bors merge

PR #3399, opened 2024-03-06
chore(dal-test): Remove the need for AWS_PKG

stack72:
We have created a new schema - swifty - which is an upframe that we can use to replace the functionality of an AWS Region. This means we create now create a test for multiple components in frames AND nested frames!

This also introduces some schema_helpers for tests for creating action funcs etc. This will make it less verbose to write schemas


lib/dal-test/src/schemas/schema_helpers.rs
nickgerace, line 1: Good call to make this module. Eventually, we could consider scoping this with a unit struct to help folks find functions (e.g. typing `Helpers::` and letting autocomplete do the rest). Example:

```rust
pub(crate) struct Helpers;

impl Helpers {
    pub fn create_identity_func() -> FuncSpec {
        IntrinsicFunc::Identity
            .to_spec()
            .expect("create identity func spec")
    }
}
```

However, this is an opinionated pattern that (I'll be transparent), I am partial to. Not saying we _should_ do it. Moreso _consider_ it for the future.

lib/dal-test/src/schemas/test_exclusive_schema_swifty.rs
```diff
@@ -0,0 +1,132 @@
...
+    ActionFuncSpec, AttrFuncInputSpec, AttrFuncInputSpecKind, PkgSpec, PropSpec, SchemaSpec,
+    SchemaVariantSpec, SchemaVariantSpecData, SiPkg, SocketSpec, SocketSpecData, SocketSpecKind,
+};
+
+pub async fn migrate_test_exclusive_schema_swifty(ctx: &DalContext) -> BuiltinsResult<()> {
+    let mut swifty_builder = PkgSpec::builder();
+
+    swifty_builder
+        .name("swifty")
+        .version("2023-05-23")
```
nickgerace, line 18: nit: we could make this today. I wonder if we could ignore the date too for the future.
cc @zacharyhamm

lib/dal-test/src/schemas/schema_helpers.rs
stack72: 100% - this was just a start so I could re-use some code - this *Will* evolve :)

lib/dal-test/src/schemas/test_exclusive_schema_swifty.rs
stack72: fixed :)

lib/dal/tests/integration_test/frame.rs
nickgerace, line 1: I trust you, so let's keep it if it makes sense to you. However, I'm curious: it looks like the test does a lot less than before. What led to this?

stack72: the test before was connecting components - it wasn't all to do with connecting to frames. Now it's connecting to frames AND it's creating a nested frame

the only potential addition that it could be would be to change the frame that a component is connected to

nickgerace: Gotcha! Makes sense. I love the nested frame addition.

For context since I was the original author, we had a weird bug where frames and components outside of frames would sometimes have an extra edge exist. The test was paranoid about this bug so it included both edges and frames.

That being said, I like your test better and if we run into that bug again, we will deal with it with a new test then.

stack72: I will extend the test to cover reparenting and making sure that the edges connect correctly :) I will have a think about that though

stack72:
bors merge

PR #3403, opened 2024-03-07
feat: enhanced NATS core/Jetstream support

fnichol:
This is work extracted from re-working Council (now essentially defunct). Highlights include:

- A new extracted `si-service` crate that helps set up and build Rust binaries (i.e. crates under `bin/`)
- Better telemetry shutdown semantics
- A small update to `si-data-nats` giving access to the underlying Jetstream APIs (temporary measure)
- A new crate `naxum` which is an approach to writing incoming NATS-based messaging applications, directly modeled from the [axum](https://crates.io/crates/axum) crate (which supports writing HTTP applications). As with axum, naxum is based on the [Tower](https://crates.io/crates/tower) middleware framework allowing us to use and reuse common middlewares such as timeout, load shedding, rate limiting, tracing, etc., etc. Note that while naxum is currently very under-documented, as it's as close to a 1:1 port of axum, axum's docs will suffice for the moment.
- A generic Delay middleware to help when developing applications which introduces artificial delay before servicing a request.

### DIFF
--- a/lib/naxum/src/error_handling.rs
+++ b/lib/naxum/src/error_handling.rs
@@ -1,0 +1,250 @@
+use std::{
+    convert::Infallible,
+    fmt,
+    future::Future,
+    marker::PhantomData,
+    task::{Context, Poll},
+};
+
+use tower::{Layer, Service, ServiceExt};
+
+use crate::{
+    extract::FromMessageHead,
+    response::{IntoResponse, Response},
+    MessageHead,
+};
+
+pub struct HandleErrorLayer<F, T> {
+    f: F,
+    _extractor: PhantomData<fn() -> T>,
+}
+
+impl<F, T> HandleErrorLayer<F, T> {
+    pub fn new(f: F) -> Self {
+        Self {
+            f,
+            _extractor: PhantomData,
+        }
+    }
+}
+
+impl<F, T> Clone for HandleErrorLayer<F, T>
+where
+    F: Clone,
+{
+    fn clone(&self) -> Self {
+        Self {
+            f: self.f.clone(),
+            _extractor: PhantomData,
+        }
+    }
+}
+
+impl<F, T> fmt::Debug for HandleErrorLayer<F, T> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("HandleErrorLayer")
+            .field("f", &format_args!("{}", std::any::type_name::<F>()))
+            .finish()
+    }
+}
+
+impl<S, F, T> Layer<S> for HandleErrorLayer<F, T>
+where
+    F: Clone,
+{
+    type Service = HandleError<S, F, T>;
+
+    fn layer(&self, inner: S) -> Self::Service {
+        HandleError::new(inner, self.f.clone())
+    }
+}
+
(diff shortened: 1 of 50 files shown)


lib/naxum/src/error_handling.rs
```diff
@@ -0,0 +1,250 @@
...
+        let future = Box::pin(async move {
+            // TODO(fnichol): at the moment the response type `Res` is generic, however we know
+            // that by the end we want our response type to be unit (i.e. `()`). Should we shut
+            // down the extra generics or allow the final response types to be dropped here?
+            // Unclear...
+            #[allow(clippy::unit_arg)]
+            match inner.oneshot(req).await {
+                Ok(_res) => Ok(()),
+                Err(err) => Ok(f(err).await.into_response()),
+            }
```
nickgerace, line 138: Let's try it!

lib/naxum/src/error_handling.rs
```diff
@@ -0,0 +1,250 @@
...
+impl_service!(T1, T2, T3, T4, T5, T6, T7);
+impl_service!(T1, T2, T3, T4, T5, T6, T7, T8);
+impl_service!(T1, T2, T3, T4, T5, T6, T7, T8, T9);
+impl_service!(T1, T2, T3, T4, T5, T6, T7, T8, T9, T10);
+impl_service!(T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11);
+impl_service!(T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12);
+impl_service!(T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12, T13);
+impl_service!(T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12, T13, T14);
+impl_service!(T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12, T13, T14, T15);
+impl_service!(T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12, T13, T14, T15, T16);
```
nickgerace, line 219: What the hell is this

lib/si-service/src/shutdown.rs
```diff
@@ -0,0 +1,95 @@
...
+    }
+}
+
+/// Gracfully shutdown a service that may be running multiple tasks and in-flight work.
+///
+/// # Platform-specific behavior
+///
+/// This function sets up a signal handler for both `SIGINT` (i.e. `Ctrl+c`) and `SIGTERM` so usage
+/// of this function with other code intercepting these signals is *highly* discouraged.
+pub async fn graceful<Fut, E>(
```
nickgerace, line 45: Love it.

nickgerace (approved):

fnichol:
bors merge

lib/naxum/src/error_handling.rs
fnichol: Haha, so much fun! That's how axum supports handler functions or extractor functions that take no args, 1 arg, 2 args...they all need to be explicitly created! I used the same approach here which does lead to some funky code

lib/si-service/src/shutdown.rs
fnichol: I predict this'll be used everywhere shortly

PR #3404, opened 2024-03-07
Centralize schema variant views (ENG-2384)

nickgerace:
## Description

Centralize schema variant view collection into the dal. The integration test using them now checks for count, but it may be able to test for more in the future.


lib/dal/tests/integration_test/mod.rs
```diff
@@ -22,5 +22,5 @@ mod frame;
 mod prop;
 mod property_editor;
 mod rebaser;
-mod sdf_mock;
+mod schema_variant_views;
```
nickgerace, line 25: This module and naming likely don't live here long term. Just wanted a place to place the test for now.

nickgerace:
bors merge

PR #3406, opened 2024-03-07
Remove ComponentKind and SchemaUiMenu (ENG-2385)

nickgerace:
## Description

This PR removes `ComponentKind` and `SchemaUiMenu`. It also restores the "schema" integration tests and deletes the "node" integration tests since the old "node" concept does not exist in the new engine.


nickgerace:
bors merge that ish

PR #3408, opened 2024-03-08
feat(si-layer-cache): Add Postgres layer to the layer cache, make writes async

zacharyhamm:
This adds a layer to the cache that sends the writes to a postgres database. All write operations will spawn a tokio task that attempts to idempotently write the key/value pair to a postgres table.

Writes to the fast disk cache are also done on a spawned task.

In practice, we will not have to join on these spawned tasks, since they will be created in a long running service like SDF, but we've implemented a lock-free join for them so that the tests are able to ensure all writes have landed before verifying the writes.


sprutton1:
bors try

sprutton1:
bors try

sprutton1:
bors try

sprutton1:
bors try

sprutton1:
bors try

sprutton1:
bors try

PR #3409, opened 2024-03-08
[wip] feat/add otel to integration tests

johnrwatson:
[Draft]

This PR adds otel to the integration tests but for some reason the traces aren't being picked up by the tracer for otel. This is a non-func change but a very good/close example for us to have in the codebase as to how you might set up tracing for a test.

I've also added a flag to the compiler for the integration tests which allows us to run something like:
`buck2 run lib/dal:test-integration -- integration_test::rebaser --nocapture -Z unstable-options --format json --report-time`

This generates a json report output from the tests which we can then pass to buildkite via the `buildkite-test-collector` crate which this PR also adds.


johnrwatson:
bors try

flake.nix
```diff
@@ -321,6 +321,7 @@
           packages =
             [
               alejandra
+              buildkite-test-collector-rust
```
nickgerace, line 324: We'll get newer version that auto-update after https://github.com/NixOS/nixpkgs/pull/294343 is unblocked and then merged.

lib/dal-test/src/helpers.rs
```diff
@@ -46,6 +49,82 @@ pub async fn workspace_signup(ctx: &DalContext) -> Result<(WorkspaceSignup, Stri
     Ok((nw, auth_token))
 }
 
+pub fn tracing_init(span_events_env_var: &'static str, log_env_var: &'static str) {
```
nickgerace, line 52: Nice!

lib/dal/BUCK
```diff
@@ -106,6 +106,7 @@ rust_test(
     ]),
     env = {
         "CARGO_PKG_NAME": "integration",
+        "RUSTC_BOOTSTRAP": "1",
```
nickgerace, line 109: Bleh but, yeah, we need it.

lib/si-test-macros/src/expand.rs
```diff
@@ -47,7 +44,12 @@ pub(crate) fn expand_test(item: ItemFn, _args: Args, fn_setup: impl FnSetup) ->
         #(#attrs)*
         fn #test_name() -> ::dal_test::Result<()> {
             use ::dal_test::WrapErr;
+            use ::std::io::Write;
+            use ::dal_test::telemetry::tracing;
 
+            // Could add a custom `name` attribute here to send through the test name
+            // SI_TEST_LOG=test_integration::integration_test::rebaser=info,off SI_TEST_LOG_SPAN_EVENTS=new,close buck2 run lib/dal:test-integration -- integration_test::rebaser --nocapture
+            #[tracing::instrument(level = "info", name="example-test-name", skip_all)]
```
nickgerace, line 52: Should we keep `example-test-name` or am I not understanding what this does?

nickgerace (approved):

lib/dal/BUCK
johnrwatson: Agree, I couldn't quite figure out how to add in-line comments within the buck file so I improved the commentary on the commit to include why I added it, hopefully that helps someone in the future

lib/si-test-macros/src/expand.rs
johnrwatson: It's a placeholder for when we figure out how to pipe through the test name into the macro, this is what the otel span has as the identifier attribute. Ideally it would be the test file path/name that is pushing the trace up but because I was not receiving the traces I had no way to actually validate any logic above was correctly setting the name (if that makes sense)

johnrwatson:
This is an example of what the pipeline needs to be adjusted to to pass these metrics up in CI:

```
---
steps:
  - group: ":test_tube: tests"
    key: tests
    steps:

      - agents:
          queue: default
          arch: x86_64
          os: linux
        command: "buck2 run root//lib/dal:test-integration -- --nocapture -Z unstable-options --format json --report-time | buildkite-test-collector"
        env:
          SI_TEST_LOG: debug
          RUST_BACKTRACE: full
          PASS_THROUGH_ENV_VARS: "CI BUILDKITE_ANALYTICS_TOKEN BUILDKITE_BRANCH BUILDKITE_MESSAGE BUILDKITE_BUILD_URL BUILDKITE_COMMIT BUILDKITE_JOB_ID"
        label: ":blue_book: :linux: test-integration (root//lib/dal:test-integration)"
        plugins:
          - "docker-compose#v4.14.0":
              run: app
              config: ".ci/docker-compose.test-integration.yml"
              mount-checkout: true
              env-propagation-list: "PASS_THROUGH_ENV_VARS"
              volumes:
                - /workdir/buck-out
                - /workdir/target
                - /workdir/tmp
        timeout_in_minutes: 125
        priority: 5
 
 ```

johnrwatson:
bors merge

johnrwatson:
bors merge

johnrwatson:
Previous failure was due to github outage/flakiness

PR #3411, opened 2024-03-08
feat(si-layer-cache): impl NATS chunking messaging

fnichol:
This change adds a new `ChunkingNats` wrapping type around an underlying
NATS Jetstream messages `futures::Stream`. The publish side takes an
arbitrarily sized `payload` of `Bytes` and internally may send multiple
NATS Jetstream messages with the payload segemented into chunks.

On the receiving side, there is a `futures::Stream` implementation which
will reassemble the chunks into single messages. Any messages that are
not marking with chunking header metadata will pass through this stream
unmodified.

Testing and integration work remains but the basics of the behavior are
there!

Co-authored-by: Adam Jacob <adam@systeminit.com>
Signed-off-by: Fletcher Nichol <fletcher@systeminit.com>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5583,18 +5583,24 @@ name = "si-layer-cache"
 version = "0.1.0"
 dependencies = [
  "async-trait",
+ "blake3",
  "buck2-resources",
+ "bytes 1.5.0",
  "criterion",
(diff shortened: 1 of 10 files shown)


fnichol:
bors merge

PR #3413, opened 2024-03-08
feat(dal, sdf-server): Restore list_variant_defs SDF Route

stack72:
Screenshot 2024-03-08 at 21 28 22

There are a number of things to node here:

1. We have no current way of understanding what the default variant is for a schema variant - so this assumes the first
2. We don't have a SchemaVariantDefinitionId so we are using the schema variant id for now
3. The only different between a schema variant and a schema variant definition is the missing input / output sockets so we may have an easier way to get that data rather than a separate route
4. It adds a basic test to ensure that the number of schemas == number of schema variant definitons

This is going to need more work later when we start to add defaults for schema_variants

### DIFF
--- a/lib/dal/src/schema/variant.rs
+++ b/lib/dal/src/schema/variant.rs
@@ -34,9 +34,9 @@ use crate::workspace_snapshot::WorkspaceSnapshotError;
 use crate::{
     pk,
     schema::variant::leaves::{LeafInput, LeafInputLocation, LeafKind},
-    AttributePrototype, AttributePrototypeId, ComponentId, DalContext, Func, FuncId, InputSocket,
-    OutputSocket, OutputSocketId, Prop, PropId, PropKind, Schema, SchemaError, SchemaId, Timestamp,
-    TransactionsError,
+    AttributePrototype, AttributePrototypeId, ComponentId, ComponentType, DalContext, Func, FuncId,
+    InputSocket, OutputSocket, OutputSocketId, Prop, PropId, PropKind, Schema, SchemaError,
+    SchemaId, Timestamp, TransactionsError,
 };
 use crate::{FuncBackendResponseType, InputSocketId};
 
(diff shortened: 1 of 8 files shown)


stack72:
bors try

lib/dal/src/schema/variant.rs
```diff
@@ -301,6 +303,36 @@ impl SchemaVariant {
...
     }
 
+    pub async fn get_default_for_schema(
+        ctx: &DalContext,
+        schema_id: SchemaId,
+    ) -> SchemaVariantResult<Self> {
+        // The first pass here is that we will just return the first schema variant found for a schema
+        // We will need to add a default edge at some point in the future - that means when we import
+        // a schema variant (if it doesn't already exist) then it gets an edge for default
+        // If we are creating a new version of a schema variant then we will be able to mark it as default
```
nickgerace, line 313: We discussed this and this is good for now. We'll need to change it sooner rather than later though. Probably you get an additional default edge when creating the first `SchemaVariant` that you can change.

lib/dal/src/schema/variant/definition.rs
```diff
@@ -762,3 +764,54 @@ pub struct SiPropValueFrom {
...
+                timestamp: default_schema_variant.timestamp.to_owned(),
+                component_type: default_schema_variant
+                    .get_type(ctx)
+                    .await?
+                    .unwrap_or(ComponentType::Component),
+                link: default_schema_variant.link.to_owned(),
+                // TODO: Schema Variants need a description
+                description: None,
+                // TODO: What is the difference between a category and a menu name?
+                menu_name: None,
```
nickgerace, line 811: I'm guessing nothing, but we'll clean this up after I get through `NodeAddMenu` removal.

lib/dal/tests/integration_test/schema/variant.rs
```diff
@@ -1,3 +1,5 @@
+mod view;
```
nickgerace, line 1: Yikes! Looks like I missed this 🤦‍♂️

lib/dal/tests/integration_test/schema/variant/view.rs
```diff
@@ -20,3 +21,21 @@ async fn list_schema_variant_views(ctx: &DalContext) {
...
+    // We are going to ensure that we get a default schema variant
+    // for each schema variant
+    let views = SchemaVariantDefinitionView::list(ctx)
+        .await
+        .expect("could not list schema variant views");
+
+    assert_eq!(
+        schema_variant_ids.len(), // expected
+        views.len()               // actual
+    )
```
nickgerace, line 40: I think this will fail if a `Schema` has more than one variant

stack72: Which is EXACTLY what we want :) It's wju I added it - because there always has to be a default for a variant :)

stack72:
bors merge

PR #3416, opened 2024-03-09
Re-organize schema views and remove node add menu (ENG-2388)

nickgerace:
## Description

This PR re-organizes `SchemaVariantViews` into `SchemaViews`. `SchemaViews` contain `SchemaVariantViews`, which is a more accurate depiction of the underlying system and prevents confusion. To satisfy use cases in the frontend, new getters and types have been added that extend and organize the raw views from the backend. It should be much easier to avoid conflating `"schemas"` with `"schemaVariants"` in the frontend now.

As a result of the aforementioned changes, the route name has changed as well as affected tests and modules.

Additionally, this commit removes `nodeAddMenu` and references to the old `Node` concept that have lingered. Since "category" is now on the `SchemaVariant `and is non-optional, we no longer need the abstraction layer.

## Additional Changes

- Add per-package benchmark when installing builtins in sdf (rounded to two decimal places upon display)
- Reduce logging when creating a `SchemaVariant` when installing builtins in sdf (`info --> debug`)
- Add additional metadata to reduced logging
- Remove unused `default_schema_variant_id` field on `Schema` since it should be an edge in the future


nickgerace:
bors merge

PR #3417, opened 2024-03-11
Fixed edge connections in backend and frontend

keeb:
Fixed connection annotations so that it converts to lowercase on both the server and the front end before selecting which edges can connect. Also added appropriate unit test to cover this case.


keeb:
bors merge

keeb:
bors merge

PR #3418, opened 2024-03-11
Reimplement actions skeleton

paulocsanz:
Actions don't run yet, but they are enqueued/dequeued, batches are created and they get processed on apply.

It doesn't have tests yet, that's also a next step in a subsequent PR.

Also fixes a bunch of misc bugs that we found along the way, that's why it's being merged even if not feature complete.

Co-Authored-By: nick@systeminit.com


.gitignore
```diff
@@ -17,6 +17,7 @@ deploy/docker-compose.env.yml
 app/web/cypress/screenshots/
 app/web/cypress/videos/
 bin/si/src/version.txt
+prelude
```
nickgerace, line 20: We may update the prelude, so we don't want to ignore it.

nickgerace:
@paulocsanz approved but I think we need to ignore the `.gitignore` change.

.gitignore
paulocsanz: It doesnt ignore if its already pushed

nickgerace: Nevermind. It already exists in the repo, so this is fine.

nickgerace: We commented at the same time lol

paulocsanz:
bors merge

PR #3419, opened 2024-03-12
feat(sdf-server, dal): Revive the get_code endpoint

stack72:
This returns the get_code endpoint for the codegen funcs on a component


stack72:
I will follow up here and add more tests to cover the different scenarios that we need the CodeView to have - different code formats etc

lib/dal-test/src/schemas/schema_helpers.rs
```diff
@@ -60,6 +60,33 @@ pub(crate) async fn build_action_func(
...
+                .code_plaintext(code)
+                .handler("main")
+                .backend_kind(FuncSpecBackendKind::JsAttribute)
+                .response_type(FuncSpecBackendResponseType::CodeGeneration)
+                .build()?,
+        )
+        .build()?;
+
+    Ok(func)
+}
```
nickgerace, line 88: Nice!

lib/dal-test/src/schemas/test_exclusive_schema_swifty.rs
```diff
@@ -43,6 +45,16 @@ pub async fn migrate_test_exclusive_schema_swifty(ctx: &DalContext) -> BuiltinsR
     // Author Resource Payload Func
     let resource_payload_to_value_func = build_resource_payload_to_value_func().await?;
 
+    // Build CodeGen Func
+    let codegen_fn_name = "test:generateCode";
+    let codegen_func_code = "async function main(input: Input): Promise < Output > {
```
nickgerace, line 50: We'll need to make sure the handler is always `main` for your function, but I don't think we need to add any additional checks or anything.

lib/dal-test/src/schemas/test_exclusive_schema_swifty.rs
```diff
@@ -104,6 +116,13 @@ pub async fn migrate_test_exclusive_schema_swifty(ctx: &DalContext) -> BuiltinsR
                         .func_unique_id(&refresh_action_func.unique_id)
                         .build()?,
                 )
+                .leaf_function(
+                    LeafFunctionSpec::builder()
+                        .func_unique_id(codegen_fn_name)
+                        .leaf_kind(LeafKind::CodeGeneration)
+                        .inputs(vec![LeafInputLocation::Domain])
+                        .build()?,
+                )
```
nickgerace, line 125: Easy to forget these. Good catch.

lib/dal/src/code_view.rs
```diff
@@ -49,18 +63,58 @@ pub struct CodeView {
...
+        };
+
+        let code_gen_entry: CodeGenerationEntry = serde_json::from_value(func_execution)?;
+        if code_gen_entry.code.is_none() || code_gen_entry.format.is_none() {
+            return Ok(None);
+        }
+
+        // Safe unwraps because of the above check
+        let format = code_gen_entry.format.as_ref().unwrap();
+        let code = code_gen_entry.code.as_ref().unwrap();
```
nickgerace, line 97: @stack72 I'd recommend using a match statement here to get that behavior.

```rust
let (code, format) = match (code_gen_entry.code, code_gen_entry.format) {
  (None, None) => return Ok(None),
  (Some, Some) => // do stuff,
  (None, None) => return Err(// new error),
};

File: lib/dal/tests/integration_test/component.rs [hunk 5]
```diff
@@ -499,3 +501,49 @@ async fn set_type(ctx: &mut DalContext) {
...
+
+    let (codegen_view, has_code) = Component::list_code_generated(ctx, component.id())
+        .await
+        .expect("unable to get codegen views");
+
+    assert_eq!(codegen_view.len(), 1);
+    assert!(has_code, "true");
+
+    // This is safe as we would have failed the above test otherwise
+    let codegen = codegen_view.clone().pop().unwrap();
```
nickgerace, line 524: Generally, we would want an expect here, but it's not a blocker.

lib/dal/src/code_view.rs
stack72: I can circle back on this and make this better - it's the same logic that existed before so I didn't change it

stack72:
bors merge

PR #3422, opened 2024-03-12
Avoid infinite loops in actions listing

paulocsanz:


app/web/src/store/actions.store.ts
```diff
@@ -136,7 +136,9 @@ export const useActionsStore = () => {
             // probably want to clean up and avoid the while loop if possible too
             const graph = _.cloneDeep(this.rawProposedActionsById);
             const actions = [];
+            const count = 0;
             while (_.keys(graph).length) {
+              if (count > 1000) throw new Error("infinite loop when flattening actions");
```
nickgerace, line 141: I think we should use the original `_.keys(graph).length` value instead.

nickgerace: I'm going back and forth on this. 1000 actions in one list would be impossible to sift through too so we would probably re-write the component anyway... hmm...

paulocsanz: I think we can deal with that problem as it happens, we have that limit in fix execution too, so it will break them

paulocsanz:
bors merge

PR #3423, opened 2024-03-12
chore: refresh NATS support

fnichol:
This change adds 2 new tools to our development environment (a NATS CLI `nats` and a top-like program `nats-top`), drops some of the default overly-verbose logging, and bumps the version of the NATS server.

### DIFF
--- a/component/nats/BUCK
+++ b/component/nats/BUCK
@@ -9,7 +9,7 @@ docker_image(
         "nats-server.conf": "component/nats",
     },
     build_args = {
-        "BASE_VERSION": "2.10.10",
+        "BASE_VERSION": "2.10.11",
     },
     run_docker_args = [
         "--publish",
--- a/component/nats/BUCK
+++ b/component/nats/BUCK
@@ -18,6 +18,5 @@ docker_image(
     run_container_args = [
         "--config",
         "nats-server.conf",
-        "-DVV",
     ],
 )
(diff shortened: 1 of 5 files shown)


component/nats/BUCK
```diff
@@ -9,15 +9,16 @@ docker_image(
...
     },
     build_args = {
-        "BASE_VERSION": "2.10.10",
+        "BASE_VERSION": "2.10.11",
     },
     run_docker_args = [
         "--publish",
         "4222:4222",
+        "--publish",
+        "8222:8222",
```
nickgerace, line 18: Probably good to have it!

flake.nix
```diff
@@ -324,6 +324,8 @@
               buildkite-test-collector-rust
               docker-compose
               jq
+              nats-top
+              natscli
```
nickgerace, line 328: Should be able to test more directly now!

nickgerace (approved):

component/nats/BUCK
fnichol: Yeah, it will get used by `nats-top` which looks like a freakin' useful one. In prod it's a tossup whether that would be on or off, but great in dev either way

flake.nix
fnichol: I feel like I've been hoarding the NATS CLI to myself--no longer!

fnichol:
bors merge

PR #3424, opened 2024-03-12
feat(sdf-server, dal): Adding an edge for default schema variants at pkg import time

stack72:
Schema / Schema Variant Edges now exist as follows:

DEFAULT: Schema -> Uses -> Schema Variant

In order to make a schema variant the default for a schema, we need to update the
correct edge from Use to Default:

UPDATE: Schema -> Default -> Schema Variant

Therefore, when we are setting a default schema variant we need to find any existing
Default edges and convert them back to uses AND we need to find the existing Use edge
between our nodes and change that to be a Default

### DIFF
--- a/lib/dal/src/action/batch.rs
+++ b/lib/dal/src/action/batch.rs
@@ -168,7 +168,7 @@ impl ActionBatch {
         workspace_snapshot
             .add_edge(
                 category_id,
-                EdgeWeight::new(change_set, EdgeWeightKind::Use)?,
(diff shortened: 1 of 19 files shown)


stack72:
bors try

stack72:
bors try

zacharyhamm:
@stack72 pushed up a change: the `Use` edge is so common that we should probably have a simpler name for the method that creates it. But also we need to include the fact that the edge is default in the merkle tree hash calculation. Looks great to me though!

stack72:
bors merge

stack72:
bors merge

PR #3428, opened 2024-03-12
Remove references and usages of "providers" in app/web

nickgerace:
This commit removes references and usages of "providers" in app/web everywhere except for function authoring. That's because we need further changes to fully restore the attribute prototype parts of the function authoring and attachment experience.


nickgerace:
bors merge

PR #3429, opened 2024-03-12
feat(dal,sdf) delete and create array/map items from prop editor

vbustamante:
CLOSES ENG-2393
CLOSES ENG-2394
CLOSES ENG-2395
CLOSES ENG-2396

### DIFF
--- a/lib/dal/src/attribute/value.rs
+++ b/lib/dal/src/attribute/value.rs
@@ -159,6 +159,8 @@ pub enum AttributeValueError {
     PropMoreThanOneChild(PropId),
     #[error("prop not found for attribute value: {0}")]
     PropNotFound(AttributeValueId),
+    #[error("trying to delete av that's not related to child of map or array: {0}")]
+    RemovingWhenNotChildOrMapOrArray(AttributeValueId),
     #[error("serde_json: {0}")]
     SerdeJson(#[from] serde_json::Error),
     #[error("store error: {0}")]
@@ -1943,5 +1945,38 @@ impl AttributeValue {
             Some(pk) => Some(FuncExecution::get_by_pk(ctx, &pk).await?),
             None => None,
         })
+    }
+
+    pub async fn remove_by_id(ctx: &DalContext, id: AttributeValueId) -> AttributeValueResult<()> {
+        let parent_av_id =
+            if let Some(parent_av_id) = AttributeValue::parent_attribute_value_id(ctx, id).await? {
+                let prop_id = AttributeValue::prop(ctx, parent_av_id).await?;
+
+                let parent_prop = Prop::get_by_id(ctx, prop_id).await?;
+
+                if !vec![PropKind::Map, PropKind::Array].contains(&parent_prop.kind) {
+                    return Err(AttributeValueError::RemovingWhenNotChildOrMapOrArray(id));
+                }
+
+                parent_av_id
+            } else {
+                return Err(AttributeValueError::RemovingWhenNotChildOrMapOrArray(id));
+            };
+
+        let av = Self::get_by_id(ctx, id).await?;
+
+        ctx.workspace_snapshot()?
+            .write()
+            .await
+            .remove_node_by_id(ctx.change_set_pointer()?, av.id)?;
+
+        ctx.enqueue_job(DependentValuesUpdate::new(
+            ctx.access_builder(),
+            *ctx.visibility(),
+            vec![parent_av_id],
+        ))
+        .await?;
+
+        Ok(())
     }
 }
--- a/lib/dal/src/attribute/value.rs
+++ b/lib/dal/src/attribute/value.rs
@@ -1947,21 +1947,54 @@ impl AttributeValue {
         })
     }
 
+    pub async fn get_parent_av_id_for_ordered_child(
+        ctx: &DalContext,
+        id: AttributeValueId,
+    ) -> AttributeValueResult<Option<AttributeValueId>> {
+        let workspace_snapshot = ctx.workspace_snapshot()?.read().await;
+
+        let ordering_node_id = match workspace_snapshot
+            .incoming_sources_for_edge_weight_kind(id, EdgeWeightKindDiscriminants::Ordinal)?
+            .first()
+            .copied()
+        {
+            Some(ordering_idx) => workspace_snapshot.get_node_weight(ordering_idx)?.id(),
+            None => return Ok(None),
+        };
+
+        let parent_av_id = if let Some(parent_av_idx) = workspace_snapshot
+            .incoming_sources_for_edge_weight_kind(
+                ordering_node_id,
+                EdgeWeightKindDiscriminants::Ordering,
+            )?
+            .first()
+            .copied()
+        {
+            let parent_av_id: AttributeValueId = workspace_snapshot
+                .get_node_weight(parent_av_idx)?
+                .id()
+                .into();
+
+            let prop_id = AttributeValue::prop(ctx, parent_av_id).await?;
+
+            let parent_prop = Prop::get_by_id(ctx, prop_id).await?;
+
+            if !vec![PropKind::Map, PropKind::Array].contains(&parent_prop.kind) {
+                return Ok(None);
+            }
+
+            parent_av_id
+        } else {
+            return Ok(None);
+        };
+
+        Ok(Some(parent_av_id))
+    }
+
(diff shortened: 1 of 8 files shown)


lib/dal/src/attribute/value.rs
```diff
@@ -1944,4 +1946,72 @@ impl AttributeValue {
             None => None,
         })
     }
+
+    pub async fn get_parent_av_id_for_ordered_child(
```
nickgerace, line 1950: Awesome!

lib/dal/src/attribute/value.rs
```diff
@@ -1944,4 +1946,72 @@ impl AttributeValue {
...
+
+        let ordering_node_id = match workspace_snapshot
+            .incoming_sources_for_edge_weight_kind(id, EdgeWeightKindDiscriminants::Ordinal)
+            .await?
+            .first()
+            .copied()
+        {
+            Some(ordering_idx) => workspace_snapshot.get_node_weight(ordering_idx).await?.id(),
+            None => return Ok(None),
+        };
```
nickgerace, line 1964: Probably `WorkspaceSnapshotGraph` or `WorkspaceSnapshot` should have this as its own method at some point.

lib/dal/src/attribute/value.rs
```diff
@@ -1944,4 +1946,72 @@ impl AttributeValue {
...
+                .get_node_weight(parent_av_idx)
+                .await?
+                .id()
+                .into();
+
+            let prop_id = AttributeValue::prop(ctx, parent_av_id).await?;
+
+            let parent_prop = Prop::get_by_id(ctx, prop_id).await?;
+
+            if !vec![PropKind::Map, PropKind::Array].contains(&parent_prop.kind) {
```
nickgerace, line 1985: I think you might be able to use a static array here since those enum variants are stack-allocated? This is a nit, so feel free to ignore.

lib/dal/src/attribute/value.rs
```diff
@@ -1944,4 +1946,72 @@ impl AttributeValue {
...
+        ctx.workspace_snapshot()?
+            .remove_node_by_id(ctx.change_set_pointer()?, av.id)
+            .await?;
+
+        ctx.enqueue_job(DependentValuesUpdate::new(
+            ctx.access_builder(),
+            *ctx.visibility(),
+            vec![parent_av_id],
+        ))
+        .await?;
```
nickgerace, line 2013: En queuing here now, great!

lib/dal/src/property_editor/values.rs
```diff
@@ -66,28 +68,68 @@ impl PropertyEditorValues {
...
+                        }
+                        _ => None,
+                    }
+                } else {
+                    None
+                };
+
+                // Ideally every attribute value with children is connected via an ordering node
+                // We don't error out on ordering not existing here because we don't have that
+                // guarantee. If that becomes a certainty we should fail on maybe_ordering==None.
```
nickgerace, line 109: Got it

lib/sdf-server/src/server/service/component/delete_property_editor_value.rs
```diff
@@ -26,74 +23,9 @@ pub async fn delete_property_editor_value(
...
-        }
-    }
-
-    if *parent_prop.kind() == PropKind::Map {
-        av.delete_by_id(&ctx).await?;
-    }
-
-    ctx.enqueue_dependent_values_update(vec![*parent_av.id()])
-        .await?;
+    AttributeValue::remove_by_id(&ctx, request.attribute_value_id).await?;
```
nickgerace, line 28: Thin routes!

nickgerace (approved):
~~Looks good to me, but we need an integration test for this. Perhaps, it could assemble the property editor values before and after a removal and we could perform some assertions based on that.~~

EDIT: hit the wrong button.

nickgerace (changes requested):
Looks good to me, but we need an integration test for this. Perhaps, it could assemble the property editor values before and after a removal and we could perform some assertions based on that.

lib/dal/tests/integration_test/property_editor.rs
```diff
@@ -32,3 +33,261 @@ async fn assemble(ctx: &DalContext) {
         .expect("could not assemble property editor schema");
     dbg!(property_editor_schema, property_editor_values);
 }
+
+#[test]
+async fn array_map_manipulation(ctx: &DalContext) {
+    // Create a component using the testing schema
+    let component = create_component_for_schema_name(ctx, "pirate", "ss poopcanoe").await;
```
nickgerace, line 40: Argh! :pirate_flag:

lib/dal/tests/integration_test/property_editor.rs
```diff
@@ -32,3 +33,261 @@ async fn assemble(ctx: &DalContext) {
...
+    AttributeValue::insert(ctx, parrot_names_value_id, Some("samantha".into()), None)
+        .await
+        .expect("two items in array");
+    AttributeValue::insert(ctx, parrot_names_value_id, Some("jessica".into()), None)
+        .await
+        .expect("three items in array");
+    AttributeValue::insert(ctx, parrot_names_value_id, Some("amanda".into()), None)
+        .await
+        .expect("four items in array");
+    AttributeValue::insert(ctx, parrot_names_value_id, Some("dr wiggles".into()), None)
```
nickgerace, line 72: Very fun

lib/dal/tests/integration_test/property_editor.rs
```diff
@@ -32,3 +33,261 @@ async fn assemble(ctx: &DalContext) {
...
+    AttributeValue::insert(
+        ctx,
+        treasure_map_value_id,
+        Some("baby back ribs".into()),
+        Some("jupiter".to_string()),
+    )
+    .await
+    .expect("five items in map");
+
+    // Grab the children for the array and check that they match what they should be
```
nickgerace, line 118: Normally, we would want to call `ctx.blocking_commit()` before performing assertions on anything we have built. That way, we can let dependent values update run and the rebaser work.

_That being said_, since this test only inserts and removes attribute values and doesn't update them, I believe this should be fine.

lib/dal/tests/integration_test/property_editor.rs
```diff
@@ -32,3 +33,261 @@ async fn assemble(ctx: &DalContext) {
...
+    // ======================================================
+
+    // Remove an item from the array prop
+    AttributeValue::remove_by_id(ctx, parrot_names_third_item.id())
+        .await
+        .expect("remove the third item in parrot_names array");
+    let parrot_names_child_ids =
+        AttributeValue::get_child_av_ids_for_ordered_parent(ctx, parrot_names_value_id)
+            .await
+            .expect("get the vec of child ids");
```
nickgerace, line 185: I'm wondering if we should instead use `PropertyEditorValues` since that is what the frontend will ask for.

Example: https://github.com/systeminit/si/blob/e18ec5484872bb5a63e13d326e32ab0f66410817/lib/dal/tests/integration_test/before_funcs.rs#L79-L84

The benefit to this: we won't be using lower level functions and will be reflecting more of SDF's common behavior. I don't think this is worth blocking the test over, but something to consider.

nickgerace (approved):
Awesome test @wendybujalski :fire:

lib/dal/tests/integration_test/property_editor.rs
vbustamante: Yeah I kept that in mind, but  sifting through the assembled object that sdf returns would be a world of pain, so for this interation I think this will suffice.

At some point I wanna add testing functions that parse both attribute values and the attribute form fields themselves into something more tree like, so that writing assertions for this data would be easier

nickgerace: > Yeah I kept that in mind, but sifting through the assembled object that sdf returns would be a world of pain, so for this interation I think this will suffice.

It'll be a pain, but we will want it. I don't think we can avoid it. Plus, there are a few helper functions on the struct already to help out.

That being said, I can see why it's not worth it for this iteration.
I do think we need to consider it as a "next time" thing though.

> At some point I wanna add testing functions that parse both attribute values and the attribute form fields themselves into something more tree like, so that writing assertions for this data would be easier

I think `materialized_view` may be what you are looking for?

lib/dal/src/attribute/value.rs
vbustamante: good catch! The linter wouldn't let us merge without fixing  this too. We changed it

vbustamante:
bors merge

PR #3433, opened 2024-03-13
Restore force change set and do not assume HEAD is nil (ENG-2400)

nickgerace:
## Description

This commit restores "force change set" functionality into the system, but it required removing and replacing assumptions about what "HEAD visibility" means... all over the system.

In the old engine, you could assume the HEAD visibility because it was a `nilId` for every tenant, workspace, etc. In the new engine, workspaces should have different default change set ids (essentially "HEAD"). Therefore, we cannot assume HEAD will be the `nilId` and we need to derive it from the workspace. This commit follows the aforementioned design goal by adjusting how builtin workspace creation works and how we derive HEAD everywhere in all critical paths of the codebase.

There's more work to be done, like completely switching over to the new `ChangeSetPointer` (to be renamed `ChangeSet`) domain and refactoring remaining visibility mutation locations, but this first pass should set us up for the future.

## Disclaimer

As mentioned above, there are many more locations to clean up. Moreover, we may want to investigate making `headChangeSetId` become non-nullable in the frontend. Perhaps, we get it from the workspace.

There's all the question of naming: head, base, default... lot of names for kinds of `ChangeSets`. Fortunately, the ubiquitous language is solid. We just need to be careful here to not conflate names. This will come as we refactor and polish the domains.

## Backend Changes

- Create a new `ChangeSetPointer` using a generated `Ulid` rather than `NONE` for the builtin workspace's default change set id
- Set up the builtin workspace internally rather than returning it
- Ensure `force_new` uses the default change set id for the workspace to determine HEAD
- Rename `update_visibility` functions to reflect real world use
- Add new `standard_id` module with an `id!` macro
- Use the new `id!` macro to remove `NONE` from `ChangeSetPointerId`
- Drop pk from `ChangeSetPointer`
- Move updating change set status to `Applied` inside `apply_to_base_change_set`
- Consolidate open change set views in backend to one location
- Add head change set id to open change set views return type
- Remove `NONE` assumptions from `DalContext` for `ChangeSetPointerId` (use default change set id instead)
- Add integration test for open change sets view

## Frontend Changes

- Derive the head change set from the backend route
- Use `headChangeSetId` returned from the `list_open_change_sets` route
- Replace `nilId` usages and HEAD assumptions with the `headChangeSetId`
- Drop pk from `ChangeSet` and update usages to use the id
- Remove duplicate `ChangeSetId` definitions

### DIFF
--- a/app/web/src/store/actions.store.ts
+++ b/app/web/src/store/actions.store.ts
@@ -4,7 +4,6 @@ import { addStoreHooks, ApiRequest } from "@si/vue-lib/pinia";
 import { trackEvent } from "@/utils/tracking";
 import { Resource } from "@/api/sdf/dal/resource";
 import { useWorkspacesStore } from "@/store/workspaces.store";
-import { nilId } from "@/utils/nilId";
 import { useChangeSetsStore } from "./change_sets.store";
 import { ComponentId } from "./components.store";
 import { useRealtimeStore } from "./realtime/realtime.store";
@@ -204,7 +203,8 @@ export const useActionsStore = () => {
         },
         actions: {
           async FETCH_QUEUED_ACTIONS() {
-            if (changeSetId === nilId()) return ApiRequest.noop;
+            if (changeSetId === changeSetsStore.headChangeSetId)
+              return ApiRequest.noop;
             return new ApiRequest<{
               actions: Record<ActionId, ProposedAction>;
             }>({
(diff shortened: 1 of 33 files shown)


app/web/src/store/actions.store.ts
```diff
@@ -258,25 +258,29 @@ export const useActionsStore = () => {
...
-                    ),
-                )?.id;
-              },
-            });
+            // FIXME(nick): we have a few options here to make this better. One idea is to ensure the headChangeSetId
+            // can never be null by getting it from the tenant. Another idea is to not send down a changeSetId at all
+            // and just rely on the backend to find the default change set id for the workspace. There are a few more
+            // options, but we can make this better.
+            if (!changeSetsStore.headChangeSetId) {
+              changeSetsStore.FETCH_CHANGE_SETS();
```
paulocsanz, line 266: What happens here? After we get the change-sets shouldnt we fetch the actions? otherwise they wont be fetched at all?

nickgerace:
@paulocsanz addressed your comment and @stack72 fixed the modal that pops up when you create a change set on head.

nickgerace:
bors merge

PR #3440, opened 2024-03-13
feat(dal, sdf-server): Restore Abandon ChangeSet Route

stack72:


lib/sdf-server/src/server/service/change_set/abandon_change_set.rs
```diff
@@ -43,22 +48,20 @@ pub async fn abandon_change_set(
...
-        .await?
-        .publish_on_commit(&ctx)
-        .await?;
+    // WsEvent::change_set_abandoned(&ctx, change_set.changeset_pk(), Some(user.pk()))
+    //     .await?
+    //     .publish_on_commit(&ctx)
+    //     .await?;
 
-    ctx.commit().await?;
+    ctx.commit_no_rebase().await?;
```
stack72, line 64: I expect this to be the area I may have not chosen correctly @nickgerace

nickgerace: Jokes aside, yes that's the right idea :smile:! You don't want to rebase because the `WorkspaceSnapshot` has not changed at all.

stack72:
bors merge

PR #3443, opened 2024-03-14
feat(layer-cache): writes are gossiped

adamhjk:
This implements the read and write side of the network cache update protocol. As things are written to disk, we also publish them to a nats jetstream topic, which all instances of LayerDB are listening to. When they get a write, they automatically place it in their memory and disk caches, making sure that data is hot in memory when possible.

As a side effect, we have changed the lower level interface of the caches so that their keys are always strings. This helps with the internal implementation, but will also help with debugging in the future (it wouldn't be super fun to look at a database full of raw bytes for primary keys). It simplified the code a bunch, and doesn't seem to have had any meaningful impact on performance.

Test coverage for the gossiping of writes is included in the integration tests for the casdb implementation.


adamhjk:
bors merge

adamhjk:
bors merge

PR #3446, opened 2024-03-14
Destroy ChangeSet struct (ENG-2405)

nickgerace:
## Description

This commit removes the `ChangeSet` struct, moves the "force new" functionality to `ChangeSetPointer` and removes relevant queries that use `ChangeSet`.

## Disclaimer

This PR does not remove migrations, does not remove `ChangeSetPk`, nor does it rename `ChangeSetPointer` to `ChangeSet`. This is the first step towards fully removing the old `ChangeSet` concept and it should help remove boilerplate for those working on the new engine.


nickgerace:
bors merge

PR #3447, opened 2024-03-14
Let you (intend to) delete things on the diagram

jhelwig:
This is more of a "mark things as having an intention to delete them", since we need the component(s) to still exist for Actions to be able to run against them, and have the Actions remove the thing that is being modeled from the outside world.


jhelwig:
bors merge

PR #3448, opened 2024-03-14
feat(layer-cache): add get_bulk method to layer-cache

sprutton1:
This can be used to get many values from the cache at once. Values that fall through to database queries will be batched into a single query.

### DIFF
--- a/lib/si-layer-cache/src/pg.rs
+++ b/lib/si-layer-cache/src/pg.rs
@@ -1,3 +1,4 @@
+use std::collections::HashMap;
 use std::{marker::PhantomData, sync::Arc};
 
 use si_data_pg::{PgPool, PgPoolConfig};
@@ -29,6 +30,7 @@ where
     pool: Arc<PgPool>,
     pub table_name: String,
     get_value_query: String,
+    get_value_many_query: String,
     insert_value_query: String,
     contains_key_query: String,
     search_query: String,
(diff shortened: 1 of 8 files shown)


lib/si-layer-cache/src/pg.rs
```diff
@@ -37,13 +39,14 @@ where
 
 impl<K> PgLayer<K>
 where
-    K: AsRef<[u8]> + Copy + Send + Sync,
+    K: AsRef<[u8]> + Copy + Send + Sync + std::hash::Hash + std::cmp::Eq,
```
zacharyhamm, line 42: Did you end up needing these extra trait bounds? I don't expect we'll ever have a key that won't be hashable or support equality, but I think these were for the FromSql stuff?

sprutton1: Both are required to insert into the hashmap on line 90 in `pg.rs`. Should I propagate the traits bounds to the other places we define k?

nickgerace (approved):

sprutton1:
bors try

lib/si-layer-cache/src/pg.rs
adamhjk: You won't need them once our code merges, since K will always be an &str or String.

sprutton1:
bors merge

PR #3449, opened 2024-03-15
feat(layer-cache): get_many returns to the PgLayer

sprutton1:
Now that the values in pg are strings this is much more straightforward to implement.


sprutton1:
bors try

PR #3454, opened 2024-03-15
feat(dal,web) show if value is set by dynamic function on prop editor backend

vbustamante:
This PR also adds commit_and_update_snapshot and connect_components_with_socket_names to the test harness, to diminish verbosity on our tests


vbustamante:
Still missing tests, but functional

lib/dal/src/attribute/value.rs
```diff
@@ -1431,6 +1434,13 @@ impl AttributeValue {
         })
     }
 
+    pub async fn is_overridden_for_id(
+        ctx: &DalContext,
+        av_id: AttributeValueId,
+    ) -> AttributeValueResult<bool> {
+        Ok(Self::component_prototype_id(ctx, av_id).await?.is_some())
+    }
```
nickgerace, line 1442: Makes sense to me

lib/dal/src/component.rs
```diff
@@ -1084,17 +1090,98 @@ impl Component {
...
+            let func_id = AttributePrototype::func_id(ctx, prototype_id).await?;
+            let func = Func::get_by_id(ctx, func_id).await?;
+
+            let this_tuple = (func_id, av_id, func.is_dynamic());
+
+            // if av has a parent and parent is controlled by dynamic func (tuple.2), that's the controller
+            // else av controls itself
+            let controlling_tuple = if let Some(parent_av_id) = maybe_parent_av_id {
+                // We can unwrap because if we're on the child, we've added the parent to result
+                let parent_tuple = result.get(&parent_av_id).unwrap().clone();
```
nickgerace, line 1127: I don't know that we want an explicit panic.

lib/dal/src/workspace_snapshot/graph.rs
```diff
@@ -1030,7 +1029,11 @@ impl WorkspaceSnapshotGraph {
             },
         );
         let filename_no_extension = format!("{}-{}", Ulid::new().to_string(), suffix);
-        let mut file = File::create(format!("/home/zacharyhamm/{filename_no_extension}.txt"))
+
+        let home_str = std::env::var("HOME").unwrap();
```
nickgerace, line 1033: We should probably not panic here.

lib/dal/src/workspace_snapshot/graph.rs
```diff
@@ -1030,7 +1029,11 @@ impl WorkspaceSnapshotGraph {
...
+        let home = std::path::Path::new(&home_str);
+
+        let mut file = File::create(home.join(format!("{filename_no_extension}.txt")))
```
nickgerace, line 1036: I don't remember the purpose of the "no extension" thing... I vaguely remember it was due to copy/pasting terminal output, but I think it's fine if you collapsed these together. This is a nit though. Feel free to ignore.

nickgerace (approved):
Other than some explicit panics, looks good to me!

lib/dal/src/workspace_snapshot/graph.rs
jhelwig: Given that this bit is debugging code, the `unwrap()` is probably fine.

lib/dal/src/component.rs
jhelwig: This should be fine. If we manage to not find something with the `get`, then on a previous iteration we somehow failed to insert into `result` at `1172`, while still finding results with the `av_queue.extend` at `1174`, and if that manages to happen, a `panic` is probably the right thing to do as we can't trust anything that's happening.

https://github.com/systeminit/si/pull/3454/files#diff-27851497f20cc332daa611bfe78589d554fbe07377859df57ce7dd53f0e999ddR1172-R1180

nickgerace: Gotcha!

nickgerace (changes requested):
I apologize for the switch up, but flipping my approval for the PR here. I thought about it a bit more, and this could regress easily, so we should add test(s) now.

Still, awesome changes! Do not hesitate to reach out if you want to talk testing.

vbustamante:
@nickgerace Yeah this probably should have been a draft. I intended to add test before merging it, but since the feature was complete I put it up for review

vbustamante:
bors merge

PR #3455, opened 2024-03-18
feat(si-layer): basic activities support

adamhjk:
This adds basic activities into layerdb. Activities represent events that we expect someone else to react to. The first two examples are a quick port of the rebaser events.

```rust
let stream = layerdb.activity_subscribe(Some([vec![ActivityPayloadDiscriminates::RebaseFinished])).await?;

layerdb.activity_publish(&rebase_request)?;
layerdb.activity_publish(&rebase_finished)?;
```

The resulting stream would only receive the second activity event. If you want to subscribe to all events, pass 'None' in as the argument to activity_subscribe.


nickgerace (approved):

adamhjk:
bors merge

PR #3458, opened 2024-03-19
feat(*): move workspace snapshots into layerdb

zacharyhamm:
Moves storage and fetch of the workspace snapshot into its own layer cache and database (WorkspaceSnapshotDb). This enables us to use an Arc without extra copying when performing read-only graph operations.

Some changes were required to the blocking_commit and commit internals to ensure we perform a rebase even if no transaction has been opened in the Dal Postgres transactions state.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4445,6 +4445,7 @@ dependencies = [
  "serde",
  "serde_json",
  "si-data-nats",
+ "si-events",
  "telemetry",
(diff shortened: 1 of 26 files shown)


zacharyhamm:
bors try

zacharyhamm:
bors try

zacharyhamm:
bors try

PR #3460, opened 2024-03-19
feat(sdf, dal): Bring back get_diff endpoint

stack72:
Bring back the sdf endpoint for get_diff.

~~**NOTE**: I don't believe this is quite right to merge yet - I still need to test the way through the system where a change happens in a new changeset and we see a diff from HEAD~~


stack72:
bors try

stack72:
bors merge

PR #3462, opened 2024-03-20
Improved Visualizer

jobelenus:
There are 3 different kinds of things you can view, all rendered as a graph.

1. The default view contains all the components you've placed on the diagram and all the nodes in the system they are connected to. We start at the components and walk the graph from there.
2. The next option contains everything. We walk from every node in the workspace, which includes assets and builtins that are not related to components. Components in the diagram are also included here.
3. The rest of the options are all the builtins themselves. We take a single schema variant and walk only its related nodes. This will never contain a component from the diagram.

I've included a handy refresh button. If you're viewing components, and change the diagram, you can hit that button to reload the page. You can search for specific nodes based on the text of their label. You can hover and click on nodes to focus on just that node and its neighbors. This is a way to focus on a piece of a large graph. And you can stop or start the animation, technically the layout algorithm is continuously optimizing the graph and moving nodes... which can make hovering or clicking on items challenging.

### DIFF
--- a/app/web/src/components/Workspace/WorkspaceVizSchemaVariant.vue
+++ b/app/web/src/components/Workspace/WorkspaceVizSchemaVariant.vue
@@ -114,7 +114,7 @@ let renderer: Sigma;
 function buildClickedNeighbors() {
   state.clickedNeighbors.clear();
   state.clickedNodes.forEach((c: string) => {
-    for (const n in graph.neighbors(c)) {
+    for (const n of graph.neighbors(c)) {
       state.clickedNeighbors.add(n);
     }
   });
@@ -155,7 +155,6 @@ onMounted(async () => {
     await vizStore.LOAD_DATA(schemaVariant.value);
     loading.value = false;
 
-    console.log("COMPONENT NODES", vizStore.nodes);
     if (vizStore.edges.length === 0 && vizStore.nodes.length === 0) {
       return;
     }
--- a/app/web/src/components/Workspace/WorkspaceVizSchemaVariant.vue
+++ b/app/web/src/components/Workspace/WorkspaceVizSchemaVariant.vue
@@ -31,10 +31,12 @@ const props = defineProps<{ schemaVariantId?: string }>();
 
 const getColor = (nodeKind: string, contentKind: string | null) => {
   const kindMap: { [key: string]: string } = {
-    Category: "#2FA",
+    Category: "#c200ff",
     Func: "#F02",
     Ordering: "#ABC",
     Prop: "#0FF",
+    AttributePrototypeArgument: "#103016",
+    FuncArgument: "#40300c",
   };
 
   const contentKindMap: { [key: string]: string } = {
--- a/app/web/src/components/Workspace/WorkspaceVizSchemaVariant.vue
+++ b/app/web/src/components/Workspace/WorkspaceVizSchemaVariant.vue
@@ -7,7 +7,7 @@
         type="dropdown"
         class="flex-1"
         :options="schemaVariantOptions"
-        placeholder="Entire Workspace"
+        placeholder="Dropped Components"
         placeholderSelectable
       />
       <VormInput
@@ -173,7 +173,11 @@ onMounted(async () => {
     }
 
     loading.value = true;
-    await vizStore.LOAD_DATA(schemaVariant.value);
+    if (schemaVariant.value) {
+      await vizStore.LOAD_VARIANTS(schemaVariant.value);
+    } else {
+      await vizStore.LOAD_COMPONENTS();
+    }
     loading.value = false;
 
     if (vizStore.edges.length === 0 && vizStore.nodes.length === 0) {
--- a/app/web/src/components/Workspace/WorkspaceVizSchemaVariant.vue
+++ b/app/web/src/components/Workspace/WorkspaceVizSchemaVariant.vue
@@ -65,6 +65,7 @@ onMounted(async () => {
   const nodes = nodesAndEdges.result.data.nodes;
   const edges = nodesAndEdges.result.data.edges;
 
+
   for (const node of nodes) {
     graph.addNode(node.id, {
       color: getColor(node.nodeKind, node.contentKind),
@@ -90,6 +91,8 @@ onMounted(async () => {
 
   const container = document.getElementById("vizDiv") as HTMLElement;
   // tslint:disable-next-line
-  const _sigma = new Sigma(graph, container);
+  const _sigma = new Sigma(graph, container, {
+    allowInvalidContainer: true,
+  });
 });
 </script>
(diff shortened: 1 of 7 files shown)


lib/sdf-server/src/server/service/graphviz.rs
```diff
@@ -263,46 +389,63 @@ pub async fn nodes_edges(
...
+                    inner
+                        .targets()
+                        .map(|targets| format!(
+                            "\nsource: {}\nto: {}",
+                            targets.source_component_id, targets.destination_component_id
+                        ))
+                        .unwrap_or("".to_string())
+                )),
+                NodeWeight::FuncArgument(inner) => Some(inner.name().to_owned()),
                 _ => None,
```
nickgerace, line 415: I suspect we will eventually want all of them right? Should those come in this PR or later?

lib/sdf-server/src/server/service/graphviz.rs
```diff
@@ -246,10 +283,99 @@ pub async fn schema_variant(
...
+    let mut added_edges = HashSet::new();
+
+    let components = Component::list(&ctx).await?;
+
+    for component in &components {
+        if !added_nodes.contains(&component.id()) {
+            added_nodes.insert(component.id());
+        } else {
+            continue;
+        }
```
nickgerace, line 312: I don't believe it is possible to see two components twice in the way that this is written.

lib/sdf-server/src/server/service/graphviz.rs
```diff
@@ -246,10 +283,99 @@ pub async fn schema_variant(
...
+
+        let node_weight = workspace_snapshot.get_node_weight_by_id(component.id()).await?;
+        let node  = GraphVizNode {
+            id: node_weight.id(),
+            content_kind: node_weight.content_address_discriminants(),
+            node_kind: node_weight.into(),
+            name: Some(component.name(&ctx).await?.to_owned()),
+        };
```
nickgerace, line 320: Okay I dig it.

lib/sdf-server/src/server/service/graphviz.rs
```diff
@@ -246,10 +283,99 @@ pub async fn schema_variant(
...
+                    .await?;
+                    rel_edges.iter().enumerate().for_each(|(_, edge)| {
+                        edges.push(GraphVizEdge {
+                            from: id,
+                            to: target.id(),
+                            edge_weight_kind: edge.kind().into(),
+                            direction: Direction::Outgoing,
+                        });
+                    });
+                }
```
nickgerace, line 342: Hmm... this is only dealing with `Components` right now, but I'm wondering if you could make it generic for _everything_ in the graph. It would likely be a simpler algorithm too (e.g. go through every single edge in the graph and send them up as views).

nickgerace (approved):

lib/sdf-server/src/server/service/graphviz.rs
jobelenus: Thats what our first hunch was as well. Unless somehow there was a circular relationship between two components?

lib/sdf-server/src/server/service/graphviz.rs
jobelenus: Do you mean like the `node_edges` endpoint?

app/web/src/components/layout/navbar/WorkspaceSettingsMenu.vue
```diff
@@ -29,20 +35,43 @@
...
   const currentWorkspace = workspacesStore.urlSelectedWorkspaceId;
   if (!currentWorkspace) return;
   window.open(`${AUTH_PORTAL_URL}/workspace/${currentWorkspace}`, "_blank");
 };
+
+const showViz = () => {
+  if (!changeSetStore.selectedChangeSetId) return false;
+  if (!featureFlagsStore.FEAT_GRAPHVIZ) return false;
+  return true;
+};
```
vbustamante, line 66: If this is not created as a `computed` field it will not have reactivity, so it would not respond to a selected changeset change, for example 

You could do something like:
```suggestion
const showViz = computed(() => {
  if (!changeSetStore.selectedChangeSetId) return false;
  if (!featureFlagsStore.FEAT_GRAPHVIZ) return false;
  return true;
});
```
and just use `showViz` instead of calling the function when refering to it

app/web/src/components/Workspace/WorkspaceVizSchemaVariant.vue
```diff
@@ -43,53 +98,265 @@ const getColor = (nodeKind: string, contentKind: string | null) => {
...
+
+async function loadData() {
+  loading.value = true;
+  if (schemaVariant.value) {
+    const variant = schemaVariant.value !== "all" ? schemaVariant.value : null;
+    await vizStore.LOAD_VARIANTS(variant);
+  } else {
+    await vizStore.LOAD_COMPONENTS();
   }
+  loading.value = false;
```
vbustamante, line 173: Our actions have automatic request status reporting, so if you call 

```
const load_variants_status = vizStore.getRequestStatus("LOAD_VARIANTS");
```

you will get back an `ApiRequestStatus` reactive object with data including the loading status (`.isPending`). You'd need to create a new computed to get status for both requests at once in a single variable, but it's a helpful pattern.

I don't know how well it plays with the vizStore structure that's in this PR, since the functions are not exported as actions (and I won't say we shouldn't merge the  PR without changing that) but it's a tradeoff that should be analyzed

lib/sdf-server/src/server/service/graphviz.rs
vbustamante: @nickgerace It starts with a list of the components in the diagram. If we don't do this filtering here, depending on the order that the components vec is in, we could add the component because it has an incoming connection from another component, and then try to add the component again on its own loop. We saw this behavior while running the system

vbustamante (approved):
This looks very helpful! Added some considerations, but the PR looks good to go

lib/sdf-server/src/server/service/graphviz.rs
vbustamante: This should probably move to the NodeWeight enum in the future too. We discussed it but decided against doing it in this PR

app/web/src/components/Workspace/WorkspaceVizSchemaVariant.vue
jobelenus: Oh right, b/c the onSuccess is a callback itself... thats just something that this pattern makes it hard to reason about

app/web/src/components/layout/navbar/WorkspaceSettingsMenu.vue
jobelenus: Good catch

jobelenus:
bors merge

PR #3466, opened 2024-03-22
chore: impl graceful shutdown & update activity streams

fnichol:
This is a combination of refactoring work, broken out into several commits (details in the commit messages)

## chore: implement graceful shutdown for layerdb

This change updates the shutdown logic for the running task components
of `LayerDb` to better support a graceful shutdown, such as when a
`SIGTERM` or `SIGINT` is sent to a running service.

There are 2 primary fixes that this work resolves:

- When the `PersisterTask` receives a shutdown signal (via a
  `CancellationToken`), it will close the `Receiver` side of its unbound
  channel which prevents any additional incoming persist requests.
  Once the reciever is closed, all remaining channel values are read and
  processed ensuring the channel is fully drained. A `TaskTracker` is
  used internally to `PersisterTask` which allows the task to track all
  spawned persist tasks and these tasks' completions will be awaited
  before the task has fully shut down. In this way we attempt to "flush"
  any remaining persisting operations before a process shuts down.
- When the `CacheUpdatesTask` receives a shutdown signal (again using a
  `CancellationToken`), the internal `ChunkedMessagesStream` will detect
  the cancellation and return a `Poll::Ready(None)` which closes the
  messages stream. This way the main run loop will find that its stream
  has closed and will cleanly return. Without this change, the incoming
  message stream will block indefinitely and this task will never
  gracefully complete.

In addition to the above changes, a graceful shutdown `Future` is
returned when a `LayerDb` is initialized. This future should be spawned
by the caller on its own `TaskTracker` which it should wait on when
shutting down. This mechanism provides all the `LayerDb` tasks their
opportunity to perform internal graceful shutdowns and will ultimately
result in the `LayerDbGracefulShutdown` future to resolve.

## feat(si-layer-cache): activity subsciptions are impl Streams

This change refactors the API for subscribing to activity streams.
Primarily, these streams are now types that implement `future::Stream`
and yeild `AckActivity` types which are deserialized `Activity` types
with a NATS Jetstream `Acker` type.


fnichol:
bors try

fnichol:
bors merge

PR #3467, opened 2024-03-22
feat(dal,sdf): unset overriden value

vbustamante:
Allow user to revert manually setting a value by detaching the av from its direct prototype, making the av be set by the default prototype for its prop


vbustamante:
bors merge

PR #3468, opened 2024-03-22
Update Components to reflect (un)deletion of Components

jhelwig:
# Update Component inputs/outputs to reflect when a Component is (un)marked as to_delete.

Data should be able to flow between components following the rules below:

| From Component `to_delete` | To Component `to_delete` | Data Propagates |
| -------------------------- | ------------------------ | --------------- |
| False                      | False                    | Yes             |
| False                      | True                     | Yes             |
| True                       | False                    | No              |
| True                       | True                     | Yes             |

Whenever a Component is marked (or un-marked) as `to_delete`, we need to make sure that both its inputs, and all downstream Components' inputs are updated to reflect the rules above. This allows us to both show the effect of "deleting" a Component on the rest of the workspace, and keep data that may be necessary to perform the delete action in Components that are marked for deletion.

# Enable delete_connection API endpoint

Removing an AttributePrototypeArgument will now cause the AttributeValue(s) that use that Prototype to automatically be updated to reflect the new function inputs.


app/web/src/store/components.store.ts
```diff
@@ -1214,7 +1195,7 @@ export const useComponentsStore = (forceChangeSetId?: ChangeSetId) => {
 
             return new ApiRequest({
               method: "post",
-              url: "diagram/restore_components",
+              url: "diagram/remove_delete_intent",
```
nickgerace, line 1198: Nice

lib/dal/tests/integration_test/component.rs
```diff
@@ -484,3 +484,425 @@ async fn set_the_universe(ctx: &mut DalContext) {
 
     assert_eq!(universe_json, materialized_view);
 }
+
+#[test]
+async fn deletion_updates_downstream_components(ctx: &mut DalContext) {
```
nickgerace, line 489: Nice!

nickgerace (approved):

jhelwig:
bors merge

PR #3470, opened 2024-03-22
feat(si-layer-cache): impl rebaser requests work queue stream

fnichol:
This is an initial implementation of a work queue stream of rebaser requests, suitable for use by the rebaser service.

Note that it's not yet clear if additional work queue streams follow this pattern, if the code is extractable into the service-related crates, or if some other code de-duplication is possible. At the moment, the newly added Rust types are doing a *lot* of heavy lifting to ensure correctness of behavior and type safety.

Also note that this NATS Jetstream setup may require further testing and refinement to confirm the desired stream sourcing behavior.

### DIFF
--- a/lib/si-layer-cache/src/activities.rs
+++ b/lib/si-layer-cache/src/activities.rs
@@ -283,3 +283,107 @@ impl Stream for ActivityStream {
         }
     }
 }
+
+#[derive(Clone)]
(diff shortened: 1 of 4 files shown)


fnichol:
bors try

adamhjk:
bors merge

PR #3471, opened 2024-03-25
chore(layer-cache): add tests for the work queue

adamhjk:
Adds a test that proves that we wind up with a load balanced work queue, by confirming that we don't get duplicate messages to the stream.


adamhjk:
bors merge

PR #3473, opened 2024-03-26
Downgrade log level for only "onto" and "to_rebase" edges (ENG-2420)

nickgerace:
This commit downgrades the log level for only "onto" and "to_rebase" edges when detecting conflicts and updates. Too noisy!


nickgerace:
bors merge

PR #3475, opened 2024-03-26
Reimplement attributes view for component debug screen

britmyerss:
This PR brings back the Debug view in the Attribute Panel, with some cleanups along the way: 

- Created `debug` modules for the various Node types on the graph: Socket, Attribute Prototype, Attribute Value, Component
- Moved logic out of SDF and into the DAL
- Moved the debug components into a Debug folder in app/web
- Fixed up the interfaces a bit there for this data
- Created a `AttributeValuePath` struct to consolidate path creation for attribute values for props and sockets (including the key/index for child attribute values of maps/arrays)
- Includes input/output socket and attribute value debug data!
- One realllllllllly basic integration test, but this should be expanded


britmyerss:
bors try

britmyerss:
bors try

britmyerss:
bors cancel

britmyerss:
bors try

app/web/src/components/AttributeDebugView.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <div class="overflow-hidden my-xs p-xs border-opacity-10 border-l-2">
     <dl class="flex flex-col gap-xs">
-      <DebugViewItem title="Attribute Value Id" :data="data.valueId" />
+      <DebugViewItem title="Attribute Value Id" :data="data.attributeValueId" />
```
nickgerace, line 4: I like this change

app/web/src/components/AttributeDebugView.vue
nickgerace, line 1: Question about the name of this file. Is it an `AttributeValue` view, an `AttributePrototype` view, both or something else?

We're overloading the usage of the word "Attribute" across the system to mean subtly different things and should tighten it up. However, the whole world is unlikely in scope for this PR.

lib/dal/src/attribute/prototype.rs
```diff
@@ -33,6 +33,7 @@ use crate::{
 };
 
 pub mod argument;
+pub mod debug;
```
nickgerace, line 36: I like the module!

lib/dal/src/attribute/prototype/debug.rs
```diff
@@ -0,0 +1,207 @@
...
+use crate::socket::input::InputSocketError;
+use crate::socket::output::OutputSocketError;
+use crate::workspace_snapshot::node_weight::NodeWeightError;
+use crate::workspace_snapshot::WorkspaceSnapshotError;
+use crate::{
+    AttributePrototype, AttributePrototypeId, AttributeValue, AttributeValueId, Component,
+    ComponentError, DalContext, Func, FuncError, FuncId, InputSocket, Prop,
+};
+use serde_json::Value;
+use telemetry::prelude::*;
```
nickgerace, line 29: We should clean up these imports. Let's put all `use <non-crate>` statements at the top, then a blank line, then all `use crate` statements.

Some folks split `use std` statements in their files too, but we usually group them together (unclear if by habit or intention, but that's a topic for another day).

lib/dal/src/attribute/prototype/debug.rs
```diff
@@ -0,0 +1,207 @@
...
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct AttributePrototypeDebugView {
```
nickgerace, line 33: I think this answers my other question.

lib/dal/src/attribute/prototype/debug.rs
```diff
@@ -0,0 +1,207 @@
...
+    pub func_id: FuncId,
+    pub func_execution: Option<FuncExecution>,
+    pub prototype_id: AttributePrototypeId,
+    pub func_name: String,
+    pub func_args: HashMap<String, Vec<serde_json::Value>>,
+    pub arg_sources: HashMap<String, Option<String>>,
+    pub all_attribute_values: Vec<AttributeValueId>,
```
nickgerace, line 40: nit: I think "all" is implied

lib/dal/src/attribute/prototype/debug.rs
```diff
@@ -0,0 +1,207 @@
...
+};
+use serde_json::Value;
+use telemetry::prelude::*;
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct AttributePrototypeDebugView {
+    pub func_id: FuncId,
+    pub func_execution: Option<FuncExecution>,
+    pub prototype_id: AttributePrototypeId,
```
nickgerace, line 36: nit: Sometimes, we will just say `id` by convention if the struct corresponds to the id provided (e.g. `Component` has a field `id` rather than `component_id` and this is usually true for "views" as well).

lib/dal/src/attribute/prototype/debug.rs
```diff
@@ -0,0 +1,207 @@
...
+    NodeWeightError(#[from] NodeWeightError),
+    #[error(transparent)]
+    OutputSocketError(#[from] OutputSocketError),
+    #[error(transparent)]
+    PropError(#[from] PropError),
+    #[error(transparent)]
+    ValueSourceError(#[from] ValueSourceError),
+    #[error(transparent)]
+    WorkspaceSnapshotError(#[from] WorkspaceSnapshotError),
+}
```
nickgerace, line 72: I would not use `transparent` here since we do not want to forward the underlying error. We should probably add shapes like the following:

```rust
#[error("name of error lower case with spaces error: {0}")]
```

Relevant doc: https://docs.rs/thiserror/latest/thiserror/

lib/dal/src/attribute/prototype/debug.rs
```diff
@@ -0,0 +1,207 @@
...
+impl AttributePrototypeDebugView {
```
nickgerace, line 73: I'd leave a blank line between the struct and the impl

lib/dal/src/attribute/prototype/debug.rs
```diff
@@ -0,0 +1,207 @@
...
+    #[instrument(level = "debug", skip_all)]
+    pub async fn new_from_attribute_value(
```
nickgerace, line 75: nit: The word `new` may imply some persistence. That's why we will often use `assemble` or `list`, depending on the object. That being said, Rust convention is that methods with the name `new` often return `Self`. I think if you renamed this to `new` and kept the parameters the same, that would work too.

lib/dal/src/attribute/prototype/debug.rs
```diff
@@ -0,0 +1,207 @@
...
+        let destination_component_id =
+            AttributeValue::component_id(ctx, attribute_value_id).await?;
+        let apa_ids = AttributePrototypeArgument::list_ids_for_prototype(ctx, prototype_id).await?;
+        let func_id = AttributePrototype::func_id(ctx, prototype_id).await?;
+        let all_attribute_values =
+            AttributePrototype::attribute_value_ids(ctx, prototype_id).await?;
+        let mut func_binding_args: HashMap<String, Vec<serde_json::Value>> = HashMap::new();
+        let mut arg_sources: HashMap<String, Option<String>> = HashMap::new();
+
+        let func_name = Func::get_by_id(ctx, func_id).await?.name;
```
nickgerace, line 89: nit: the order and flow of these could be changed for readability (e.g. group methods that use `prototype_id` separately from those that use `attribute_value_id`. Maybe separate those groups with new lines and maybe comments).

lib/dal/src/attribute/prototype/debug.rs
```diff
@@ -0,0 +1,207 @@
...
+        for apa_id in apa_ids {
+            let apa = AttributePrototypeArgument::get_by_id(ctx, apa_id).await?;
+            let expected_source_component_id = apa
+                .targets()
+                .map(|targets| targets.source_component_id)
+                .unwrap_or(destination_component_id);
+
+            if apa.targets().map_or(true, |targets| {
+                targets.destination_component_id == destination_component_id
+            }) {
```
nickgerace, line 102: I dig it

lib/dal/src/attribute/prototype/debug.rs
```diff
@@ -0,0 +1,207 @@
...
+        let all_attribute_values =
+            AttributePrototype::attribute_value_ids(ctx, prototype_id).await?;
+        let mut func_binding_args: HashMap<String, Vec<serde_json::Value>> = HashMap::new();
+        let mut arg_sources: HashMap<String, Option<String>> = HashMap::new();
+
+        let func_name = Func::get_by_id(ctx, func_id).await?.name;
+
+        let func_execution =
+            Some(FuncExecution::get_latest_execution_by_func_id(ctx, &func_id).await?);
+        for apa_id in apa_ids {
```
nickgerace, line 93: nit: feel free to ignore. I am not a fan of abbreviations, but this leans less into convention and more into our codebase style. We generally avoid abbreviations to prevent convention, help new hires and new contributors, and prevent shoring of information behind abbreviated abstractions.

Relevant doc (kinda?): https://rust-lang.github.io/api-guidelines/naming.html

lib/dal/src/attribute/prototype/debug.rs
```diff
@@ -0,0 +1,207 @@
...
+
+                let func_arg_id =
+                    AttributePrototypeArgument::func_argument_id_by_id(ctx, apa_id).await?;
+                let func_arg_name = ctx
+                    .workspace_snapshot()?
+                    .get_node_weight_by_id(func_arg_id)
+                    .await?
+                    .get_func_argument_node_weight()?
+                    .name()
+                    .to_owned();
```
nickgerace, line 126: This should be behind a method in `FuncArgument` rather than using the snapshot directly.

lib/dal/src/attribute/prototype/debug.rs
```diff
@@ -0,0 +1,207 @@
...
+                            let mut values = vec![];
+
+                            for av_id in other_source
+                                .attribute_values_for_component_id(
+                                    ctx,
+                                    expected_source_component_id,
+                                )
+                                .await?
+                            {
+                                let attribute_value = AttributeValue::get_by_id(ctx, av_id).await?;
```
nickgerace, line 156: A smell for me when match arms get too big is when you have a scope that contains non-trivial iterator or recursive logic in them. I'd look into breaking this up.

Example:
```rust
match foo {
    Some(v) => println!("{v}"),
    None => {
        for i in big_ass_vec {
            // a bunch of code
        }
```

lib/dal/src/attribute/prototype/debug.rs
```diff
@@ -0,0 +1,207 @@
...
+                                    }
+                                    ValueIsFor::Prop(_) => {
+                                        let prop_id = AttributeValue::prop_id_for_id(
+                                            ctx,
+                                            attribute_value.id(),
+                                        )
+                                        .await?;
+                                        Prop::path_by_id(ctx, prop_id).await?
+                                    }
+                                    _ => continue,
```
nickgerace, line 173: nit: I think since there are only three `ValueIsFor` enum variants, I'd just explicitly write it here in case we add another one in the future and want the developer doing so to look at each location rather than getting masked by `_`. This is unlikely, but a good habit... depending on context.

lib/dal/src/attribute/value.rs
```diff
@@ -2125,4 +2128,108 @@ impl AttributeValue {
 
         Ok(input_socket_ids)
     }
+    #[instrument(level = "info", skip_all)]
+    /// Get the morale equivalent of the prop path for a given attribute value id.
+    /// This includes the key/index in the path, unlike the path for a Prop which doesn't
+    /// include the key/index
+    pub async fn get_path_for_id(
```
nickgerace, line 2135: For style, I'd put the comment above the macro.

lib/dal/src/attribute/value.rs
```diff
@@ -2125,4 +2128,108 @@ impl AttributeValue {
...
+        let mut work_queue = VecDeque::from([attribute_value_id]);
+
+        while let Some(mut attribute_value_id) = work_queue.pop_front() {
+            let prop_name = match AttributeValue::prop_id_for_id(ctx, attribute_value_id).await {
+                Ok(prop) => Prop::get_by_id(ctx, prop).await?.name,
+                Err(_) => {
+                    // This means it's an input or output socket, which has no path and should just be empty string
+                    continue;
+                }
+            };
```
nickgerace, line 2149: Is the only error possible here the one that you listed? If not, then we should handle the specific error or refactor the `prop_id_for_id` function to return a different type to help you determine that (e.g. `Result<Option<T>, E>>` instead of `Result<T, E>`).

nickgerace (changes requested):
Alright @britmyerss, you wanted nits... you got nits!

lib/dal/src/attribute/value.rs
```diff
@@ -2125,4 +2128,108 @@ impl AttributeValue {
...
+                            .workspace_snapshot()?
+                            .incoming_sources_for_edge_weight_kind(
+                                attribute_value_id,
+                                EdgeWeightKindDiscriminants::Ordinal,
+                            )
+                            .await?
+                            .pop()
+                            .ok_or(AttributeValueError::NoOrderingNodeForAttributeValue(
+                                attribute_value_id,
+                            ))?;
```
nickgerace, line 2169: This should be its own function for `AttributeValue`. We do not want to expose the snapshot.

lib/dal/src/attribute/value.rs
```diff
@@ -2125,4 +2128,108 @@ impl AttributeValue {
...
+
+                        let order_node_weight = ctx
+                            .workspace_snapshot()?
+                            .get_node_weight(order_of_thing_node)
+                            .await?
+                            .get_ordering_node_weight()?;
+
+                        let order = order_node_weight.order();
```
nickgerace, line 2177: Same here as above.... Maybe on `WorkspaceSnapshot` itself though since "ordering nodes" do not belong to SI-specific entities.

lib/dal/src/attribute/value.rs
```diff
@@ -2125,4 +2128,108 @@ impl AttributeValue {
...
+
+                        // Find the index of the item in the array to add to path
+                        let index = order
+                            .iter()
+                            .position(|&id| id == attribute_value_id.into())
+                            .ok_or(AttributeValueError::ParentAttributeValueMissing(
+                                attribute_value_id,
+                            ))?;
+                        let name = format!("{0}[{1}]", prop_name, &index.to_string());
+                        parts.push_front(name.to_owned());
```
nickgerace, line 2187: What is being assembled here? It looks like the path for display purposes in the frontend.

lib/dal/src/attribute/value.rs
```diff
@@ -2125,4 +2128,108 @@ impl AttributeValue {
...
+                                .edges_directed_for_edge_weight_kind(
+                                    attribute_value_id,
+                                    Direction::Incoming,
+                                    EdgeWeightKindDiscriminants::Contain,
+                                )
+                                .await?
+                                .pop()
+                                .ok_or(AttributeValueError::ParentAttributeValueMissing(
+                                    attribute_value_id,
+                                ))?;
```
nickgerace, line 2202: Same comment as above with not exposing the graph

lib/dal/src/attribute/value.rs
```diff
@@ -2125,4 +2128,108 @@ impl AttributeValue {
...
+                            if let EdgeWeightKind::Contain(contain_key) = edge_weight.0.kind() {
+                                contain_key.to_owned()
+                            } else {
+                                None
+                            }
+                        } {
+                            Some(key) => {
+                                let name = format!("{0}[{1}]", prop_name, &key.to_owned());
+                                //let name = prop_name + &key.to_owned();
```
nickgerace, line 2211: I'd delete this comment

lib/dal/src/attribute/value.rs
```diff
@@ -2125,4 +2128,108 @@ impl AttributeValue {
...
+            } else {
+                // no more parents, need to add the prop_name to the parts because now we're at the root!
+                parts.push_front(prop_name);
+            }
+        }
+        if !parts.is_empty() {
+            Ok(Some(PropPath::new(parts).with_replaced_sep("/")))
+        } else {
+            Ok(None)
+        }
```
nickgerace, line 2233: I'd explain what's going on here in a comment and leave an additional space. I think what I've figured out is that we are assembling the parts here. I'm curious, why not use `PropPath` itself during the iterations rather than string operations?

lib/dal/src/attribute/value/debug.rs
```diff
@@ -0,0 +1,131 @@
...
+use crate::workspace_snapshot::node_weight::NodeWeightError;
+use crate::workspace_snapshot::WorkspaceSnapshotError;
+
+use crate::ComponentError;
+use telemetry::prelude::*;
+
+use crate::{func::execution::FuncExecution, AttributeValue, AttributeValueId, DalContext, Prop};
+use crate::{AttributePrototypeId, FuncError, FuncId, PropKind};
+
+use super::ValueIsFor;
```
nickgerace, line 31: Same comment about imports as before

lib/dal/src/attribute/value/debug.rs
```diff
@@ -0,0 +1,131 @@
...
+    #[error(transparent)]
+    NodeWeightError(#[from] NodeWeightError),
+    #[error(transparent)]
+    OutputSocketError(#[from] OutputSocketError),
+    #[error(transparent)]
+    PropError(#[from] PropError),
+    #[error(transparent)]
+    ValueSourceError(#[from] ValueSourceError),
+    #[error(transparent)]
+    WorkspaceSnapshotError(#[from] WorkspaceSnapshotError),
```
nickgerace, line 82: Same comment about `transparent` as before

lib/dal/src/attribute/value/debug.rs
```diff
@@ -0,0 +1,131 @@
...
+            value_is_for,
+            func_name: attribute_prototype_debug_view.func_name,
+            func_args: attribute_prototype_debug_view.func_args,
+            arg_sources: attribute_prototype_debug_view.arg_sources,
+            value,
+            prop_kind,
+            materialized_view,
+        };
+        Ok(view)
+    }
```
nickgerace, line 130: Now I'm a bit confused. What's the difference between this and the other view?

lib/dal/src/component/debug.rs
```diff
@@ -0,0 +1,177 @@
...
+/// the components attributes. Used for constructing a debug view of the component and also for
+/// cloning a component
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct ComponentDebugView {
+    pub name: String,
+    pub schema_variant_id: SchemaVariantId,
+    pub attribute_tree: HashMap<AttributeValueId, Vec<AttributeValueId>>,
+    // pub input_sockets: Vec<AttributeDebugView>,
+    // pub output_sockets: Vec<AttributeDebugView>,
```
nickgerace, line 33: We should delete the commented out stuff unless accompanied by a `NOTE`, `TODO` or `FIXME`.

lib/dal/src/component/debug.rs
```diff
@@ -0,0 +1,177 @@
...
+}
+
+#[remain::sorted]
+#[derive(Error, Debug)]
+pub enum ComponentDebugViewError {
+    #[error(transparent)]
```
nickgerace, line 39: Same comment about `transparent`

lib/dal/src/component/debug.rs
```diff
@@ -0,0 +1,177 @@
...
+        let debug_view = ComponentDebugView {
+            name,
+            schema_variant_id: schema_variant.id(),
+            attribute_tree: child_values,
+        };
+
+        Ok(debug_view)
+    }
+    #[instrument(level = "info", skip_all)]
+    pub async fn get_attribute_value_tree_for_component(
```
nickgerace, line 103: I'd leave a space here.

lib/dal/src/component/debug.rs
```diff
@@ -0,0 +1,177 @@
...
+    #[error(transparent)]
+    WorkspaceSnapshotError(#[from] WorkspaceSnapshotError),
+}
+
+impl ComponentDebugView {
+    #[instrument(level = "info", skip_all)]
+    pub async fn new(ctx: &DalContext, component: &Component) -> ComponentDebugViewResult<Self> {
+        let schema_variant = Component::schema_variant(component, ctx).await?;
+
+        let child_values =
```
nickgerace, line 86: Good idea to break this up!

lib/dal/src/component/debug.rs
```diff
@@ -0,0 +1,177 @@
...
+    #[error(transparent)]
+    UlidDecode(#[from] ulid::DecodeError),
+    #[error(transparent)]
+    WorkspaceSnapshotError(#[from] WorkspaceSnapshotError),
+}
+
+impl ComponentDebugView {
+    #[instrument(level = "info", skip_all)]
+    pub async fn new(ctx: &DalContext, component: &Component) -> ComponentDebugViewResult<Self> {
+        let schema_variant = Component::schema_variant(component, ctx).await?;
```
nickgerace, line 84: nit: You don't need the whole schema variant. We may be able to just fetch the id from the node.

lib/dal/src/component/debug.rs
```diff
@@ -0,0 +1,177 @@
...
+                            .into();
+                        child_attribute_values_with_keys_by_id
+                            .insert(child_id, (target_idx, key.to_owned()));
+                    }
+                }
+
+                let maybe_ordering =
+                    AttributeValue::get_child_av_ids_for_ordered_parent(ctx, attribute_value_id)
+                        .await
+                        .ok();
```
nickgerace, line 140: Are we eating a potential error here?

lib/dal/src/component/debug.rs
```diff
@@ -0,0 +1,177 @@
...
+    }
+    #[instrument(level = "info", skip_all)]
+    pub async fn get_attribute_value_tree_for_component(
+        ctx: &DalContext,
+        component_id: ComponentId,
+    ) -> ComponentDebugViewResult<HashMap<AttributeValueId, Vec<AttributeValueId>>> {
+        let mut child_values = HashMap::new();
+        // Get the root attribute value and load it into the work queue.
+        let root_attribute_value_id = Component::root_attribute_value_id(ctx, component_id).await?;
+        let workspace_snapshot = ctx.workspace_snapshot()?;
```
nickgerace, line 110: I'm a bit conflicted. I see other `AttributeValue` methods called in here, so that's a sign to create more methods to sandbox the snapshot. However, it makes more sense in this function than before because this is more of a raw usage scenario. I'd still investigate breaking them up.

lib/dal/src/component/debug.rs
```diff
@@ -0,0 +1,177 @@
...
+
+                // Load the work queue with the child attribute value.
+                work_queue.push_back(child_attribute_value_id);
+
+                // Cache the  prop values to eventually insert into the child property editor values map.
+                child_attribute_value_ids.push(child_attribute_value.id());
+            }
+            child_values.insert(attribute_value_id, child_attribute_value_ids);
+        }
+        return Ok(child_values);
```
nickgerace, line 175: No need to do this. Just `Ok(child_values)`. Remember, Rust operates in patterns by default, so no need to make a statement here!

lib/dal/src/component/view/debug.rs
nickgerace, line 1: Nice

lib/dal/tests/integration_test/component/debug.rs
```diff
@@ -0,0 +1,68 @@
+use dal::component::debug::ComponentDebugView;
+use dal::prop::PropPath;
+use dal::{AttributeValue, Component, DalContext, Prop};
+use dal_test::test;
+use dal_test::test_harness::{commit_and_update_snapshot, create_component_for_schema_name};
+use pretty_assertions_sorted::assert_eq;
+
+#[test]
+async fn get_debug_view(ctx: &mut DalContext) {
```
nickgerace, line 9: Let's leave some spaces and comments. This is difficult to read.

Usually, the convention we have to group things by `commit` when in doubt. Additionally, grouping by "preparing the environment" vs. performing mutations and assertions can be helpful.

lib/dal/tests/integration_test/component/debug.rs
nickgerace, line 1: Yay! Testing!

lib/sdf-server/src/server/service/component/debug.rs
```diff
@@ -1,11 +1,16 @@
...
 
 use serde::{Deserialize, Serialize};
 
 use super::ComponentResult;
 use crate::server::extract::{AccessBuilder, HandlerContext};
-use dal::{Component, ComponentId, SchemaVariantId, Visibility};
+use dal::attribute::value::debug::AttributeDebugView;
+use dal::component::debug::ComponentDebugView;
+use dal::{AttributeValueId, Component, ComponentId, SchemaVariantId, Visibility};
+use telemetry::prelude::*;
```
nickgerace, line 13: Same comment as before

lib/sdf-server/src/server/service/component/debug.rs
```diff
@@ -57,55 +39,58 @@ pub struct DebugComponentRequest {
...
+                }
+            }
+        }
+    }
+    attributes.sort_by_key(|view| view.path.to_lowercase());
+
+    //grab input socket info next
+    //grab output socket info next
+
+    dbg!(transform_start.elapsed());
```
nickgerace, line 93: Let's delete these or put this behind an `info` or `debug` statement with context for what it's benchmarking.

lib/sdf-server/src/server/service/component/debug.rs
```diff
@@ -57,55 +39,58 @@ pub struct DebugComponentRequest {
...
+                        name,
+                        path: child_avd.path.to_owned(),
+                        debug_data: child_avd,
+                    });
+                    cache.push(child_av);
+                }
+            }
+        }
+    }
+    attributes.sort_by_key(|view| view.path.to_lowercase());
```
nickgerace, line 88: My instinct is to put this behind a dal struct and then test that in integration tests. We want `sdf` routes thin in lieu of not having API-driven testing infrastructure. Up to you though if you think this is covered in tests.

lib/dal/src/attribute/value.rs
vbustamante: That is correct. This function is supposed to assemble a path to the attribute value, for display in the debugging panel. So if its an array item, it would be something like `/root/domain/things/thing[0]` with `0` being the index of the item, `thing` being the prop name for entry and `things` the prop name for the array.

For a map its the same thing but with key instead of index

lib/dal/src/attribute/prototype/debug.rs
jobelenus: We don't have a linter that will sort those for us?

nickgerace: IDEs typically do. There are a few features needed to do this well via automation and at least one of them is not `stable` yet: https://rust-lang.github.io/rustfmt/?version=master&search=group#imports_granularity

lib/dal/src/attribute/value.rs
nickgerace: Gotcha. Is there a reason why we wouldn't use `PropPath` instead of strings? Is it because it doesn't work with array items?

vbustamante: Exactly. Prop paths are different from the attribute value paths this creates since we show keys and indices

nickgerace: I see!! String manipulation makes me nervous... I wonder if an `AttributeValuePath` type entity would be useful.

britmyerss: Yeah I debated making an `AttributeValuePath` entity, because I was honestly surprised we didn't have this yet. But it felt strange to formalize it for just this debug view and I figured if we end up needing it later we can always formalize it then and replace this implementation with that one.

nickgerace: I won't be a blocker here, but I would personally formalize it because there may be a day 2 bug or feature extension here that relies on using/changing/extending the string manipulation.

It's a nit though.

app/web/src/components/AttributeDebugView.vue
britmyerss: yeah I asked myself the same thing here. I decided to keep it as is both because the whole world is not in scope for this, and because this actually is a debug view of the attributes and we're displaying the `AttributePrototype` info from the perspective of the `AttributeValues` - but your questions below, I separated this on the backend because the next step is to bring back input/output sockets, which will also display the same `AttributePrototype` debug data for them AND I want to surface this on the viz diagram eventually.

lib/dal/src/attribute/prototype/debug.rs
britmyerss: will do! yeah my IDE does some of this formatting and merging for me but I wasn't sure what 'more right' vs 'less right' looked like :)

lib/dal/src/attribute/prototype/debug.rs
britmyerss: @vbustamante something for us to talk through so I can understand better!

lib/dal/src/attribute/prototype/debug.rs
britmyerss: Funny enough, I had it as `assemble` originally!  Following my other comment, we'll also need to generate this for the socket debug views, so I had envisioned a `new_from_socket` func.... but maybe I just go back to `assemble` after-all and use the `AttributeValue` for that Socket to drive it 🤷‍♀️

app/web/src/components/AttributeDebugView.vue
nickgerace: Okay that sounds good to me.

lib/dal/src/attribute/prototype/debug.rs
nickgerace: Whatever works best :smile:

I think either `new` or `assemble` or `assemble_<other-words>` works!

britmyerss:
bors try

lib/si-pkg/src/spec/attribute_value.rs
nickgerace, line 1: These changes are excellent! Love the use of the constant for the separator.

lib/dal/tests/integration_test/component/debug.rs
```diff
@@ -0,0 +1,78 @@
...
+        .copied()
+        .expect("get first value id");
+
+    let attribute_path = AttributeValue::get_path_for_id(ctx, rigid_designator_value_id)
+        .await
+        .expect("can't get the path");
+    assert_eq!(
+        attribute_path,
+        Some(rigid_prop_path.with_replaced_sep("/").to_string())
+    );
```
nickgerace, line 77: Love it!

nickgerace:
Really awesome work @britmyerss 🥳

britmyerss:
bors try

britmyerss:
bors merge

PR #3477, opened 2024-03-27
feat: add target for cypress e2e tests into BUCK

johnrwatson:
This PR adds a new toolchain into our BUCK build system to allow us to execute Cypress end to end tests via BUCK directly.

In a follow up PR, the test will then be executed in CI.

The added target is runnable and buildable, depending on how you would like to execute it.

You can explicitly invoke as follows:
`buck2 run app/web:e2e-test -- --tests 'cypress/e2e/**' --web-endpoint http://localhost:8080`

This will run all our e2e tests against your local stack. These values are the defaults and can be excluded entirely on invocation if you desire, i.e.
`buck2 run app/web:e2e-test`

When executed, the target will check a few things before executing, these are listed below:
- The stack is responsive on the given `--web-endpoint` within 60s
- Cypress installation is valid
- You are on an supported OS / Architecture

During execution, it will then output the log contents and videos from cypress into the BUCK out path, which is usually something in the form of:
`si/buck-out/v2/gen/root/<uuid>/app/web/__e2e-test__/<content here>`

**Variables**
Some variables are required to allow these tests to execute successfully and each test has slightly differing requirements. In the header of each of the cypress typescript test files you will see a block like follows:
```
const AUTH0_USERNAME = Cypress.env('VITE_AUTH0_USERNAME') || import.meta.env.VITE_AUTH0_USERNAME;
const AUTH0_PASSWORD = Cypress.env('VITE_AUTH0_PASSWORD') || import.meta.env.VITE_AUTH0_PASSWORD;
const AUTH_API_URL = Cypress.env('VITE_AUTH_API_URL') || import.meta.env.VITE_AUTH_API_URL;
const SI_WORKSPACE_ID = Cypress.env('VITE_SI_WORKSPACE_ID') || import.meta.env.VITE_SI_WORKSPACE_ID;
```

This basically means, either have the variables in your local shell environment or within the .env files for vite, i.e. `.env.local` usually. I've included some psuedocode in `app/web/cypress/support/commands.ts` as to how we might refactor this to centralise the variables in one place, but I got into a bit of a tizzy with cypress coinage so I may have to come back to this in a follow up PR to implement the new test into CI.

### DIFF
--- a/app/web/cypress/e2e/authentication/login.cy.ts
+++ b/app/web/cypress/e2e/authentication/login.cy.ts
@@ -1,15 +1,23 @@
-Cypress._.times(import.meta.env.VITE_SI_CYPRESS_MULTIPLIER ? import.meta.env.VITE_SI_CYPRESS_MULTIPLIER : 1, () => {
+const SI_CYPRESS_MULTIPLIER = Cypress.env('VITE_SI_CYPRESS_MULTIPLIER') || import.meta.env.VITE_SI_CYPRESS_MULTIPLIER || 1;
+const AUTH0_USERNAME = Cypress.env('VITE_AUTH0_USERNAME') || import.meta.env.VITE_AUTH0_USERNAME;
+const AUTH0_PASSWORD = Cypress.env('VITE_AUTH0_PASSWORD') || import.meta.env.VITE_AUTH0_PASSWORD;
+const AUTH_API_URL = Cypress.env('VITE_AUTH_API_URL') || import.meta.env.VITE_AUTH_API_URL;
+const SI_WORKSPACE_ID = Cypress.env('VITE_SI_WORKSPACE_ID') || import.meta.env.VITE_SI_WORKSPACE_ID;
+const UUID = Cypress.env('VITE_UUID') || import.meta.env.VITE_UUID || "local";
+
+Cypress._.times(SI_CYPRESS_MULTIPLIER, () => {
   describe("login", () => {
     beforeEach(() => {
       cy.visit("/");
     });
 
     it("log_in", () => {
-      cy.loginToAuth0(import.meta.env.VITE_AUTH0_USERNAME, import.meta.env.VITE_AUTH0_PASSWORD);
-      cy.visit("/");
-      cy.sendPosthogEvent(Cypress.currentTest.titlePath.join("/"), "test_uuid", import.meta.env.VITE_UUID ? import.meta.env.VITE_UUID: "local");
+      cy.loginToAuth0(AUTH0_USERNAME, AUTH0_PASSWORD);
+      cy.visit(AUTH_API_URL + '/workspaces/' + SI_WORKSPACE_ID + '/go');
+      cy.sendPosthogEvent(Cypress.currentTest.titlePath.join("/"), "test_uuid", UUID);
       // check that you're on head i.e. that you were redirected correctly
-      cy.url().should("contain", "head");
+      cy.wait(4000)
+      cy.url().should("contain", SI_WORKSPACE_ID);
     });
 
   });
--- a/app/web/cypress/e2e/web-functionality/get-summary.cy.ts
+++ b/app/web/cypress/e2e/web-functionality/get-summary.cy.ts
@@ -1,23 +1,26 @@
 // @ts-check
 ///<reference path="../global.d.ts"/>
 
+const AUTH0_USERNAME = Cypress.env('VITE_AUTH0_USERNAME') || import.meta.env.VITE_AUTH0_USERNAME;
+const AUTH0_PASSWORD = Cypress.env('VITE_AUTH0_PASSWORD') || import.meta.env.VITE_AUTH0_PASSWORD;
+const SI_WORKSPACE_URL = Cypress.env('VITE_SI_WORKSPACE_URL') || import.meta.env.VITE_SI_WORKSPACE_URL;
+const SI_WORKSPACE_ID = Cypress.env('VITE_SI_WORKSPACE_ID') || import.meta.env.VITE_SI_WORKSPACE_ID;
+const UUID = Cypress.env('VITE_UUID') || import.meta.env.VITE_UUID || "local";
+
 describe('web', () => {
-    beforeEach(function () {
-      cy.loginToAuth0(import.meta.env.VITE_AUTH0_USERNAME, import.meta.env.VITE_AUTH0_PASSWORD);
+  beforeEach(function () {
+    cy.loginToAuth0(AUTH0_USERNAME, AUTH0_PASSWORD);
+  });
 
+  it('get_summary', () => {
+    // Go to the Synthetic Workspace
+    cy.visit(SI_WORKSPACE_URL + '/w/' + SI_WORKSPACE_ID + '/head');
+    cy.sendPosthogEvent(Cypress.currentTest.titlePath.join("/"), "test_uuid", UUID);
+    cy.intercept('GET', SI_WORKSPACE_URL + '/api/qualification/get_summary?visibility_change_set_pk=00000000000000000000000000', (req) => {
+      // Log the intercepted request URL and response status code
+      cy.log(`Request to ${req.url}`, req.response.statusCode);
+      // Assert that the status code is 200
+      expect(req.response.statusCode).to.eq(200);
     });
-  
-    it('get_summary', () => {
-  
-      // Go to the Synthetic Workspace
-      cy.visit(import.meta.env.VITE_SI_WORKSPACE_URL + '/w/' + import.meta.env.VITE_SI_WORKSPACE_ID + '/head')
-      cy.sendPosthogEvent(Cypress.currentTest.titlePath.join("/"), "test_uuid", import.meta.env.VITE_UUID ? import.meta.env.VITE_UUID: "local");
-      cy.intercept('GET', import.meta.env.VITE_SI_WORKSPACE_URL + '/api/qualification/get_summary?visibility_change_set_pk=00000000000000000000000000', (req) => {
-        // Log the intercepted request URL and response status code
-        cy.log(`Request to ${req.url}`, req.response.statusCode);
-        // Assert that the status code is 200
-        expect(req.response.statusCode).to.eq(200);
-      });
-
-    })
-})
+  });
+});
--- a/app/web/src/store/auth.store.ts
+++ b/app/web/src/store/auth.store.ts
@@ -150,6 +150,9 @@ export const useAuthStore = defineStore("auth", {
         if (errCode === "WORKSPACE_NOT_INITIALIZED") {
           // db is migrated, but workspace does not exist, probably because it has been reset
           const _reconnectReq = await this.AUTH_RECONNECT();
+          // WHAT HAD HAPPENED WAS...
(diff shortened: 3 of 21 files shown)


app/web/cypress/e2e/authentication/login.cy.ts
```diff
@@ -1,15 +1,23 @@
-Cypress._.times(import.meta.env.VITE_SI_CYPRESS_MULTIPLIER ? import.meta.env.VITE_SI_CYPRESS_MULTIPLIER : 1, () => {
+const SI_CYPRESS_MULTIPLIER = Cypress.env('VITE_SI_CYPRESS_MULTIPLIER') || import.meta.env.VITE_SI_CYPRESS_MULTIPLIER || 1;
+const AUTH0_USERNAME = Cypress.env('VITE_AUTH0_USERNAME') || import.meta.env.VITE_AUTH0_USERNAME;
+const AUTH0_PASSWORD = Cypress.env('VITE_AUTH0_PASSWORD') || import.meta.env.VITE_AUTH0_PASSWORD;
+const AUTH_API_URL = Cypress.env('VITE_AUTH_API_URL') || import.meta.env.VITE_AUTH_API_URL;
+const SI_WORKSPACE_ID = Cypress.env('VITE_SI_WORKSPACE_ID') || import.meta.env.VITE_SI_WORKSPACE_ID;
+const UUID = Cypress.env('VITE_UUID') || import.meta.env.VITE_UUID || "local";
```
nickgerace, line 6: This target will solely be "runnable" and not "testable" as a result. That's totally fine with me, but just wanted to point it out in case that will no longer suffice.

prelude-si/macros.bzl
```diff
@@ -98,4 +104,4 @@ load(
 tilt_docker_compose_pull = _tilt_docker_compose_pull
 tilt_docker_compose_stop = _tilt_docker_compose_stop
 tilt_down = _tilt_down
-tilt_up = _tilt_up
+tilt_up = _tilt_up
```
johnrwatson, line 107: This looks weird, will amend

prelude-si/tilt.bzl
```diff
@@ -71,7 +71,7 @@ tilt_docker_compose_pull = rule(
     attrs = {
         "docker_compose_file": attrs.string(
             default = "docker-compose.yml",
-            doc = """The Tiltfile to run.""",
+            doc = """The Docker Compose to run.""",
```
johnrwatson, line 74: Just a typo in previous art

app/web/cypress/e2e/authentication/login.cy.ts
```diff
@@ -1,15 +1,23 @@
...
     it("log_in", () => {
-      cy.loginToAuth0(import.meta.env.VITE_AUTH0_USERNAME, import.meta.env.VITE_AUTH0_PASSWORD);
-      cy.visit("/");
-      cy.sendPosthogEvent(Cypress.currentTest.titlePath.join("/"), "test_uuid", import.meta.env.VITE_UUID ? import.meta.env.VITE_UUID: "local");
+      cy.loginToAuth0(AUTH0_USERNAME, AUTH0_PASSWORD);
+      cy.visit(AUTH_API_URL + '/workspaces/' + SI_WORKSPACE_ID + '/go');
+      cy.sendPosthogEvent(Cypress.currentTest.titlePath.join("/"), "test_uuid", UUID);
       // check that you're on head i.e. that you were redirected correctly
-      cy.url().should("contain", "head");
+      cy.wait(4000)
```
nickgerace, line 19: Is this the beginning of the test that ensures SI is ready to go? If so, this will work for now, but it would be great if there was a signal or hook of some kind.

app/web/cypress/e2e/web-functionality/get-summary.cy.ts
```diff
@@ -1,23 +1,26 @@
 // @ts-check
 ///<reference path="../global.d.ts"/>
 
+const AUTH0_USERNAME = Cypress.env('VITE_AUTH0_USERNAME') || import.meta.env.VITE_AUTH0_USERNAME;
+const AUTH0_PASSWORD = Cypress.env('VITE_AUTH0_PASSWORD') || import.meta.env.VITE_AUTH0_PASSWORD;
+const SI_WORKSPACE_URL = Cypress.env('VITE_SI_WORKSPACE_URL') || import.meta.env.VITE_SI_WORKSPACE_URL;
+const SI_WORKSPACE_ID = Cypress.env('VITE_SI_WORKSPACE_ID') || import.meta.env.VITE_SI_WORKSPACE_ID;
+const UUID = Cypress.env('VITE_UUID') || import.meta.env.VITE_UUID || "local";
```
nickgerace, line 8: My Python skills are rusty, but it feels like all files should be able to import these in a new file via relative path. Thoughts?

app/web/src/store/auth.store.ts
```diff
@@ -150,6 +150,9 @@ export const useAuthStore = defineStore("auth", {
         if (errCode === "WORKSPACE_NOT_INITIALIZED") {
           // db is migrated, but workspace does not exist, probably because it has been reset
           const _reconnectReq = await this.AUTH_RECONNECT();
+          // WHAT HAD HAPPENED WAS...
+          // in a local scenario where prd auth portal has a workspace, and local does not, the first attempt this will hit an infinite loop trying to find the workspace, totally stuck
+          // on a second attempt, it appears, the workspace gets created and user exits the loop
           // TODO: react to failure here?
```
nickgerace, line 156: This is messy. Good comment, but it feels like there's a longer term fix waiting here once we pay more attention to the auth stack again.

prelude-si/e2e/e2e_test.py
```diff
@@ -0,0 +1,199 @@
...
+
+
+class PlatformArchitecture(BaseEnum):
+    Aarch64 = "aarch64"
+    X86_64 = "x86_64"
+
+
+class PlatformOS(BaseEnum):
+    Darwin = "darwin"
+    Linux = "linux"
```
nickgerace, line 38: This is a nit, so feel free to ignore. Is there value in detecting this here? Both our README and Nix flake are only compatible with this platform matrix anyway.

johnrwatson: Fair comment, I was thinking there was a chance it might make it through to this point on a strange variant, or maybe we raw-port the python onto alpine or something as an execution environment.

I.e. by invoke like
```
python3 ./e2e_test.py
```

app/web/src/store/auth.store.ts
johnrwatson: Goodness I know, we found the TODO when deving this out + decided to add a bit more colour as it really wasn't clear what was going on.

I am genuinely not sure what the long term fix would be here apart from the obvious response of "handle this case"

app/web/cypress/e2e/web-functionality/get-summary.cy.ts
johnrwatson: Absolutely, all this needs to be centralised in a cy.command so that the user doesn't need to read the exact file to import the necessary vars.

I've added some pseduo-code here as to how we might do this here but couldn't quite get it over the line in this PR:
https://github.com/systeminit/si/pull/3477/files#diff-46bd59ba837fc6b94e01a7f8aa51ed980790920e2ddb8d49702c84c7400f71fbR72-R86

Ideally these tests are simple & friendly to run so it's imperative we resolve things like the above to make them so.

prelude-si/macros.bzl
johnrwatson: Looks like a nil diff, leaving it

app/web/cypress/e2e/authentication/login.cy.ts
johnrwatson: Not quite, two things going on here:
1. I believe after the `cy.visit(AUTH_API_URL + '/workspaces/' + SI_WORKSPACE_ID + '/go');` 
 that the /restore_session (or similar) endpoint showed some latency so this was a basic stabilisation workaround to try and make sure the session (and webpage) had been established in the browser before trying to assert that it was on the correct workspace. It would definitely be better to wait for a specific signal from the webpage (like a hook/interception of a specific API call response)
 
2. We did notice some weirdness as to which changeset was loaded into by default by the web app. Currently the application tries to work it out via the session storage keys as to which changeset the user was last on, if the key isn't found it defaults to HEAD. Sometimes in the tests however the last changeset would be B in the session storage key but when the user was routed through to auto it would redirect to HEAD, or vice versa. The check on line 20 was therefore modified to only check it was on the correct workspace, not specifically the correct changeset/HEAD.

johnrwatson:
bors merge

nickgerace (approved):

app/web/cypress/e2e/authentication/login.cy.ts
nickgerace: Gotcha. Feels like for both scenarios we getting around some logic we need to tighten up. However, I think it's unrealistic to chase those down today as the system is rapidly changing, so this little wait works for me!'

app/web/cypress/e2e/web-functionality/get-summary.cy.ts
nickgerace: Okay sounds good. Will follow up offline for how we can track that.

app/web/src/store/auth.store.ts
nickgerace: That's exactly the fix. We'll do it.

prelude-si/e2e/e2e_test.py
nickgerace: Gotcha. Let's keep it.

app/web/cypress/e2e/authentication/login.cy.ts
```diff
@@ -1,15 +1,23 @@
-Cypress._.times(import.meta.env.VITE_SI_CYPRESS_MULTIPLIER ? import.meta.env.VITE_SI_CYPRESS_MULTIPLIER : 1, () => {
+const SI_CYPRESS_MULTIPLIER = Cypress.env('VITE_SI_CYPRESS_MULTIPLIER') || import.meta.env.VITE_SI_CYPRESS_MULTIPLIER || 1;
+const AUTH0_USERNAME = Cypress.env('VITE_AUTH0_USERNAME') || import.meta.env.VITE_AUTH0_USERNAME;
+const AUTH0_PASSWORD = Cypress.env('VITE_AUTH0_PASSWORD') || import.meta.env.VITE_AUTH0_PASSWORD;
+const AUTH_API_URL = Cypress.env('VITE_AUTH_API_URL') || import.meta.env.VITE_AUTH_API_URL;
+const SI_WORKSPACE_ID = Cypress.env('VITE_SI_WORKSPACE_ID') || import.meta.env.VITE_SI_WORKSPACE_ID;
```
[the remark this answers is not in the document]
nickgerace: I actually think you want this "as is". I'd want explicit passing from the executor rather than some implicit logic to go find it.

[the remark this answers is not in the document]
nickgerace: That being said... we should centralize it.
https://github.com/systeminit/si/pull/3477#discussion_r1541477047

PR #3478, opened 2024-03-27
Move encrypted secrets into LayerDB (ENG-2415)

nickgerace:
## Description

This PR moves encrypted secrets into LayerDB.

Before this PR, encrypted secrets were store on the existing PG table whose migration is from the `old-engine`. This table implemented standard model and was both tenancy-scoped and visibility-scoped. While referential secrets made it on the `new-engine` graph structure, their encrypted bits remained in the old table.

Now that LayerDB is in play, we are able to leverage it to get encrypted secrets working between change sets as well as to work with them faster than before.

## Primary Changes

- Add new LayerDB table for encrypted secrets
- Generate keys for the table with the tenancy, `SecretId`, and a generated `Ulid` from the current change set
- Concentrate secrets into two primary objects: `Secret` and `EncryptedSecret` (while retaining `DecryptedSecret` for sending information to cyclone and for testing)
  - `Secret` contains metadata for the secret as well as a key to the encrypted contents
  - `EncryptedSecret` contains the encrypted contents
- Ensure that when encrypted contents are updated, we insert a new `EncryptedSecret` and update the key on the `Secret`
- Split secrets functionality into separate private modules
- Add integration test for updating both the metadata and the encrypted contents for a secret
- Refactor secrets unit tests to be split across relevant modules and integration tests to match the new paradigm

## Secondary Changes

- Privatize the `func::before` module and export its core function and error type
- Restore the qualification check in the before funcs test for added coverage
- Remove unused identity func helper module
- Add `remain::sorted` to `LayeredEventKind`


nickgerace:
bors merge

nickgerace:
bors merge

PR #3479, opened 2024-03-27
fix(dal): make sure you can manually override maps and arrays

vbustamante:
Previously, the rebaser was skipping any unordered children of ordered containers.

Since an attribute value for a map has an ordering node connection, when computing its children, the rebaser was only considering those edges to those nodes as updates. Since when we do an override we’re binding an attributePrototype directly to the attribute value, without the ordering connection, the rebaser was skipping it.

But this was working fine for string attribute values, for instance, since they are considered unordered containers (they’re not containers at all, but since they don’t have the ordering node that’s how we treat them).
 
This PR also fixes an issue in which adding an ordered node to a fresh graph fails, which was forcing unit tests to be more complex than they needed to be

### DIFF
--- a/app/web/src/components/AttributesPanel/AttributesPanelItem.vue
+++ b/app/web/src/components/AttributesPanel/AttributesPanelItem.vue
@@ -308,6 +308,7 @@
         />
         <Icon
           v-if="
+            sourceOverridden &&
             currentValue !== null &&
             !propPopulatedBySocket &&
             !propControlledByParent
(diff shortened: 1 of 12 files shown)


vbustamante:
bors try

zacharyhamm:
I was there when the changes to graph.rs happened so makes sense to me! As a note to @jhelwig, the conflict/update detection for ordered containers ignored all edges that were not in the ordering, so we failed to produce NewEdge etc. updates if you added a Prototype edge from an AttributeValue that was a container type (object, array, map), making it impossible to set the component-level prototype for that edge. @vbustamante maybe add a comment right before the call to find_ordered inside find_unordered that explains that we have to also look for conflicts/updates for the edges that are not in the order.

vbustamante:
> @vbustamante maybe add a comment right before the call to find_ordered inside find_unordered that explains that we have to also look for conflicts/updates for the edges that are not in the order.

Even better! find_ordered does not exist anymore

vbustamante:
bors merge

PR #3480, opened 2024-03-27
feat(web): Show "set by socket" option on dropdown for maps and arrays.

vbustamante:
image

This resets the value of a map, array or object back to the default function for the schema variant, as the x button does for text fields (string, number)


nickgerace (approved):

vbustamante:
bors merge

PR #3481, opened 2024-03-28
feat: add xvfb into ci-base

johnrwatson:
Adding a virtual framebuffer/a display server implementing X11 within the ci-base as there are no FE packages in there at the minute.

I imagine this might not be the final change here as we will probably need chrome or some additional browser packages for cypress.

Will work forward as needed.

Not adding into the nix flake as any normal developer machine will have this dep already satisfied


johnrwatson:
bors merge

PR #3486, opened 2024-03-29
feat(sdf-server): Reimplement create_variant endpoint

stack72:
Fixes: ENG-2426

Screenshot 2024-03-29 at 00 47 00

This behaves differently than it used to. TL;DR, the way that we are going to implement create_variant now is to create a schema, create a schema variant and set the schema variant as the default for the schema

This used to only happen when "Update Asset" was triggered after the stub was created. But, as SchemaVariantDefinitions no longer exist, we are now going to always default to creating a new SchemaVariant so that we are continually testing our systems ability to put items fast on the graph

This is scenario 1 of bringing back the system authoring


stack72:
bors try

stack72:
bors merge

PR #3490, opened 2024-04-01
Ensure actions work with secrets and move apply change set logic to dal (ENG-2410)

nickgerace:
## Description

This PR adds an integration test for ensuring that actions work with secrets. For both testing and for code reuse, the apply logic has been moved to the dal.

## Additional Changes

- Move the upsertion of "create" actions to `Component::new`
- Rename `ChangeSetPointerError` to `ChangeSetError` and `ChangeSetPointerResult` to `ChangeSetResult` for two reasons
  - First reason: this is part of the ongoing refactor to replace the new change set concept with the old one
  - Second reason: to avoid confusion with the introduction of the `ChangeSetApplyError` and `ChangeSetApplyResult` types
- Reduce and refactor sdf change set service errors
- Standardize package version string for test exclusive schemas
- Standardize package "created by" string for test exclusive schemas
- Make the secret defining test exclusive schema more generic and rename "fake" to "dummy" (the previous name was confusing given the name of the sockets for the fallout and starfield schemas)
- Return an error when a qualification map attribute value does not have an ordering node

### DIFF
--- a/lib/dal-test/src/lib.rs
+++ b/lib/dal-test/src/lib.rs
@@ -753,7 +753,7 @@ async fn migrate_local_builtins(
     schema::migrate_pkg(&ctx, SI_AWS_EC2_PKG, None).await?;
     schemas::migrate_test_exclusive_schema_starfield(&ctx).await?;
     schemas::migrate_test_exclusive_schema_fallout(&ctx).await?;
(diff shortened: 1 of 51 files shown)


stack72:
bors try

nickgerace:
bors merge

PR #3492, opened 2024-04-01
feat(sdf, dal): Restore get_resource endpoint to the system

stack72:
This can't be merged yet  because refresh operations overwrite the /root/resource prop which means this will always return an empty resource to the UI


nickgerace (approved):
Hot take: unless the result of `get_resource` blocks other functionality in the frontend, I think this is safe to merge despite the overwrite issue. Up to you though!

stack72:
Going to merge and will follow up with a test when refresh doesn't overwrite /root/resource

stack72:
bors merge

PR #3113, opened 2024-01-03
Switch to the new engine

nickgerace:
## Description

Since July 2023, we have been dreaming of this moment. The changes in `new-engine` all hit `main`. This has been a journey of many squashed commits, bad git merges, pain, sweat and tears. That all changes today. Welcome to the new engine.

## History

This PR is a continuation of #2879, which was closed because the branch was renamed to `new-engine`.

PR #3494, opened 2024-04-02
Annihilate all unsued migrations and queries in the dal (ENG-2435)

nickgerace:
## Description

This is it. Since merging the new engine onto main (#3113), we have kept around migrations and queries to help restore functionality in the system. We're now at a point where we can destroy the bulk of our old migrations and queries. You all served us well.

## Additional or Noteworthy Changes

- Move and rename all remaining queries to subdirectories based on which tables they reference
- Remove dead code related to migrations and queries
- Modify func-related and validation-related migrations to work since they are still used a hybrid fashion
- Make `FuncBinding` methods private that can be made private
- Fix "todo" for missing builtin workspace
- Fix "get_diff" integration test final assertion


lib/dal/src/migrations/U0042__funcs.sql
nickgerace, line 1: Remove `Func` and leave `FuncBinding` and `FuncBindingReturnValue`

lib/dal/src/migrations/U2420__validation_resolvers.sql
```diff
@@ -1,7 +1,3 @@
-DROP TABLE validation_prototypes CASCADE;
-DELETE FROM standard_models where table_name = 'validation_prototypes';
-DROP TABLE validation_resolvers CASCADE;
-
```
nickgerace, line 4: No need to drop the old tables if they are never migrated

lib/dal/src/migrations/U3010__modify_func_bindings.sql
```diff
@@ -1,10 +1,9 @@
 --- we no longer use a sha256 hash on a function's code, blake3 is used instead
 ALTER TABLE func_bindings
-  ADD COLUMN func_id ident,
-  ADD COLUMN code_blake3 TEXT,
-  DROP COLUMN code_sha256;
+    ADD COLUMN func_id     ident,
+    ADD COLUMN code_blake3 TEXT,
+    DROP COLUMN code_sha256;
 
-DROP TABLE func_binding_belongs_to_func CASCADE;
```
nickgerace, line 7: This was the primary change of this file

lib/dal/tests/integration_test/component/get_diff.rs
```diff
@@ -147,9 +149,10 @@ async fn get_diff_component_change_comp_type(ctx: &mut DalContext) {
...
+        Some(" {\n   \"si\": {\n     \"color\": \"#ffffff\",\n     \"name\": \"this is a new component\",\n-    \"type\": \"component\"\n+    \"type\": \"configurationFrameDown\"\n   },\n   \"domain\": {\n     \"name\": \"this is a new component\",\n     \"possible_world_a\": {\n       \"wormhole_1\": {\n         \"wormhole_2\": {\n           \"wormhole_3\": {}\n         }\n       }\n     },\n     \"possible_world_b\": {\n       \"wormhole_1\": {\n         \"wormhole_2\": {\n           \"wormhole_3\": {\n             \"naming_and_necessity\": \"not hesperus\"\n           }\n         }\n       }\n     },\n     \"universe\": {\n       \"galaxies\": []\n     }\n   }\n }".to_string()), // expected
```
nickgerace, line 155: It's a bit strange that this was passing before. The new assertion is the right shape.

nickgerace:
bors merge

PR #3496, opened 2024-04-02
Drop ActionKind from ActionPrototype edge (ENG-2434)

nickgerace:
## Description

This PR drops `ActionKind` from the `ActionPrototype` variant within `EdgeWeightKind`. This information is available within the `ActionPrototype` node and we should not store duplicate information unless strictly necessary.


nickgerace:
bors merge

PR #3474, opened 2024-03-26
feat(dal,sdf,app): Reimplement variant_defs endpoints

stack72:
As we are moving away from the concept of SchemaVariantDefinitions, we want to move the metadata that was part of a SchemaVariantDefinition to be part of the SchemaVariant

This will allow us to start bringing back the authoring workflow without the need to have the concept of a SchemaVariantDefinition

This PR also re-implements the get_variant_def endpoint as well as renaming variant_def to be variant in the API routes

PR #3497, opened 2024-04-02
Remove unused schema variant def module and unused auth prototype code (ENG-2426)

nickgerace:
## Description

`SchemaVariantDefinition` no longer exists in the new engine, so this PR removes its remains from the dal.

## Additional Note

This is a follow-up to #3474.


nickgerace:
bors merge

PR #3498, opened 2024-04-02
feat(ci): enable release and debug builds

adamhjk:
This PR enables release and debug builds for buck2. By default, we will always do debug builds (or, more accurately, whatever the rustc compiler thinks the defaults should be).

If you want to build for release or debug explicitly, you can do that with:

```
$ buck2 build @//mode/release //bin/sdf:sdf
```

And the result will be compiled with release optimizations. You can also use `@//mode/debug` if you want. Those files are just ways of DRY-ing up the configuration options - you could also pass them directly on the command line.

We can add as many configuration options as we want, and as many toolchain configurations as we want, by extending the options in the `//config` tree, adding a new toolchain statement, and extending the select statement for the `:rust` alias.

### DIFF
--- a/.buckconfig
+++ b/.buckconfig
@@ -14,8 +14,9 @@ fbsource = none
 
 [parser]
 target_platform_detector_spec = \
-  target:root//...->prelude//platforms:default \
-  target:prelude-si//...->prelude//platforms:default
+  target:root//...->prelude-si//platforms:default \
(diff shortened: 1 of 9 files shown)


nickgerace (approved):

stack72:
bors merge

adamhjk:
bors merge

PR #3505, opened 2024-04-03
Replace FuncVariant with FuncKind and consolidate list func logic

nickgerace:
## Description

This PR replaces `FuncVariant` with `FuncKind`. `FuncKind` is being used in the new engine and `FuncVariant` is a very similar, but leftover, enum from the func authoring experience in the old engine. These changes do not guarantee that creating funcs will work, but listing and getting funcs in the frontend should continue to work.

This PR also consolidates the `"list_func"` route logic into one location in the dal. It is backed with an integration test as a result.

## Miscellaneous Changes

- Add "kind" to the `Func` struct since it is already on the node weight
- Move `FuncKind` to its own submodule
- Remove unnecessary conversions of `FuncBackendKind <> FuncVariant` and hardcode `FuncBackendKind` in relevant routes
- Separate `FuncKind` and `CustomizableFuncKind` in the frontend (the latter is a subset of the former)
- Add ability to get default `SchemaVariantId` without fetching the entire `Schema` or `SchemaVariant`
- Replace usages of "transparent" in error types with populated fields as well as generally clean up error types
- Fix usage of `schema_name` in dal test helper

### DIFF
--- a/lib/dal/src/func.rs
+++ b/lib/dal/src/func.rs
@@ -4,7 +4,7 @@ use si_events::ContentHash;
 use std::collections::HashMap;
 use std::string::FromUtf8Error;
 use std::sync::Arc;
-use strum::{AsRefStr, Display, IntoEnumIterator};
+use strum::IntoEnumIterator;
 use telemetry::prelude::*;
 use thiserror::Error;
 use ulid::Ulid;
@@ -29,11 +29,14 @@ pub mod binding;
 pub mod binding_return_value;
 pub mod execution;
 pub mod intrinsics;
+pub mod view;
 
 mod before;
+mod kind;
 
 pub use before::before_funcs_for_component;
 pub use before::BeforeFuncError;
+pub use kind::FuncKind;
 
 #[remain::sorted]
 #[derive(Error, Debug)]
@@ -60,8 +63,8 @@ pub enum FuncError {
     Transactions(#[from] TransactionsError),
     #[error("could not acquire lock: {0}")]
     TryLock(#[from] tokio::sync::TryLockError),
-    #[error("unable to determine the function type")]
-    UnknownFunctionType,
+    #[error("unable to determine the function type for backend kind ({0}) and backend response type ({1})")]
+    UnknownFunctionType(FuncBackendKind, FuncBackendResponseType),
     #[error("utf8 error: {0}")]
     Utf8(#[from] FromUtf8Error),
     #[error("workspace snapshot error: {0}")]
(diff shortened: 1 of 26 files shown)


lib/dal/src/func.rs
```diff
@@ -127,8 +114,10 @@ pub enum FuncKind {
 #[derive(Deserialize, Serialize, Debug, Clone, PartialEq, Eq)]
 pub struct Func {
     pub id: FuncId,
-    pub timestamp: Timestamp,
     pub name: String,
+    pub kind: FuncKind,
```
nickgerace, line 118: Already was on the node weight! Now you have it on the struct

lib/dal/src/func/kind.rs
```diff
@@ -0,0 +1,60 @@
...
+            FuncBackendKind::JsAuthentication => Ok(FuncKind::Authentication),
+            FuncBackendKind::JsSchemaVariantDefinition => Ok(FuncKind::SchemaVariantDefinition),
+            FuncBackendKind::JsValidation => {
+                warn!(
+                    ?func_backend_kind,
+                    ?func_backend_response_type,
+                    "found JsValidation func backend kind, marking as unknown"
+                );
+                Ok(FuncKind::Unknown)
+            }
```
nickgerace, line 43: Replaced `dbg` with log

lib/dal/src/func/kind.rs
```diff
@@ -0,0 +1,60 @@
...
+            | FuncBackendKind::Integer
+            | FuncBackendKind::Map
+            | FuncBackendKind::Object
+            | FuncBackendKind::String
+            | FuncBackendKind::Unset
+            | FuncBackendKind::Validation => Ok(FuncKind::Intrinsic),
+            _ => Err(FuncError::UnknownFunctionType(
+                func_backend_kind,
+                func_backend_response_type,
+            )),
```
nickgerace, line 57: Expanded the error type with what was in the dbg

lib/dal/src/func/view.rs
```diff
@@ -0,0 +1,4 @@
+mod list;
+
+pub use list::FuncSummary;
+pub use list::FuncSummaryError;
```
nickgerace, line 4: This is a relatively empty module, but we know we will need it for the "get func" stuff soon

lib/dal/tests/integration_test/func.rs
```diff
@@ -0,0 +1,35 @@
...
+use dal_test::test;
+use pretty_assertions_sorted::assert_eq;
+
+#[test]
+async fn summary(ctx: &mut DalContext) {
+    let schema = Schema::find_by_name(ctx, "starfield")
+        .await
+        .expect("could not find schema")
+        .expect("schema not found");
+    let schema_variant_id = SchemaVariant::get_default_id_for_schema(ctx, schema.id())
```
nickgerace, line 12: Using the new, efficient method because we do not need the full `SchemaVariant`

lib/dal/tests/integration_test/func.rs
```diff
@@ -0,0 +1,35 @@
...
+    let found_func_for_all = all_funcs
+        .iter()
+        .find(|f| f.name() == func_name)
+        .expect("could not find func");
+    let found_func_for_schema_variant = funcs_for_schema_variant
+        .iter()
+        .find(|f| f.name() == func_name)
+        .expect("could not find func");
+
+    assert_eq!(found_func_for_all, found_func_for_schema_variant);
```
nickgerace, line 34: Not the greatest test in the world and will get slower as we add more funcs to test exclusive schemas, but it is very fast now, does the job and even if it does get slower, it operates in memory and will likely be imperceptible.

nickgerace:
bors merge

PR #3506, opened 2024-04-03
fix(dal): Ensure we commit before we run a refresh action

stack72:
Fixes ENG-2433

Right now, when a create action has successfully completed, we run a refresh operation. The lack of a blocking commit meant that we were trying to run a refresh operation without the system having the correct data in place for a refresh to happen


paulocsanz (approved):

stack72:
bors merge

stack72:
bors retry

PR #3508, opened 2024-04-03
chore(dal): Adding a test for setting an update action

stack72:
Fixes ENG-2432

This test creates a component, merges to head, sees the create action happen, forks head, queues an update action, merged to head then checks the component resource view


stack72:
bors try

stack72:
bors merge

PR #3510, opened 2024-04-04
chore: enable debug symbols and clean up chatty output

adamhjk:
This PR tweaks the compilation profile for our "default" debug builds so that they actually include debug symbols. This helps when you want to use rust-gdb on a buck2 build binary or running process (they work perfectly now, including source lines!)

I also turned down the level for two chatty lines in the cyclone server and the raw telemetry settings, which make the test output usable again.


adamhjk:
bors merge

PR #3512, opened 2024-04-04
feat(web,dal,sdf) move joi validations to the backend

vbustamante:
image


lib/dal/src/job/definition.rs
```diff
@@ -1,7 +1,21 @@
...
 //mod refresh;
 
 pub use action::{ActionRunnerItem, ActionsJob};
 pub use dependent_values_update::DependentValuesUpdate;
 //pub use refresh::RefreshJob;
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Copy)]
+pub enum AttributeValueBasedJobIdentifier {
+    DependentValuesUpdate,
+    ComputeValidation,
```
britmyerss, line 13: I'd love to understand more why this isn't part of dependent values update, vs a separate job

vbustamante: Dependent values update does not run for the direct values that have been updated. The ideal place to have it run synchronously would be inside of AttributeValue::set_real_values(), that is hit by both the DVU and the functions that set the first values. So that's were we enqueue the job

vbustamante:
bors merge

vbustamante:
bors merge

vbustamante:
bors merge

PR #3514, opened 2024-04-04
feat(sdf): Restore Modules tab

stack72:
The functionality allows listing remote modules only at this time. It also [temporarily] removes the ability to install the package or to manage a builtin


stack72:
bors merge

PR #3515, opened 2024-04-05
Implement macro to enforce strong typing/direction in edges

paulocsanz:
Creates `implement_add_edge_to` macro that defines functions to specify strongly typed connections with the direction hardcoded.

This is a quick win implementation to solve a couple of footguns the system had, it does not intend on fixing all footguns. Since now is not the time to workshop a perfect solution this quick win felt like the way to go.

This unblocks copy pasting since it enumerates all connections in the system and makes it greppable.

Please try to use the `implement_add_edge_to`  macro for new connections instead of `workspace_snapshot.add_edge` directly.

The current macro does not support all use cases, like category nodes, and definitely needs expansion.

It checks if edge weight kind in the function call matches the discriminant defined in the macro call.

Example:

Before:
```
    // Component (this) --> Schema Variant
    workspace_snapshot
        .add_edge(
            id, // Ulid
            EdgeWeight::new(change_set, EdgeWeightKind::new_use())?,
            schema_variant_id, // Ulid
        )
        .await?;
```

After:
```
    implement_add_edge_to!(
        source_id: ComponentId,
        destination_id: SchemaVariantId,
        add_fn: add_edge_to_schema_variant,
        discriminant: EdgeWeightKindDiscriminants::Use,
        result: ComponentResult,
    );

    // Enforces that the source is of type ComponentId and destination is of type SchemaVariantId
    // Runtime checks that EdgeWeightKind variant is Use
    Component::add_edge_to_schema_variant(
        ctx,
        component.id,
        schema_variant_id,
        EdgeWeightKind::new_use(),
    )
    .await?;
```

Feedback is needed here!

Benefits that this creates:
- Enforces connection direction
- Enforces connected objects types
- Runtime checks that the EdgeWeightKind passed is of the expected EdgeWeightKindDiscriminant
- Makes possible connections enumerable


jhelwig (dismissed):
I think we should sync up with a small group to figure out what we should do for this, and how far we need to take things right now. These changes make things harder to read and understand what's happening for me.

lib/dal/src/standard_connection.rs
```diff
@@ -0,0 +1,65 @@
...
+    #[error("transactions error: {0}")]
+    Transactions(#[from] TransactionsError),
+    #[error("workspace snapshot error: {0}")]
+    WorkspaceSnapshot(#[from] WorkspaceSnapshotError),
+}
+
+pub type HelperResult<T> = Result<T, HelperError>;
+
+#[macro_export]
+macro_rules! implement_add_edge_to {
```
vbustamante, line 22: I like this new name

lib/dal/src/socket/input.rs
```diff
@@ -238,26 +234,29 @@ impl InputSocket {
...
+        Self::add_edge_to_attribute_prototype(
+            ctx,
+            socket.id,
+            attribute_prototype.id(),
+            EdgeWeightKind::Prototype(None),
+        )
+        .await?;
 
-        Ok(Self::assemble(id.into(), content))
+        Ok(socket)
```
vbustamante, line 259: Why do some of these "create_edge + assemble and return"  blocks get reordered? Does that change functionality?

vbustamante (approved):
Looks good, just a question about some refactorings i didn't get 1000%

lib/dal/src/socket/input.rs
paulocsanz: Not really, it's mostly that we can use the typed socket.id instead of calling id.into(), it's also side-effect from the previous implementation that required `&self` to opperate. Since this changed almost 30 files I wasn't that thorough to ensure the diff was minimized.

But should not change functionality

vbustamante: ahh gotcha! looks good

lib/dal/src/standard_connection.rs
nickgerace: Same!

lib/dal/src/standard_connection.rs
```diff
@@ -0,0 +1,65 @@
...
+    InvalidEdgeWeight(EdgeWeightKind, EdgeWeightKindDiscriminants),
+    #[error("edge weight error: {0}")]
+    EdgeWeight(#[from] EdgeWeightError),
+    #[error("transactions error: {0}")]
+    Transactions(#[from] TransactionsError),
+    #[error("workspace snapshot error: {0}")]
+    WorkspaceSnapshot(#[from] WorkspaceSnapshotError),
+}
+
+pub type HelperResult<T> = Result<T, HelperError>;
```
nickgerace, line 19: This is a nitpick, but I wonder if we want to name this `AddEdgeError` or something similar? I'm wondering if it would help with traces and logging.

lib/dal/src/standard_connection.rs
```diff
@@ -0,0 +1,65 @@
...
+macro_rules! implement_add_edge_to {
+    (
+        source_id: $source_id:ty,
+        destination_id: $destination_id:ty,
+        add_fn: $add_fn:ident,
+        discriminant: $discriminant:expr,
+        result: $result:ty,
+    ) => {
+        paste::paste! {
+            /// Inserts edge from source to destination with specified weight to the graph
```
nickgerace, line 31: Nice!

lib/dal/tests/integration_test/old-engine-testing-layout/internal/attribute/prototype.rs
```diff
@@ -515,8 +515,7 @@ async fn remove_component_specific(ctx: &DalContext) {
                 assert!(
                     AttributePrototype::get_by_id(ctx, confirm_deletion_prototype_id)
                         .await
-                        .expect("could not get attribute prototype by id")
-                        .is_none()
```
nickgerace, line 519: This code isn't compiling right? Looks like a find and replace function caught it. We should probably remove this change, but it doesn't matter.

lib/dal/src/socket/input.rs
```diff
@@ -37,10 +38,14 @@ pub enum InputSocketError {
...
     #[error("func error: {0}")]
     Func(#[from] FuncError),
+    #[error("helper error: {0}")]
+    Helper(#[from] HelperError),
     #[error("layer db error: {0}")]
     LayerDb(#[from] LayerDbError),
     #[error("node weight error: {0}")]
     NodeWeight(#[from] NodeWeightError),
+    #[error("schema variant error: {0}")]
+    SchemaVariant(#[from] Box<SchemaVariantError>),
```
nickgerace, line 48: I see the `Box` usages in the PR. I like trying to stick to one pattern (my preference is to use neither `Box` nor `String`, but to create smaller error types), but is there a chance of infinite recursion with error types even with using `Box`?

lib/dal/src/standard_connection.rs
```diff
@@ -0,0 +1,65 @@
...
+            pub async fn $add_fn(ctx: &$crate::DalContext, source_id: $source_id, destination_id: $destination_id, weight: $crate::EdgeWeightKind) -> $result<()> {
+                if $crate::EdgeWeightKindDiscriminants::from(weight.clone()) != $discriminant {
```
nickgerace, line 33: Do we have to clone here?

lib/dal/tests/integration_test/old-engine-testing-layout/internal/attribute/prototype.rs
paulocsanz: Nope, we were returning Option in get_by_id, but we always returned Some. So I made AttributePrototype::get_by_id work like other get_by_ids in the new engine, returning an error if missing. And this was the only place looking for a non existant prototype. So instead of none is error.

lib/dal/src/socket/input.rs
paulocsanz: String is not a good pattern, Box has no risk of infinite recursion. It will hold the depth just like String. We should not be using String as it removes context for no added safety.

This is a very common pattern for the rust community

lib/dal/src/standard_connection.rs
paulocsanz: Good question, no it doesn't. Fixed! Thank you for the thorough review, this is not a small PR.

paulocsanz:
bors try

lib/dal/src/socket/input.rs
nickgerace: Thanks for the explanation!

paulocsanz:
bors merge

PR #3519, opened 2024-04-05
feat(dal): Adding a new CategoryNode for Module

stack72:
ENG-2447

This allows us to start creating Modules on the graph


stack72:
bors merge

PR #3520, opened 2024-04-05
feat(dal): Adds a Module Node to the graph for each of the installed builtins

stack72:
This isn't connected to anything at this time. I have also not removed the InstalledPkgs impl OR the storing in the old database structure

### DIFF
--- a/lib/dal/src/layer_db_types/content_types.rs
+++ b/lib/dal/src/layer_db_types/content_types.rs
@@ -32,6 +32,7 @@ pub enum ContentTypes {
     Func(FuncContent),
     FuncArgument(FuncArgumentContent),
     InputSocket(InputSocketContent),
+    Module(ModuleContent),
     Prop(PropContent),
     Schema(SchemaContent),
     SchemaVariant(SchemaVariantContent),
@@ -85,6 +86,7 @@ impl_into_content_types!(Func);
 impl_into_content_types!(FuncArgument);
 impl_into_content_types!(InputSocket);
 impl_into_content_types!(OutputSocket);
+impl_into_content_types!(Module);
 impl_into_content_types!(Prop);
 impl_into_content_types!(Schema);
 impl_into_content_types!(SchemaVariant);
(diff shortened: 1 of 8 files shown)


stack72:
bors try

lib/dal/src/layer_db_types/content_types.rs
```diff
@@ -258,6 +260,18 @@ pub struct InputSocketContentV1 {
     pub connection_annotations: Vec<ConnectionAnnotation>,
 }
 
+#[derive(Debug, Clone, EnumDiscriminants, Serialize, Deserialize, PartialEq)]
+pub enum ModuleContent {
+    V1(ModuleContentV1),
+}
+
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq)]
+pub struct ModuleContentV1 {
```
britmyerss, line 269: do we also need: 
`display_name` `name` etc?

stack72: That's not part of the module - that's part of the Schema / Schema Variant. The module just has a name and a root_hash.

britmyerss: `export interface LocalModuleSummary {
  name: string;
  hash: ModuleHash;
  isBuiltin: boolean;
}

export interface LocalModuleDetails {
  name: string;
  version: string;
  description: string;
  createdAt: IsoDateString;
  createdBy: string;
  schemas: string[];
  funcs: PkgFuncView[];
  hash: ModuleHash;
  kind: "module" | "workspaceExport";
}`
I think there's a name/description/etc. for that too - here's from app/web

nickgerace: I wonder if modules implicitly only have one schema (not saying how it should always be, just how it is now) which is influencing what's returned in that route?

britmyerss: I don't think so.. even now we have multiple schemas in a module

stack72:
bors try

stack72:
bors merge

PR #3521, opened 2024-04-05
Rename existing Actions to make room for new Actions architecture

jhelwig:
In order to keep the ability to run actions while the new Actions architecture is implemented, this renames most of the old things to `DeprecatedAction...`. We will still need to adjust things like API routes when it comes time to implement them on top of the new architecture, but this will largely allow us to avoid having to do a big delete + rename at the end to avoid keeping an `ActionNew` around.


jobelenus (approved):

paulocsanz:
bors merge

PR #3522, opened 2024-04-05
feat(dal): Add an edge from Module to package contents on import

stack72:
This will allow us to be able to understand what makes up a package at import time - we are creating connections to funcs, schemas and schema variants

There are still some TODOs in this PR:

* Before we import funcs, schemas or schema variants, we need to check the graph to see if there's already an edge to them

That will be followed up with as it doesn't change the way that the import_pkg works


nickgerace (approved):

stack72:
bors try

stack72:
bors try

stack72:
bors merge

PR #3523, opened 2024-04-06
Ensure updated encrypted secrets trigger dependent values update

nickgerace:
## Description

Currently, if we update the encrypted contents of a secret (or any contents of a secret), all `AttributeValues` dependent on the source `AttributeValue` containing the corresponding `SecretId` will not be updated. Example: a qualification that ensures credentials are valid may depend on the `AttributeValue` in the `"/root/secrets"` tree corresponding to a specific `Secret`. If you update the `Secret`, the `SecretId` contained within the `AttributeValue` stays the same, so the qualification is not re-ran.

With this PR, only `AttributeValues` that _directly_ contain the `SecretId` for a given `Secret` have an outgoing edge to that `Secret`. This creates an interesting "one off" design where updating property editor values for a Secret results in a slightly different backend call. Given that `Secrets` already exhibit divergent behavior in the frontend based on the `Prop`'s widget, this behavior was trivial to sandbox.

## Accepted Risk

As mentioned above, this is a "one off" scenario for updating property editor values where SDF has to know to call `AttributeValue::update_for_secret` instead of `AttributeValue::update`.

Why do we need this? The following must be true with the current `Secrets` design in the new engine: if we want updating encrypted contents to trigger dependent values update for relevant values, **_there must be an edge that connects the `Secret` to  the directly set `AttributeValue` or to a uniquely identifying node for that `AttributeValue` (e.g. its `AttributePrototype`)_**.

Thus, the solution in this PR accepts the risk of having two update calls since `AttributeValues` for `Secrets` must have different treatment _somewhere_ in the system given the aforementioned constraint.

### DIFF
--- a/app/web/src/components/AttributesPanel/AttributesPanel.vue
+++ b/app/web/src/components/AttributesPanel/AttributesPanel.vue
@@ -178,6 +178,7 @@ function updateSiProp(key: keyof typeof siValues) {
       propId: prop.propId,
       componentId: component.id,
       value: newVal,
+      isForSecret: false,
     },
   });
 }
--- a/app/web/src/components/AttributesPanel/AttributesPanelItem.vue
+++ b/app/web/src/components/AttributesPanel/AttributesPanelItem.vue
@@ -809,6 +809,7 @@ const setSource = (source: AttributeValueSource) => {
         propId: props.attributeDef.propId,
         componentId,
         value,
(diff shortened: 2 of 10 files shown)


nickgerace:
bors merge

nickgerace:
bors cancel

nickgerace:
To return soon!

PR #3525, opened 2024-04-07
feat(dal, sdf): Update list_pkgs to use Modules on the graph

stack72:
We are going to read the modules on the graph as the list packages now

Screenshot 2024-04-07 at 21 14 53


stack72:
bors merge

PR #3526, opened 2024-04-07
feat(dal, sdf): Reinstate get_pkg using the Modules on the graph

stack72:
This used to load the package from disk and we don't need to do that anymore now that we have all the details on the graph


stack72:
bors merge

PR #3528, opened 2024-04-08
chore(layerdb): performance and correctness fixes

adamhjk:
This rebuilds the activity and event systems to better use NATS jetstream. We were previously wrapping this up in streams, which was causing us to not correctly deal with things like ack's, and perhaps other small details that existed in the client side (like puback's). This was causing us to fall afoul of Jetstreams internal checks, which would make things no longer work. We now pull the streams directly, and publish the results down channels. The end result is you can't really make mistakes on things like acking, and everything seems to work as the jetstream docs say it should.

Additionally, redb was simply too slow. After thinking about why, I realized we kept trying to use a database, when all we really needed was a file cache. So I switched to one, and now disk cache performance is stable and easy to replicate with something as simple as rsync if we want to.

I also added a pair of stress tests for the LayerEvent stream (cache update events) and RebaseRequests in the work queue.

The LayerEvent tests are focused on large payloads - we send 35 CAS events with a payload of ~10mb each (350mb total). The test completes in 3.5 seconds on my machine.

The RebaseRequest stress test is more about pure volume, since I don't want to fire up a rebaser and actually do any work. This test measures how long it takes for two consumers to pull 10,000 rebase requests. It fails if it takes more than 10 seconds. It currenctly takes more like 3.


adamhjk:
bors merge

PR #3529, opened 2024-04-08
feat(sdf, dal): Restore install_pkg API Endpoint

stack72:
There may be some reactivity things that we need to follow up on but this is largely working now

### DIFF
--- a/lib/sdf-server/src/server/service/pkg/install_pkg.rs
+++ b/lib/sdf-server/src/server/service/pkg/install_pkg.rs
@@ -1,22 +1,21 @@
-use std::str::FromStr;
+use axum::extract::OriginalUri;
+use axum::http::Uri;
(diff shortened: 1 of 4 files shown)


lib/sdf-server/src/server/service/pkg/install_pkg.rs
```diff
@@ -86,8 +85,8 @@ pub async fn install_pkg(
 async fn install_pkg_inner(
```
britmyerss, line 85: This can happen in a separate pass (no need to block the PR) - but would be good to pull the module nomenclature into these endpoints

stack72: yeah my plan was to rename these to *_module ;) That PR will follow shortly!

stack72:
bors merge

PR #3530, opened 2024-04-08
chore(sdf): Rename pkg routes to be modules routes

stack72:
This means all of the /pkg/ are not /module/ as this is more in keeping with our nomenclature


stack72:
bors merge

PR #3533, opened 2024-04-08
chore(rebaser): fix graceful shutdown

adamhjk:
The rebaser was failing to shut down when it was killed from a signal rather than a straight ctrl-c. The side effect is that you likely have multiple rebasers running.


stack72:
bors merge

PR #3534, opened 2024-04-08
Summarize and restyle proposed actions modal

jobelenus:
NOTE: the 3 sets of icons is mocked to test the styles works when we've got multiple things in the list. There is really only one action, as per the rest of the screenshot.

image
styling-dog

### DIFF
--- a/app/web/src/components/ApplyChangeSetButton.vue
+++ b/app/web/src/components/ApplyChangeSetButton.vue
@@ -40,7 +40,7 @@
                 size="sm"
                 tone="success"
                 loadingText="Applying Changes"
-                label="Override Approval And Apply"
+                label="Continue"
                 :requestStatus="applyChangeSetReqStatus"
                 :disabled="statusStoreUpdating"
                 @click="applyChangeSet"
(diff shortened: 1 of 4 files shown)


app/web/src/utils/defaultmap.ts
wendybujalski, line 1: useful, love it

app/web/src/components/ApplyChangeSetButton.vue
wendybujalski, line 1: Looks good! Only one minor note - try to avoid arbitrary and numeric tailwind class styles when possible. Our predefined semantic sizes _(2xs, xs, sm, md, lg, xl, 2xl)_ are easier to understand and keep consistent

For example, `mr-2` can just be `mr-xs` and `mx-1` can just be `mx-2xs`

As for `mx-[.66em]` I assume there was a good reason you needed that specific margin size!

wendybujalski (approved):
Awesome work, looks great!

app/web/src/components/ApplyChangeSetButton.vue
jobelenus: Yes, that one was annoying. With the `-1` or `-2` the edges didn't line up with the elements directly below it :twitchingeye:

jobelenus:
bors merge

PR #3536, opened 2024-04-09
bug(rebaser): only rebase one request at a time per change set

adamhjk:
This PR fixes the issue where the rebaser would rebase things as fast as it possibly could. Instead, it creates a queue for every change set ID it sees, and only processes one message at a time for each change set.

It currently doesn't support removing those queues - once we see it, the worker will stay resident forever. That's probably not a big deal, given that we mostly have single users and the proliferation of change sets isn't so high. But it's obviously not the optimal solution for long.

This is a stop-gap PR while Fletcher works on the long term solution, which likely involves some clever use of Jetstream's tuning options to ensure only one message can be in flight at a time, and that those messages get replayed if needed.

### DIFF
--- a/lib/rebaser-server/src/server/core_loop.rs
+++ b/lib/rebaser-server/src/server/core_loop.rs
@@ -10,12 +10,16 @@ use si_data_pg::PgPool;
 use si_layer_cache::activities::rebase::RebaseStatus;
 use si_layer_cache::activities::ActivityRebaseRequest;
 use si_layer_cache::LayerDbError;
+
+use std::collections::HashMap;
 use std::sync::Arc;
 use std::time::Instant;
+use tokio_util::task::TaskTracker;
+use ulid::Ulid;
 
 use telemetry::prelude::*;
 use thiserror::Error;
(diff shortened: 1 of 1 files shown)


adamhjk:
bors merge

lib/rebaser-server/src/server/core_loop.rs
```diff
@@ -71,24 +75,27 @@ pub(crate) async fn setup_and_run_core_loop(
...
     info!("setup complete, entering core loop");
-    core_loop_infallible(ctx_builder, stream, shutdown_watch_rx).await;
+    run_me(ctx_builder, stream, shutdown_watch_rx).await;
     info!("exited core loop");
 
     Ok(())
 }
 
-async fn core_loop_infallible(
+async fn run_me(
```
nickgerace, line 84: I much prefer my old name for it

adamhjk: I mean, I think your name is probably better! I had meant to refactor it, actually. :)

PR #3537, opened 2024-04-09
WIP: feat(dal): Move resourceToPayload to intrinsics

stack72:
We use this func everywhere so we are moving it to be an intrinsic

### DIFF
--- a/lib/dal/src/pkg/import.rs
+++ b/lib/dal/src/pkg/import.rs
@@ -59,7 +59,7 @@ pub struct ImportOptions {
     pub is_builtin: bool,
 }
 
-const SPECIAL_CASE_FUNCS: [&str; 2] = ["si:resourcePayloadToValue", "si:normalizeToArray"];
+const SPECIAL_CASE_FUNCS: [&str; 1] = ["si:normalizeToArray"];
 
 #[allow(clippy::too_many_arguments)]
 async fn import_change_set(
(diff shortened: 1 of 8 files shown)


nickgerace:
bors merge

nickgerace:
bors cancel

lib/dal/src/pkg/import.rs
```diff
@@ -88,6 +88,8 @@ async fn import_change_set(
             || SPECIAL_CASE_FUNCS.contains(&func_spec.name())
             || func_spec.is_from_builtin().unwrap_or(false)
         {
+            dbg!("Is an intrinsic");
+            dbg!(&func_spec.name());
```
nickgerace, line 92: We should probably remove these

lib/dal/src/pkg/import.rs
```diff
@@ -59,7 +59,7 @@ pub struct ImportOptions {
     pub is_builtin: bool,
 }
 
-const SPECIAL_CASE_FUNCS: [&str; 2] = ["si:resourcePayloadToValue", "si:normalizeToArray"];
+const SPECIAL_CASE_FUNCS: [&str; 1] = ["si:normalizeToArray"];
```
nickgerace, line 62: Hmmm maybe this should be an intrinsic too in another PR

stack72:
Spoke with @zacharyhamm - we need to make some changes here to build a backend for this func a la https://github.com/systeminit/si/tree/main/lib/dal/src/func/backend

PR #3540, opened 2024-04-09
feat: use xxh3 hash instead of blake3 for merkle tree hash and content hashes

zacharyhamm:
Use a 128 bit hash instead of a 256 bit one. Removes a few megabytes from the initial graph.

NOTE: this is a breaking change, you'll need to clear your databases and start over.


zacharyhamm:
bors try

zacharyhamm:
bors retry

johnrwatson:
bors retry

PR #3543, opened 2024-04-09
chore(dal): Simplify test migrations

vbustamante:
import_pkg by default will import all schemas in pkg if the list is not passed, so we can get away with passing the default import options on tests.


vbustamante:
bors merge

PR #3544, opened 2024-04-09
chore(dal): Remove installed_pkgs migration

stack72:
This is no longer needed since things are now on the graph


stack72:
bors merge

PR #3548, opened 2024-04-10
Find func associations for code generation and qualification funcs

nickgerace:
## Description

This PR adds the ability to find func associations for code generation and qualification funcs. It may have also done it for attribute funcs, but that remains to be seen.

There's one integration test added as part of this commit, but it's hardly enough for the authoring domain. Baby steps.

### DIFF
--- a/lib/dal/src/prop.rs
+++ b/lib/dal/src/prop.rs
@@ -351,10 +351,13 @@ impl Prop {
         self.id
     }
 
+    /// Returns `Some` with the parent [`PropId`](Prop) or returns `None` if the parent is a
+    /// [`SchemaVariant`].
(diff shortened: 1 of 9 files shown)


nickgerace:
bors merge

lib/dal/src/prop.rs
```diff
@@ -375,7 +378,7 @@ impl Prop {
                     _ => return Err(PropError::PropParentInvalid(prop_id)),
                 },
             ),
-            None => Ok(None),
+            None => Err(PropError::PropIsOrphan(prop_id)),
```
nickgerace, line 381: Found this along the way. Could've been painful in the future, now there's an error.

PR #3554, opened 2024-04-10
Restore func associations for actions

nickgerace:
## Description

This PR restores func associations for actions. In addition, it refactors the schema variant domain to make the core module easier to navigate. That was fueled by listing schema variants for auth funcs moving to the `SchemaVariant` impl as well as the addition of a new function to list schema variants for action funcs.

This PR also includes an integration test for each association. They're non-comprehensive, but are a step in the right direction.


nickgerace:
bors merge

PR #3556, opened 2024-04-10
Ignore deleted values in DVU

paulocsanz:
Ideally we should be dequeueing values that don't exist, but that's a can of worms since we can't ensure it will happen, as removing some edge may delete something because it gets garbage collected.

This should be ok, and not hide any bugs, and the warning is there so we can be informed of further consequence. Right now it will pretty much only happen in tests and copy pasting.

If the warning gets too verbose, we can remove it, or downgrade to a debug statement, but let's see how it behaves first.

### DIFF
--- a/lib/dal/src/attribute/value/dependent_value_graph.rs
+++ b/lib/dal/src/attribute/value/dependent_value_graph.rs
@@ -1,5 +1,6 @@
 use petgraph::prelude::*;
 use std::collections::{hash_map::Entry, HashMap, VecDeque};
+use telemetry::prelude::*;
 use tokio::{fs::File, io::AsyncWriteExt};
 use ulid::Ulid;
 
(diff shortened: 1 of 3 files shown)


jhelwig:
Ignoring missing AttributeValues from the initial set that's passed in makes sense, but there should never be missing AttributeValues that are part of the work-queue, which comes from generating the graph of downstream values.

paulocsanz:
bors try

lib/dal/src/attribute/value/dependent_value_graph.rs
```diff
@@ -51,8 +52,19 @@ impl DependentValueGraph {
...
         while let Some(current_attribute_value_id) = work_queue.pop_front() {
+            if workspace_snapshot
+                .try_get_node_index_by_id(current_attribute_value_id)
+                .await?
+                .is_none()
+            {
+                warn!("Attribute Value {current_attribute_value_id} missing, skipping it in DependentValueGraph");
+                continue;
+            }
+
```
jhelwig, line 67: Please add a comment here along the lines of
```
// It's possible that one or more of the initial AttributeValueIds provided by the enqueued DependentValuesUpdate job
// may have been removed from the snapshot between when the DVU job was created and when we're processing
// things now. This could happen if there are other modifications to the snapshot before the DVU job starts
// executing, as the job always operates on the current state of the change set's snapshot, not the state at the time
// the job was created.
```

I don't think the warning is necessary. It could be dropped down to `debug`.

jhelwig (changes requested):

lib/dal/src/attribute/value/dependent_value_graph.rs
paulocsanz: Done

jhelwig (approved):

paulocsanz:
bors merge

PR #3557, opened 2024-04-10
Add Function Execution DB 

sprutton1:
This yet another db inside of layer db. What's special about this one is that it only interacts with postgres and does not use the in-memory or on-disk layers at all (so no NATs bits either).  Also, I'm attempting to use an ugly version of a composite key to ensure we can look these executions up from a few different angles. Currently, the key is `component_id + prototype_id + action_id` . This allows us to do lookups for specific actions and lookups for all action executions that ever happened for a given component/prototype. I _think_ there is a cleaner way to do this, so I'm all ears if anyone has thoughts.

Note that this doesn't implement the actual struct that will be stored. That will come next.

### DIFF
--- a/lib/dal/src/context.rs
+++ b/lib/dal/src/context.rs
@@ -37,7 +37,7 @@ use crate::{
 };
 use crate::{EncryptedSecret, Workspace};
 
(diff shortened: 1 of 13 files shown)


sprutton1:
bors try

sprutton1:
bors merge

PR #3561, opened 2024-04-11
fix(web): changesetId on modules store

vbustamante:
The module modals were being instantiated at load, and never getting the updated changeset  (since we need to await the fetch_changesets call to set the active changeset id - we get the id from the URL but need to confirm it before calling requests) - thus module requests were sending in the nil changeset id ("0000...") and failing

While we could get the changeset id from the url directly, the way that it is setup makes it so that setting selectedChangesetId based on the URL directly redirects to HEAD automatically instead of throwing an error when the url has an invalid changeset ID. 

The solution on this PR makes the calls functional, and it would be better to move the visibility params to something implicit in the API request lib anyway, which this code sort of points towards.

### DIFF
--- a/app/web/src/store/module.store.ts
+++ b/app/web/src/store/module.store.ts
@@ -2,9 +2,8 @@ import { defineStore } from "pinia";
 import * as _ from "lodash-es";
 import { addStoreHooks, ApiRequest } from "@si/vue-lib/pinia";
 import { DiagramInputSocket, DiagramOutputSocket } from "@/api/sdf/dal/diagram";
-import { Visibility } from "@/api/sdf/dal/visibility";
-import { nilId } from "@/utils/nilId";
 import { useWorkspacesStore } from "@/store/workspaces.store";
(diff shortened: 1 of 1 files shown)


app/web/src/store/module.store.ts
```diff
@@ -101,12 +100,27 @@ export type RemoteModuleDetails = RemoteModuleSummary & {
   };
 };
 
+// Gather the current changeset ID, since our components don't appear to be
+// reacting to the visibility changes that happen on the setup code
+// TODO: Generalize this and make the api client pass these arguments implicitly
+async function getVisibilityParams(forceChangeSetId?: ChangeSetId) {
```
jobelenus, line 106: nit: doesn't need to be `async`?

vbustamante: True. This is leftover from a version that would await for some requests and didn't end up getting implemented. Fixed it

vbustamante:
bors merge

PR #3567, opened 2024-04-11
bug(*): need change set written after all

adamhjk:
This restores the change set written event, which it turns out we do need - the rebaser still handles when data is "committed", and therefore when it is visible in different parts of the UI. Long term, it would be great to figure out how to thread more specific events through, so that we don't have a cascade of calls that are needless from the frontend. But for now, this will work.


adamhjk:
bors merge

PR #3574, opened 2024-04-12
Remove `PropParent` and fix `SchemaVariant::all_funcs`

nickgerace:
## Description

This PR removes the `PropParent` enum and replaces it with explicit methods. In addition, `Props` can now only have "ordered" `Prop` parents, which has been the only way we've created `Props` anyway.

This commit also fixes `SchemaVariant::all_funcs` to account for sockets. The method has been optimized it to batch load funcs after all `FuncIds` have been collected.

## Misc Changes

- Remove unneeded `dbg!` from the rebaser
- Remove unused `import_builtins` method from the old engine
- Allow `AsRef<str>` parameter for `IntrinsicFunc::maybe_from_str`
- Tweak ergonomics for downstream methods related to `SchemaVariant::all_funcs`
- Replace remaining references to providers in import logic
- Add `find_for_prop` integration test for `AttributePrototype` in addition to tests related to `SchemaVariant::all_funcs`

### DIFF
--- a/lib/dal/src/context.rs
+++ b/lib/dal/src/context.rs
@@ -830,23 +830,6 @@ impl DalContext {
         Ok(is_in_our_tenancy)
     }
 
-    /// Copies every single row from `Self::builtin()` to our tenancy on head change-set
-    /// Needed to remove universal tenancy while packages aren't a thing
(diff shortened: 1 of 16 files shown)


nickgerace:
bors merge

PR #3575, opened 2024-04-12
feat: separate func executions and messages into different tables

sprutton1:
Okay, we have now broken out the `FuncExecutions` from the `FuncExecutionMessages`. This allows us to get the status of a given set of executions without having also retrieve the arbitrarily-sized log messages. We are somewhat breaking the key->value layer-db paradigm a bit with `FuncExecutions`. The key and value still exist, but we also have other keys that we believe will be relevant for filtering (give me all executions for a given changeset, component, etc.).

The structs created to represent these two things are somewhat hokey at the moment, but I didn't want to take them much further until we better understand how this will be used by nodes on the graph. The pieces here should be enough to extend to a glorious future.


sprutton1:
bors try

lib/si-layer-cache/src/db.rs
```diff
@@ -17,28 +17,28 @@ use crate::{
...
 pub mod cas;
 pub mod encrypted_secret;
-pub mod function_execution;
+pub mod func_execution;
 pub mod serialize;
 pub mod workspace_snapshot;
 
 #[derive(Debug, Clone)]
-pub struct LayerDb<CasValue, EncryptedSecretValue, FunctionExecutionValue, WorkspaceSnapshotValue>
+pub struct LayerDb<CasValue, EncryptedSecretValue, FuncExecutionValue, WorkspaceSnapshotValue>
```
adamhjk, line 32: Why do we need this to be a generic bound? if it only holds one FuncExectionValue, then it doesn't need to be another generic parameter.

lib/si-layer-cache/src/db.rs
```diff
@@ -47,12 +47,12 @@ where
     instance_id: Ulid,
 }
 
-impl<CasValue, EncryptedSecretValue, FunctionExecutionValue, WorkspaceSnapshotValue>
-    LayerDb<CasValue, EncryptedSecretValue, FunctionExecutionValue, WorkspaceSnapshotValue>
+impl<CasValue, EncryptedSecretValue, FuncExecutionValue, WorkspaceSnapshotValue>
+    LayerDb<CasValue, EncryptedSecretValue, FuncExecutionValue, WorkspaceSnapshotValue>
```
adamhjk, line 51: Same here

lib/si-layer-cache/src/db/func_execution.rs
```diff
@@ -0,0 +1,213 @@
...
+
+pub const PARTITION_KEY: &str = KEYWORD_PLURAL;
+pub const DBNAME: &str = KEYWORD_PLURAL;
+pub const CACHE_NAME: &str = KEYWORD_PLURAL;
+pub const SORT_KEY: &str = KEYWORD_SINGULAR;
+pub const EXECUTIONS_TABLE_NAME: &str = KEYWORD_PLURAL;
+pub const MESSAGES_TABLE_NAME: &str = "func_execution_messages";
+
+#[derive(Debug, Clone)]
+pub struct FuncExecutionDb<V>
```
adamhjk, line 21: This doesn't need to be generic. I

lib/si-layer-cache/src/db/func_execution.rs
```diff
@@ -0,0 +1,213 @@
...
+where
+    V: Serialize + DeserializeOwned + Clone + Send + Sync + 'static,
+{
+    pub cache: LayerCache<Arc<V>>,
```
adamhjk, line 25: This should be LayerCache<Arc<FuncExecution> or whatnot

PR #3576, opened 2024-04-12
fix: remove unnecessary gerneric bounds for func execution

sprutton1:
We don't actually need the bounds around the func execution db to be generic as we know the type we want to use and we don't interact with the DB layers that need it.


sprutton1:
bors try

sprutton1:
bors merge

PR #3577, opened 2024-04-12
Reimplement copy-paste in the new engine

paulocsanz:


lib/dal/src/component.rs
```diff
@@ -1017,10 +1181,17 @@ impl Component {
...
         destination_component_id: ComponentId,
         destination_input_socket_id: InputSocketId,
-    ) -> ComponentResult<(AttributeValueId, AttributePrototypeArgumentId)> {
+    ) -> ComponentResult<Option<(AttributeValueId, AttributePrototypeArgumentId)>> {
+        let destination_component = Component::get_by_id(ctx, destination_component_id).await?;
+        for connection in destination_component.incoming_connections(ctx).await? {
+            if connection.from_component_id == source_component_id && connection.from_output_socket_id == source_output_socket_id && connection.to_component_id == destination_component_id && connection.to_input_socket_id == destination_input_socket_id {
+                return Ok(None);
+            }
+        }
```
nickgerace, line 1193: Nice! I presume this solved the duplicate edges bug?

lib/dal/src/attribute/prototype.rs
```diff
@@ -580,4 +580,81 @@ impl AttributePrototype {
...
+                        .into());
+                    }
+                },
+                other => {
+                    return Err(WorkspaceSnapshotError::UnexpectedEdgeSource(
+                        other.id(),
+                        id.into(),
+                        EdgeWeightKindDiscriminants::Prototype,
+                    )
+                    .into());
```
nickgerace, line 641: I like the checks!

lib/dal/src/component.rs
```diff
@@ -1403,6 +1579,59 @@ impl Component {
...
+                    message: None,
+                    logs: Vec::new(),
+                    last_synced: Some(Utc::now().to_rfc3339()),
+                },
+            )
+            .await?;
+
+        pasted_comp
+            .set_name(ctx, &format!("{} - Copy", self.name(ctx).await?))
+            .await?;
```
nickgerace, line 1618: I could be wrong, but won't the component's name be `Copy - Copy` with how this is written?

lib/dal/src/component/debug.rs
```diff
@@ -215,74 +212,6 @@ impl ComponentDebugData {
...
-                // Load the work queue with the child attribute value.
-                work_queue.push_back(child_attribute_value_id);
-
-                // Cache the  prop values to eventually insert into the child property editor values map.
-                child_attribute_value_ids.push(child_attribute_value.id());
-            }
-            child_values.insert(attribute_value_id, child_attribute_value_ids);
-        }
-        Ok(child_values)
+        Ok(AttributeValue::tree_for_component(ctx, component_id).await?)
```
nickgerace, line 215: Nice!

lib/dal/tests/integration_test/component.rs
```diff
@@ -906,3 +910,81 @@ async fn undoing_deletion_updates_inputs(ctx: &mut DalContext) {
...
+    let mut view = pasted_pirate_component
+        .materialized_view(ctx)
+        .await
+        .expect("unable to get materialized view of component")
+        .expect("no view found");
+
+    // Non deterministic
+    *view
+        .pointer_mut("/resource/last_synced")
+        .expect("no resource last_synced found") = serde_json::Value::Null;
```
nickgerace, line 965: The whole test is non deterministic or just this value?

As an aside, we probably want to restore the old `ComponentViewProperties` builder (or something like it) that could drop individual fields, if you remember it. It was a way to remove timestamp fields and non-deterministic fields from the `ComponentView`. We probably want something similar for the materialized view.

stack72:
Ok, I took this for a test drive and found the following:

* Sometimes that positions get. bit strange for the pasted components - this isn't hugely important, it was just confusing

* The Qualifications for the copied component are not the correct ones:
A keypair has 2 qualifications:
- Prop Validations
- Can create Key Pair

Screenshot 2024-04-19 at 11 29 07

When I copy a keypair, the qualifications tab shows me:
- Prop Validations
- si:setObject

Screenshot 2024-04-19 at 11 28 58

The codegen doesn't change to reflect the name of the new component - even when I change it, codegen doesn't re-run

Screenshot 2024-04-19 at 11 26 59

lib/dal/tests/integration_test/component.rs
paulocsanz: last_synced uses the clock, so it may not be deterministic

It would be cool doing that with materialized view yeah

paulocsanz: Removed the entire resource

lib/dal/src/attribute/value.rs
```diff
@@ -428,50 +427,24 @@ impl AttributeValue {
...
 
-        if spawn_dependent_values_update {
-            ctx.enqueue_dependent_values_update(vec![attribute_value_id])
-                .await?;
-        }
+        ctx.enqueue_dependent_values_update(vec![attribute_value_id])
+            .await?;
 
         Ok(())
     }
```
nickgerace, line 446: I'd explicitly call this out in the commit message including why we did it.

nickgerace (approved):

paulocsanz:
bors merge

PR #3578, opened 2024-04-12
Re-introduce saving attribute funcs in the new engine

nickgerace:
## Description

Currently, `save_func` works in the new engine. However, we need it to work for attribute funcs and advanced use cases. This commit is a step in that direction with the goal of ensuring of collecting and saving func associations works when saving an attribute func.

## Relevant Changes

- Fix func argument mutation, removal and additional procedures when working with attribute funcs
- Fix action kind selection in the authoring screen
- Add integration tests for all func saving scenarios (with attribute funcs getting a full-length test)
- Fix code gen and qualification input checkboxes

## Misc Changes

- Convert stdout lines to logs when importing packages
- Make errors clearer when importing packages
- Replace default schema variant log with debug log

## Known Issues

- Changing the type of a func argument for an attribute func does not work
- Changing "run on assets of type" and similar values does not work
- Using the attribute bindings panel for attribute funcs may not work as intended
- Changing builtin funcs is allowed

### DIFF
--- a/lib/dal/src/attribute/prototype.rs
+++ b/lib/dal/src/attribute/prototype.rs
@@ -34,7 +34,7 @@ use crate::workspace_snapshot::node_weight::{
 };
 use crate::workspace_snapshot::WorkspaceSnapshotError;
 use crate::{
-    attribute::prototype::argument::AttributePrototypeArgumentId, id, implement_add_edge_to,
+    attribute::prototype::argument::AttributePrototypeArgumentId, implement_add_edge_to, pk,
     AttributeValue, AttributeValueId, ComponentId, DalContext, FuncId, HelperError, InputSocketId,
     OutputSocketId, PropId, SchemaVariant, SchemaVariantError, SchemaVariantId, Timestamp,
     TransactionsError,
@@ -98,7 +98,9 @@ pub enum AttributePrototypeEventualParent {
     SchemaVariantFromProp(SchemaVariantId, PropId),
 }
 
-id!(AttributePrototypeId);
+// TODO(nick): switch to the "id!" macro once the frontend doesn't use the old nil id to indicate
+// that the argument is a new one.
+pk!(AttributePrototypeId);
 
 #[derive(Deserialize, Serialize, Debug, Clone, PartialEq, Eq)]
 pub struct AttributePrototype {
--- a/lib/dal/src/func/associations.rs
+++ b/lib/dal/src/func/associations.rs
@@ -1,20 +1,24 @@
 use serde::{Deserialize, Serialize};
+use strum::EnumDiscriminants;
 use telemetry::prelude::*;
 use thiserror::Error;
 
-use crate::attribute::prototype::argument::value_source::ValueSource;
-use crate::attribute::prototype::argument::{
-    AttributePrototypeArgument, AttributePrototypeArgumentError, AttributePrototypeArgumentId,
-};
+use crate::attribute::prototype::argument::AttributePrototypeArgumentError;
 use crate::attribute::prototype::{AttributePrototypeError, AttributePrototypeEventualParent};
-use crate::func::argument::{FuncArgument, FuncArgumentError, FuncArgumentId};
+use crate::func::argument::{FuncArgument, FuncArgumentError};
 use crate::func::view::FuncArgumentView;
 use crate::func::FuncKind;
 use crate::schema::variant::leaves::LeafInputLocation;
 use crate::{
-    AttributePrototype, AttributePrototypeId, ComponentId, DalContext, Func, FuncId, InputSocketId,
-    OutputSocketId, PropId, SchemaVariant, SchemaVariantError, SchemaVariantId,
+    AttributePrototype, ComponentId, DalContext, DeprecatedActionKind, DeprecatedActionPrototype,
+    DeprecatedActionPrototypeError, Func, FuncId, SchemaVariant, SchemaVariantError,
+    SchemaVariantId,
 };
+
+mod view;
+
+pub use view::AttributePrototypeArgumentView;
+pub use view::AttributePrototypeView;
 
 #[remain::sorted]
 #[derive(Error, Debug)]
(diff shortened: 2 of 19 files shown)


lib/dal/src/attribute/prototype.rs
```diff
@@ -98,7 +98,9 @@ pub enum AttributePrototypeEventualParent {
     SchemaVariantFromProp(SchemaVariantId, PropId),
 }
 
-id!(AttributePrototypeId);
+// TODO(nick): switch to the "id!" macro once the frontend doesn't use the old nil id to indicate
+// that the argument is a new one.
+pk!(AttributePrototypeId);
```
nickgerace, line 103: You'll see this a few times in the diff. Essentially, the frontend indicates that the func is new by using the `nilId`. For nodes on the graph, we've scrubbed the usage of `nilId` where possible, but we need we will be friends with it for now since we are trying to touch func authoring flow as little as possible.

lib/dal/src/func/associations.rs
```diff
@@ -166,8 +87,25 @@ impl FuncAssociations {
...
+                // TODO(nick): right now, we just grab the first one and it decides the action kind for all of them.
+                // This should be configurable on a "per prototype" basis in the future.
+                let mut kind = DeprecatedActionKind::Create;
+                for action_prototype_id in action_prototype_ids {
+                    let action_prototype =
+                        DeprecatedActionPrototype::get_by_id_or_error(ctx, action_prototype_id)
+                            .await?;
+                    kind = action_prototype.kind;
+                    break;
+                }
```
nickgerace, line 102: I don't love this because different action prototypes can technically have different action kinds for the same func. However, this is how it worked in the old engine and we are in the middle of an actions re-write, so restoring old behavior is fine for now.

lib/dal/src/func/authoring.rs
```diff
@@ -1,6 +1,16 @@
...
+//! [`FuncAssociations`] subsystem is less so. Essentially, [`FuncAssociations`] are a "bag" that
+//! come alongside [`FuncView::assemble`](crate::func::view::FuncView::assemble) and are mutated via
+//! [`FuncAuthoringClient::save_func`].
+//!
+//! The existence, difference or absence of an entity or field within a [`FuncAssociations`] bag
+//! dictates what we must do during mutation. New argument? Create one and send it back up. Removed
+//! argument? Delete it and send it back up. Mutated argument? Change it and send it back up. The
+//! payload is the same on both sides.
+//!
+//! _Note_: all submodules are private since the [`FuncAuthoringClient`] is the primary interface.
```
nickgerace, line 13: It's one big loop.

lib/dal/src/func/authoring/save.rs
```diff
@@ -18,13 +30,14 @@ pub(crate) async fn save_func(
...
 
-    // Disallow the ability to modify builtins.
-    if func.builtin {
-        return Err(FuncAuthoringError::NotWritable);
-    }
+    // TODO(nick): we should eventually either return an error or make it a no-op.
+    // For now, we need configurable builtins to ensure that the system is working.
+    // if func.builtin {
+    //     return Err(FuncAuthoringError::NotWritable);
+    // }
```
nickgerace, line 40: Temporary measure. It wasn't working as intended and broke the UI. Not only that, we are using it for testing func authoring. This is mainly for the user to not break their environments. We will re-enable this once things settle.

lib/dal/src/schema/variant.rs
```diff
@@ -904,7 +906,9 @@ impl SchemaVariant {
             SchemaVariant::find_leaf_item_prop(ctx, schema_variant_id, leaf_kind).await?;
 
         if component_id.is_some() {
-            unimplemented!("component context not supported for leaf functions");
+            // NOTE(nick): replaced a "unimplemented" here with an error, but we will need to think
+            // about whether we want this in the future.
+            return Err(SchemaVariantError::ComponentContextNotSupportedForLeafFunctions);
```
nickgerace, line 911: Do not want a production panic

lib/dal/src/schema.rs
```diff
@@ -175,10 +175,9 @@ impl Schema {
...
         let workspace_snapshot = ctx.workspace_snapshot()?;
 
-        info!(
-            "Setting {} as the default schema variant for {}",
-            schema_variant_id.clone(),
-            self.id.clone()
+        debug!(
+            %schema_variant_id, "setting the default schema variant for schema: {}",
+            self.id
         );
```
nickgerace, line 181: Should be a tad quieter

lib/sdf-server/src/server/server.rs
```diff
@@ -433,17 +432,22 @@ async fn install_builtins(
...
+                    Err(PkgError::PackageAlreadyInstalled(hash)) => {
+                        count += 1;
+                        warn!(%hash, "pkg {pkg_name} already installed ({count} of {total} installed)");
+                    }
+                    Err(err) => error!(?err, "pkg {pkg_name} install failed"),
                 }
             }
             Err(err) => {
-                println!("Pkg {pkg_name} Install failed, {err}");
+                error!(?err, "pkg {pkg_name} install failed with server error");
```
nickgerace, line 450: Cleaning up logs, standardizing them and making errors clearer

lib/dal/src/func/associations.rs
```diff
@@ -166,8 +82,28 @@ impl FuncAssociations {
...
+                    Some(action_prototype_id) => {
+                        let action_prototype = DeprecatedActionPrototype::get_by_id_or_error(
+                            ctx,
+                            *action_prototype_id,
+                        )
+                        .await?;
+                        action_prototype.kind
+                    }
+                    None => DeprecatedActionKind::Create,
+                };
```
nickgerace, line 100: I don't love this because different action prototypes can technically have different action kinds for the same func. However, this is how it worked in the old engine and we are in the middle of an actions re-write, so restoring old behavior is fine for now.

nickgerace:
bors merge

PR #3579, opened 2024-04-12
Fix Frames, a journey

jobelenus:
1. do not rely on browser hover when dropping components into frame. perform the math to see what we are inside.
2. this lets us remove the hack in cypress tests
3. fix to missing parentId (was renamed to parentComponentId)
4. `set_component_position` endpoint needed to change for the new engine
5. fixed a swap of width & height bug on resizing
6. send a web event from the `set_component_position` endpoint
7. frame movement now supports deeply nested children
8. nudging is debounced now, so we don't flood the system with changes
9. creating or attaching a component inside a frame expands the child to the available size and fully bounded by the parent
10. so long as there are no other children, in that case, it makes the frame smaller and does not attempt a "best fit".
Art Frame GIF by Mowgli420


jobelenus:
bors merge

stack72:
bors retry

PR #3580, opened 2024-04-13
bug(layerdb): race to create broadcast channels

adamhjk:
We were using the wrong kind of lock a (RwLock rather than a Mutex) around the creation of broadcast channels in the activity multiplexer. When we went to create a new subscription, we would acquire a Read lock when checking for the existence of the broadcast channel, and if it didn't exist, continue on to create a new one by acquiring a Write lock later in the function.

This, of course, was racy - multiple consumers could acquire a failing Read lock condition, and then pass through to the critical section where we create a new broadcast channel - the result being multiple consumers of the underlying NATS jetstream would be created, only one of which was actually sending the messages we wanted.

I believe this is the source of the occasional deadline misses in rebase finished requests - at the very least, this code was easy to trigger with the test coverage in this PR, which leads me to believe it was also the source of our other issues.

Let's see how it goes. :)

### DIFF
--- a/lib/si-layer-cache/src/activities.rs
+++ b/lib/si-layer-cache/src/activities.rs
@@ -1,9 +1,4 @@
-use std::{
-    collections::HashMap,
-    fmt::{self, Debug},
-    str::FromStr,
-    sync::Arc,
-};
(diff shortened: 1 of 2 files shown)


adamhjk:
bors merge

PR #3581, opened 2024-04-18
fix(dal): Ensure we set if the component has a resource

stack72:
We defaulted to false but since we started being able to create resources we can now return if the resource exists


stack72:
bors merge

PR #3585, opened 2024-04-22
feat: add layer cache dbname to init container config

sprutton1:
We need these values in place in order to set the layer cache db name in pgbouncer and the related services.


sprutton1:
bors merge

PR #3587, opened 2024-04-22
fix(dal): Ensure we don't list duplicate schema variants

stack72:


lib/dal/src/schema/variant.rs
```diff
@@ -459,7 +459,10 @@ impl SchemaVariant {
 
         // Now add the default schema variant to the list as this is a different edge kind
         let default_schema_variant = Self::get_default_for_schema(ctx, schema_id).await?;
-        schema_variants.push(default_schema_variant);
+        // We will only add the default schema variant IF there is no existing schema variant found
+        if !schema_variants.contains(&default_schema_variant) {
+            schema_variants.push(default_schema_variant);
+        }
```
nickgerace, line 465: I think we can move these lines altogether since `EdgeWeightKindDiscriminants::Use` (passed into `outgoing_targets_for_edge_weight_kind_by_index`) includes both `Use { default: true }` and `Use { default: false }`.

nickgerace (changes requested):

lib/dal/src/schema/variant.rs
nickgerace: I'd change line `425` to be `EdgeWeightKindDiscriminants::Use` to be more clear too. Might help others that run into the issue.

stack72:
bors merge

lib/dal/src/schema/variant.rs
```diff
@@ -419,10 +419,12 @@ impl SchemaVariant {
         let mut schema_variants = vec![];
         let parent_index = workspace_snapshot.get_node_index_by_id(schema_id).await?;
 
+        // We want to use the EdgeWeightKindDiscriminants rather than EdgeWeightKind
+        // this will bring us back all use and default edges!
         let node_indices = workspace_snapshot
             .outgoing_targets_for_edge_weight_kind_by_index(
                 parent_index,
-                EdgeWeightKind::new_use().into(),
+                EdgeWeightKindDiscriminants::Use,
```
nickgerace, line 427: Awesome!!

PR #3589, opened 2024-04-23
Feat/max bytes

sprutton1:
This normalizes the rebaser config to match how pinga is configured today. We shouldn't be doing this an instead should be removing all of the bits the rebaser doesn't use, but it's a bigger lift to refactor how we create the services context. This also adds the `max_bytes` setting wherever we use jetstream. It is currently set to 1mb, but this is easily tuneable.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4104,6 +4104,7 @@ dependencies = [
  "clap",
  "color-eyre",
  "rebaser-server",
+ "si-std",
  "telemetry-application",
  "tokio",
  "tokio-util",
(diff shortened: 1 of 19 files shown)


sprutton1:
bors merge

sprutton1:
bors merge

sprutton1:
bors merge

sprutton1:
bors try

sprutton1:
bors merge

sprutton1:
bors merge

PR #3590, opened 2024-04-23
fix: defaults in rebaser config overwrite local settings

sprutton1:
Missed some defaults in my earlier PR for the rebase config.


sprutton1:
bors merge

PR #3592, opened 2024-04-23
fix(web): don't inject route outside of createStore

zacharyhamm:
`useRoute` cannot be used outside of a Vue component (it's enough to put it inside of the createStore). Quiet the warning about inject in the wrong context.


zacharyhamm:
@britmyerss not entirely sure of the behavior so you might want to click test this one

PR #3593, opened 2024-04-23
Attach, Detach, and Nested Frames, Oh My!

britmyerss:
This PR changes how we handle frames in the DAL. Previously, when connecting a Component to a Frame (up or down), we would create Attribute Prototype Arguments signifying which Input Socket was connected to which Output Socket for a given set of Components, in addition to creating a FrameContains edge between those components.  With this PR, we are relying solely on the FrameContains edge between Components and walking the lineage tree to determine the connection between Input and Output Sockets.  This means we no longer need an Attribute Prototype Argument between sockets, as the connection is 'inferred' from the lineage. As of this change, Input Sockets can only be driven through inference from a single Output Socket, though this can be expanded in the future as the need arises.

This PR also includes 6(!) new test exclusive schemas that can be used to write integration tests with interesting relationships. 'Even Lego' schemas have Input Sockets with even-numbered names, and Output Sockets with odd-numbered names. 'Odd Lego' schemas have Input Sockets with odd-numbered names and Output Sockets with even-numbered names.  There are small, medium, and large flavors of each Even/Odd Lego - allowing us to easily test the various permutations of inferred socket connections.  Small Legos have sockets with names "1", and "2", Medium has "1", "2", "3", "4", and Large Legos have "1","2","3","4","5",and "6".

I've also added an entry to the Component Debug panel to show when a Socket has an inferred connection.

Follow up PR will add WS Events for the frame attachment/detachment and fix up the UI to fill in the socket when it has an inferred connection.

There's definitely room for more tests but that can also follow after this!

<div><br/>via <a href="https://giphy.com/disney/">Disney</a> on <a href="https://giphy.com/gifs/disney-walt-disney-animation-studios-feast-frozen-fever-ZdilASsUTJW4o">GIPHY</a></div>


lib/dal/src/attribute/value.rs
```diff
@@ -549,93 +550,121 @@ impl AttributeValue {
         // progress). To handle this here, we grab a read lock, which will be
         // locked for writing in the dependent values update job while the
         // execution result is being written to the graph.
+
+        // BRIT HERE
```
zacharyhamm, line 554: HELLO BRIT (probably should remove this)

lib/dal/src/attribute/value/dependent_value_graph.rs
```diff
@@ -198,7 +252,7 @@ impl DependentValueGraph {
                     .value_depends_on(parent_attribute_value_id, current_attribute_value_id);
             }
         }
-
+        info!("Dependent value graph: {:?}", dependent_value_graph);
```
jhelwig, line 255: Drop to `debug!` level?

lib/dal/src/attribute/value.rs
```diff
@@ -549,93 +550,121 @@ impl AttributeValue {
...
-                            ]
+        if apa_ids.is_empty() {
+            if let Some(implicit_input) =
+                Self::get_inferred_input_value(ctx, attribute_value_id).await?
+            {
+                let input_func = AttributePrototype::func_id(ctx, prototype_id).await?;
+
+                match Func::get_by_id(ctx, input_func).await? {
+                    Some(func) => {
+                        if func.name != "si:identity" {
```
zacharyhamm, line 575: We have a helper so we don't have to hardcode this. `InstrinsicFunc::Identity.name()`

lib/dal/src/attribute/value.rs
britmyerss: ah yes thank you! 🤦‍♀️

lib/dal/src/attribute/value.rs
```diff
@@ -549,93 +550,121 @@ impl AttributeValue {
...
-                                apa_id,
-                            ),
-                        )? {
-                        ValueSource::StaticArgumentValue(static_argument_value_id) => {
-                            vec![
-                                StaticArgumentValue::get_by_id(ctx, static_argument_value_id)
-                                    .await?
-                                    .value,
-                            ]
+        if apa_ids.is_empty() {
```
zacharyhamm, line 567: In the case of sockets with an Arity of many, do we want to allow both manually drawn connections and inferred connections from the frame relationships?

lib/dal-test/src/lib.rs
```diff
@@ -736,7 +736,12 @@ async fn migrate_local_builtins(
     schemas::migrate_test_exclusive_schema_validated_input(&ctx).await?;
     schemas::migrate_test_exclusive_schema_validated_output(&ctx).await?;
     schemas::migrate_test_exclusive_schema_bad_validations(&ctx).await?;
-
+    schemas::migrate_test_exclusive_schema_large_odd_lego(&ctx).await?;
+    schemas::migrate_test_exclusive_schema_large_even_lego(&ctx).await?;
+    schemas::migrate_test_exclusive_schema_medium_even_lego(&ctx).await?;
+    schemas::migrate_test_exclusive_schema_medium_odd_lego(&ctx).await?;
+    schemas::migrate_test_exclusive_schema_small_odd_lego(&ctx).await?;
+    schemas::migrate_test_exclusive_schema_small_even_lego(&ctx).await?;
```
nickgerace, line 744: Wheeeee

app/web/src/components/Debug/SocketDebugView.vue
```diff
@@ -40,6 +40,22 @@
           <p v-else>No input sources</p>
         </template>
       </DebugViewItem>
+      <DebugViewItem title="Inferred Connection(s)">
```
nickgerace, line 43: I love this

lib/dal/src/attribute/prototype/argument.rs
```diff
@@ -543,17 +544,47 @@ impl AttributePrototypeArgument {
...
-        let prototype_id = apa.prototype_id(ctx).await?;
+        let attribute_prototype_argument =
+            Self::get_by_id(ctx, attribute_prototype_argument_id).await?;
+        attribute_prototype_argument.remove_inner(ctx).await?;
+        Ok(())
+    }
+
+    /// Removes the [`AttributePrototypeArgument`] corresponding to the provided ID, but is a
+    /// "no-op" if it cannot be found before removal.
+    pub async fn remove_or_no_op(
```
nickgerace, line 560: Did you get this from #3578? Curious to why the diff would show this

lib/dal/src/attribute/value.rs
```diff
@@ -549,93 +550,121 @@ impl AttributeValue {
...
+            if let Some(implicit_input) =
+                Self::get_inferred_input_value(ctx, attribute_value_id).await?
+            {
+                let input_func = AttributePrototype::func_id(ctx, prototype_id).await?;
+
+                match Func::get_by_id(ctx, input_func).await? {
+                    Some(func) => {
+                        if func.name != "si:identity" {
+                            None
+                        } else {
```
jhelwig, line 577: Is the goal here to forbid having functions other than `si:identity` on sockets that get implicit connections? Wondering why the `if` check at all, instead of only having the `else` case.

lib/dal/src/attribute/value.rs
nickgerace: Nah let's keep it 😤

lib/dal/src/attribute/value/dependent_value_graph.rs
zacharyhamm, line 255: Should probably drop this (could end up being a ton of data sent to the otelcol)

lib/dal/src/attribute/value.rs
```diff
@@ -549,93 +550,121 @@ impl AttributeValue {
...
+                    let func_arg_id =
+                        AttributePrototypeArgument::func_argument_id_by_id(ctx, apa_id).await?;
+                    let func_arg_name = ctx
+                        .workspace_snapshot()?
+                        .get_node_weight_by_id(func_arg_id)
+                        .await?
+                        .get_func_argument_node_weight()?
+                        .name()
+                        .to_owned();
+                    // brit down here
```
nickgerace, line 620: hey brit

lib/dal/src/attribute/value.rs
```diff
@@ -549,93 +550,121 @@ impl AttributeValue {
...
-                        )? {
-                        ValueSource::StaticArgumentValue(static_argument_value_id) => {
-                            vec![
-                                StaticArgumentValue::get_by_id(ctx, static_argument_value_id)
-                                    .await?
-                                    .value,
-                            ]
+        if apa_ids.is_empty() {
+            if let Some(implicit_input) =
+                Self::get_inferred_input_value(ctx, attribute_value_id).await?
```
nickgerace, line 569: I dig it

lib/dal/src/attribute/value.rs
```diff
@@ -549,93 +550,121 @@ impl AttributeValue {
...
+                    None => None,
+                };
+            };
+        } else {
+            for apa_id in apa_ids {
+                let apa = AttributePrototypeArgument::get_by_id(ctx, apa_id).await?;
+                let expected_source_component_id = apa
+                    .targets()
+                    .map(|targets| targets.source_component_id)
+                    .unwrap_or(destination_component_id);
```
nickgerace, line 593: I'm a bit confused at what's happening here: what (and why) are we falling back to if we don't have targets?

lib/dal/src/attribute/value.rs
jhelwig: Mixing implicit & explicit connections like that seems like it'll lead to a lot of complexity/confusion. I think it'd be a lot clearer for users if we say that there are either inferred connections, or manual connections, but not both on the same socket. If they would like both manual data, and what would have been coming from the inferred connection, then it's possible to make both connections manually. Having the inferred one still exist when there's a manual connection means that the inferred one can't ever be overridden.

lib/dal/src/attribute/value.rs
```diff
@@ -726,6 +756,74 @@ impl AttributeValue {
...
+        info!(
+            " AV: Finding inferred input value for input value {}",
+            input_attribute_value_id
+        );
+
+        Ok(
+            match Component::find_inferred_connection_to_input_socket(ctx, input_socket_match).await
+            {
+                Ok(Some(output_match)) => {
+                    info!("AV: found output socket {:?}", output_match);
```
nickgerace, line 790: We should probably chop down (either in level or get rid of) some of these logging statements

jobelenus:
This passes muster as far as the pinia store goes 🖖

lib/dal/src/component.rs
```diff
@@ -1399,10 +1462,435 @@ impl Component {
...
+    ) -> ComponentResult<Option<OutputSocketMatch>> {
+        let mut found_match: Option<OutputSocketMatch> = None;
+        let component_id = input_socket_match.component_id;
+        let children = Component::get_children_for_id(ctx, component_id).await?;
+        //load up the children and look for matches
+        let mut work_queue: VecDeque<Vec<ComponentId>> = VecDeque::new();
+        work_queue.push_front(children);
+
+        'parents: while let Some(children) = work_queue.pop_front() {
+            for child_component in children {
```
nickgerace, line 1716: Generally speaking, using a named loop is a bit of a smell that the logic is too complex for one function (cc @zacharyhamm since I think you've mentioned this before). I'd consider breaking this out into another function for processing each set of children. The return type could influence how you break out of the `while` loop.

lib/dal/src/component/frame.rs
```diff
@@ -32,79 +40,158 @@ pub type FrameResult<T> = Result<T, FrameError>;
...
-    /// Provides the ability to attach a child [`Component`] to a parent frame.
-    pub async fn attach_child_to_parent(
+    /// Provides an ability to remove the existing ['Component']'s parent``
+    #[instrument(level = "info", skip_all)]
+    pub async fn orphan_child(ctx: &DalContext, child_id: ComponentId) -> FrameResult<()> {
+        if let Some(parent_id) = Component::get_parent_by_id(ctx, child_id).await? {
+            Self::detach_child_from_parent_inner(ctx, parent_id, child_id).await?;
+        }
+        Ok(())
+    }
```
nickgerace, line 50: Okay, this makes sense to make

lib/dal/src/component/frame.rs
```diff
@@ -32,79 +40,158 @@ pub type FrameResult<T> = Result<T, FrameError>;
...
     }
 
-    async fn attach_child_to_parent_symbolic(
+    async fn attach_child_to_parent_inner(
         ctx: &DalContext,
         parent_id: ComponentId,
         child_id: ComponentId,
     ) -> FrameResult<()> {
+        // detect cycles here?
         Component::add_edge_to_frame(ctx, parent_id, child_id, EdgeWeightKind::FrameContains)
```
nickgerace, line 79: What cycles are you worried about?

lib/dal/tests/integration_test/frame.rs
```diff
@@ -104,6 +110,853 @@ async fn convert_component_to_frame_and_attach_no_nesting(ctx: &mut DalContext)
     );
 }
 
+#[test]
+async fn simple_frames(ctx: &mut DalContext) {
```
nickgerace, line 114: Nice!

lib/dal/tests/integration_test/frame.rs
```diff
@@ -104,6 +110,853 @@ async fn convert_component_to_frame_and_attach_no_nesting(ctx: &mut DalContext)
...
+        .expect("could not get attribute value")
+        .materialized_view(ctx)
+        .await
+        .expect("could not get materialized view")
+        .expect("is some");
+
+    assert_eq!(one_input_mat_view, serde_json::json!("1"));
+}
+#[test]
+async fn simple_down_frames_nesting(ctx: &mut DalContext) {
```
nickgerace, line 469: Thank you

lib/dal/tests/integration_test/frame.rs
```diff
@@ -104,6 +110,853 @@ async fn convert_component_to_frame_and_attach_no_nesting(ctx: &mut DalContext)
...
+        .await
+        .expect("could not get attribute value")
+        .materialized_view(ctx)
+        .await
+        .expect("could not get materialized view")
+        .expect("is some");
+    assert_eq!(one_input_mat_view, serde_json::json!("4"));
+}
+#[test]
+async fn simple_up_frames_some_nesting(ctx: &mut DalContext) {
```
nickgerace, line 709: You fixed up frames!

lib/dal/tests/integration_test/frame.rs
```diff
@@ -104,6 +110,853 @@ async fn convert_component_to_frame_and_attach_no_nesting(ctx: &mut DalContext)
...
+    assert_eq!(odd_up_frame_output_av, output_match.attribute_value_id);
+
+    let one_input_mat_view = AttributeValue::get_by_id(ctx, one_input_match.attribute_value_id)
+        .await
+        .expect("could not get attribute value")
+        .materialized_view(ctx)
+        .await
+        .expect("could not get materialized view")
+        .expect("materialized view is some");
+    assert_eq!(one_input_mat_view, serde_json::json!("2"));
```
nickgerace, line 957: I like the use of materialized views to validate data passing.

lib/dal/tests/integration_test/schema/variant.rs
```diff
@@ -232,6 +232,33 @@ async fn all_funcs(ctx: &DalContext) {
         all_funcs.as_slice(),
     );
     assert_eq!(expected, actual);
+    // let schema = Schema::find_by_name(ctx, "large even lego")
+    //     .await
+    //     .expect("unable to get schema")
+    //     .expect("schema not found");
+    // let schema_variant_id = schema
+    //     .get_default_schema_variant(ctx)
```
nickgerace, line 240: I'd just delete this

lib/dal/src/attribute/value.rs
britmyerss: Yeah I think we *could* do it in the future, but not without loads more UX to make it easier to understand what is happening.

nickgerace:
I love the overall idea! Just a few comments. Great work.

lib/dal/src/attribute/value.rs
britmyerss: yeah this came out of a conversation with Zack, I think we can remove the if.

lib/dal/src/attribute/value.rs
britmyerss: This was as it was, so without loading this context in, I'm not sure.

britmyerss:
bors try

britmyerss:
bors try

britmyerss:
bors retry

lib/dal/src/attribute/prototype/argument.rs
britmyerss: I actually don't know what happened here... let's chat

stack72:
bors retry

britmyerss:
bors merge

britmyerss:
bors merge

britmyerss:
bors merge

britmyerss:
bors try

britmyerss:
Bors merge

britmyerss:
Bors retry

johnrwatson:
bors try

stack72:
bors try

britmyerss:
bors merge

nickgerace:
bors merge that shit

PR #3594, opened 2024-04-23
feat(binary): move nix-invoked buck build to use release mode by default

johnrwatson:
Docker build and omnibus use nix-invoked builds to create the binaries/packages.

This moves those publishable Artifacts to use release build by default.


johnrwatson:
bors merge

PR #3599, opened 2024-04-23
feat: add port for node-exporter

johnrwatson:
Node exporter wasn't available on the nodes interfaces so that it could be consumed by prometheus.

This adds it onto the host interface so metrics can be pulled up the observability stack.


johnrwatson:
bors merge

PR #3601, opened 2024-04-24
feat(dal, sdf): Reimplement clone_variant

stack72:
This also adds the work to clone a Func and checks if the func name is already in use for that variant


stack72:
bors try

stack72:
bors merge

PR #3602, opened 2024-04-24
Fix incorrect modify removed item conflict detection

nickgerace:
## Description

This PR fixes a bug in conflict and update detection where a modify removed item conflict was found instead of a remove edge update. This occurred because the lamport clock used in the comparison was the same.

Interestingly, this bug was not picked up in unit tests nor do unit tests performing equivalent logic to the new integration test catch the change. Thus, a new integration test that catches the bug has been added.


lib/dal/src/workspace_snapshot/graph.rs
```diff
@@ -1150,7 +1150,7 @@ impl WorkspaceSnapshotGraph {
                 if to_rebase_item_weight
                     .vector_clock_write()
                     .entry_for(to_rebase_vector_clock_id)
-                    >= onto_last_saw_to_rebase
```
nickgerace, line 1153: This caused the bug.

nickgerace:
bors try

johnrwatson:
bors try

nickgerace:
bors try

nickgerace:
bors merge

PR #3603, opened 2024-04-24
Revert "feat: set moka max_size to 12gb to ensure we don't starve the host"

sprutton1:
This was not considered well enough. We actually make 4 distinct caches, so collectively we are setting the total memory to 48gb. Reverting so we can do this properly.

This reverts commit f108aec47d5cc32be89e0404d84339b349d69e6e.


sprutton1:
bors merge

PR #3606, opened 2024-04-24
Actions V2 Part 2

jhelwig:
## Split core dependency graph tracking/helpers into its own struct

This will be shared between the AttributeValue dependency graph, and the (upcoming) Action dependency graph.

## Calculate dependency graph of Actions


lib/dal/src/action/dependency_graph.rs
```diff
@@ -0,0 +1,122 @@
...
+        // things they are feeding data into.
+        for component_id in actions_by_component_id.keys().copied() {
+            let component = Component::get_by_id(ctx, component_id).await?;
+            for incoming_connection in component.incoming_connections(ctx).await? {
+                component_dependencies
+                    .entry(component_id)
+                    .or_default()
+                    .insert(incoming_connection.from_component_id);
+            }
+            // TODO: Handle implicit/inferred dependencies from Frames once that work lands?
```
britmyerss, line 67: You should be able to do the same as above, only via `component.inferrred_connections` but you'll need to filter for deletes as I couldn't make this check too low level without hiding the `to_delete` flag on the incoming inferred connection

jhelwig: `component.inferred_connections` has both the incoming, and outgoing inferred connections, right? So, it would need to be limited to inferred connections where the `to_component_id` is the same as `component.id()`. Looks easy enough, though. 👍🏻

jhelwig:
bors merge

PR #3610, opened 2024-04-24
chore(db): add pgbouncer to our dev setup

adamhjk:
This PR adds pgbouncer to our dev setup. It moves the "real" postgres instances to 7432 (si) and 8432 (si-test) respectively, replacing them with two different pgbouncers. This change is primarily to ensure that we avoid any threat of pool exhaustion in the test suite, but has the added benefit of making the dev stack closer to our production deployment (which would also include pgbouncer.)

PgBouncer configures a pool per hostname+database. Since each of our dal integration tests creates its own database, it thereby also creates its own pool. To manage this, we create a pool size of 10 per hostname+database tuple, which we will then reap after 10 idle seconds. The result is that the current test suite peaks somewhere around 300 tests, but very rapidly shrinks. We should have plenty of headroom.

The only tricky thing here is that we alter the migration paths for tests, as we migrate with the underlying database connection directly. When we do, we must forcibly disconnect the bouncer - so if you happen to be logged in to postgres, get ready to be logged out when we need to make the original template database for tests.

### DIFF
--- a/.ci/DEV.md
+++ b/.ci/DEV.md
@@ -1,0 +1,38 @@
+# How to run the CI stack locally
+
+First, you need to prepare a local copy of the code you want to test.
+
+```
+$ cp ./si ./si-fixme
+$ chown -R 2000:2000 ./si-fixme
+```
+
+This is to create a working copy of the source that is owned by the same user
+as the internal 'ci' user.
+
+## Run the CI platform
+
+```
+$ docker-compose -f .ci/docker-compose.test-integration.yml up
+```
+
+This will start all the services, but the 'app' service will fail. This is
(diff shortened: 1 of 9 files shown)


adamhjk:
bors try

adamhjk:
bors try

adamhjk:
bors try

adamhjk:
bors try

adamhjk:
bors try

adamhjk:
bors try

adamhjk:
bors try

adamhjk:
bors try

adamhjk:
bors try

adamhjk:
bors merge

PR #3611, opened 2024-04-24
feat(dal,web): mark as updating during rebases

zacharyhamm:
Ensures we also make the status bar reactive to rebase operations


zacharyhamm:
bors retry

PR #3616, opened 2024-04-25
ENG-2502 edge set response id

jobelenus:
`componentStore.selectedEdge` fails without saving over the tempID.


jobelenus:
bors merge

PR #3618, opened 2024-04-25
feat(dal): check for cycles when connecting components

zacharyhamm:
Ensures we check for graph cycles when connecting components (to prevent connecting components in a cycle), and when attaching children to parent frames.

### DIFF
--- a/lib/dal/src/workspace_snapshot.rs
+++ b/lib/dal/src/workspace_snapshot.rs
@@ -31,6 +31,7 @@ pub mod update;
 pub mod vector_clock;
 
 use si_pkg::KeyOrIndex;
+use std::sync::atomic::AtomicBool;
 use std::sync::Arc;
 use tokio::sync::{RwLock, RwLockReadGuard, RwLockWriteGuard};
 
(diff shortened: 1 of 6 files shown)


lib/dal/src/workspace_snapshot.rs
```diff
@@ -125,13 +126,41 @@ pub struct WorkspaceSnapshot {
...
 }
 
+impl std::ops::Drop for CycleCheckGuard {
+    fn drop(&mut self) {
+        if Arc::strong_count(&self.cycle_check) <= 2 {
+            self.cycle_check
+                .store(false, std::sync::atomic::Ordering::Relaxed);
+        }
+    }
+}
```
nickgerace, line 155: Clever! Not dumb :smile:

lib/dal/tests/integration_test/cycle_check_guard.rs
```diff
@@ -0,0 +1,14 @@
+use dal::DalContext;
+use dal_test::test;
+
+#[test]
+async fn cycle_check_guard_test(ctx: &DalContext) {
```
nickgerace, line 5: Nice!

nickgerace (approved):

PR #3619, opened 2024-04-25
chore(dal): Cleanup workspace export code from export

stack72:


lib/dal/src/pkg/export.rs
```diff
@@ -1435,42 +1175,21 @@ impl PkgExporter {
...
-                //             .build()?,
-                //     );
-                // }
-
-                // pkg_spec_builder.change_set(
-                //     head_builder
-                //         .funcs(remove_duplicate_func_specs(&head_funcs))
-                //         .build()?,
-                // );
+                info!("Workspace exports not supported");
```
vbustamante, line 1178: We should probably just fail here instead of returning an OK. 
adding a new PkgError variant that returns what the info! above does will do the trick

stack72: ok! Happy to do that :)

stack72:
bors merge

vbustamante (approved):
Great!

PR #3620, opened 2024-04-25
Basic multiplayer experience for moving and resizing elements on the diagram

jobelenus:
The changes include:
1. Make sure all changes to position and size runs through a single function. That refactor is most of the diff. Nearly every single position change also had a line of code to POST the same position. We can encapsulate that logic to prevent errors and introduce clarity.
2. Adding a new receiver for WsEvents in SDF that reflects the payload back out to all clients.
3. Sending websocket messages onDragMove and onResize events.
multiplayer

### DIFF
--- a/app/web/src/components/ModelingDiagram/ModelingDiagram.vue
+++ b/app/web/src/components/ModelingDiagram/ModelingDiagram.vue
@@ -211,12 +211,14 @@ import tinycolor from "tinycolor2";
 import { LoadingMessage, getToneColorHex } from "@si/vue-lib/design-system";
 import { connectionAnnotationFitsReference } from "@si/ts-lib/src/connection-annotations";
 import { windowListenerManager } from "@si/vue-lib";
+import { el } from "date-fns/locale";
 import { useCustomFontsLoaded } from "@/utils/useFontLoaded";
 import DiagramGroup from "@/components/ModelingDiagram/DiagramGroup.vue";
 import {
   ComponentId,
   EdgeId,
   useComponentsStore,
+  FullComponent,
 } from "@/store/components.store";
 import DiagramGroupOverlay from "@/components/ModelingDiagram/DiagramGroupOverlay.vue";
 import { DiagramCursorDef, usePresenceStore } from "@/store/presence.store";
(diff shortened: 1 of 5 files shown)


lib/dal/src/component.rs
```diff
@@ -2150,16 +2150,29 @@ pub struct ComponentUpdatedPayload {
...
-    width: Option<i32>,
-    height: Option<i32>,
-    user_pk: Option<UserPk>,
+    pub change_set_id: ChangeSetId,
+    pub component_id: ComponentId,
+    pub x: i32,
+    pub y: i32,
+    pub width: Option<i32>,
+    pub height: Option<i32>,
+    pub user_pk: Option<UserPk>,
```
vbustamante, line 2159: Do these really need to be public? I'm not catching the use case in this PR

britmyerss (approved):
this looks good to me!

app/web/src/components/ModelingDiagram/ModelingDiagram.vue
```diff
@@ -1281,25 +1286,77 @@ const draggedElementsPositionsPreDrag = ref<
...
+          userPk: authStore.userPk,
+          componentId,
+          changeSetId: changeSetsStore.selectedChangeSetId,
+          x: position.x,
+          y: position.y,
+          width: size?.width || null,
+          height: size?.height || null,
+        },
+      });
+    }
```
nickgerace, line 1357: Makes sense!

lib/dal/src/component.rs
jobelenus: Mmm, excellent call out. The previous construction needed it. When I take it away, I get:
```
error[E0616]: field `change_set_id` of struct `ComponentSetPositionPayload` is private
   --> lib/sdf-server/src/server/service/ws/workspace_updates.rs:235:116
    |
235 | ...                   let event = WsEvent::reflect_component_position(self.workspace_pk, payload.change_set_id, payload).await?;
    |                                                                                                  ^^^^^^^^^^^^^ private field
```
Should I make _just_ that one `pub`? Probably better to make an accessor `change_set_id()` right?

lib/sdf-server/src/server/service/ws/workspace_updates.rs
```diff
@@ -110,6 +110,8 @@ mod workspace_updates {
     #[derive(Serialize, Deserialize, Debug, Clone)]
     #[serde(tag = "kind", content = "data")]
     pub enum WebsocketEventRequest {
+        #[serde(rename_all = "camelCase")]
+        ComponentSetPosition(ComponentSetPositionPayload),
```
nickgerace, line 114: Clever

lib/dal/src/component.rs
nickgerace: I think they were needed in the past, but not now. Either way, it's probably no big deal since the mutability of the payload struct is "meh": you either send it or receive it and it's not long lived. Good catch though. I think you're right.

nickgerace (approved):

lib/dal/src/component.rs
vbustamante: @jobelenus Yeah that sounds about right! Getters only  for the things that need to be public. We've been naming then just the name of the field,  so change_set_id() would be the pattern

jobelenus:
bors merge

PR #3621, opened 2024-04-25
Remove revert func

nickgerace:
## Description

This PR removes used and unused references to "revert func". We do not know the product direction for this feature in the new engine yet and will re-evaluate it in the future.

This is an atomic removal of anything related to reverting funcs or checking if funcs are "revertible".


nickgerace:
bors merge

PR #3623, opened 2024-04-26
Implement component deletion

paulocsanz:


app/web/src/store/actions.store.ts
```diff
@@ -303,43 +303,90 @@ export const useActionsStore = () => {
...
                   action_batch_id: update.batchId,
                 });
 
-                this.LOAD_ACTION_BATCHES();
+                const batchIndex = this.actionBatches.findIndex((batch) => batch.id === update.batchId);
+                const batch = this.actionBatches[batchIndex];
+                if (!batch) {
+                  // Short term fix for reactivity issue on apply, since the
+                  // first load won't have the actions since the rebaser isnt done
+                  setTimeout(() => this.LOAD_ACTION_BATCHES(), 500);
```
paulocsanz, line 339: This is horrifying but was the only way to consistently get the data after change-set apply. It's a general bug after apply, where data is loaded before the rebaser finishes so everything is stale, but it made actions pretty much unusable, so this is a temporary fix until we solve it for everything.

Documented on https://linear.app/system-initiative/issue/BUG-105/no-reactivity-after-applying-a-change-set

paulocsanz:
bors try

paulocsanz:
bors try

paulocsanz:
bors try

app/web/src/store/actions.store.ts
nickgerace: I think a 500ms post execution makes sense to me as a patch until we solve it fully!

app/web/src/store/components.store.ts
```diff
@@ -1131,21 +1133,29 @@ export const useComponentsStore = (forceChangeSetId?: ChangeSetId) => {
...
                 componentIds,
                 ...visibilityParams,
               },
+              onSuccess: (response) => {
+                for (const [componentId, hasResource] of _.entries(response)) {
+                  if (!hasResource) {
+                    delete this.rawComponentsById[componentId];
+                  }
+                }
+              },
```
nickgerace, line 1150: Good idea!

app/web/src/store/components.store.ts
```diff
@@ -1360,6 +1381,20 @@ export const useComponentsStore = (forceChangeSetId?: ChangeSetId) => {
...
+              eventType: "ComponentDeleted",
+              callback: (data) => {
+                if (data.changeSetId !== changeSetId) return;
+                delete this.rawComponentsById[data.componentId];
+
+                // remove invalid component IDs from the selection
+                const validComponentIds = _.intersection(
+                  this.selectedComponentIds,
+                  _.keys(this.rawComponentsById),
+                );
```
nickgerace, line 1394: Ah, good point.

dev/Tiltfile
```diff
@@ -156,8 +156,8 @@ sdf_target = "//bin/sdf:sdf"
 local_resource(
     "sdf",
     labels = ["backend"],
-    cmd = "buck2 build @//mode/release {}".format(sdf_target),
-    serve_cmd = "buck2 run @//mode/release {}".format(sdf_target),
+    cmd = "buck2 build @//mode/debug {}".format(sdf_target),
+    serve_cmd = "buck2 run @//mode/debug {}".format(sdf_target),
```
nickgerace, line 160: Was this for temporary local dev?

nickgerace: I wonder if we can add a `buck2 run dev:up-debug` target in the future...

lib/dal/src/deprecated_action/prototype.rs
```diff
@@ -344,13 +315,48 @@ impl DeprecatedActionPrototype {
...
+        // manually cast it here, it's horrifying, but func editing is not 100% in a place where we
+        // can just update all actions
+        if let Some(payload) = component_view
+            .as_mut()
+            .and_then(|v| v.pointer_mut("/resource/payload"))
+        {
+            if let serde_json::Value::String(string) = payload.clone() {
+                *payload = serde_json::from_str::<serde_json::Value>(&string)?;
+            }
+        }
```
nickgerace, line 359: Fair point. We'll coordinate on this!

nickgerace (approved):
Other than the tilt thing, this looks awesome. Clever use of the existing tests too. Love it :fire:

dev/Tiltfile
paulocsanz: Totally missed that, sorry. But I feel we need more optimizations on rebaser than sdf, so I wonder if we could make sdf compile faster.

It's taking a long time here, everything else compiles quickly

paulocsanz: Fixed

paulocsanz:
bors merge

PR #3625, opened 2024-04-26
BUG-90 fix state mutation in getter

jobelenus:
Replacing some of `this` with state in the getter. But mostly this is about deleting the two mutations within the getter and putting that logic down in the subscription callback
squash


jobelenus:
bors merge

PR #3627, opened 2024-04-26
Restore all of func authoring

nickgerace:
## Description

image

This is it. This PR restores all of func authoring. The features affected include, but are not limited to the following: test func execution, save and exec func, save func fixes, attribute prototype argument associations fixes, delete func and more. Let's talk about it.

All routes related to func authoring return with this PR. You can now delete funcs, perform test executions on them, and save and exec funcs. Among these restored features comes fixes to how associations are handled, general save func fixes and other quality of life improvements. This also includes the next pass at evaluating and generating input types for `Props`, leaf nodes and `Components`.

There _will_ be bugs. Please see the **Disclaimers** section for more information.

## Primary Changes

- Restore `save_and_exec` func with `save_func` and `execute` (func) as the backing functions
- Restore `exec_func` with `dummy_execute_func` as the backing function
- Restore `list_input_sources` used for attribute bindings as their own sub-module
- Restore `delete_func`
- Fix input types for leaf functions
- Polish edges and existing bits of remaining func authoring logic (i.e. restored `WsEvents`)
- Remove unused modules, including but not limited to the following: attribute view, prop tree, reconciliation prototype and index map

## Secondary Changes

- Add ability to list ordered child props
- Ensure ordering of props at a single "level" is stable for the property editor
- Add ability for props to know their input types via `ts_types` (the implementation is not fully correct because type collection is inverted, but it is close)
- Remove unused modules, including but not limited to the following: attribute view, prop tree, reconciliation prototype and index map

## Known Issues

- Action prototypes via `save_func` are re-created when they shouldn't be and we should be handling this logic in a different way
- Auth prototypes via `save_func` are re-created when they shouldn't be and we should be handling this logic in a different way
- Input source and attribute binding parameter collection and display functionality is not polished
- Creating or updating attribute bindings may not function as intended
- `AttributePrototypeArgumentBags` need to be able to specify a `PropId` (replaces the implicit `InternalProviderId` field from the old engine)
- Reloading after `save_func` or updating attachments sometimes does not persist

## Disclaimer

Due to the scope of this change, it is likely that features contain rough edges or missing elements (especially test execution and "save_and_exec"). The goal of this commit is to unblock future work related to func authoring.

Usually, I'd aim for more certainty with the quality of my PRs, but the opportunity cost of holding up individual
piecemeal authoring features is greater than the cost of the team swarming and polishing the remaining bits and bugs. Moreover, the func authoring experience is relatively sandboxed, so the rest of the system should not regress if the func authoring experience is untouched.


app/web/src/components/FuncEditor/AttributeBindings.vue
```diff
@@ -200,9 +200,8 @@ const prototypeViews = computed(() =>
 
       const args = proto.prototypeArguments.map((arg) => ({
         name: funcArgumentsIdMap?.value[arg.funcArgumentId]?.name ?? "none",
-        prop: arg.inputSocketId
-          ? funcStore.inputSocketIdToSourceName(arg.inputSocketId) ?? "none"
-          : "none",
+        // TODO(nick): re-evaluate this when prop id is added back to the attribute prototype argument bag.
+        prop: arg.inputSocketId ?? "none",
```
nickgerace, line 204: You'll see a few changes like this. Essentially, the `AttributePrototypeArgumentBag` that is sent down needs a way to tell users "hey, I want my input from a `Prop`". In the old engine, it didn't need to do this. In fact, it had to do something worse: it found the implicit `InternalProviders` (remember those?) and then mapped them back to the `Props`. Now, we can just use the `Props` directly... just not in this PR due to its size and scope.

app/web/src/components/FuncEditor/FuncTest.vue
```diff
@@ -369,10 +369,14 @@ const prepareTest = async () => {
...
+          // TODO(nick): restore the idea behind how this worked. Now, prototype arguments likely
+          // need to EITHER get an inputSocketId or a propId. Probably, in the old engine, the
+          // "internalProviderId" meant either explicit or implicit. Since implicit internal providers
+          // are dead, we need to use the prop directly.
+          const inputSocket = funcStore.inputSocketForId(arg.inputSocketId ?? "");
+
+          if (inputSocket) {
+            return `${inputSocket.name}`;
           }
         }
```
nickgerace, line 381: Similar to the above comment: this is known to be broken.

app/web/src/store/func/funcs.store.ts
```diff
@@ -147,23 +147,9 @@ export const useFuncStore = () => {
             return undefined;
           },
 
-        proprForInputSocketId:
+        inputSocketForId:
```
nickgerace, line 150: This was the implicit `InternalProvider` to `Prop` conversion getter. It is now dead.

lib/dal-test/src/schemas/test_exclusive_schema_starfield.rs
```diff
@@ -73,6 +73,7 @@ pub(crate) async fn migrate_test_exclusive_schema_starfield(
           // Force the entries arg to be an Array (and return an empty array if the arg is absent/undefined/null).
           if (entries === undefined) return galaxies;
           if (entries === null) return galaxies;
+          if (entries.length == 0) return galaxies;
```
nickgerace, line 76: Just to make the test easier. Probably a good change anyway.

app/web/src/store/func/types.ts
```diff
@@ -78,7 +78,6 @@ export interface InputSourceProp {
   propId: string;
   kind: PropKind;
   schemaVariantId: string;
-  inputSocketId?: string;
```
nickgerace, line 81: Again, the `AttributePrototypeArgumentBag` will need a `PropId` field.

lib/dal/src/func.rs
```diff
@@ -445,34 +453,50 @@ impl Func {
...
 
-    pub async fn remove(ctx: &DalContext, id: FuncId) -> FuncResult<()> {
-        // to remove a func we must remove all incoming edges to it. It will then be
-        // garbage collected out of the graph
-
-        let workspace_snapshot = ctx.workspace_snapshot()?;
-
-        let arg_node_idx = workspace_snapshot.get_node_index_by_id(id).await?;
+    /// Deletes the [`Func`] and returns the name.
+    pub async fn delete_by_id(ctx: &DalContext, id: FuncId) -> FuncResult<String> {
```
nickgerace, line 457: The return of `delete_func`

lib/dal/src/func.rs
```diff
@@ -445,34 +453,50 @@ impl Func {
...
-                    change_set,
-                    user,
-                    arg_node_idx,
-                    EdgeWeightKind::new_use().into(),
-                )
-                .await?;
-        }
-
-        // Removes the actual node from the graph
         workspace_snapshot.remove_node_by_id(change_set, id).await?;
```
nickgerace, line 497: Handles the edge deletions too

lib/dal/src/func/associations.rs
```diff
@@ -341,47 +358,47 @@ impl FuncAssociations {
...
-    //         let ts_type = prop.ts_type(ctx).await?;
-    //
-    //         if !per_variant_types.contains(&ts_type) {
-    //             per_variant_types.push(ts_type);
-    //         }
-    //     }
-    //
-    //     Ok(per_variant_types.join(" | "))
-    // }
+    async fn compile_leaf_function_input_types(
```
nickgerace, line 361: Let's compile some leaf input types. We should be able to do something similar for all attribute funcs, especially now that `Props` have the `ts_types` method.

lib/dal/src/func/authoring.rs
```diff
@@ -116,7 +132,26 @@ impl FuncAuthoringClient {
         create::create_func(ctx, kind, name, options).await
     }
 
-    /// Saves a [`Func`] and returns the [result](SavedFunc).
+    /// Performs a "dummy" [`Func`] execution and returns the [result](DummyExecutionResult).
+    #[instrument(name = "func.authoring.dummy_execute_func", level = "info", skip_all)]
+    pub async fn dummy_execute_func(
```
nickgerace, line 137: Next pass: make it clear in the SDF routes that the `execute` route is for _test_ executions and has nothing to do with `save_and_exec`.

We'll also align all the naming to use either "dummy" or "test", but probably "test".

lib/dal/src/func/authoring/dummy_execute.rs
```diff
@@ -0,0 +1,54 @@
+use base64::engine::general_purpose;
+use base64::Engine;
+
+use crate::func::authoring::{DummyExecutionResult, FuncAuthoringResult};
+use crate::func::binding::FuncBinding;
+use crate::{ComponentId, DalContext, Func, FuncId};
+
+pub(crate) async fn dummy_execute_func(
```
nickgerace, line 8: Test execution returns

lib/dal/src/func/authoring/execute.rs
```diff
@@ -0,0 +1,45 @@
+use telemetry::prelude::warn;
+
+use crate::func::authoring::{FuncAuthoringError, FuncAuthoringResult};
+use crate::func::FuncKind;
+use crate::{AttributePrototype, AttributeValue, DalContext, Func, FuncId};
+
+pub(crate) async fn execute_func(ctx: &DalContext, id: FuncId) -> FuncAuthoringResult<()> {
```
nickgerace, line 7: `save_and_exec` returns

lib/dal/src/func/authoring/execute.rs
```diff
@@ -0,0 +1,45 @@
...
+    Ok(())
+}
+
+async fn update_values_for_func(ctx: &DalContext, func: &Func) -> FuncAuthoringResult<()> {
+    let attribute_prototype_ids = AttributePrototype::list_ids_for_func_id(ctx, func.id).await?;
+
+    if attribute_prototype_ids.is_empty() {
+        warn!(%func.id, "nothing to execute: no attribute prototype ids found for attribute func id");
+        return Ok(());
+    }
```
nickgerace, line 30: This was an error in the past, but a no-op should suffice in case the prototypes fall out.

lib/dal/src/func/authoring/execute.rs
```diff
@@ -0,0 +1,45 @@
...
+    let func = Func::get_by_id_or_error(ctx, id).await?;
+
+    match func.kind {
+        FuncKind::Attribute => update_values_for_func(ctx, &func).await?,
+        FuncKind::Action => {
+            // TODO(nick): fully restore or wait for actions v2. Essentially, we need to run
+            // every prototype using the func id for every component.
+            warn!("skipping action execution...");
+            return Ok(());
+        }
```
nickgerace, line 17: Here's the thing: do we want to _literally_ create stuff on `save_and_exec`? We should talk about this. That's why it's a warning right now.

lib/dal/src/func/authoring/save.rs
```diff
@@ -128,10 +128,15 @@ pub(crate) async fn save_func(
...
         FuncKind::Intrinsic | FuncKind::SchemaVariantDefinition | FuncKind::Unknown => {
-            return Err(FuncAuthoringError::NotWritable)
+            return Err(FuncAuthoringError::NotWritable(func.id))
         }
     }
 
+    WsEvent::func_saved(ctx, func.id)
+        .await?
+        .publish_on_commit(ctx)
+        .await?;
```
nickgerace, line 138: With this in place, we should probably remove the return payload of `save_func` and `save_and_exec` func. After all, we want `layerdb` to handle persistence and the rebaser to finish for us to know when things are ready to be fetched.

lib/dal/src/func/authoring/save.rs
```diff
@@ -331,15 +336,16 @@ async fn save_attr_func_proto_arguments(
     arguments: Vec<AttributePrototypeArgumentBag>,
 ) -> FuncAuthoringResult<()> {
     let mut id_set = HashSet::new();
+
     for arg in &arguments {
```
nickgerace, line 340: I go back and forth on handling this struct. We are probably due for a re-write that doesn't use `NONE` IDs at all, so I am holding off. Let's get feature parity first unless this becomes annoying... well, more than it already is.

lib/dal/src/func/binding.rs
```diff
@@ -351,6 +355,51 @@ impl FuncBinding {
         let (context, rx) = FuncDispatchContext::new(ctx);
         Ok((func, execution, context, rx))
     }
+
+    pub async fn execute_dummy(
```
nickgerace, line 359: Let's let `FuncBinding` own dummy execution since we need its private methods and exposing them is likely not a good idea.

lib/dal/src/func/binding.rs
```diff
@@ -351,6 +355,51 @@ impl FuncBinding {
...
+            .await
+            .map_err(Box::new)?;
+
+        let func_binding = Self::new(ctx, args, func_id, func_backend_kind).await?;
+
+        let (func, _execution, context, mut rx) = func_binding.prepare_execution(ctx).await?;
+
+        // TODO(nick): if we could do this without a rollback that would be great, especially since
+        // only have to work with the graph in memory.
+        ctx.rollback().await?;
```
nickgerace, line 377: We likely do not need this in the future. Just a port of the original code that works because `FuncBinding` stuff still uses the old tables.

lib/dal/src/func/binding.rs
```diff
@@ -366,3 +415,11 @@ impl WsEvent {
...
     }
 }
+
+// TODO(nick): centralize this.
+async fn publish_immediately(ctx: &DalContext, ws_event: WsEvent) -> WsEventResult<()> {
+    let subject = format!("si.workspace_pk.{}.event", ws_event.workspace_pk());
+    let msg_bytes = serde_json::to_vec(&ws_event)?;
+    ctx.nats_conn().publish(subject, msg_bytes.into()).await?;
+    Ok(())
+}
```
nickgerace, line 425: We should never be doing this. Just a port of the original code. Will need to cull in a second pass.

lib/dal/src/input_sources.rs
```diff
@@ -0,0 +1,171 @@
...
+    pub schema_variant_id: SchemaVariantId,
+    pub prop_id: PropId,
+    pub kind: PropKind,
+    pub name: String,
+    pub path: String,
+}
+
+#[derive(Deserialize, Serialize, Debug)]
+#[serde(rename_all = "camelCase")]
+pub struct InputSources {
```
nickgerace, line 58: `list_input_sources` is restored

lib/dal/src/prop.rs
```diff
@@ -874,28 +874,43 @@ impl Prop {
         Ok(())
     }
 
-    #[allow(dead_code)]
-    async fn get_content(
-        ctx: &DalContext,
-        prop_id: PropId,
-    ) -> PropResult<(ContentHash, PropContentV1)> {
+    /// List [`Props`](Prop) for a given list of [`PropIds`](Prop).
+    pub async fn list_content(ctx: &DalContext, prop_ids: Vec<PropId>) -> PropResult<Vec<Self>> {
```
nickgerace, line 878: Batch collect props

lib/dal/src/prop.rs
```diff
@@ -926,4 +941,99 @@ impl Prop {
...
+                        .await?
+                    {
+                        let child_prop =
+                            Self::get_by_id_or_error(ctx, child_prop_weight.id().into()).await?;
+                        ordered_child_props.push(child_prop);
+                    }
+                }
+                ordered_child_props
+            }
+        };
```
nickgerace, line 978: Collect child props in order. This contains safety rails in case you accidentally call this from a `Prop` whose type is a scalar.

lib/dal/src/prop.rs
```diff
@@ -926,4 +941,99 @@ impl Prop {
...
+        Ok(ordered_child_props)
+    }
+
+    // TODO(nick): this is straight up broken and inverted. You need to encapsulate the child type
+    // in the parent type. The problem is that this was ported from a recursive function in the old
+    // engine and is now a work queue. The collection pattern is ironically inverted: now, we start
+    // at the highest level and descend (instead of immediately descending and popping the result
+    // back up).
+    pub async fn ts_type(&self, ctx: &DalContext) -> PropResult<String> {
+        let mut work_queue = VecDeque::new();
```
nickgerace, line 989: The comment does not lie! This needs a second pass, but it is close.

lib/dal/src/prop.rs
```diff
@@ -926,4 +941,99 @@ impl Prop {
...
+                    for child in &Self::direct_child_props_ordered(ctx, prop_id).await? {
+                        // We serialize the object key as a JSON string because its
+                        // the easiest way to ensure we create a valid TS interface
+                        // even with keys that are not valid javascript identifiers.
+                        // (e.g., we escape quotes in the prop name this way)
+                        let name_value = serde_json::to_value(&child.name)?;
+                        let name_serialized = serde_json::to_string(&name_value)?;
+                        object_interface
+                            .push_str(format!("{name_serialized}: {ts_type};\n").as_str());
+                        work_queue.push_back((child.id, child.kind));
```
nickgerace, line 1017: Port of the original, for context

lib/dal/src/property_editor/schema.rs
```diff
@@ -31,38 +29,24 @@ impl PropertyEditorSchema {
...
-                        EdgeWeightKindDiscriminants::Use,
-                    )
-                    .await?
-                {
-                    if let NodeWeight::Prop(child_prop_weight) = workspace_snapshot
-                        .get_node_weight(child_prop_node_index)
-                        .await?
-                    {
-                        let child_prop_id: PropId = child_prop_weight.id().into();
+                let child_props = Prop::direct_child_props_ordered(ctx, prop_id).await?;
```
nickgerace, line 42: Fixes any potential mis-ordering in the property editor

lib/dal/src/schema/variant.rs
```diff
@@ -335,6 +335,16 @@ impl SchemaVariant {
         Ok(prop_ids)
     }
 
+    /// Returns all [`Props`](Prop) for a given [`SchemaVariantId`](SchemaVariant).
+    pub async fn all_props(
```
nickgerace, line 339: Not using this anymore, but may in a follow up. I'd normally prefer dead code to be culled, but it may become immediately useful again.

lib/dal/tests/integration_test/func.rs
```diff
@@ -74,3 +77,245 @@ async fn duplicate(ctx: &mut DalContext) {
     assert_eq!(duplicated_func.handler, func.handler);
     assert_eq!(duplicated_func.code_base64, func.code_base64);
 }
+
+#[test]
+async fn create_and_delete_attribute_func_with_arguments(ctx: &mut DalContext) {
```
nickgerace, line 82: Deletion feels good

lib/dal/tests/integration_test/func.rs
```diff
@@ -74,3 +77,245 @@ async fn duplicate(ctx: &mut DalContext) {
...
+
+    let root_prop_id = SchemaVariant::get_root_prop_id(ctx, schema_variant_id)
+        .await
+        .expect("could not get root prop id");
+    let root_prop = Prop::get_by_id_or_error(ctx, root_prop_id)
+        .await
+        .expect("could not get prop by id");
+
+    // TODO(nick): check that the ts type is right!
+    let _ts_type = root_prop.ts_type(ctx).await.expect("could not get ts type");
```
nickgerace, line 320: The comment is right!

lib/dal/tests/integration_test/func/authoring/dummy_execute.rs
```diff
@@ -0,0 +1,101 @@
+use dal::func::authoring::FuncAuthoringClient;
+use dal::func::backend::js_action::DeprecatedActionRunResult;
+use dal::func::view::FuncView;
+use dal::{DalContext, Func};
+use dal_test::helpers::{create_component_for_schema_name, ChangeSetTestHelpers};
+use dal_test::test;
+use veritech_client::ResourceStatus;
+
```
nickgerace, line 8: Surprisingly, dummy execution feels good too

lib/sdf-server/src/server/service/func.rs
```diff
@@ -65,70 +65,24 @@ pub type FuncResult<T> = Result<T, FuncError>;
...
+        .route("/save_and_exec", post(save_and_exec::save_and_exec))
         .route("/save_func", post(save_func::save_func))
-    //         .route("/delete_func", post(delete_func::delete_func))
-    //         .route("/save_and_exec", post(save_and_exec::save_and_exec))
-    //         .route("/execute", post(execute::execute))
-    //         .route(
-    //             "/list_input_sources",
-    //             get(list_input_sources::list_input_sources),
-    //         )
+}
```
nickgerace, line 81: Life.

lib/dal/src/func/authoring/execute.rs
britmyerss: Yeah, we threw a warning on the front end for create and deletes. I think that's ok because if you're authoring new actions, it seems reasonable that we'd need to do some manual cleanup. Or, if you target localstack, less risk, but either way I think we need this.

nickgerace: Good to know! Can you do it on HEAD only or any change set?

nickgerace:
bors merge this monstrosity

nickgerace:
bors merge

nickgerace:
bors cancel

nickgerace:
bors merge

PR #3628, opened 2024-04-26
fix(dal): Ensure that we export prop widget_options

stack72:


stack72:
bors merge

stack72:
bors merge

PR #3629, opened 2024-04-26
Added to_delete to inferred connections and ensure output sockets can drive inferred and manual connections simultaneously

britmyerss:
This PR moves around the check for deletes so we can pass the flag appropriately on inferred connections. I also added a test to make sure output sockets can pass values to children and manually connected components

<div><br/>via <a href="https://giphy.com/channel/tkyle/">T. Kyle</a> on <a href="https://giphy.com/gifs/realitytvgifs-rhonj-danielle-staub-tgif-xzDIIZCozVXI4">GIPHY</a></div>

### DIFF
--- a/lib/dal/src/attribute/value.rs
+++ b/lib/dal/src/attribute/value.rs
@@ -531,10 +531,6 @@ impl AttributeValue {
         let destination_component_id =
             AttributeValue::component_id(ctx, attribute_value_id).await?;
         let value_is_for = AttributeValue::is_for(ctx, attribute_value_id).await?;
-        info!(
-            "Attribute value Id {} value is for {:?}",
-            attribute_value_id, value_is_for
-        );
         let apa_ids = AttributePrototypeArgument::list_ids_for_prototype(ctx, prototype_id).await?;
         let mut func_binding_args: HashMap<String, Vec<serde_json::Value>> = HashMap::new();
         if apa_ids.is_empty() {
@@ -722,7 +718,7 @@ impl AttributeValue {
         })
     }
 
-    #[instrument(level = "info", skip_all)]
+    #[instrument(level = "debug", skip_all)]
     async fn get_inferred_input_value(
         ctx: &DalContext,
         input_attribute_value_id: AttributeValueId,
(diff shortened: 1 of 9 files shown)


britmyerss:
bors try

lib/dal/src/component.rs
```diff
@@ -1520,7 +1523,37 @@ impl Component {
 
         Ok(result)
     }
-
+    /// Checks the destination and source component to determine if data flow between them
+    /// Both "deleted" and not deleted Components can feed data into
+    /// "deleted" Components. **ONLY** not deleted Components can feed
+    /// data into not deleted Components.
+    pub async fn should_data_flow_between_components(
```
britmyerss, line 1530: added a helper to consolidate the logic

lib/dal/src/component.rs
```diff
@@ -1732,24 +1765,35 @@ impl Component {
...
     }
     /// For a given [`InputSocketMatch`], find the single inferred [`OutputSocketMatch`] that is driving it
     /// if it exists. This walks up or down the component lineage tree depending on the [`ComponentType`]
     /// and finds the closest matching [`OutputSocket`]
-    #[instrument(level = "debug", skip_all)]
-    pub async fn find_inferred_connection_to_input_socket(
+    ///
+    /// Note: this does not check for whether data should actually flow between components
+    #[instrument(level = "debug", skip(ctx))]
+    pub async fn find_potential_inferred_connection_to_input_socket(
```
britmyerss, line 1789: I renamed some of these functions as now, whether there is an inferred connection is a separate concern from whether the data should flow through.

lib/dal/src/component.rs
```diff
@@ -1921,13 +1975,29 @@ impl Component {
...
+                    .await?
             {
+                // add the check for to_delete on either to or from component
+                // Both "deleted" and not deleted Components can feed data into
+                // "deleted" Components. **ONLY** not deleted Components can feed
+                // data into not deleted Components.
+                let destination_component = Self::get_by_id(ctx, to_component_id).await?;
+                let source_component =
+                    Self::get_by_id(ctx, output_socket_match.component_id).await?;
+                let to_delete = !Self::should_data_flow_between_components(
```
britmyerss, line 1988: derives whether data should flow through based on the `to_delete` state of either end of the connection so we can handle these the same on the front end.

lib/dal/tests/integration_test/frame.rs
```diff
@@ -376,6 +380,99 @@ async fn simple_frames(ctx: &mut DalContext) {
     }
 }
 
+#[test]
+async fn output_sockets_can_have_both(ctx: &mut DalContext) {
```
britmyerss, line 384: new test! lots more testing to do here to make sure we're handling deletes correctly, nesting, I've got a plan but that can come later

lib/dal/src/attribute/value.rs
```diff
@@ -744,42 +740,37 @@ impl AttributeValue {
...
-                            .await
-                            .map_err(Box::new)?;
-                    let destination_component =
-                        Component::get_by_id(ctx, output_match.component_id)
-                            .await
-                            .map_err(Box::new)?;
-
-                    if destination_component.to_delete()
-                        || (!destination_component.to_delete() && !source_component.to_delete())
+                    if !Component::should_data_flow_between_components(
```
stack72, line 753: Love this refactoring!!

lib/dal/src/attribute/value/dependent_value_graph.rs
```diff
@@ -163,52 +158,46 @@ impl DependentValueGraph {
...
-                        {
-                            work_queue.push_back(input_socket_match.attribute_value_id);
-                            dependent_value_graph.value_depends_on(
-                                input_socket_match.attribute_value_id,
-                                current_attribute_value_id,
-                            );
+            // Also check if this value is an output socket as the attribute value
+            // might have implicit dependendcies based on the ancestry (aka frames/nested frames)
+            // note: we filter out non-deleted targets if the source component is set to be deleted
+            if let ValueIsFor::OutputSocket(_) = value_is_for {
```
stack72, line 164: Do we no longer need to check if apas.is_empty?

britmyerss: correct! if the attribute value for an output socket changes, we want to queue inferred AND explicit connections (so drawing edges and having components inside of down frames for example). Doing this check is what was preventing data from flowing to both components in a frame and components connected to the frame's output socket

britmyerss:
bors merge

PR #3630, opened 2024-04-26
fix(sdf, dal, web): Change Asset Panel to use schemaID

stack72:
We were using SchemaVariantId to load the Variant details. This meant when we needed to clone OR to create a new variant, the screen would have redirected

We are now going to use SchemaId as the identitifer and we will then load the default_schema_variant for the schema


stack72:
bors merge

PR #3633, opened 2024-04-29
feat(dal): Update SummaryComponent to suggest if on latest schema variant

stack72:
We can use this to understand if we can upgrade a component to a new version of a schema variant


stack72:
bors merge

PR #3636, opened 2024-04-29
feat(dal): add exclusive outgoing edge conflicts to rebase

zacharyhamm:
Many nodes assume complete control of their direct, outgoing neighbors, and allowing them to be silently merged with other neighbors from another changeset would create an incorrect graph. For example, an AttributeValue should only ever have one outgoing Prototype edge defining the "component specific" prototype for that value. If another user merges in a changeset with an updated prototype edge that we have never seen in our change set, that should be considered a conflict, since otherwise the result of the rebase would be an AttributeValue with *two* Prototype edges, an incorrect state. The same applies for "Contains" edges and many others. This change will produce an `ExclusiveEdgeMismatch` conflict when the snapshot being rebased (the `to_rebase`) has a container with edges not seen by the `onto`, *if* that container defines the edge kinds as exclusive edges. Be sure to add these exclusions the node weight when adding a new edge kind, if it makes sense to do so.

### DIFF
--- a/lib/dal/src/workspace_snapshot/node_weight.rs
+++ b/lib/dal/src/workspace_snapshot/node_weight.rs
@@ -9,7 +9,6 @@ use si_events::{
 use strum::EnumDiscriminants;
 use thiserror::Error;
 
-use crate::func::execution::FuncExecutionPk;
 use crate::workspace_snapshot::vector_clock::VectorClockId;
 use crate::{
     change_set::{ChangeSet, ChangeSetError},
@@ -19,6 +18,7 @@ use crate::{
     },
     PropKind,
 };
+use crate::{func::execution::FuncExecutionPk, EdgeWeightKindDiscriminants};
 
 use crate::func::FuncKind;
 pub use action_node_weight::ActionNodeWeight;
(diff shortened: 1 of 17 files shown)


lib/dal/src/workspace_snapshot/node_weight.rs
```diff
@@ -457,6 +457,36 @@ impl NodeWeight {
...
+    /// relative to another changeset in order to have a correctly constructed
+    /// graph. For example, only one set of children of a given attribute value
+    /// should "win" in a rebase operation, otherwise there could be duplicate
+    /// child values for an attribute value. This method will be called during
+    /// conflict detection in order to produce a conflict if the change set
+    /// being rebased has unseen edges of this type for a given "container"
+    /// node. If edge kinds are not returned here, those unseen edges will be
+    /// silently merged with the `onto` changeset's edges. This a "business"
+    /// logic problem, rather than a purely graph-theoretical one.
+    pub const fn exclusive_outgoing_edges(&self) -> &[EdgeWeightKindDiscriminants] {
```
jhelwig, line 470: Mostly just an observation:

I think this fairly clearly points us down the path of having more granular (rather than generic) edge weight kinds. It would be very difficult to express that _only_ `Use` edges that point to a specific `ContentAddressDiscriminant` `ContentNodeWeight` should be "exclusive", and that any other `Use` edges are allowed to an arbitrary number. At the very least, any singleton relationship would need to have a unique `EdgeWeightKindDiscriminant` among all of the edge weight kinds used with a particular `NodeWeightDiscriminant`.

zacharyhamm: That was 100% a thought I had while implementing this. We likely need more distinction among edge weight kinds

PR #3638, opened 2024-04-29
ENG-2508 update edge logic

jobelenus:
Edges no longer have an inherent ID. They are defined by what they connect. Diagram::assemble now returns edges and inferredEdges and we need to reconcile both of those in the UI.

1. Change Vue to use edgeId that is a compound key made up by {to_socket_ulid}_{from_socket_ulid} for all edges
2. Change SDF to remove id  and edgeId from Diagram::assemble
3. In Vue add inferred edges to the store edgesById where the rest of the edges live.
4. Change SDF to accept two params {to_socket_ulid} and {from_socket_ulid} to delete an edge and to restore an edge
5. On detach component from parent remove inferred edges between the two components.
6. On attach component to parent fire the component updated WsEvent (which fires the big hammer).
7. isInvisible should reflect whether an edge is inferred or not.

knife_fingers

### DIFF
--- a/app/web/src/components/ModelingDiagram/DiagramEdge.vue
+++ b/app/web/src/components/ModelingDiagram/DiagramEdge.vue
@@ -28,7 +28,7 @@
         stroke: strokeColor,
         strokeWidth: 2,
         hitStrokeWidth: 10,
-        listening: !edge.def.isInvisible,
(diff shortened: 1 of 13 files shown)


jobelenus:
bors merge

jobelenus:
bors merge

PR #3641, opened 2024-04-29
chore(pg): update hostnames in self signed cert

adamhjk:
This updates the hostnames in the self signed cert, in preparation for adding pgbouncer in CI and Dev.


adamhjk:
bors merge

PR #3642, opened 2024-04-29
More Frames Fixes

britmyerss:
This PR contains the following fixes: 

- accurately infer if an input socket is manually configured or not (I was previously not filtering correctly by destination component)
- grab the correct Attribute Prototype Argument when showing inputs in the debug panel
- Display more helpful info in the debug panel. I added the value source, value id source, whether or not the data should pass through, and I now include inferred inputs to attribute prototypes (which matters only for input sockets) 
- Fixed a false positive integration test, that was falsely passing due to the above bug on inferring whether an input socket is manually inferred


britmyerss:
bors try

britmyerss:
bors merge

PR #3643, opened 2024-04-29
Polish func authoring: attribute bindings, test executions, and more

nickgerace:
## Description

This PR is a follow-up to #3627 and it focuses polishes a lot of the func authoring experience.

## Attribute Bindings Changes

Updating and adding attribute bindings for components has been polished. Adding, editing and removing attribute bindings for components should work as intended. However, there are remaining issues with attribute bindings for schema variants, but they will be touched on later in the commit.

There are now an integration test for creating a new prototype and prototype argument for an existing attribute func with an existing attribute binding.

---
image
---
image
---
image

## `FuncAssociations` Changes

A key change to getting attribute bindings in a better state as well as overall improvements was to restore the idea of "implicit internal providers" for attribute func associations. Essentially, a `prop_id` field now takes a similar role to before. The collection and mutation of the field is now backed by new and old integration tests alike.

## Func Test Execution Changes

All naming related to the "test execution" of funcs has been realigned. The "critical" section layer of func execution has been sandboxed in order to avoid transaction rollback scenarios (i.e. no bindings, return values nor execution recorded in old database tables).

---
image

## Func Deletion and Detachment Changes

Detaching funcs and deleting funcs works. Func deletion will be unsuccessful if there are existing associations, but we do not make it clear in the UI when this happens.

## Known Issues:

- Working with attribute bindings at the schema variant level sees an "attribute prototype has been orphaned" error
- Attribute funcs corresponding to attribute bindings at the component level do not appear to run on apply
- Editing an attribute func in the interfere does not show existing options in the modal
- Static argument values are not available in the modal for selection
- It is not obvious when func associations existing blocks func deletion


app/web/src/components/FuncEditor/AttributeBindings.vue
```diff
@@ -200,8 +200,9 @@ const prototypeViews = computed(() =>
 
       const args = proto.prototypeArguments.map((arg) => ({
         name: funcArgumentsIdMap?.value[arg.funcArgumentId]?.name ?? "none",
-        // TODO(nick): re-evaluate this when prop id is added back to the attribute prototype argument bag.
-        prop: arg.inputSocketId ?? "none",
+        prop: arg.propId
+          ? funcStore.propIdToSourceName(arg.propId) ?? "none"
+          : "none",
```
nickgerace, line 205: How this worked in the old engine: this was the "implicit internal provider to prop name" converter.

app/web/src/components/FuncEditor/AttributeBindings.vue
```diff
@@ -82,7 +82,7 @@ import { computed, inject, ref, Ref, watch } from "vue";
 import { VButton } from "@si/vue-lib/design-system";
 import {
   AttributeAssociations,
-  AttributePrototypeView,
+  AttributePrototypeBag,
```
nickgerace, line 85: Renamed to match backend.

app/web/src/components/FuncEditor/AttributeBindingsModal.vue
```diff
@@ -179,12 +179,10 @@ const inputSourceOptions = computed<Option[]>(() => {
...
 
-  // TODO(nick): restore the true idea behind this. Likely requires more refactoring of prototype
-  // arguments in the backend to support both inputSocketId and propId.
   const propOptions =
     funcStore.inputSourceProps[selectedVariantId]?.map((prop) => ({
-      label: `Attribute: ${prop.path}${prop.name}`,
-      value: nilId(),
+      label: `Attribute: ${prop.path}`,
+      value: prop.propId,
     })) ?? [];
```
nickgerace, line 186: Now matches how output socket options work since implicit internal providers are gone.

app/web/src/components/FuncEditor/FuncTest.vue
```diff
@@ -297,7 +297,7 @@ const testComponentDisplayName = computed(() => {
 });
 
 const testStatus = computed(() => {
-  const status = funcStore.getRequestStatus("EXECUTE").value;
+  const status = funcStore.getRequestStatus("TEST_EXECUTE").value;
```
nickgerace, line 300: Naming alignment.

lib/dal/src/context.rs
```diff
@@ -1160,7 +1160,7 @@ async fn rebase(
         )
         .await?;
     info!("got response from rebaser: {:?}", start.elapsed());
-    info!(
+    debug!(
```
nickgerace, line 1163: De-noising.

lib/dal/src/func/authoring/save.rs
```diff
@@ -193,27 +193,32 @@ async fn save_auth_func_prototypes(
...
         };
         id_set.insert(attribute_prototype_id);
 
-        save_attr_func_proto_arguments(ctx, prototype_view.id, prototype_view.prototype_arguments)
-            .await?;
+        // Use the attribute prototype id variable rather than the one off the iterator so that we
+        // don't use the nil one by accident.
+        save_attr_func_proto_arguments(
+            ctx,
+            attribute_prototype_id,
```
nickgerace, line 218: The comment is saying "this destroyed Nick for longer than he'd like to admit".

lib/dal/src/func/binding/critical_section.rs
```diff
@@ -0,0 +1,104 @@
...
+use crate::func::binding::{FuncBindingError, FuncBindingResult};
+use crate::{Func, FuncBackendKind};
+
+/// Perform function execution to veritech for a given [`Func`] and [`FuncDispatchContext`].
+///
+/// This function does not persist data to "layerdb". If data persistence for execution is required,
+/// use [`FuncBinding::create_and_execute`](crate::FuncBinding::create_and_execute).
+///
+/// _Warning:_ this function should be used with caution!
+pub(crate) async fn execute_critical_section(
```
nickgerace, line 29: I went back and forth on splitting this out. Ultimately, the idea of "here's my func, args, and before funcs, let's roll" is a solid bounded context. However, we _really_ do not want multiple ways to execute funcs, so "test execute func" and `FuncBinding::create_and_execute` should be the only two things that call this function. Let's keep an eye on this moving forward.

lib/dal/tests/integration_test/func/associations.rs
```diff
@@ -133,6 +134,115 @@ async fn for_attribute(ctx: &mut DalContext) {
...
+            "domain",
+            "possible_world_a",
+            "wormhole_1",
+            "wormhole_2",
+            "wormhole_3",
+            "rigid_designator",
+        ]),
+    )
+    .await
+    .expect("could not find prop by path");
```
nickgerace, line 218: Let's see where it's coming from and where it's going.

lib/dal/tests/integration_test/func/associations.rs
```diff
@@ -41,17 +41,7 @@ async fn for_action(ctx: &mut DalContext) {
...
-        .await
-        .expect("could not perform find by name")
-        .expect("no schema found");
-    let schema_variant_id = schema
-        .get_default_schema_variant_id(ctx)
-        .await
-        .expect("could not perform get default schema variant")
-        .expect("default schema variant not found");
-
+async fn for_attribute_with_input_socket_input(ctx: &mut DalContext) {
```
nickgerace, line 44: The diff didn't make this obvious, but this test only looked at an attribute func with a single input from an `InputSocket`. The new test in this module is similar, but has a `Prop` as the single input.

lib/dal/tests/integration_test/func/authoring/save_func/attribute.rs
```diff
@@ -448,6 +458,209 @@ async fn delete_add_modify(ctx: &mut DalContext) {
     }
 }
 
+#[test]
+async fn create_attribute_prototype_with_argument(ctx: &mut DalContext) {
```
nickgerace, line 462: This is the big one. We have to do some gymnastics for the assertions, but it reflects the creation of new attribute bindings when one already exists.

nickgerace:
bors merge

PR #3644, opened 2024-04-29
chore(pg): update certificates for dev

adamhjk:
this updates all the right certs for dev, and keeps the rootca around so that we can generate new certificates later if we need it.


adamhjk:
bors merge

adamhjk:
bors merge

PR #3646, opened 2024-04-30
feat(dal, sdf): Restore variant edit with new version and variant save

stack72:
This restores the ability to use exec_variant (now called update_variant) but only when components exist on the graph. Next up will be the ability to edit a schema variant when there are no components on the graph


stack72:
bors merge

PR #3654, opened 2024-04-30
Remove badge for nightly CI builds

nickgerace:
## Description

This PR removes the badge for nightly CI builds. With the new engine, we believe we do not need nightly CI runs (for now) because equivalent tests are expected to be much faster to run. Combine that with the fact that nightly runs have not been ran in nearly two months and it's evident that it's time to retire the badge.


nickgerace:
bors merge

PR #3655, opened 2024-04-30
Make links between documents easier to read

nickgerace:
## Description

This PR makes links between documents easier to read by using naturally flowing sentences rather than hard references by file names.


nickgerace:
bors merge

PR #3657, opened 2024-04-30
feat(module-index): Import workspace endpoint on module index

vbustamante:
We're merging the modules index changes first since they're additive, and we can test the export/import loop more thoroughly - wiping the local database also wipes the local module index


vbustamante:
bors merge

PR #3660, opened 2024-04-30
Fix attribute bindings not appearing in the asset screen

nickgerace:
## Description

This PR fixes attribute bindings not appearing in the asset screen by using the schema variant id instead of the schema id.

Screenshot 2024-04-30 at 4 36 48 PM


nickgerace:
bors merge

PR #3665, opened 2024-05-01
feat: only run pganalyze migration if not in rds

sprutton1:
The user we run `sdf` under does not have the appropriate permissions to run the pganalyze migration in prod. We do not want to escalate permissions just to run this migration. This change checks if the `rds_superuser` role exists and if we have it. If it doesn't or it does and we have it, it will run the migration. This cause no disruption to the local dev experience while allowing us to perform migrations in rds without extra hoop jumping.

### DIFF
--- a/lib/dal/src/migrations/U0644__pganalyze_setup.sql
+++ b/lib/dal/src/migrations/U0644__pganalyze_setup.sql
@@ -7,52 +7,56 @@ BEGIN
     ) THEN
         RAISE NOTICE 'Role ''pganalyze'' already exists. Skipping pganalyze setup.';
     ELSE
-        CREATE USER pganalyze WITH PASSWORD 'tSYwfHHdBBlfWlxa' CONNECTION LIMIT 5;
-        GRANT pg_monitor TO pganalyze;
+        IF NOT EXISTS (
+            SELECT FROM pg_catalog.pg_roles
+            WHERE rolname = 'rds_superuser'
+        ) OR pg_has_role(current_user, 'rds_superuser') THEN
+            CREATE USER pganalyze WITH PASSWORD 'tSYwfHHdBBlfWlxa' CONNECTION LIMIT 5;
+            GRANT pg_monitor TO pganalyze;
 
-        CREATE SCHEMA IF NOT EXISTS pganalyze;
-        GRANT USAGE ON SCHEMA pganalyze TO pganalyze;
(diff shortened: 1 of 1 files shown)


lib/dal/src/migrations/U0644__pganalyze_setup.sql
```diff
@@ -7,52 +7,56 @@ BEGIN
...
-        -- For pganalyze buffer cache report
-        CREATE EXTENSION IF NOT EXISTS pg_buffercache WITH SCHEMA public;
-        CREATE OR REPLACE FUNCTION pganalyze.get_buffercache() RETURNS SETOF public.pg_buffercache AS
-        $$
-        /* pganalyze-collector */ SELECT * FROM public.pg_buffercache;
-        $$ LANGUAGE sql VOLATILE SECURITY DEFINER;
+        IF NOT EXISTS (
+            SELECT FROM pg_catalog.pg_roles
+            WHERE rolname = 'rds_superuser'
+        ) OR pg_has_role(current_user, 'rds_superuser') THEN
```
johnrwatson, line 13: Does this not need to be and AND?

**We want the migration to run if:**
- It's never run before, and
- the user has the right permission

I believe as is, the above will:
- Run locally always on fresh DB
- Run in RDS always on fresh DB (whether the user permission is set or not)

---
I think that locally, the user won't have the rds-superuser perm (as you'd expect) so we may need to adjust that logic if we swap it to an AND

johnrwatson (commented):
Query on logical statement

lib/dal/src/migrations/U0644__pganalyze_setup.sql
sprutton1: This first IF will eval to `true` when the role does not exist, meaning it will always migrate locally. It will be `false` on rds, so it will then check if the user has the role. If they do, it will run. If not, it won't.

johnrwatson: Ok sweet, sounds good to me!

johnrwatson (approved):
LGTM based on comments

sprutton1:
@nickgerace I know you had comments on the previous attempt at this, so if you could give it a look I would appreciate it.

nickgerace (approved):
Love it!

nickgerace:
bors merge

PR #3667, opened 2024-05-01
Batch Set Component Position

jobelenus:
This way we get only 1 job for the rebaser no matter how many components we are moving at a given time.


jobelenus:
bors merge

PR #3668, opened 2024-05-01
feat: add simple deploy workflow for tools-prod

sprutton1:
Adds a dead simple deploy button with a drop down. This will either deploy all services or a specific one. All it does is tell the relevant autoscaling group to refresh it's instance. There is nothing here about deploying a specific version, it simply resets the cloudfront cache for the artifact store and then the ASGs pull the latest `stable` version on startup. Don't overthink it.


sprutton1:
bors merge

PR #3671, opened 2024-05-01
chore(test): remove dbg and println

adamhjk:
This PR removes dbg and println statements from the test suites, as they were mostly just generating noise (in particular in CI). Let's try and keep this tidy in the future :)


adamhjk:
bors merge

PR #3672, opened 2024-05-01
fix(dal, sdf): Ensure update_asset works in all scenarios

stack72:
In order to support workspace level imports, the import code had a Changeset based thingmap. That thingmap was causing issues in variant export because some funcs were being added to the map with the incorrect changeset id

We no longer use the import_workspace code path so we only ever import a single changeset at a time

We have no rid the world of a Changeset scoped Thingmap and this means the update_variant is a happier world and we can now happily upgrade variants!


stack72:
bors try

stack72:
bors merge

PR #3675, opened 2024-05-01
fix(web): Ensure that the count of changes in the changeset doesn't include diffs

stack72:
We were showing a Pill counter beside "Apply Changes" that was Actions + diffs - we need to remove the diffs as these are no longer listed in the changeset


stack72:
bors merge

PR #3677, opened 2024-05-01
Fix test executions for attribute funcs and missing payload value

nickgerace:
## Description

This PR fixes test executions for attribute funcs. It also scopes test execution abilities to solely attribute, code generation and qualification funcs.

The existing code assumes that there was only one binding for attribute funcs. Attribute funcs may have multiple bindings, which means that they can have different input sources (`AttributePrototypeArguments`) for the same `FuncArguments`. For example: if you have a `FuncArgument` that takes in a string called `toddhoward`, it is possible for one binding to get that value from `/root/si/name` and another binding to get that value from an input socket who provides a string.

The existing code also assumed that there was only one `FuncArgument`. Now, we derive the arguments from the backend and avoid walking the result of `component/json` (as well as assuming that the `FuncArgument` name will match the name of a given `Prop`).

This PR also fixes a bug where `/root/resource/payload` would not be returned from the `component/json` route due to a dropped field during de/serialization.

## Additional Fixes
- Placeholder text colors for `VormInput` now work in both light and dark mode
- Split selector details out into a new component: `FuncTestSelector`
- Split argument preparation for prototype execution into its own method with a corresponding integration test
- Removed unused code related to fetching latest func executions
- Add new `/attribute` endpoint and alongside a route to provide prototype arguments

## Screenshots: Attribute Func Test (Light Mode)

image

---

image

---

image

## Screenshots: Attribute Func Test (Dark Mode)

image

---

image

---

image

## Screenshots: Post Test Execution of an Attribute Func (Dark Mode)

image

## Screenshots: Code Generation Func Test with Execution (Light Mode)

image

---

image

---

image

## Screenshots: Qualification Func Test with Execution (Dark Mode)

image

---

image

---

image

## Screenshots: No Func Test Tab for Action or Authentication Funcs (Light Mode)

image

---

image

### DIFF
--- a/app/web/src/components/FuncEditor/FuncDetails.vue
+++ b/app/web/src/components/FuncEditor/FuncDetails.vue
@@ -32,12 +32,7 @@
           <Stack class="p-2 border-b dark:border-neutral-600" spacing="xs">
             <div class="flex gap-1 flex-wrap">
               <VButton
-                v-if="
-                  testPanelEnabled &&
-                  funcStore.selectedFuncDetails &&
-                  funcStore.selectedFuncDetails?.associations?.type !==
-                    'authentication'
-                "
+                v-if="enableTestPanel"
                 class="--tone-action"
                 icon="save"
                 size="md"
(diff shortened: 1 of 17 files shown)


nickgerace:
bors merge

stack72:
:shipit:

PR #3684, opened 2024-05-02
Take new frames into account when tracking actions ordering

paulocsanz:
We defined actions ordering through regular component connections, now that frames don't create them, actions ordering was not taking frames into account.

Also fixes pasting components into a frame.


paulocsanz:
bors merge

paulocsanz:
bors merge

PR #3685, opened 2024-05-02
Make func test selector placeholder refer to bindings

nickgerace:
This commit makes the func test selector placeholder for attribute funcs refer to bindings. The word "prototype" does not appear in this user domain and should not be used here.


nickgerace:
bors merge

PR #3687, opened 2024-05-02
Forbid editing functions on head

paulocsanz:
- No CRDT on head, since it should be set from an apply
- Stop creating two change-sets (saving on the edit and then on unmount)
- Create new change-set with the updated func, leaving head untouched


paulocsanz:
bors merge

PR #3691, opened 2024-05-02
feat(web): Removal of SiChip to denote if something is builtin or not

stack72:
This is a controversial one - TL;DR, I don't believe a user cares about something being builtin or not. All a builtin is, is something we give a user at application startup time. It was more important in the old engine where there was no concept of lineage

In this new engine, we can allow editing of all builtins EXCEPT the intrinsics because they are important to the system. This feels less confusing to me, as a user now

Screenshot 2024-05-02 at 23 20 01


stack72:
bors merge

PR #3692, opened 2024-05-03
Fix tooltip on attributes panel, add some tracing for rebaser loop for better visibility

britmyerss:
Before this change, we'd see tooltips like this (a KeyPair in a Region): 
image

After this change, it'd look like this: 
image


app/web/src/components/AttributesPanel/AttributesPanelItem.vue
```diff
@@ -861,6 +860,8 @@ const sourceTooltip = computed(() => {
       return `${propName.value} is set by a dynamic function`;
     } else if (propHasSocket.value) {
       return `${propName.value} is set via an empty socket`;
+    } else if (propControlledByParent.value) {
+      return `${propName.value} is set via a function from an ancestor`;
```
vbustamante, line 864: The `PropControlledByParent` check needs to execute before the `propSetByDynamicFunc`, since everything that is controlled by parent is set by dynamic func but the opposite is not true

lib/rebaser-server/src/server/rebase.rs
```diff
@@ -55,7 +55,7 @@ pub(crate) async fn perform_rebase(
         to_rebase_workspace_snapshot_address,
         onto_workspace_snapshot.id().await
     );
-    info!("after snapshot fetch and parse: {:?}", start.elapsed());
+    debug!("after snapshot fetch and parse: {:?}", start.elapsed());
```
vbustamante, line 58: Not very fond of these changes being commited on what could be a frontend only PR, but this is a nit

app/web/src/components/AttributesPanel/AttributesPanelItem.vue
vbustamante: Just noticed this fact means you can simplify `propIsEditable`

vbustamante (changes requested):
The behaviour mostly checks out, separating out the state variable is also a good call. Just a little detail to be fixed

app/web/src/components/AttributesPanel/AttributesPanelItem.vue
britmyerss: perfect, thank you!!

britmyerss:
bors merge

britmyerss:
bors retry

vbustamante (approved):

vbustamante:
bors merge

PR #3694, opened 2024-05-03
Updating encrypted contents triggers dependent values update

nickgerace:
## Description

image

Welcome.  This PR ensures that updating the encrypted contents of a secret results in dependent values update being triggered. Sit down with some coffee, this is a bit of a weird one.

In the frontend, the encrypted contents for secrets are updated centrally, which means that if multiple attribute values "use" that secret, they must be processed in dependent values update. This presents us with a few design challenges unique to the new engine, but none that are/were unsolveable.

The first architectural challenge was to make an implicit relationship (i.e. setting `SecretId` as a string on a prop with a "secret" widget kind) into an explicit one. The solution: add a new value source for `AttributePrototypeArguments`. The new value source indicates how the sole argument should be assembled when executing from prototype func (i.e. `si:identity`). With this change, we now have an explicit relationship linking the attribute subsystem to the secret.

> _A notable result of the aforementioned solution is that the encrypted secret key is now passed through the attribute subsystem rather than the `SecretId` alone. We'll touch back on this when talking about the third architectural challenge._

The second architectural challenge was to allow `SecretIds` to be passed into dependent values update. Part of the long term vision for dependent values update is to be able to take any node on a `WorkspaceSnapshotGraph`, determine what depends on it and execute relevant functions. Working with Jacob Helwig (@jhelwig) and Zachary Hamm (@zacharyhamm) led to the solution in this PR: ensure the dependency graph can take in arbitrary values, but continue using attribute values for the remainder of dependent values update. There is a little blast radius here given the proximity to `ComputeValidations` and job de-duplication code, but the bulk of the changes are mostly self-contained.

The third architectural challenge was to inform the frontend (via the property editor) of which secret had been selected for a given property. The solution: send the `SecretId` along with the encrypted secret key in the attribute subsystem. In the frontend, the list of secrets available purely uses the `SecretIds`, which are stable. We could have used solely the key and provided a way to find the secret via the key. We also could have sent up the entire object when assembling the property editor values. However, the implementation chosen involves scrubbing the encrypted secret key from the property editor values. This reduces friction in the frontend since `app/web` is solely concerned with `SecretIds` and not the encrypted secret keys themselves.

There are a lot more changes in this PR, including moving the `before_func` domain underneath the `secrets` domain, but that's the breakdown of the primary objective and implementation.

## Misc Changes

- Move `before_funcs` domain underneath the `secrets` domain
- Expand `ValueSource` display implementation
- Rename `Visibility::new_head` to `Visibility::new_head_fake` to indicate intent alongside a doc comment
- Allow `DependentValuesUpdate` to take any node as input, though only contents nodes for `Secrets` and `AttributeValues` are evaluated at present


lib/dal/src/attribute/prototype/debug.rs
```diff
@@ -148,6 +152,18 @@ impl AttributePrototypeDebugView {
...
+                        vec![FuncArgDebugView {
+                            value,
+                            name: func_arg_name.clone(),
+                            value_source: value_source.to_string(),
+                            value_source_id: secret_id.into(),
+                            path: None,
+                            socket_source_kind: None,
+                            is_used: arg_used,
+                        }]
+                    }
```
nickgerace, line 166: Show the key in the debug view.

lib/dal/src/attribute/value.rs
```diff
@@ -670,6 +672,11 @@ impl AttributeValue {
                                         .value,
                                 ]
                             }
+                            ValueSource::Secret(secret_id) => {
+                                vec![Secret::get_serialized_key_for_id(ctx, secret_id)
+                                    .await
+                                    .map_err(Box::new)?]
+                            }
```
nickgerace, line 679: When executing from prototype function for direct, dependent `AttributeValues` of the secret, we will use the key and not the id.

lib/dal/src/attribute/value/dependent_value_graph.rs
```diff
@@ -14,77 +15,127 @@ use crate::{
...
+                        // We only want to add the direct dependent attribute values. We also need
+                        // to mark them as exceptions to the "independent values" collection.
+                        ContentAddressDiscriminants::Secret => {
+                            let direct_dependents =
+                                Secret::direct_dependent_attribute_values(ctx, initial_id.into())
+                                    .await
+                                    .map_err(Box::new)?;
+                            self.exceptions_to_independent_values
+                                .extend(direct_dependents.clone());
+                            values.extend(direct_dependents);
```
nickgerace, line 103: Two subtle things here:

1. Unlike the above, we _do not_ push the initial value into the `values` if it is a `SecretId`.
2. We are adding the first set of values as _exceptions_ to the `independent_values()` method. I don't know how I feel about the shape or naming, but this feels like the best way (as of now) to ensure that the DVU job is not affected by these graph generation changes.

lib/dal/src/attribute/value/dependent_value_graph.rs
```diff
@@ -364,7 +410,20 @@ impl DependentValueGraph {
...
+                        .contains(&attribute_value_id)
+                    {
+                        None
+                    } else {
+                        Some(attribute_value_id)
+                    }
+                } else {
+                    None
+                }
+            })
```
nickgerace, line 426: Exceptions in action. As of now, these are only the direct dependents of the secret. They are technically not "independent" despite the graph relationship.

lib/dal/src/func.rs
```diff
@@ -37,15 +37,12 @@ pub mod summary;
 pub mod view;
 
 mod associations;
-mod before;
```
nickgerace, line 40: Move before funcs to the secrets domain so that everything related to secrets is co-located. This PR does this for tests too.

lib/dal/src/job/definition/dependent_values_update.rs
```diff
@@ -134,8 +130,7 @@ impl DependentValuesUpdate {
     ) -> DependentValueUpdateResult<JobCompletionState> {
         let start = tokio::time::Instant::now();
 
-        let mut dependency_graph =
-            DependentValueGraph::for_values(ctx, self.attribute_values.clone()).await?;
+        let mut dependency_graph = DependentValueGraph::new(ctx, self.nodes.clone()).await?;
```
nickgerace, line 133: Only functional change of DVU is how the graph is generated.

lib/dal/src/secret.rs
```diff
@@ -291,6 +326,69 @@ impl Secret {
...
+            .await?
+            .set_value_from_secret_id(ctx, secret_id)
+            .await?;
+
+        // TODO(nick): start here when back. This blocks compilation because we need to de-couple
+        // how the "attribute value jobs" queue works with the ids. They may need to accept ulids.
+        //
+        // ctx.enqueue_dependent_values_update(vec![secret_id]).await?;
+
+        Ok(())
```
nickgerace, line 389: This is the next thing to solve. After that, we should be good.

lib/dal/src/secret.rs
```diff
@@ -291,6 +326,69 @@ impl Secret {
...
+        let func_argument = FuncArgument::find_by_name_for_func(ctx, func_argument_name, func_id)
+            .await?
+            .ok_or(SecretError::FuncArgumentNotFound(
+                func_id,
+                func_argument_name.to_owned(),
+            ))?;
+        AttributePrototypeArgument::new(ctx, attribute_prototype.id(), func_argument.id)
+            .await?
+            .set_value_from_secret_id(ctx, secret_id)
+            .await?;
```
nickgerace, line 382: This is where we make the secret relationship explicit and on the graph.

lib/dal/src/secret.rs
```diff
@@ -291,6 +326,69 @@ impl Secret {
...
+    pub async fn attach_for_attribute_value(
+        ctx: &DalContext,
+        attribute_value_id: AttributeValueId,
+        secret_id: Option<SecretId>,
+    ) -> SecretResult<()> {
+        // First, check if the caller would like to unset the secret.
+        let secret_id = match secret_id {
+            Some(provided_secret_id) => provided_secret_id,
+            None => {
+                AttributeValue::use_default_prototype(ctx, attribute_value_id).await?;
```
nickgerace, line 348: I'm still thinking through whether or not the user would want a scenario where they'd want to override the default "I get my secret from an input socket" prototype, but they would still want it to "unset". We may actually just want to unset the value and keep the prototype.

lib/sdf-server/src/server/service/component/update_property_editor_value.rs
```diff
@@ -78,3 +78,26 @@ pub async fn update_property_editor_value(
...
+            let secret_id: SecretId = serde_json::from_value(value.to_owned())?;
+            Secret::attach_for_attribute_value(ctx, attribute_value_id, Some(secret_id)).await?;
+            return Ok(());
+        }
+    }
+
+    // Regardless of whether this is for a secret, unsetting works the same way.
+    AttributeValue::update(ctx, attribute_value_id, maybe_value).await?;
+    Ok(())
+}
```
nickgerace, line 103: We will also want to align this with how we handle "unsetting" in the attach method.

nickgerace (commented):
~~@zacharyhamm @jhelwig would love some thoughts on some comments I've left here as well as the overall draft PR when you have the chance. Thanks!~~

EDIT: deleted out of date comments and replaced relevant ones with the newer diffs. The PR is now ready for review.

lib/dal/src/attribute/value/dependent_value_graph.rs
```diff
@@ -14,77 +15,131 @@ use crate::{
...
+                        // processed in DVU.
+                        ContentAddressDiscriminants::Secret => {
+                            let direct_dependents =
+                                Secret::direct_dependent_attribute_values(ctx, initial_id.into())
+                                    .await
+                                    .map_err(Box::new)?;
+                            self.values_that_need_to_be_processed
+                                .extend(direct_dependents.clone());
+                            values.extend(direct_dependents);
+                        }
```
nickgerace, line 108: This is where we determine "what attribute values directly dependent on this secret"

lib/dal/src/secret.rs
nickgerace: This is solved

lib/dal/src/attribute/value/dependent_value_graph.rs
nickgerace: This is outdated

lib/dal/src/attribute/value/dependent_value_graph.rs
nickgerace: This is outdated

lib/dal/src/attribute/value/dependent_value_graph.rs
nickgerace: This is outdated

lib/dal/src/attribute/value/dependent_value_graph.rs
```diff
@@ -249,12 +229,86 @@ impl DependentValueGraph {
...
+                        // processed in DVU.
+                        ContentAddressDiscriminants::Secret => {
+                            let direct_dependents =
+                                Secret::direct_dependent_attribute_values(ctx, initial_id.into())
+                                    .await
+                                    .map_err(Box::new)?;
+                            self.values_that_need_to_execute_from_prototype_function
+                                .extend(direct_dependents.clone());
+                            values.extend(direct_dependents);
+                        }
```
nickgerace, line 298: This is where we determine "what attribute values directly dependent on this secret"

lib/dal/src/attribute/prototype/debug.rs
```diff
@@ -148,6 +152,17 @@ impl AttributePrototypeDebugView {
                             ),
                         )?;
                 let values_for_arg = match value_source {
+                    ValueSource::Secret(secret_id) => {
+                        vec![FuncArgDebugView {
+                            value: Secret::payload_for_prototype_debug_view(secret_id)?,
```
nickgerace, line 157: Redact the key during debug view.

lib/dal/src/job/definition/dependent_values_update.rs
```diff
@@ -144,7 +139,9 @@ impl DependentValuesUpdate {
 
         // Remove the first set of independent_values since they should already have had their functions executed
         for value in dependency_graph.independent_values() {
-            dependency_graph.remove_value(value);
+            if !dependency_graph.values_needs_to_execute_from_prototype_function(value) {
+                dependency_graph.remove_value(value);
+            }
```
nickgerace, line 144: Okay, I slightly lied. We need to process attribute values that directly dependent on the secret because the secret itself is not an attribute value.

lib/dal/src/attribute/value.rs
nickgerace: This is outdated

lib/dal/src/attribute/prototype/debug.rs
nickgerace: This is outdated

lib/dal/src/attribute/value.rs
```diff
@@ -687,6 +689,11 @@ impl AttributeValue {
                                         .value,
                                 ]
                             }
+                            ValueSource::Secret(secret_id) => {
+                                vec![Secret::payload_for_prototype_execution(ctx, secret_id)
+                                    .await
+                                    .map_err(Box::new)?]
+                            }
```
nickgerace, line 696: This is where the payload is used for the sole argument to the identity function. This is only used for _direct_ dependent `AttributeValues`.

lib/dal/src/property_editor/values.rs
```diff
@@ -213,11 +214,47 @@ impl PropertyEditorValues {
...
+                PropertyEditorError::PropertyEditorValueNotFoundByPropId(secret_object_prop_id),
+            )?;
+
+            if let Some(property_editor_value) = self.values.get_mut(&attribute_value_id.into()) {
+                if Value::Null != property_editor_value.value {
+                    property_editor_value.value = Secret::payload_for_property_editor_values(
+                        property_editor_value.value.clone(),
+                    )?;
+                }
+            }
```
nickgerace, line 254: Remove the key from the values here. This has negligible performance impact in the real world from looking at round trip times from live routes.

lib/dal/src/secret.rs
```diff
@@ -291,6 +318,66 @@ impl Secret {
...
+    ///   value. If restoring the [`AttributeValue`] to its original state is desired, use
+    ///   [`AttributeValue::use_default_prototype`].
+    ///
+    /// This method will enqueue
+    /// [`DependentValuesUpdate`](crate::job::definition::DependentValuesUpdate).
+    pub async fn attach_for_attribute_value(
+        ctx: &DalContext,
+        attribute_value_id: AttributeValueId,
+        secret_id: Option<SecretId>,
+    ) -> SecretResult<()> {
```
nickgerace, line 336: This is the way to "attach" secrets to a given attribute value.

lib/dal/src/secret.rs
```diff
@@ -291,6 +318,65 @@ impl Secret {
...
+    /// - If no [`Secret`] is provided, then the [`AttributeValue`] will be updated to its original
+    ///   state via [`AttributeValue::use_default_prototype`].
+    ///
+    /// This method will enqueue
+    /// [`DependentValuesUpdate`](crate::job::definition::DependentValuesUpdate).
+    pub async fn attach_for_attribute_value(
+        ctx: &DalContext,
+        attribute_value_id: AttributeValueId,
+        secret_id: Option<SecretId>,
+    ) -> SecretResult<()> {
```
nickgerace, line 335: This is the way to "attach" secrets to a given attribute value.

lib/dal/src/attribute/value/dependent_value_graph.rs
```diff
@@ -51,50 +48,112 @@ impl WorkQueueValue {
...
+                                Secret::direct_dependent_attribute_values(ctx, initial_id.into())
+                                    .await
+                                    .map_err(Box::new)?;
+                            self.values_that_need_to_execute_from_prototype_function
+                                .extend(direct_dependents.clone());
+                            values.extend(
+                                direct_dependents
+                                    .iter()
+                                    .map(|d| WorkQueueValue::Initial(*d)),
+                            );
```
nickgerace, line 125: This is where we find attribute values that _directly_ depend on a secret.

lib/dal/src/secret.rs
```diff
@@ -291,6 +318,65 @@ impl Secret {
...
+        // First, check if the caller would like to unset the secret.
+        let secret_id = match secret_id {
+            Some(provided_secret_id) => provided_secret_id,
+            None => {
+                AttributeValue::use_default_prototype(ctx, attribute_value_id).await?;
```
nickgerace, line 340: We use the default prototype here rather than passing `None` to `AttributeValue::update` because we neither want nor need existing prototype arguments and the user cannot override the prototype for values corresponding to `/root/secrets/<secret>`.

nickgerace:
I've deleted and re-posted comments many times on this PR due to frequent rebases and small design changes, so I may have missed a few spots in the subsequent self-reviews. Thanks for your patience.

nickgerace:
bors merge

nickgerace:
bors merge

nickgerace:
bors merge

nickgerace:
bors merge

PR #3695, opened 2024-05-03
BUG-138 diagram outliner fix recursion

jobelenus:
The problem is that the computed `searchFiltersWithCounts` uses the fn `filterComponentArrayBySearchStringAndFilters` and that fn uses `searchFiltersWithCounts` within it. So I found a way to remove the reference from within the fn.


jobelenus:
bors merge

PR #3699, opened 2024-05-03
Remove invite_user feature flag and default ossrelease to true

britmyerss:
OSS Release is a bit more intertwined, so not taking this on now, but removed the invite_user feature flag and defaulted oss release to true so we can remove the flag from posthog!


wendybujalski (approved):
Looks good! I don't see anywhere else with the INVITE_USER flag so we should be all set.

PR #3700, opened 2024-05-04
feat(rebaser): re-impl server w/activity stream & per-change set queues

fnichol:
This work is a collection of changes which are broken out into commits but summarized here:

## feat(rebaser): re-impl server w/activity stream & per-change set queues

This is a rather large change which re-implements the Rebaser server
code using several new architectural patterns and frameworks.

The Rebaser server now processes a filtered activity stream (provided by
the LayerDb infrastructure) of rebaser requests and for each message
checks to see if there is a corresponding running task for the
associated change set. If not a new "micro-service"/"actor" is spawned
which consumes requests for the given change set from a NATS Jetstream
work-queue in a one-at-a-time fashion. The work-queue can be shared
across multiple running instances of Rebaser servers, meaning that a
group of Rebasers can share the coordinated processing of all messages
while simultaneously allowing work across multiple change sets to run
concurrently.

## feat(naxum): impl Ack middleware which manages acking each message

This change adds a new `Ack` middleware for use in Naxum-based services.
It is used with a incoming NATS Jetstream messages and deals with 3
primary scenarios:

- When the handler function completes successfully (that is a
  `Result::Ok`, a `StatusCode` that is not a server error, etc.), the
  corresponding message will be double acked (that is, acked and then
  awaited for confirmation by the NATS server).
- When the handler function completes unsuccessfully (that is a
  `Result::Err`, a server error `StatusCode`, etc.), the corresponding
  messaged with by negatively acked via a `Nack` message. Depending on
  the Jetstream and consumer semantics this typically triggers the
  message to be redelivered. This default behavior means that in a work
  queue scenario, any `Err` returned by a handler would be considered a
  failure and will be redelivered for a retry.
- When the handler function runs for a long-enough period, importantly
  near to the `ack_wait` period, a concurrent task will send `Pending`
  for the associated message. This way the handler author does not have
  to think about sending progress, cleaning up this task, etc.

## chore(si-data-nats): instrument jetstream::Context

This change adds initial instrumentation around the JetStream `Context`
type using the same strategy as the instrumentation over core NATS APIs.
This increased "loggability" will help us to determine slow paths, the
speed of responses/acks/setting up streams and consumers/etc.

## chore(dal-test): disable OpenTelemetry in integration tests

The OpenTelemetry setup in our integration tests are close to operating,
while still having issues when executing. The largest problem to
overcome is dealing with failing/panicking tests that don't easily
provide a way to the OpenTelemetry collector to flush and clean up
appropriately.

Future work may try to track tests in flight and infer when the last
test is done so as to wait for a clean shutdown of OpenTelemetry, but
this is very difficult considering there is no low level configuration
of the Rust test suite harness. In other words, it'll be something of a
hack in the end...


fnichol:
bors try

fnichol:
bors merge

stack72:
bors retry

PR #3703, opened 2024-05-06
BUG-142 debounce & batch prop edits at 1 second

jobelenus:
We probably don't need this if we can avoid using materialized view which generates conflicts and Adam's "send WsEvent after rebaser commits" is completed.


nickgerace (approved):

jobelenus:
bors try

jobelenus:
bors try

jobelenus:
bors try

jobelenus:
bors try

PR #3705, opened 2024-05-06
Add metrics to pinga jobs

paulocsanz:
Improves telemetry/metrics of `execute_job_task`.

Logs the number of concurrent tasks and limit every time a new job is spawned in pinga. Also logs if pinga is at full capacity, which means no new tasks will be consumed until one is finished.

We also are logging the status of execution, jobs without status are still being executed, so we can easily keep track of what is going on at each instance.

### DIFF
--- a/lib/pinga-server/src/server.rs
+++ b/lib/pinga-server/src/server.rs
@@ -1,4 +1,9 @@
-use std::{future::IntoFuture, io, sync::Arc};
+use std::{
+    future::IntoFuture,
+    io,
+    sync::atomic::{self, AtomicUsize},
+    sync::Arc,
+};
 
 use dal::job::definition::compute_validation::ComputeValidation;
 use dal::{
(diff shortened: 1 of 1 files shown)


paulocsanz:
bors merge

lib/pinga-server/src/server.rs
```diff
@@ -419,6 +438,8 @@ async fn execute_job_task(
     metadata: Arc<ServerMetadata>,
```
britmyerss, line 438: it's the lines write above this, but if we add telemetry to this func, and record the fields you added, we'd be able to do some fancy graphing in honeycomb to show how these numbers change over time!

lib/pinga-server/src/server.rs
```diff
@@ -361,17 +366,26 @@ async fn receive_job_requests(
     Ok(())
 }
 
+const CONCURRENT_TASKS: AtomicUsize = AtomicUsize::new(0);
+
```
britmyerss, line 370: same thing here! Something like 
``` #[instrument (level="info", skip(rx)]```

lib/pinga-server/src/server.rs
paulocsanz: Sorry, not sure I understand

lib/pinga-server/src/server.rs
paulocsanz: Why would we want a span for the task that spawns task?

paulocsanz: Something like this?

image

britmyerss: yep!!

lib/pinga-server/src/server.rs
britmyerss: oh sorry - this IS in the instrument tag. ignore me!

nickgerace (approved):
:rocket:

lib/pinga-server/src/server.rs
paulocsanz: Done!

paulocsanz:
bors merge

paulocsanz:
bors merge

paulocsanz:
bors merge

PR #3707, opened 2024-05-06
bug(dal): send nats events after rebase finishes

adamhjk:
We should only send WsEvents after we rebase. This PR fixes it, without needing to implement passhtrough of WsEvents to the rebaser (yet).


adamhjk:
bors merge

fnichol (approved):
Yeah, wow, such a small change that it feels like that must be it!

Also, the fix is very reasonable

PR #3708, opened 2024-05-06
feat(dal): support children of "dynamic" objects as inputs to other functions

zacharyhamm:
This makes it possible to take the children of an object set by a custom function as the input to another function. It ensures we do not place the child values directly onto the dependent value graph, but instead we place the "controlling" attribute value onto the graph if one is discovered (we will resolve the input to the correct child value in function execution).

Many of our assets depend on this, since they take children of the resource_value tree as inputs to other functions.

Closes BUG-135


zacharyhamm:
bors try

zacharyhamm:
bors try

PR #3710, opened 2024-05-06
Wendy/eng 2530 new 2xs text size for 10px current 2xs text size should be

wendybujalski:
- Adjusted text sizes - `2xs` is now `0.6rem` and `3xs` is `0.5rem`
- Used new text size to make the SiChip on FuncList smaller
- Cleaned up and compactified AssetFuncAttachDropdown and NewFuncDropdown
- New `compact` style for DropdownMenu for menus that should have less padding
- New `FontSizes` export from `size_utils.ts` to use when passing a text size as a prop
- Fixes to prevent SidebarSubpanelTitles from changing size based on contents
- Removed redundant unused Icons from `icon_set.ts`


wendybujalski:
bors merge

PR #3713, opened 2024-05-07
chore(dal): do less work when walking child values in dvu

zacharyhamm:
It's only necessary to add an object's direct children to the workqueue (the next level down will be handled in the next iteration of the queue)


nickgerace (approved):
Great optimization!

PR #3715, opened 2024-05-07
feat(sdf): Adding WsEvents when props change for a component

stack72:
This fixes reactivity for changing the input type for a prop, colour changes to the component, adding / updating or removing array and map elements


stack72:
bors merge

PR #3717, opened 2024-05-07
Implement actions v2 job definition

paulocsanz:


paulocsanz:
bors try

paulocsanz:
bors try

paulocsanz:
bors try

lib/dal/src/action.rs
```diff
@@ -231,4 +304,74 @@ impl Action {
...
+            .await?
+            .ok_or(ActionError::ComponentNotFoundForAction(id))?;
+
+        let prototype_id = Action::prototype_id(ctx, id)
+            .await?
+            .ok_or(ActionError::PrototypeNotFoundForAction(id))?;
+        let prototype = ActionPrototype::get_by_id(ctx, prototype_id).await?;
+
+        Action::set_state(ctx, id, ActionState::Running).await?;
+
```
britmyerss, line 329: I think we want to commit here so that this status is rebased on head

paulocsanz: That's a good point, I may move this to the job definition to make it less confusing, as commiting inside of the regular dal methods feels weird.

Good catch!

lib/dal/src/action.rs
```diff
@@ -184,6 +232,31 @@ impl Action {
         Ok(result)
     }
 
+    pub async fn prototype_id(
+        ctx: &DalContext,
+        action_id: ActionId,
+    ) -> ActionResult<Option<ActionPrototypeId>> {
```
jhelwig, line 238: Why an `Option`? If there's an `Action` without an `ActionPrototype` that seems like things are very broken, and not really recoverable, or we're in the middle of creating the `Action` where we should already know which `ActionPrototype` we'd be setting things up with.

lib/dal/src/action.rs
jhelwig: Yup, definitely want to commit in this phase of running, and completely agree with Paulo that this probably belongs in the job definition.

lib/dal/src/action.rs
paulocsanz: That makes sense will change.

Btw why is component optional for the action?

jhelwig: Possibly premature future planning: I can envision Actions that are scoped to the entire workspace (having `ActionPrototype` linked to the workspace itself, in addition to the ones linked to the `SchemaVariant`s).

paulocsanz:
bors try

paulocsanz:
bors try

paulocsanz:
bors try

paulocsanz:
bors cancel

paulocsanz:
bors try

paulocsanz:
bors merge

PR #3721, opened 2024-05-07
Fix: don't create pending components on head

jobelenus:
WsEvents are firing *BEFORE* the POST returns and when a new changeset is created by the backend, we end up re-creating componentStore, so the store for HEAD never runs onSuccess below and we end up with pending components on HEAD that never go away

To fix it; don't create pending components if we're on HEAD


jobelenus:
bors merge

PR #3726, opened 2024-05-07
Upgrade: when deleting a frame, pop confirmation dialog with all fram…

jobelenus:
…e children, delete all children with the frame


jobelenus:
bors merge

PR #3727, opened 2024-05-07
feat(sdf, dal, web): Send a WsEvent after a component connected request

stack72:
This allows us to reload the diagram correctly


stack72:
bors merge

PR #3733, opened 2024-05-08
feat: add graceful shutdown to pool_noodle, cleanup logs lines

sprutton1:


sprutton1:
bors try

lib/deadpool-cyclone/examples/deadpool-cyclone-ping-pool/main.rs
```diff
@@ -33,7 +34,9 @@ async fn main() -> Result<(), Box<(dyn std::error::Error + 'static)>> {
     };
 
     let spec = spec()?;
-    let mut pool: PoolNoodle<LocalUdsInstance, _> = PoolNoodle::new(10, spec.clone());
+    let (shutdown_broadcast_tx, _) = broadcast::channel(16);
```
paulocsanz, line 37: `tokio_util::sync::CancellationToken` may be a better fir for the shutdown channel, but it's not a blocker

sprutton1: I used broadcast as that's what Veritech already uses for its graceful shutdown bits, so it was easiest to just leverage what was already there

PR #3734, opened 2024-05-08
fix(layer-cache): Prevent jetstream consumers from exceeding their inactivity threshold

stack72:
The default timeout was 5 seconds so we want to prevent any slowness of instances causing a disconnect


stack72:
bors merge

PR #3735, opened 2024-05-08
Fix: batch connect components since frame edges are exclusive and concurrent rebases conflict

jobelenus:


jobelenus:
bors merge

PR #3737, opened 2024-05-08
Frame fixes, and support up frames with many children

britmyerss:
Fixes include: 
- make edge ids unique in the web so they don't disappear 
- remove the Frame Contains edge before updating the attribute value for the impacted input socket

Lastly, we (once again) support Up Frames inferring their inputs from multiple children, if and only if each child with matching output sockets only has one match. 

<div><br/>via <a href="https://giphy.com/election2020/">Election 2020</a> on <a href="https://giphy.com/gifs/2020-iowa-democratic-party-liberty-and-justice-celebration-IgdrPBTDxRBi0tK6FH">GIPHY</a></div>

### DIFF
--- a/app/web/src/store/components.store.ts
+++ b/app/web/src/store/components.store.ts
@@ -233,7 +233,7 @@ const edgeFromRawEdge =
   (isInferred: boolean) =>
   (e: RawEdge): Edge => {
     const edge = structuredClone(e) as Edge;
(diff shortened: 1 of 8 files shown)


stack72:
bors try

britmyerss:
bors try

britmyerss:
bors cancel

PR #3739, opened 2024-05-09
Fix secret attribute value containing an object instead of a string

nickgerace:
## Description

This PR is a follow-up to #3694 and fixes the attribute values corresponding to individual secrets containing an object instead of a string.

The encrypted secret key is now passed through the attribute subsystem alone. Since the property editor still needs to reference which encrypted secret key belongs to which secret, we determine the origin secret dynamically when assembling the property editor (~3.8ms wall clock time for 1000 secrets with a 3970X system on `Pop!_OS 22.04 LTS` with Linux kernel `6.8.0`).

As a result of these changes, we were able to remove intermediary types with a lot of footguns as well as move the encrypted secret key to the new `SecretNodeWeight` kind. Overall performance and consistency in results should be net positive.

## Other Changes

Another change this PR brings is the inclusion of our first integration test... "bench". We want to ensure that looking up which secret an encrypted secret key is derived from is fast, so the new test will fail if the function takes too long. We may want to use this logic to create tests where similar performance scrutiny is needed.

Finally, we are now using the `create_xxhash_type` macro to ensure that the `EncryptedSecretKey` works with `si-layer-cache`.

## Q&A

Let's talk!

### Why now? Doesn't this already work from #3694?

Passing around object types as strings has potential to break or result in dropped or mangled values in dependent values update. Not only that, but the footguns working with transitory structs are greater than just passing the key as a string downstream and then deriving the secrets from the keys since there is a one-to-one relationship between each secret and key.

### Are you worried about performance?

Secrets collections have the potential to be multitudinous in nature. This PR adds a performance-baseline "bench" for ensuring that finding the secret for the corresponding key happens in an acceptable timeframe.

### Does this break existing users?

If all `Secrets` are deleted in all `ChangeSets` for all `Workspaces`, then re-using a database from a deployment of latest `main` should work. However, we do not yet guarantee backwards compatibility with recent `main`, though that will likely change shortly.

### DIFF
--- a/lib/dal/src/attribute/prototype/debug.rs
+++ b/lib/dal/src/attribute/prototype/debug.rs
@@ -22,7 +22,7 @@ use crate::workspace_snapshot::node_weight::NodeWeightError;
 use crate::workspace_snapshot::WorkspaceSnapshotError;
 use crate::{
     AttributePrototype, AttributePrototypeId, AttributeValue, AttributeValueId, Component,
-    ComponentError, DalContext, Func, FuncError, FuncId, Secret, SecretError,
+    ComponentError, DalContext, Func, FuncError, FuncId, SecretError,
 };
 
 type AttributePrototypeDebugViewResult<T> = Result<T, AttributePrototypeDebugViewError>;
@@ -54,6 +54,8 @@ pub enum AttributePrototypeDebugViewError {
     PropError(#[from] PropError),
     #[error("secret error: {0}")]
     SecretError(#[from] SecretError),
+    #[error("serde json error: {0}")]
+    SerdeJsonError(#[from] serde_json::Error),
     #[error("value source error: {0}")]
     ValueSourceError(#[from] ValueSourceError),
     #[error("workspace snapshot error: {0}")]
@@ -154,7 +156,9 @@ impl AttributePrototypeDebugView {
                 let values_for_arg = match value_source {
                     ValueSource::Secret(secret_id) => {
                         vec![FuncArgDebugView {
-                            value: Secret::payload_for_prototype_debug_view(secret_id)?,
+                            value: serde_json::to_value(format!(
+                                "[REDACTED KEY FOR SECRET (ID: {secret_id})]"
+                            ))?,
                             name: func_arg_name.clone(),
                             value_source: value_source.to_string(),
                             value_source_id: secret_id.into(),
--- a/lib/dal/src/secret.rs
+++ b/lib/dal/src/secret.rs
@@ -25,6 +25,7 @@
     while_true
 )]
 
+use chrono::Utc;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
 use si_crypto::{SymmetricCryptoError, SymmetricCryptoService, SymmetricNonce};
@@ -36,6 +37,7 @@ use sodiumoxide::crypto::box_::{PublicKey, SecretKey};
 use sodiumoxide::crypto::sealedbox;
 use std::collections::HashMap;
 use std::fmt;
+use std::str::FromStr;
 use std::sync::Arc;
 use telemetry::prelude::*;
 use thiserror::Error;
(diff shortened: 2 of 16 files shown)


lib/dal/src/attribute/prototype/debug.rs
```diff
@@ -154,7 +156,9 @@ impl AttributePrototypeDebugView {
                 let values_for_arg = match value_source {
                     ValueSource::Secret(secret_id) => {
                         vec![FuncArgDebugView {
-                            value: Secret::payload_for_prototype_debug_view(secret_id)?,
+                            value: serde_json::to_value(format!(
+                                "[REDACTED KEY FOR SECRET (ID: {secret_id})]"
```
jhelwig, line 160: Where all is the `AttributePrototypeDebugView` used? Putting on my Security Engineer Hat makes me want to say that this should be `[REDACTED]` without the `secret_id` so it can't expose that the same value is used in however many places (and exactly what those places are).

nickgerace: It is used in the frontend to debug what values are used. I believe knowing the `SecretId` is fine because it refers to the `Secrets` available in the `Secret` modal and picker in the frontend.

nickgerace: However, I am willing to drop this too. It has been mostly used for development purposes.

jhelwig: Yeah, given that we expose which secret is used (not the value itself, but which secret), and exactly where each is used as part of the normal UI, it should be fine.

:removes tin-foil hat:

nickgerace:
bors merge

lib/dal/src/secret.rs
```diff
@@ -557,29 +601,65 @@ impl Secret {
...
         lambda(&mut secret)?;
+        if before != SecretContentV1::from(secret.clone()) {
+            match ctx.history_actor() {
+                HistoryActor::SystemInit => {}
+                HistoryActor::User(id) => {
+                    secret.updated_by = Some(*id);
+                }
+            }
+            secret.timestamp.updated_at = Utc::now();
+        }
```
nickgerace, line 639: Now, we update the history actor and "updated_at" timestamp every time that something has changed. This is a bonus benefit of the PR.

lib/dal/src/secret.rs
```diff
@@ -593,34 +673,15 @@ impl Secret {
...
-}
-
-// NOTE(nick): keep this and its fields private! It should only be used by methods within the Secret
-// impl. I will be sad if it neither remains private nor remains used in a limited fashion.
-#[derive(Debug, Serialize, Deserialize)]
-struct PayloadForPrototypeDebugView {
-    secret_id: SecretId,
-    encrypted_secret_key: String,
-}
-
```
nickgerace, line 623: Goodbye footguns!

lib/dal/src/workspace_snapshot/node_weight/secret_node_weight.rs
```diff
@@ -0,0 +1,198 @@
...
+#[derive(Clone, Serialize, Deserialize)]
+pub struct SecretNodeWeight {
+    id: Ulid,
+    lineage_id: LineageId,
+    content_address: ContentAddress,
+    merkle_tree_hash: MerkleTreeHash,
+    vector_clock_first_seen: VectorClock,
+    vector_clock_recently_seen: VectorClock,
+    vector_clock_write: VectorClock,
+    encrypted_secret_key: EncryptedSecretKey,
```
nickgerace, line 27: The purpose of the node weight is to put the key onto the node. We will need to have quick access to all secrets by key, so this ensures we can remain entirely in memory.

lib/dal/tests/integration_test/secret/bench.rs
```diff
@@ -0,0 +1,50 @@
...
+
+    // Now that we have a graph with many secrets, let's run the function and cache the result.
+    let list_ids_by_key_instant = tokio::time::Instant::now();
+    let _map = Secret::list_ids_by_key(ctx)
+        .await
+        .expect("could not list ids by key");
+    let list_ids_by_key_instant_elapsed = list_ids_by_key_instant.elapsed();
+
+    // Ensure that the result meets our expectations for wall clock time.
+    assert!(Duration::from_millis(10) > list_ids_by_key_instant_elapsed);
```
nickgerace, line 45: Performance upper bound for listing all secrets by key. This is using 1000 secrets. The method takes ~3.8ms on the machine described in the PR description.

lib/si-events-rs/src/encrypted_secret.rs
```diff
@@ -1,3 +1,3 @@
-/// This key uses a [`blake3::Hash`] wrapper, [`si_hash::Hash`], in order to get de/ser and display
-/// implementation benefits.
-pub type EncryptedSecretKey = si_hash::Hash;
+use crate::create_xxhash_type;
+
+create_xxhash_type!(EncryptedSecretKey);
```
nickgerace, line 3: Align with other hash types.

PR #3744, opened 2024-05-09
Dispatch Actions V2 after successful rebase of default change set for workspace

jhelwig:
Whenever we have successfully rebased the default change set for a workspace, we need to dispatch any eligible Actions (and mark them as dispatched). By checking which Actions are eligible to dispatch on successful rebase, we ensure that any changes to the default change set (typically `HEAD`, or `main`) will automatically cause relevant Actions to run, such as an Action being removed from the graph because it succeeded, or new Actions being queued from merging a change set into the default change set.

### DIFF
--- a/lib/rebaser-server/src/change_set_requests/handlers.rs
+++ b/lib/rebaser-server/src/change_set_requests/handlers.rs
@@ -2,14 +2,20 @@
 
 use std::result;
 
-use dal::{Tenancy, Visibility, WorkspacePk, WsEvent};
+use dal::{
+    Tenancy, Visibility, Workspace, WorkspaceError, WorkspacePk, WorkspaceSnapshot,
+    WorkspaceSnapshotError, WsEvent,
+};
 use naxum::{
     extract::State,
     response::{IntoResponse, Response},
 };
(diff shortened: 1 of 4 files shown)


lib/rebaser-server/src/change_set_requests/handlers.rs
```diff
@@ -67,6 +79,20 @@ pub async fn process_request(State(state): State<AppState>, msg: InnerMessage) -
             }
         });
 
+    // Dispatch eligible actions if the change set is the default for the workspace.
+    // Actions are **ONLY** ever dispatched from the default change set for a workspace.
+    if RebaseStatusDiscriminants::Success == rebase_status.clone().into() {
+        if let Some(workspace_pk) = ctx.tenancy().workspace_pk() {
+            if let Some(workspace) = Workspace::get_by_pk(&ctx, &workspace_pk).await? {
+                if let Some(change_set) = ctx.change_set().ok() {
```
jhelwig, line 87: My kingdom for stable if-let chains.

jhelwig:
bors try

jhelwig:
bors merge

PR #3746, opened 2024-05-09
feat(dal,rebaser-server): rebaser debounces dvu jobs

zacharyhamm:
Send the set of values we intend to run DependentValuesUpdate on along with the rebase request if a non-blocking commit is requested. If values appear in the rebase request, send them to a task which will dispatch DependentValuesUpdate jobs in 5 second intervals, blocking until they finish, so that they run serially for a change set.

Note, DalContext `commit` will no longer send DependentValuesUpdate jobs to pinga (no other job types are affected). Instead it sends them to the rebaser and the rebaser will send them to pinga. `blocking_commit` still processes the job queue the same way as before, since we expect this behavior in tests. Our tests should be restructured to not require the blocking commits.

### DIFF
--- a/lib/dal/src/job/queue.rs
+++ b/lib/dal/src/job/queue.rs
@@ -1,4 +1,7 @@
-use std::{collections::HashMap, collections::HashSet, collections::VecDeque, sync::Arc};
+use std::{
+    collections::{hash_map::Entry, HashMap, HashSet, VecDeque},
+    sync::Arc,
+};
 use tokio::sync::Mutex;
 use ulid::Ulid;
 
@@ -11,7 +14,7 @@ type AttributeValueBasedJobs = Arc<
     Mutex<
         HashMap<
             AttributeValueBasedJobIdentifier,
-            HashMap<(ChangeSetId, AccessBuilder), HashSet<Ulid>>,
+            HashMap<ChangeSetId, (HashSet<Ulid>, AccessBuilder)>,
         >,
     >,
 >;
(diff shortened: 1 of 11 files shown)


lib/dal/src/job/queue.rs
```diff
@@ -11,7 +14,7 @@ type AttributeValueBasedJobs = Arc<
     Mutex<
         HashMap<
             AttributeValueBasedJobIdentifier,
-            HashMap<(ChangeSetId, AccessBuilder), HashSet<Ulid>>,
+            HashMap<ChangeSetId, (HashSet<Ulid>, AccessBuilder)>,
```
nickgerace, line 17: That makes more sense

lib/rebaser-server/src/change_set_requests/handlers.rs
```diff
@@ -80,6 +80,13 @@ pub async fn process_request(State(state): State<AppState>, msg: InnerMessage) -
         .await
         .map_err(HandlerError::SendRebaseFinished)?;
 
+    // only enqueue values if rebase succeeded. if it failed, there's no work to do
+    if let RebaseStatus::Success { .. } = rebase_status {
+        if let Some(values) = message.payload.dvu_values {
+            state.dvu_debouncer.enqueue_values(values);
+        }
+    }
+
```
nickgerace, line 89: Seemed scary upon first read, but it makes sense!

lib/rebaser-server/src/dvu_debouncer.rs
```diff
@@ -0,0 +1,189 @@
...
+            workspace_id,
+            change_set_id,
+            ctx_builder,
+        };
+
+        let debouncer_clone = debouncer.clone();
+        let debouncer_clone_2 = debouncer.clone();
+
+        tokio::task::spawn(async { dvu_task(debouncer_clone, dvu_rx).await });
+        tokio::task::spawn(async { debouncer_task(debouncer_clone_2, value_rx).await });
```
nickgerace, line 87: image

nickgerace (commented):
This is awesome!!

PR #3750, opened 2024-05-10
Update secret prop type to fix copy/paste

paulocsanz:


sprutton1:
bors try

stack72:
This won't fix the bug we are seeing as this is at asset creation time

paulocsanz:
@stack72 asset update could be used to fix the issue, but yeah, this is important so new workspaces are not broken

paulocsanz:
bors merge

PR #3751, opened 2024-05-10
feat: paralleize pool_noodle work

sprutton1:
Let's parallelize the pool_noodle work. I was hesitant to do this originally due to experiences in earlier versions of the firecracker scripts that lead to bad times. I tested this pretty thoroughly and didn't come across any issues other than a warm CPU.

The test was a region with 120 ec2 instances in it. I change the region and start the clock. Here are the results:

<google-sheets-html-origin>
pool size | workers | time
-- | -- | --
100 | 10 | 1:24
100 | 5 | 1:52
10 | 5 | 1:36
10 | 10 | 1:34

I'm setting this to `10` for now. At some point I'd like to make this configurable, but I want to get this out there and see how it goes.

### DIFF
--- a/lib/si-pool-noodle/src/pool_noodle.rs
+++ b/lib/si-pool-noodle/src/pool_noodle.rs
@@ -112,11 +112,13 @@ where
 
         let _ = tokio::spawn(Self::handle_shutdown(me.clone(), stop.clone()));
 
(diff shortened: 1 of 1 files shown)


sprutton1:
bors merge

PR #3754, opened 2024-05-10
fix: layer cache write order

sprutton1:
Not awaiting the write layer task means they could happen out of order, which is bad. We also want to log errors if the persister fails to persist.


stack72:
bors retry

PR #3760, opened 2024-05-10
feat(dal): only run update_from_prototype_function in DVU

zacharyhamm:
To reduce more chances of conflict, remove all calls to update_from_prototype_function for dependent functions, restricting their use to the DependentValuesUpdate job (except in the case of the function execution call on the customize screen, which we have to decide how best to handle).


nickgerace (commented):
Good idea! Will think on how to best handle execution... I think I have some ideas.

PR #3762, opened 2024-05-10
Variant needs to include componentType, so when we place the first non-frame inside a frame, we know not to reposition it

jobelenus:


jobelenus:
bors merge

PR #3764, opened 2024-05-10
feat(dal,sdf): Close and Put On hold actions v2 endpoints

vbustamante:
- Close actions by deleting them from the graph
- put actions on hold by updating the status
- send all actions to frontend instead of just enqueued ones
- fix some issues in the actions_v2 tests

### DIFF
--- a/lib/sdf-server/src/server/service/action/cancel.rs
+++ b/lib/sdf-server/src/server/service/action/cancel.rs
@@ -1,0 +1,37 @@
+use axum::extract::Query;
+use dal::action::{Action, ActionState};
+use dal::{ActionId, Visibility};
+use serde::{Deserialize, Serialize};
+
+use super::ActionResult;
+use crate::server::extract::{AccessBuilder, HandlerContext};
+use crate::service::action::ActionError;
+
+#[derive(Deserialize, Serialize, Debug)]
+#[serde(rename_all = "camelCase")]
+pub struct PutOnHoldRequest {
+    id: ActionId,
+    #[serde(flatten)]
+    pub visibility: Visibility,
+}
+
+pub async fn cancel(
+    HandlerContext(builder): HandlerContext,
+    AccessBuilder(request_ctx): AccessBuilder,
+    Query(request): Query<PutOnHoldRequest>,
+) -> ActionResult<()> {
+    let ctx = builder.build(request_ctx.build(request.visibility)).await?;
+
+    let action = Action::get_by_id(&ctx, request.id).await?;
+
+    match action.state() {
+        ActionState::Running | ActionState::Dispatched => {
+            return Err(ActionError::InvalidActionCancellation(request.id))
+        }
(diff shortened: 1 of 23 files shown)


lib/dal/src/schema/variant.rs
```diff
@@ -871,10 +871,11 @@ impl SchemaVariant {
...
         for prototype_idx in action_prototype_node_idxs {
-            let weight = workspace_snapshot
-                .get_node_weight(prototype_idx)
-                .await?
-                .get_action_prototype_node_weight()?;
+            let NodeWeight::ActionPrototype(weight) =
+                workspace_snapshot.get_node_weight(prototype_idx).await?
+            else {
+                continue;
+            };
```
vbustamante, line 878: This was breaking component creation when they had both new and old creation actions

lib/sdf-server/src/server/service/action/cancel.rs
```diff
@@ -0,0 +1,37 @@
...
+    let ctx = builder.build(request_ctx.build(request.visibility)).await?;
+
+    let action = Action::get_by_id(&ctx, request.id).await?;
+
+    match action.state() {
+        ActionState::Running | ActionState::Dispatched => {
+            return Err(ActionError::InvalidActionCancellation(request.id))
+        }
+        ActionState::Failed | ActionState::OnHold | ActionState::Queued => {}
+    }
```
vbustamante, line 32: Would like @jhelwig's opinion on this one, i think this makes sense

lib/sdf-server/src/server/service/action/put_on_hold.rs
```diff
@@ -0,0 +1,38 @@
...
+
+    let action = Action::get_by_id(&ctx, request.id).await?;
+
+    match action.state() {
+        ActionState::Running
+        | ActionState::Dispatched
+        | ActionState::Failed
+        | ActionState::OnHold => return Err(ActionError::InvalidOnHoldTransition(request.id)),
+        ActionState::Queued => {}
+    }
```
vbustamante, line 33: Likewise for these ones - are these conditions correct?

lib/sdf-server/src/server/service/action/cancel.rs
jhelwig: Asking about only being able to cancel Actions that are Failed, On Hold, or Queued? (Cannot do it for Dispatched, or Running.) Yeah, that's correct.

vbustamante: Yup, thanks!

lib/sdf-server/src/server/service/action/put_on_hold.rs
jhelwig: I think putting a Failed action on hold makes sense. It would effectively be re-queueing it immediately into the On Hold state.

paulocsanz:
bors try

paulocsanz:
bors try

PR #3765, opened 2024-05-13
Up Frames can connect to multiple child components and fix data propagation on attach/detach/delete

britmyerss:
Frames are a bit messy, but functional! I will be refactoring this into modules in another pass this week, but I want to get this out there for now. 

This PR includes the following changes: 

- Up Frames can now take inputs from multiple children (for example, when turning a Butane Component into an Up Frame and dropping multiple Docker Images inside) 
- When you attach a component to a parent, we will now run queue up only those input socket's Attribute Values for Dependent Values Update that have different inputs as a result of the attachment (this fixes a bug where we were not always re-running the right functions, which was the original bug)
- When you orphan a component, we first cache the current impacted input sockets so we can rerun them after we remove the FrameContains edge (another bug, as we were rerunning funcs THEN removing the edge, womp womp)
- Note: This takes significant advantage of the latest change to Dependent Values Update as any Input Socket's Attribute Value that gets queued for DVU will probably (always?) have a dynamic function that sets the value, meaning it will run during the next DVU job
- When we remove the Component on delete (because it doesn't have a resource), we also orphan it, making sure that any impacted input sockets are queued up for DVU
- Add a complex test for multiple components as children of an Up Frame, with nesting, moves, and deletes, making sure the data flows as expected
- Added some test helpers to make it faster to write tests for frames as we continue to cover the various permutations. 
- Add some tracing for the rebaser loop so we can see the count of updates and conflicts better in Jaeger and Honeycomb. I needed this to debug an issue while I was working on this and left it in as this will be valuable for us in the short term (and more valuable as we enrich the metadata we return)

<div><br/>via <a href="https://giphy.com/election2020/">Election 2020</a> on <a href="https://giphy.com/gifs/2020-iowa-democratic-party-liberty-and-justice-celebration-IgdrPBTDxRBi0tK6FH">GIPHY</a></div>

### DIFF
--- a/lib/dal-test/src/helpers.rs
+++ b/lib/dal-test/src/helpers.rs
@@ -8,8 +8,8 @@ use dal::property_editor::schema::{
 use dal::property_editor::values::{PropertyEditorValue, PropertyEditorValues};
 use dal::property_editor::{PropertyEditorPropId, PropertyEditorValueId};
 use dal::{
-    Component, ComponentId, DalContext, InputSocket, KeyPair, OutputSocket, Prop, Schema,
-    SchemaVariant, SchemaVariantId, User, UserClaim, UserPk,
+    AttributeValue, Component, ComponentId, DalContext, InputSocket, KeyPair, OutputSocket, Prop,
+    Schema, SchemaVariant, SchemaVariantId, User, UserClaim, UserPk,
 };
 use itertools::enumerate;
 use jwt_simple::algorithms::RSAKeyPairLike;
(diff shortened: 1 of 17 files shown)


jhelwig (approved):

britmyerss:
bors try

lib/dal-test/src/helpers.rs
```diff
@@ -158,6 +158,85 @@ pub async fn connect_components_with_socket_names(
...
+        .get(&input_socket.id())
+        .expect("got input socket av");
+    let input_socket_av = AttributeValue::get_by_id(ctx, input_socket_match.attribute_value_id)
+        .await
+        .expect("got input av");
+    input_socket_av.view(ctx).await.expect("got view")
+}
+
+/// Gets the [`Value`] for a specific [`Component`]'s [`OutputSocket`] by the [`OutputSocket`] name
+pub async fn get_component_output_socket_value(
```
nickgerace, line 191: Great helper ideas

lib/dal/src/component.rs
```diff
@@ -2256,102 +2316,249 @@ impl Component {
         }
         Ok(connections)
     }
+    /// Finds all inferred outgoing connections for the [`Component`]
+    /// A connection is inferred if it's output sockets is driving
+    /// another component's [`InputSocket`] as a result of lineage via
+    /// the FrameContains edge.
+    #[instrument(level = "debug", skip(ctx))]
+    pub async fn inferred_outgoing_connections(
```
nickgerace, line 2324: Roger that

lib/dal/src/context.rs
```diff
@@ -1162,7 +1177,17 @@ async fn rebase(
...
+        "si.change_set.id",
+        rebase_request.to_rebase_change_set_id.to_string(),
+    );
+    span.record(
+        "si.workspace.pk",
+        tenancy
+            .workspace_pk()
+            .unwrap_or(WorkspacePk::NONE)
+            .to_string(),
+    );
```
nickgerace, line 1190: Oh this will be helpful!

Also... optional tenant id probably needs to get removed. Unrelated to this PR, just noticed it on line 1188.

nickgerace (commented):
This is awesome!! Great work @britmyerss

PR #3766, opened 2024-05-13
Fix attaching two auth or action funcs to the same variant

nickgerace:
## Description

This commit contains two fixes that present themselves as the same issue: attaching auth or action funcs to the same variant silently fails. The individual fixes are related in logic changes, but are different overall. Essentially, we were not handling the "existing vs. removed" `SchemaVariantId` logic for the associations properly. Now, we ensure that existing associations are not accidentally dropped.


nickgerace:
bors merge

PR #3769, opened 2024-05-13
feat(dal, web): Add the ability to upgrade a component via the web

stack72:


lib/dal/src/component.rs
```diff
@@ -2805,17 +2867,143 @@ impl Component {
...
+                new_component.id(),
+            ));
+        }
 
-        // now lets delete the edges from this component
-        // Find all of the attributes for the component
-        //
+        new_component.clone_attributes_from(ctx, self.id()).await?;
+        new_component
+            .set_name(ctx, &original_component_name.clone())
```
zacharyhamm, line 2888: name and type are part of the attribute value tree. Does it work without them?

stack72:
bors merge

PR #3770, opened 2024-05-13
chore(rebaser): enable release mode for rebaser in debug build

adamhjk:
When using `buck2 run //dev:up-debug`, the rebaser is now run in release mode regardless. If you really want everything in debug mode, there is another target `buck2 run //dev:up-debug-all`.

This makes migrations take much longer, but recompile times are cut by roughly 1/3 for SDF.


adamhjk:
bors merge

PR #3773, opened 2024-05-14
Fix the (not so) Mysterious Up Frame Bug

britmyerss:
When we execute an attribute's function, we gather up any inferred inputs by walking the graph and finding inferred connections.  

As input sockets share prototypes across components, if there are multiple instances of the component on the graph, where one has an explicit edge drawn and one is supposed to take an inferred input, the existence of prototype arguments as a result of the edge being drawn meant we were never getting inferred inputs for the second component.  

This fixes the logic, such that we look to find inferred inputs only after we haven't found any explicit inputs for that attribute value. 

<div><br/>via <a href="https://giphy.com/gifs/fx-charlie-always-sunny-l0IykOsxLECVejOzm">GIPHY</a></div>

### DIFF
--- a/lib/dal/src/attribute/value.rs
+++ b/lib/dal/src/attribute/value.rs
@@ -634,95 +634,96 @@ impl AttributeValue {
         // Gather the raw func bindings args into a map.
         let mut func_binding_args: HashMap<String, Vec<Value>> = HashMap::new();
         let apa_ids = AttributePrototypeArgument::list_ids_for_prototype(ctx, prototype_id).await?;
-        if apa_ids.is_empty() {
+        for apa_id in apa_ids {
+            let apa = AttributePrototypeArgument::get_by_id(ctx, apa_id).await?;
+            let expected_source_component_id = apa
+                .targets()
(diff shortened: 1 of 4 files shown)


lib/dal/src/attribute/value.rs
```diff
@@ -634,95 +634,96 @@ impl AttributeValue {
...
+                    .and_modify(|values| values.extend(values_for_arg.clone()))
+                    .or_insert(values_for_arg);
+            }
+        }
+        // if we haven't found func binding args by now, let's see if there are any inferred inputs
+        // Note: a given input only takes args through inferences (i.e. frames) if it doesn't have any
+        // explicitly configured args
+
+        if (func_binding_args.is_empty()) {
+            let inferred_inputs = Self::get_inferred_input_values(ctx, attribute_value_id).await?;
```
britmyerss, line 722: we used to only get inferred input values if we didn't find any attribute prototype arguments for the given prototype. Now, we will search for inferred input values if we don't find any func binding args for this actual attribute value id

britmyerss:
bors cancel

PR #3775, opened 2024-05-14
Component position & parentage need to be sent together for rebaser. Send WsEvents to populate components & edges appropriately.

jobelenus:
OK—the front end now has all the info required to display sockets correctly through attach & detach. And the conflicts that happen when position and parentage are sent in separate rebases is avoided. This is the way.


jobelenus:
bors merge

PR #3776, opened 2024-05-14
Component updated called when to_delete=True. Component deleted already called

jobelenus:
Undoing a boo-boo


jobelenus:
bors merge

PR #3780, opened 2024-05-14
Fix detach for attribute and code gen

nickgerace:
## Description

This commit fixes detaching for attribute and code generation funcs.

For attribute funcs, we needed fallback to removing the prototype at the schema variant level if an attribute prototype could not be found at the component level (i.e. the prototype did not correspond to an attribute value). There is now an integration test tracking this behavior.

For code generation funcs, we needed to ensure that we were using the detach ref. No new integration tests were added for this behavior because the bug was purely a frontend usage bug.

### DIFF
--- a/app/web/src/components/FuncEditor/FuncDetails.vue
+++ b/app/web/src/components/FuncEditor/FuncDetails.vue
@@ -177,6 +177,7 @@
               editingFunc.associations &&
               editingFunc.associations.type === 'codeGeneration'
             "
+            ref="detachRef"
             v-model="editingFunc.associations"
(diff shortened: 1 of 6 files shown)


app/web/src/components/FuncEditor/FuncDetails.vue
```diff
@@ -177,6 +177,7 @@
               editingFunc.associations &&
               editingFunc.associations.type === 'codeGeneration'
             "
+            ref="detachRef"
```
nickgerace, line 180: Sneaky!

nickgerace:
bors merge

PR #3784, opened 2024-05-14
chore: move message decryption into Veritech server

fnichol:
This rather large change move the decryption of function execution requests from the Cyclone server to the Veritech server. Sensitive string redaction still takes place in the Cyclone server.

Prior to this change, portions of the function execution requests would be encrypted before being passed into a Veritech client call using a a public/private keypair called the "Cylone encryption key". The reason for this naming is that the request portions would be decrypted in a Cyclone server before being passed to a language server for final execution.

In order to remove any chance of a user-authored function being able to access such a decryption key, the request decryption activity is now handled back in the Veritech server before it sends the now fully decrypted request over a web socket to a Cyclone server. As a result, there is zero requirement for *any* message cryptography in the Cyclone server.

The second task that was performed in the Cyclone server was secret substring redaction, using a `CycloneSecretStrings` Rust type. With this change, the secret substring redaction still takes place in the Cyclone server but now this is realized by adding the set of "sensitive" secrets as part of every Cyclone request. This is how the Cyclone server knows which strings to redact while not participating in the original decryption of the data.

As a result of the change in responsibility, these encryption/decryption keys are now known as "Veritech encryption" and "Veritech decryption" keys. Consequently all variables, CLI arguments, and files in our source tree have been updated to eliminate any possibility of future confusion.


fnichol:
bors try

lib/veritech-server/src/config.rs
```diff
@@ -52,6 +55,8 @@ pub struct Config {
 
     cyclone_spec: CycloneSpec,
 
+    decryption_key_path: CanonicalFile,
```
sprutton1, line 58: These aren't always paths. We pass them in as base64 strings in prod. SDF Server takes these in using the CryptoConfig object. Can we do that here, too? It will let us keep using our [generated config file](https://github.com/systeminit/si/blob/main/component/init/configs/service.toml).

```
#[builder(default = "CryptoConfig::default()")]
    crypto: CryptoConfig,
```

fnichol:
bors retry

sprutton1:
bors try

fnichol:
bors merge

PR #3786, opened 2024-05-14
experiment(web): Remove component spinner on update

stack72:
We are experimenting with the removal of the spinner icon for component upgrades and changing the colour and opacity of the component


stack72:
bors merge

PR #3787, opened 2024-05-14
chore(dal): remove `Debug` implementation on `DalContext`

fnichol:
This change removes the automatically derived `Debug` implementation from the `DalContext` type.

At present, this type is incredibly complex, with many complex parts. A developer who wishes to "debug" this type will not be helped by calling `dbg!(ctx)` as the output may grow into the megabytes/gigabytes. Furthermore this type can be accidentally debugged when using a tracing `#[instrument()]` call. For any function or method, the `instrument` macro has an implicit "register all function parameters" behavior, meaning that:

```rust
fn do_something(ctx: &DalContext) {}
```

will create a span called `do_something` with a field called `ctx`. The `ctx` field would contain the contents of the `impl std::fmt::Debug for DalContext` output. Removing the `Debug` implementation on `DalContext` makes this accidental behavior not possible as it results in a Rust compilation failure.

### DIFF
--- a/lib/dal-test/src/lib.rs
+++ b/lib/dal-test/src/lib.rs
@@ -21,9 +21,18 @@
     while_true
 )]
 
+use std::{
+    borrow::Cow,
+    env, fmt,
+    future::IntoFuture,
+    path::{Path, PathBuf},
+    sync::{Arc, Once},
+};
+
 use buck2_resources::Buck2Resources;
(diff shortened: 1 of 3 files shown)


nickgerace (approved):
Good idea!

fnichol:
Let's see how this goes...

fnichol:
bors merge

PR #3789, opened 2024-05-14
fix(dal): debug view should not assume all values have executed

zacharyhamm:
We no longer execute all functions for all values on component creation, so we can't assume every value's prototype function has in fact executed.


nickgerace (commented):
Looks good to me! Doesn't have to be this PR, but I wonder if we should have an integration test covering the debug call.

zacharyhamm:
bors retry

PR #3790, opened 2024-05-14
feat(dal,rebaser): store pending dvu values on snapshot

zacharyhamm:
Instead of sending the set of values to use as the starting point for calculating the dependent value graph in the DependentValuesUpdate job, place those values on the graph off a new category node. Then, when the job executes them, remove them from the graph.

### DIFF
--- a/lib/dal/src/workspace_snapshot.rs
+++ b/lib/dal/src/workspace_snapshot.rs
@@ -34,15 +34,16 @@ use futures::executor;
 use si_pkg::KeyOrIndex;
 use std::sync::atomic::AtomicBool;
 use std::sync::Arc;
-use tokio::sync::{RwLock, RwLockReadGuard, RwLockWriteGuard};
 use tokio::task::JoinError;
 
 use petgraph::prelude::*;
 pub use petgraph::Direction;
 use si_data_pg::PgError;
 use si_events::{ulid::Ulid, ContentHash, WorkspaceSnapshotAddress};
+use strum::IntoEnumIterator;
 use telemetry::prelude::*;
 use thiserror::Error;
+use tokio::sync::{RwLock, RwLockReadGuard, RwLockWriteGuard};
 
 use crate::action::{Action, ActionError};
 use crate::change_set::{ChangeSet, ChangeSetError, ChangeSetId};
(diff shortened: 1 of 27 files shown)


lib/dal/src/workspace_snapshot.rs
```diff
@@ -1278,4 +1241,119 @@ impl WorkspaceSnapshot {
...
+        &self,
+        change_set: &ChangeSet,
+        value_id: impl Into<Ulid>,
+    ) -> WorkspaceSnapshotResult<()> {
+        let value_id = value_id.into();
+        let (dv_category_id, existing_value_id) =
+            self.find_existing_dependent_value(value_id).await?;
+
+        if existing_value_id.is_some() {
+            return Ok(());
```
jhelwig, line 1280: Didn't we want to "touch" the existing node somehow to force conflict resolution?  In the case where there's a DVU running that hasn't had the removal of these go through the rebaser, and we modify something (that won't be visible to the running DVU), we don't want to lose the fact that things still need to run. The conflict should always be a "modified a deleted" (or "deleted a modified") one, and the resolution would be to keep the modified version. In the case of a "conflicting modify", we can deduplicate down to one (doesn't really matter which). Same if we detect that two change sets tried to add the same one when neither started from a base that had it.

lib/dal/src/workspace_snapshot.rs
```diff
@@ -232,52 +233,14 @@ impl WorkspaceSnapshot {
...
-            graph.root(),
-            EdgeWeight::new(change_set, EdgeWeightKind::new_use())?,
-            module_node_index,
-        )?;
-        graph.add_edge(
-            graph.root(),
-            EdgeWeight::new(change_set, EdgeWeightKind::new_use())?,
-            action_node_index,
-        )?;
+        for category_node_kind in CategoryNodeKind::iter() {
```
fnichol, line 236: That feels like this will scale better over time

fnichol (approved):
Looking good to me and agreed we can follow up on @jhelwig's comment

zacharyhamm:
forgot i wanted to ensure this was backwards compat

PR #3791, opened 2024-05-14
fix(dal): Ensure that we requeue any actions for the upgraded component

stack72:
If the user upgrades the component, ensure we don't enqueue a create action but also enqueue anything the old component had queued


stack72:
bors merge

stack72:
bors retry

stack72:
bors retry

PR #3793, opened 2024-05-14
fix(dal): do less work in replace references

zacharyhamm:
Replace references was doing too much work. This restores it to almost its original logic, without breaking the multiply parented nodes tests (or any others). I believe the issue fixed by the more complicated logic was actually out of date node indexes, and the extra walks were just duplicating work.


zacharyhamm:
bors try

PR #3797, opened 2024-05-15
Implement actions v2 end to end with the old UI

paulocsanz:


nickgerace (approved):

lib/rebaser-server/src/change_set_requests/handlers.rs
```diff
@@ -94,6 +94,7 @@ pub async fn process_request(State(state): State<AppState>, msg: InnerMessage) -
     if RebaseStatusDiscriminants::Success == rebase_status.clone().into() {
         if let Some(workspace) = Workspace::get_by_pk(&ctx, &workspace_pk.into()).await? {
             if workspace.default_change_set_id() == ctx.visibility().change_set_id {
+                ctx.update_snapshot_to_visibility().await?;
```
paulocsanz, line 97: The actions were only being dispatched the second time they appeared in a commit to head, since the first time the rebaser had added them to head, but the ctx didn't have them since it wasn't in the latest snapshot

jhelwig:
The changes to `set_state` in `action.rs` look reasonable. For future thought: It would be nice if we had a way to quickly implement [`modify`](https://github.com/systeminit/si/blob/d8e9f81416c7de9997e691e23f92cb1bb13a433d/lib/dal/src/component.rs#L1750) for various models quickly. It gets a bit trickier with having to handle model-specific logic (like what happens in `Component`), though.

paulocsanz:
Totally agree with that modify point, we should definitely add some standard-model like system to make that ergonomic, even if it doesn't use macros.

I would also love to expand implement_add_edge_to to allow fetching the data from the edge instead of only adding it

paulocsanz:
bors merge

PR #3798, opened 2024-05-15
fix: remove decryption key referencce in setup script

sprutton1:
Whoops. Missed a decryption key reference.


fnichol (approved):
Oh yeah, that's a nonstarter now!

PR #3799, opened 2024-05-15
feat(layerdb): support eviction for workspace snapshots

adamhjk:
This adds workspace snapshot eviction. The layerdb instance that calls evict will delete the entry from memory, disk, and postgres, and send an eviction event to all other layerdb instances. Those instances will then evict from memory and disk.

If we find ourselves adding a lot more eviction style calls, we need to refactor the layered event structure - right now, it always expects a value, and it's theoretically type safe to pass the wrong kind of layeredevent between eviction and writing. (It may be type safe, but it would be very hard to do in practice, since you never assemble these things directly anyway)

### DIFF
--- a/lib/dal/src/change_set.rs
+++ b/lib/dal/src/change_set.rs
@@ -566,6 +566,7 @@ impl ChangeSet {
 
         Ok(None)
     }
+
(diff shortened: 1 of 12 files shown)


fnichol:
bors merge

PR #3802, opened 2024-05-15
Remove recommendation to run old engine

nickgerace:
## Description

Now that we can deploy SI with SI, we can remove the recommendation to run the old engine. This PR also contains polishing outdated bits of documentation and converts the architecture documentation into something less confusing.


docs/TROUBLESHOOTING.md
```diff
@@ -31,15 +31,15 @@ These external services are deployed via [`docker compose`](https://docs.docker.
 Thus, it is worth checking if you are running these services to avoid conflicts when running SI.
 Potentially conflicting services include, but are not limited to, the following:
 
-* PostgreSQL DB
-* OpenTelemetry
+* Jaeger
 * NATS
```
stack72, line 35: should we mention that this is using Jetstream? I think that's the only change!

stack72 (approved):

docs/TROUBLESHOOTING.md
nickgerace: Good call out since, if I recall correctly, they use two separate ports

nickgerace:
bors cancel

PR #3803, opened 2024-05-15
Remove blocking commits from action

paulocsanz:


lib/dal/src/component.rs
```diff
@@ -2089,41 +2089,6 @@ impl Component {
             .clone_attributes_from(ctx, self.id(), true, true)
             .await?;
 
-        // Enqueue creation actions
```
britmyerss, line 2092: this is what was causing the duplicate creates - we were enqueuing a create action in both the paste AND on creating the component

paulocsanz: Great catch! I totally missed that

lib/dal/src/job/definition/action.rs
```diff
@@ -128,20 +127,24 @@ impl TryFrom<JobInfo> for ActionJob {
...
 ) -> JobConsumerResult<(Option<DeprecatedActionRunResult>, Vec<String>)> {
+    let span = Span::current();
     let component_id = Action::component_id(ctx, id)
         .await?
         .ok_or(ActionError::ComponentNotFoundForAction(id))?;
 
     let prototype_id = Action::prototype_id(ctx, id).await?;
     let prototype = ActionPrototype::get_by_id(ctx, prototype_id).await?;
-
+    span.record("si.action.kind", &tracing::field::debug(&prototype.kind));
```
paulocsanz, line 146: Nice!

jhelwig (approved):

britmyerss:
bors merge

PR #3805, opened 2024-05-16
Support authoring for Actions V2

britmyerss:
This PR does the following things: 

- Create both types of action funcs (regardless of flag, this way new actions funcs are on the graph in the v2 flavor for tools prod)
- Only export v1 or v2 action funcs based on the flag (no need to duplicate what's exported as we can import either kind) 
- Fixes a bug with authoring action funcs where changing the action kind results in duplicate prototypes
- prevents users from creating duplicate action func kinds for a given schema variant

### DIFF
--- a/lib/dal/src/action/prototype.rs
+++ b/lib/dal/src/action/prototype.rs
@@ -5,13 +5,13 @@ use strum::Display;
 use thiserror::Error;
 
 use crate::{
-    func::backend::js_action::DeprecatedActionRunResult,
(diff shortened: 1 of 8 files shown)


stack72 (approved):

britmyerss:
bors try

britmyerss:
bors try

britmyerss:
bors try

britmyerss:
bors try

britmyerss:
bors merge

PR #3806, opened 2024-05-16
Action Prototypes Have Edges!

britmyerss:
When importing action functions, if there's a refresh and create function imported, we create an edge FROM the `ActionPrototype` for Create TO the `ActionPrototype` for Refresh. 

This allows us to enqueue the refresh function after the create successfully runs without special casing it. 

Todo: Any existing workspace will not have this edge created, so we'll need to talk about how to ensure we're backwards compatible. 

Todo 2: TESTS!

### DIFF
--- a/lib/dal/src/workspace_snapshot/edge_weight.rs
+++ b/lib/dal/src/workspace_snapshot/edge_weight.rs
@@ -23,6 +23,11 @@ pub enum EdgeWeightKind {
     Action,
     /// A function used by a [`SchemaVariant`] to perform an action that affects its resource
     ActionPrototype,
+    /// Used to indicate that an [`ActionPrototype`] should always run
+    /// after a successful action run
+    ///
+    /// For example, we currently always run a refresh action after a successful create
+    ActionRunsAfter,
     /// A function defined for a secret defining [`SchemaVariant`] to be executed before funcs on
     /// components that have a secret of that kind
     AuthenticationPrototype,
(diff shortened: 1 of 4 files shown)


paulocsanz (approved):

lib/dal/src/workspace_snapshot/edge_weight.rs
```diff
@@ -23,6 +23,11 @@ pub enum EdgeWeightKind {
     Action,
     /// A function used by a [`SchemaVariant`] to perform an action that affects its resource
     ActionPrototype,
+    /// Used to indicate that an [`ActionPrototype`] should always run
+    /// after a successful action run
+    ///
+    /// For example, we currently always run a refresh action after a successful create
+    ActionRunsAfter,
```
jhelwig, line 30: Gonna nit-pick the name here. I think the edge `A -> B` would be `When A finishes, it triggers B to run`. But the name `ActionRunsAfter` implies a relationship that is the opposite direction from the edge I described `A -- ActionRunsAfter --> B` would be `When B finishes, it triggers A to run`. I think the `ActionRunsAfter` name also brings up the question of whether the feature is "you can specify that this action always runs after any of A, B, or 
C succeed", or if it is "you can specify that whenever this action succeeds, it will cause the following action(s) to run". (Either way can be implemented on top of edges in either direction, but it's probably best to start out with the user facing feature, and the underlying data to match in their directionality.)

PR #3810, opened 2024-05-17
fix(dal): use correct logic in detect_conflicts_and_updates

zacharyhamm:
Our logic in `detect_conflicts_and_updates` was incorrect. A big reason for this was that we were not marking changesets as seen in the tests, and we were also not marking forked / new change sets as *seen by themselves*. Since they had no vector clock entry for the change set itself, we could not reliably detect all the times that onto saw to_rebase or to_rebase saw onto.

For example, if I open up change set B, and make changes to it, in the rebase the changes I make will be the "onto" changeset and the not-yet-modified snapshot pointed to by B will be the "to rebase" changeset. Since onto was forked from to_rebase, onto should have a vector clock entry for to rebase (B). But we were not marking the snapshot for B as "seen" by B when we forked B from the base change set. Thus, we had no way of reliably knowing what things in the "to rebase" were there when we started editing it to produce the "onto", so we couldn't accurately produce the right conflicts and updates.

Note that this means that all currently open changesets are broken, since they don't have a vector clock entry for the change set that they are on. Before we merge this we should decide if we want to try and fix open changesets in prod/tools-prod, or if they can be abandoned. Newly forked changesets will be correct.

### DIFF
--- a/lib/dal/src/workspace_snapshot/graph.rs
+++ b/lib/dal/src/workspace_snapshot/graph.rs
@@ -996,7 +996,8 @@ impl WorkspaceSnapshotGraph {
         debug!("only to rebase edges: {:?}", &only_to_rebase_edges);
         debug!("only onto edges: {:?}", &only_onto_edges);
 
-        let root_seen_as_of_onto = self
+        // This is the last time that to_rebase knows about onto having seen the snapshot
+        let to_rebase_last_saw_onto = self
             .get_node_weight(self.root_index)?
             .vector_clock_recently_seen()
             .entry_for(onto_vector_clock_id);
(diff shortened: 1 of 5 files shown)


zacharyhamm:
bors try

lib/dal/src/workspace_snapshot/graph.rs
```diff
@@ -996,7 +996,8 @@ impl WorkspaceSnapshotGraph {
         debug!("only to rebase edges: {:?}", &only_to_rebase_edges);
         debug!("only onto edges: {:?}", &only_onto_edges);
 
-        let root_seen_as_of_onto = self
+        // This is the last time that to_rebase was written to by the onto change set
```
jhelwig, line 999: Shouldn't this say `... the last time that to_rebase knows about onto seeing the state of the graph`? `onto` won't necessarily have made any changes other than recording that it saw things.

lib/dal/src/workspace_snapshot/graph.rs
```diff
@@ -1013,45 +1014,63 @@ impl WorkspaceSnapshotGraph {
...
 
-            if to_rebase_edge_weight
+            // This is an edge that is only to_rebase. So either:
+            // -- Onto has seen this edge:
+            //     -- So, if to_rebase has modified the target of the edge since onto last saw the target,
+            //      we should produce a ModifyRemovedItem conflict,
+            //     -- OR, if to_rebase has *not* modified the target, we should produce a RemoveEdge update.
+            // -- Onto has never seen this edge:
+            //     -- So, either we should just silently let the edge stay in to_rebase, OR, we should produce
+            //          an exclusive edge conflict
```
jhelwig, line 1024: Worth pointing out that the reason this scenario would happen is if the edge were added to `to_rebase` **after** `onto` was forked from `to_rebase`?

lib/dal/src/workspace_snapshot/graph.rs
```diff
@@ -1063,39 +1082,53 @@ impl WorkspaceSnapshotGraph {
...
-                        // Edge first seen by `onto` > "seen as of" on `to_rebase` graph for `onto`'s entry on
-                        // root node: Item is new.
-                        // Other case where item is new: the `to_rebase` has never seen anything from
-                        // the `onto` change set. All the items are new.
-                        updates.push(Update::NewEdge {
-                            source: to_rebase_container_index,
-                            destination: only_onto_edge_info.target_node_index,
-                            edge_weight: onto_edge_weight.clone(),
-                        });
+                // if the edge was first seen by onto *before* onto last saw to rebase?
```
jhelwig, line 1098: Left over note-to-self?

zacharyhamm: yes!

jhelwig (approved):

PR #3812, opened 2024-05-19
fix(dal): Fix up autocompletion types for Func editor

stack72:
Ensure that we set the correct autocorrect types for action and attribute funcs


stack72:
bors try

stack72:
bors merge

stack72:
bors retry

PR #3815, opened 2024-05-20
feat(ci): invalidation of tools CDN on web refresh

johnrwatson:
Adds a basic CDN invalidation when we deploy to our tooling environment, see passing runs here
- https://github.com/systeminit/si/actions/runs/9163133049 (web only)
- https://github.com/systeminit/si/actions/runs/9163157549 (all services)

(Basically it prevents us from needing to remember to do it against tools. Production will be managed via SI directly in actions)


johnrwatson:
bors merge

PR #3816, opened 2024-05-20
fix(web): Don't show that a root prop has a value overridden

stack72:
Domain and Secrets are props attached to the root and are not set directly, therefore, we were telling a user they had overridden them which was misleading

Screenshot 2024-05-20 at 20 06 04

No icons in place for props directly connected to the root


stack72:
bors merge

PR #3817, opened 2024-05-20
Do not dispatch Actions with pending dependent value updates

jhelwig:
While there are AttributeValues for any given Component that are still involved in a DependentValuesUpdate, we want to hold off on dispatching any Actions for those Components. This will let us make sure that the Action will be operating with the most up-to-date data possible in the model.

Now that we're tracking the root nodes for the DependentValuesUpdate graph of AttributeValues to update, we can directly check the Components that are involved in the pending DependentValuesUpdate to see if an Action would operate on an involved Component.

### DIFF
--- a/lib/dal/src/action.rs
+++ b/lib/dal/src/action.rs
@@ -1,4 +1,4 @@
-use std::collections::VecDeque;
+use std::collections::{HashSet, VecDeque};
 
 use petgraph::prelude::*;
 use serde::{Deserialize, Serialize};
@@ -11,6 +11,7 @@ use veritech_client::ResourceStatus;
 use crate::{
     action::dependency_graph::ActionDependencyGraph,
     action::prototype::{ActionKind, ActionPrototype, ActionPrototypeError},
+    attribute::value::{AttributeValueError, DependentValueGraph},
     func::backend::js_action::DeprecatedActionRunResult,
     func::execution::{FuncExecution, FuncExecutionError, FuncExecutionPk},
     id, implement_add_edge_to,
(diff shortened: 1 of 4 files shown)


jhelwig:
bors try

lib/dal/src/action.rs
```diff
@@ -545,19 +548,57 @@ impl Action {
...
+        let mut dvu_component_ids: HashSet<ComponentId> = HashSet::new();
+        for av_id in &dependent_value_graph.all_value_ids() {
+            dvu_component_ids.insert(AttributeValue::component_id(ctx, *av_id).await?);
+        }
 
         for possible_action_id in action_dependency_graph.independent_actions() {
             let action = Action::get_by_id(ctx, possible_action_id).await?;
 
             if action.is_eligible_to_dispatch() {
+                if let Some(action_component_id) = Action::component_id(ctx, action.id()).await? {
```
britmyerss, line 593: would we ever want this to get more granular? 

I haven't thought this all of the way through, but if a qualification hasn't finished yet for a component, we could still run the action?

jhelwig: If we were to want to do that in the future, we'd want to extend the graph to also contain what kind of thing each AV is for, and record the information as we're building up the graph.

To complicate things further: While we don't do this now, with the way we've structured things, it is possible for the result of a qualification to be an input to another function (it's possible to expose qualification results on an output socket, or even to use them in another Prop). So, while it's possible we might not want to wait on the qualification(s) in that case, we'd still need to build up the full graph, and consider each AttributeValue individually as to whether or not to include it's Component ID in the list of blockers (instead of filtering the initial set of "root" AV IDs).

So, the highlighted check wouldn't change; we'd be populating the set of ComponentIds with more consideration as to what that individual AttributeValue "is".

PR #3818, opened 2024-05-20
fix(dal): Ensure that CodeGen and Qualifications can be executed

stack72:
We were making the asusmption that executable func types were limited to Attribute funcs when it was infact FuncBackendKind::JsAttribute which includes codegen and qualifications

The logic doesn't need to filter the runs to components on the graph, our DVU enqueue will do that for us as the attribute_value_ids for the prototype will find all the values across all components for that proto


stack72:
bors merge

PR #3820, opened 2024-05-21
fix(dal): Ensure that we don't allow changing a func to an existing kind

stack72:
If you have a Create func for an asset type, and we change a func type to be Create, then we want to error on save of the func because it's not allowed to have 2 different funcs of the same type for a schema variant


stack72:
bors merge

PR #3821, opened 2024-05-21
Include node ID and kind in Conflict & Update records

jhelwig:
Previously, we were only including the (internal to the graph) `NodeIndex` in the structs for representing conflicts & updates. While this is all that's strictly necessary for performing the updates and (eventually) having SDF report what the conflicts are, having more contextual information such as the node weight kind, and the ID of the node makes things a bit easier for debugging issues when looking at spans & traces.


jhelwig:
bors merge

stack72:
bors will have failed due to CI having a termination signal in it - 

```

# Received cancellation signal, interrupting
--
  | Terminated
  | 🚨 Error: The command exited with status 143
  | user command error: The plugin docker command hook exited with status 143


```

stack72:
ooohhhhh it does a retry now - I forgot!!

PR #3825, opened 2024-05-22
chore(*): layer-db tracks function execution

nickgerace:
This PR makes layer-db track function execution in func_runs() and
func_run_logs() tables. It also removes the following concepts:

* Func Bindings
* Func Binding Return Values
* Func Executions
* Actions v1

We have not cleaned up the old standard model tables, but other than
that, we're gtg.


fnichol:
bors try

stack72:
The main issue I see here is that a failed action will still set a resource prop - it's empty but it's still set nonetheless and therefore the component can never be deleted

johnrwatson:
bors cancel

adamhjk:
bors try

adamhjk:
bors merge

PR #3830, opened 2024-05-22
Update prototype function instead of removing prototype

paulocsanz:


paulocsanz:
bors try

lib/dal/src/func/authoring/save.rs
```diff
@@ -411,7 +411,8 @@ async fn remove_or_reset_attribute_prototype(
             return Ok(());
         }
     }
-    AttributePrototype::remove(ctx, attribute_prototype_id).await?;
+    let identity_func_id = Func::find_intrinsic(ctx, IntrinsicFunc::Identity).await?;
+    AttributePrototype::update_func_by_id(ctx, attribute_prototype_id, identity_func_id).await?;
```
nickgerace, line 415: @paulocsanz I *think* we may want this to be `Unset` instead, but I would have to check the calling locations of `remove_or_reset_attribute_prototype` to be sure.

paulocsanz:
bors merge

PR #3833, opened 2024-05-22
fix(dal): add recently seen clocks to edge weights

zacharyhamm:
EdgeWeights had no "recently seen" vector clocks. This lead to situations where we produced a conflict but should not have, because we only had the "first seen" clock to work with. This adds the recently seen vector clock to the graph. However, this cannot be done without a new version of the WorkspaceSnapshotGraph, so migration logic has been implemented to automatically migrate the snapshot on first read.

### DIFF
--- a/lib/dal/src/workspace_snapshot.rs
+++ b/lib/dal/src/workspace_snapshot.rs
@@ -31,20 +31,24 @@ pub mod update;
 pub mod vector_clock;
 
 use futures::executor;
-use si_pkg::KeyOrIndex;
+use std::collections::HashMap;
 use std::sync::atomic::AtomicBool;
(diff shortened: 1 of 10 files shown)


lib/dal/src/workspace_snapshot.rs
```diff
@@ -745,6 +767,89 @@ impl WorkspaceSnapshot {
         })
     }
 
+    /// Attempts to migrate a snapshot graph to the latest version. Should only
```
britmyerss, line 770: Clever! So we don't actually need a version mapping explicitly then? What if we need to make another change to EdgeWeight? Or anything else about the WorkspaceSnapshotGraph needs to change?

PR #3834, opened 2024-05-22
Move check for Conflict::ExclusiveEdgeMismatch to operate on the net changes for the outgoing edges

jhelwig:
Previously, the check was not taking into account that while there may be an `Update::NewEdge` that is creating an outgoing edge of the same `EdgeWeightKind` as an existing edge, there might also be an `Update::RemoveEdge` that is removing that pre-existing edge.

We now check to make sure that after all additions & removals of outgoing edges for a specific `EdgeWeightKind`, the net result is that we have no more than a single edge.

The following exclusive edge mismatch conditions have been removed as they are more nuanced than "Node X can have at most 1 of edge kind Y", and will be addressed at a later date.

* `AttributeValue` - `Contain`: `AttributeValue`s can have multiple outgoing `Contain` edges. The restriction on whether or not a specific `Contain` edge is a semantic conflict with another depends on information about the `Prop` tree structure, and looking at the `key` portion of the `Contain` edge weight.
* `Component` - `FrameContains`: The exclusivity of the `FrameContains` edge is that a `Component` can only have one *incoming* `FrameContains` edge. An arbitrary number of *outgoing* `FrameContains` are allowed.
* `Func` - `Use`: A `Func` can have multiple outgoing `Use` edges, and is a requirement for any `Func` that has multiple `FuncArgument`s.
* `Prop` - `Prototype`: As with `AttributeValue` and `Contain`, this requires looking at the prop tree structure, and taking into consideration the `key` portion of the `Prototype` edge weight.
* `Prop` - `Use`: As with `FrameContains`, the direction of the exclusivity is on *incoming* edges, not on the number of *outgoing* edges.

### DIFF
--- a/lib/dal/src/workspace_snapshot.rs
+++ b/lib/dal/src/workspace_snapshot.rs
@@ -69,7 +69,7 @@ use crate::{
 use self::node_weight::{NodeWeightDiscriminants, OrderingNodeWeight};
 
 pk!(NodeId);
-#[derive(Debug, Copy, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
 pub struct NodeInformation {
     pub index: NodeIndex,
     pub node_weight_kind: NodeWeightDiscriminants,
--- a/lib/dal/src/workspace_snapshot/conflict.rs
+++ b/lib/dal/src/workspace_snapshot/conflict.rs
@@ -5,7 +5,7 @@ use crate::{workspace_snapshot::NodeInformation, EdgeWeightKindDiscriminants};
 /// Describe the type of conflict between the given locations in a
 /// workspace graph.
 #[remain::sorted]
-#[derive(Debug, Copy, Clone, PartialEq, Eq, Deserialize, Serialize)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Deserialize, Serialize, Hash)]
 pub enum Conflict {
     ChildOrder {
         onto: NodeInformation,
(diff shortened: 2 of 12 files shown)


keeb:
I gave this PR a spin tonight - the following regression use case failed

1. Create butane component on graph
2. Turn butane into up frame
3. Add 2 docker images to graph, not into frame
4. Drag 2 docker images onto frame
NOTICE: conflicts and that DVU does not run properly.

The rebase request shows the conflict in tracing as the following.

image

The following **PATCH** fixes the issue. I do not know if this is the right thing to do, I am only noting that it fixes the particular problem that has been surfaced

```diff --git a/lib/dal/src/workspace_snapshot/node_weight/component_node_weight.rs b/lib/dal/src/workspace_snapshot/node_w
eight/component_node_weight.rs
index a1a5f9806..d821b1048 100644
--- a/lib/dal/src/workspace_snapshot/node_weight/component_node_weight.rs
+++ b/lib/dal/src/workspace_snapshot/node_weight/component_node_weight.rs
@@ -174,7 +174,6 @@ impl ComponentNodeWeight {
     pub const fn exclusive_outgoing_edges(&self) -> &[EdgeWeightKindDiscriminants] {
         &[
             EdgeWeightKindDiscriminants::Use,
-            EdgeWeightKindDiscriminants::FrameContains,
             EdgeWeightKindDiscriminants::Root,
             EdgeWeightKindDiscriminants::SocketValue,
         ]
```

jhelwig:
@keeb Yup, that change is the right thing to do. We need that to be an exclusive *incoming* edge, not an exclusive *outgoing* edge. We need to extend the edge checks to handle a number of other semantic conflicts beyond purely exclusive incoming/outgoing edges.

jhelwig:
bors try

jhelwig:
bors try

jhelwig:
bors try

jhelwig:
bors try

keeb:
bors merge

PR #3835, opened 2024-05-22
Actions V2 Fixes

britmyerss:
Fixed 2 bugs in this PR: 
- an issue with removing delete actions from the graph on successful action run (due to a race in removing all existing actions when removing a component, and the surgical 'remove this action because it finished') 
- when we set attribute values, (for example setting the resource), we were always deleting and recreating the Attribute Prototype. With this change we will first check if there's a component specific attribute prototype and if so, wipe the Arguments and repopulate them. Then, create one if it doesn't already exist. 

<div><br/>via <a href="https://giphy.com/nickelodeon/">Nickelodeon</a> on <a href="https://giphy.com/gifs/nickelodeon-theatre-the-loud-house-and-action-EukOAWKKs0QoW3ps7Z">GIPHY</a></div>

### DIFF
--- a/lib/dal/src/action.rs
+++ b/lib/dal/src/action.rs
@@ -9,19 +9,23 @@ use thiserror::Error;
 use veritech_client::{OutputStream, ResourceStatus};
 
 use crate::{
-    action::dependency_graph::ActionDependencyGraph,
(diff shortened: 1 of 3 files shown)


britmyerss:
bors merge

PR #3836, opened 2024-05-23
Add regression test for qualification views

nickgerace:
This was found during func execution refactoring (#3825) since other tests rely on qualification view generation working.


nickgerace:
bors merge

PR #3839, opened 2024-05-23
fix(dal): Ensure that we copy Socket data on export schema variant

stack72:
We were not merging the socket data when exporting a schema variant and thus dropping all of the export details


stack72:
bors merge

PR #3840, opened 2024-05-23
dev: Adding Localstack to our devstack

stack72:
We no longer need AWS credentials to be able to use the system - we can use the AWS_ENDPOINT http://0.0.0.0:4566 and we are good to go!


stack72:
bors merge

stack72:
bors retry

PR #3841, opened 2024-05-23
fix(dal): fix snapshot migration to be more reliable

zacharyhamm:
Migrates the builtin change set eagerly on boot, and migrates change sets lazily in sdf, when the change set is opened. Instead of the previous strategy of waiting for the rebaser to update the pointer with the snapshot that SDF updated, the first service to read the snapshot moves the change set pointer ahead itself.


zacharyhamm:
bors try

PR #3843, opened 2024-05-24
Remove func editor panel

paulocsanz:
Nobody was using it, and it was creating a bunch of edge-cases related to shared function bindings and things like that.

There is value in being able to list all functions and being able to search for them. Also seeing where they are used, but this is not a flow we have been using, so if it becomes valuable we should reimplement it properly.

In any case, if you are from the future: feel free to revert this PR to tweak it


paulocsanz:
bors merge

PR #3845, opened 2024-05-24
Split attribute func mutation into separate routes

nickgerace:
## Description

This PR splits attribute func mutation into separate routes. Specifically, this affects attribute prototypes, attribute prototype arguments and func arguments. A "Bindings" tab has been added to show all attribute bindings for a given selected func rather than just the bindings specific to the selected asset.

This PR also addresses bug fixes like func arguments and attribute prototype arguments being out of sync as well as creating the attribute prototypes necessary when attaching an existing Func to another asset.

## Known Issues

Known issues include, but are not limited to, the following: detach button not working, func test execution not working and general polish related to func executions in layerdb and how that may interact with the func authoring experience.


app/web/src/components/FuncEditor/AttributeBindings.vue
nickgerace, line 1: We need to fix reactivity in the bindings listed

app/web/src/components/FuncEditor/FuncArguments.vue
```diff
@@ -12,54 +12,60 @@
         placeholder="New argument name..."
         :disabled="disabled"
       />
-      <SelectMenu
+      <!-- <SelectMenu
```
nickgerace, line 15: Fix these selects. Hopefully, we don't have to use `Option` and can mimic other select boxes.

app/web/src/components/FuncEditor/FuncArguments.vue
```diff
@@ -12,54 +12,60 @@
...
           v-model="arg.name"
           type="text"
           noLabel
           placeholder="Argument name"
           :disabled="disabled"
-          @blur="saveArguments"
+          @blur="updateArgument(arg.id, arg.name, arg.kind, arg.elementKind)"
         />
-        <SelectMenu
+        <!-- <SelectMenu
```
nickgerace, line 39: Fix these selects. Hopefully, we don't have to use `Option` and can mimic other select boxes.

nickgerace (commented):
Last thing: we should refactor to get the removed independent func panel as well as the new func execution system

lib/dal/src/func/authoring/save.rs
```diff
@@ -237,14 +236,7 @@ async fn update_attribute_associations(
...
-    arguments: Vec<FuncArgumentBag>,
 ) -> FuncAuthoringResult<()> {
-    // First, modify the arguments because they dictate what we can do within the
-    // attribute subsystem.
-    save_attr_func_arguments(ctx, func, arguments).await?;
-
-    // Now that we know what func arguments exist and have been modified, we can work
-    // within the attribute subsystem.
     let backend_response_type =
         save_attr_func_prototypes(ctx, func, prototypes, true, None).await?;
```
nickgerace, line 241: **This is the reason why this PR exist. The `arguments` and `prototypes` can get out of sync quickly. `FuncArguments` should be handled independently and then `prototypes` (which include `prototypeArguments`, what that are called in `app/web`) can use said `FuncArguments`.**

app/web/src/components/FuncEditor/FuncArguments.vue
nickgerace, line 1: We may need to "debounce" these new API requests like we do for `save_func`.

fnichol:
> We should **git rebase** to get the removed independent func panel as well as the new func execution system

Yep, done now!

app/web/src/components/AssetFuncAttachModal.vue
britmyerss, line 1: Here's where next steps are. 

When the user selects an existing Attribute Func, we need to load in that Func's Args, and show a drop down for each with the available input sources for the selected schema variant. Then, instead of returning the new associations, (or maybe when we return the new associations, dealer's choice) we need to call `CREATE_ATTRIBUTE_PROTOTYPE` and pass in required data.  This same thing can be accomplished from the bindings tab when viewing an existing attribute func, so there's a bit of a pattern to follow there.

app/web/src/components/FuncEditor/AttributeBindings.vue
```diff
@@ -100,52 +99,65 @@ const props = defineProps<{
...
-    funcArgumentId: id,
-  })),
+  prototypeArguments: funcArguments.value
+    ? funcArguments.value.map(({ id }) => ({
+        funcArgumentId: id,
+      }))
+    : [],
+});
+
+// todo: Don't remove existing configured args when pulling in the new ones
```
britmyerss, line 122: nice to have: right now when we edit a binding, we're wiping any currently configured args. We should instead keep the currently configured ones, letting them configure any new ones if they exist (and change existing if necessary)

app/web/src/components/FuncEditor/AttributeBindings.vue
```diff
@@ -217,14 +225,14 @@ const prototypeViews = computed(() =>
...
   if (props.schemaVariantId) {
-    return {
-      ...associations.value,
-      prototypes: associations.value.prototypes.filter(
-        (proto) =>
-          funcStore.schemaVariantIdForAttributePrototype(proto) !==
-          props.schemaVariantId,
-      ),
-    };
+    const prototype = associations.value.prototypes.find(
```
britmyerss, line 228: When viewing an Attribute Func, if the user clicks the `Detach` button, we need to instead of removing the association from what's in memory and saving the whole thing, call removeBinding.

app/web/src/store/func/funcs.store.ts
britmyerss, line 1: It's a bit ugly, but instead of separately fetching the attribute prototypes for an Attribute Func, I'm relying on the existing Func Details to hydrate what's required.  I started to break this out but for expediency just left it here. So we still have a hammer for how we get the attribute prototypes for attribute funcs, but no longer rely on the `save_func` hammer for updating/removing/adding them.

app/web/src/store/func/funcs.store.ts
```diff
@@ -618,11 +786,14 @@ export const useFuncStore = () => {
 
               if (this.selectedFuncId) {
                 // only fetch if we don't have this one already in our state,
+                // or if the func kind is attribute
                 // otherwise we can overwrite functions with their previous value
                 // before the save queue is drained.
                 if (
-                  typeof this.funcDetailsById[this.selectedFuncId] ===
-                    "undefined" &&
+                  (typeof this.funcDetailsById[this.selectedFuncId] ===
```
britmyerss, line 793: I think this does what we need given I'm using the FuncSaved event still when Attribute Prototypes are modified, but I'd appreciate a second opinion :)

lib/dal/src/func/authoring/save.rs
```diff
@@ -40,17 +40,21 @@ pub(crate) async fn update_associations(
...
-                arguments,
-            } => update_attribute_associations(ctx, func, prototypes, arguments).await,
-            invalid => {
-                return Err(FuncAuthoringError::InvalidFuncAssociationsForFunc(
-                    invalid, func.id, func.kind,
-                ))
-            }
-        },
+        FuncKind::Attribute => Ok(()),
+        // don't update attribute associations this way
```
britmyerss, line 44: we can remove this commented out code but left it for context. Save_func no longer updates Attribute Prototypes

lib/dal/tests/integration_test/func/authoring/func_argument.rs
```diff
@@ -0,0 +1,235 @@
...
+            found_array_func_argument.kind  // actual
+        );
+        assert_eq!(
+            array_func_argument_element_kind,       // expected
+            found_array_func_argument.element_kind  // actual
+        );
+
+        (found_string_func_argument.id, found_array_func_argument.id)
+    };
+    // note from brit: unclear if this test was ever passing, but because this new
```
britmyerss, line 136: left this here for context and to get a second opinion :)

lib/dal/src/func/authoring/save.rs
britmyerss: So after picking up this PR - we needed to take this a step further as while we separated out the Func Arguments, we also needed to separate out how we handle prototype arguments. This was because when adding a new argument to  a shared attribute function, since we were still saving ALL prototype arguments for all schema variants, the saves would fail if you only configured the new prototype argument for ONE schema variant and not the others.  

I broke out CRUD routes for create/update/remove attribute prototypes so we can individually save each schema variants prototype without having to configure the prototypes for all schema variants at once. 

Note: the 'associations' for an attribute func are still returned when we get func details, I didn't break out separate list calls for expediency, but I expect this domain to evolve and didn't want to go to far in architecting to account for the current experience we provide.

nickgerace:
Future folks: we are not ignoring the comments in this PR, but... we are. We plan to fix the rest on `main`.

nickgerace:
bors try

nickgerace:
bors merge

PR #3846, opened 2024-05-24
bug(dal): fix back compat for workspace snapshots

adamhjk:
In theory, now we should get some default options.

Maybe don't fuck with weights?

Love,
Adam, Fletcher, Zack and Brit


adamhjk:
bors merge

PR #3847, opened 2024-05-24
Fix: action kind for actions and for funcs has drifted apart on the backend 

jobelenus:
Address it here


stack72:
We don't actually need this as it was picked up elsewhere - sorry!!

PR #3849, opened 2024-05-24
feat: added log aggregation in grafana for local services

johnrwatson:
Four main changes:
- Adds a `/tmp/si-logs/<service>` log file when running tilt. Deletes/Recreates this folder just before starting tilt. 
- `tees` the logs of the rust services into those files
- Adds promtail + loki so that the service logs for the rust services can be aggregated in local grafana.
- Sorts out the groupings of the services within the tiltfile

Kudos should be directed to my partner in crime @keeb

### DIFF
--- a/dev/docker-compose.platform.yml
+++ b/dev/docker-compose.platform.yml
@@ -1,6 +1,20 @@
 ---
+# configures the system to use the loki docker driver
+# see https://grafana.com/docs/loki/latest/send-data/docker-driver/
+x-loki-logged-service: &loki-logged-service
+    logging:
+      driver: loki
+      options:
+        loki-url: "http://loki:3100/loki/api/v1/push"
+        loki-batch-size: "400"
+        loki-retries: "1"
+        max-file: "2"
+        max-size: "10m"
+
 services:
   db:
+    # uncomment below to get these logs in grafana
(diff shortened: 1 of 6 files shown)


dev/docker-compose.platform.yml
```diff
@@ -118,7 +158,7 @@ services:
       - jaeger
 
   localstack:
+    #<<: *loki-logged-service
     image: localstack/localstack
     ports:
       - "4566:4566"
-      - "4510-4559:4510-4559"
```
johnrwatson, line 124: removed as these are not needed for local development and were just hogging the tilt UI space

prelude-si/tilt.bzl
```diff
@@ -124,12 +124,16 @@ def _invoke_tilt(ctx: AnalysisContext, subcmd: str) -> list[[DefaultInfo, RunInf
         ctx.attrs.tiltfile,
     )
 
+    RunInfo(["rm", "-rf", "/tmp/si-logs"])
```
johnrwatson, line 127: Nothing like an `rm -rf` to wake you up in the morning

keeb: This one is tactical

johnrwatson:
bors merge

PR #3850, opened 2024-05-24
feat(web) - Actions V2 UI mock

wendybujalski:
https://github.com/systeminit/si/assets/8495209/8e903b5d-aed6-4668-874f-d2d55f071090


app/web/src/components/Actions/ActionCard.vue
```diff
@@ -13,43 +13,49 @@
...
+                  'opacity-30',
+                  themeClasses('text-warning-500', 'text-warning-300'),
+                ]
+              : themeClasses('text-warning-400', 'text-warning-300'),
+          )
         "
+        size="lg"
       />
+      <template v-else-if="actionFailed">
+        <Icon
```
jobelenus, line 43: We were gonna only show this on "proposed", not "history", right? So history can never have 3 icons...

jobelenus (approved):
SIIIIICK

app/web/src/components/Actions/ActionCard.vue
wendybujalski: That's all within a `<template>` from lines 16-55 that is only for `actionProposed`

I do have a "Failed" result as part of the `ActionHistoryResult` but we can remove it if we end up not using it.

PR #3851, opened 2024-05-25
Do not attempt to re-use pre-existing Prototype nodes when setting the function for an AttributeValue

jhelwig:
The change to introduce this behavior happened because the `ExclusiveEdgeMismatch` machinery was not detecting edge additions that had corresponding edge removals for the pre-existing exclusive edge.

Re-using the prototype was not fully cleaning up the pre-existing arguments, so we were ending up creating things like a `si:setObject` prototype with multiple arguments. This would result in an array being passed to `si:setObject`, and the execution blowing up as an array can't be (de)serialized as a map.

Now that the `ExclusiveEdgeMismatch` machinery has been updated to correctly handle the case where the exclusive constraint for an outgoing edge is not violated after all of the additions & removals are considered, the additional logic to try to re-use & clean up the old `AttributePrototype` is no longer necessary.

### DIFF
--- a/lib/dal/src/attribute/value.rs
+++ b/lib/dal/src/attribute/value.rs
@@ -1901,31 +1901,10 @@ impl AttributeValue {
                 IntrinsicFunc::from(prop_node.kind())
             }
         };
-        let associated_component_id = AttributeValue::component_id(ctx, attribute_value_id).await?;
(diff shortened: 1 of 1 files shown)


jhelwig:
bors merge

PR #3855, opened 2024-05-26
Build & use reverse component dependency graph for finding dependencies of Destroy Actions

jhelwig:
We need to ensure that things are destroyed in the reverse order of what we would have created them in.


jhelwig:
bors merge

PR #3856, opened 2024-05-26
Retry fetching workspace snapshot for a change set if content for the snapshot address is not found

jhelwig:
There is a race condition:
  * Process A consults change set X for current pointer address of the workspace snapshot
  * Rebaser moves change set X on to point to a new snapshot
  * Rebaser garbage collects snapshot change set X moved on from
  * Process A tries to retrieve snapshot at pointer address it got
  * Process A is unable to retrieve garbage collected snapshot

Rather than immediately failing, we now have a limited number of retries where we'll fetch what the address has moved on to, and attempt to fetch the content at that address, with a small sleep between each retry.


keeb:
bors merge

PR #3862, opened 2024-05-28
chore(dal): refactor FuncRunner to support instrumentation

fnichol:
This change adds rather extensive instrumentation coverage around all
function execution types. As this code lives in the DAL, we have access
to data such as change set and workspace identifiers so this is a very
nice location to attach this relevant data.

The largest refactoring step was to take the 2 spawned activities and
split them into explicit types to make future tracking and cancelling
easier.


fnichol:
(zoom pr'd with @nickgerace)

bors merge

PR #3864, opened 2024-05-28
Handle Some(serde_json::Value::Null) more gracefully when setting nested values

jhelwig:
Rather than trying to use `si:setString`, `si:setBool`, or any similar function with a `serde_json::Value::Null`, we now translate this to behave as though we had gotten a `None`, and use `si:unset`.

Ideally we would handle things in such a way that we can differentiate between "there is no value here" (`None`), and "the value here is `null`" (`Some(serde_json::Value::Null)`), but this is hard enough to do within JavaScript/JSON on its own, let alone round tripping data in & out of JSON multiple times before dealing with it.

### DIFF
--- a/lib/dal/src/attribute/value.rs
+++ b/lib/dal/src/attribute/value.rs
@@ -1874,6 +1874,7 @@ impl AttributeValue {
         attribute_value_id: AttributeValueId,
         value: Option<Value>,
     ) -> AttributeValueResult<()> {
+        let mut normalized_value = value.to_owned();
         let prop_id = match AttributeValue::is_for(ctx, attribute_value_id).await? {
             ValueIsFor::Prop(prop_id) => prop_id,
             _ => {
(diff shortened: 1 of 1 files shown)


fnichol (approved):
Well yep, that certainly is subtle, and matches what we discovered the other week.

If the tests pass then I think we're good here

jhelwig:
Well, that's an interesting test failure. Gonna have to look at that one in a bit.

```
Message:  assertion failed: `(left == right)`
--
  |  
  | Diff < left / right > :
  | [
  | QualificationView {
  | title: "Prop Validations",
  | output: [],
  | description: None,
  | link: None,
  | result: Some(
  | QualificationResult {
  | status: Success,
  | title: None,
  | link: None,
  | sub_checks: [
  | QualificationSubCheck {
  | description: "Component has 0 invalid value(s).",
  | status: Success,
  | },
  | ],
  | },
  | ),
  | qualification_name: "validations",
  | finalized: true,
  | },
  | QualificationView {
  | title: "test:qualificationDummySecretStringIsTodd",
  | <        output: [
  | <            QualificationOutputStreamView {
  | <                stream: "output",
  | <                line: "Output: {\n  \"protocol\": \"result\",\n  \"status\": \"success\",\n  \"executionId\": \"tomcruise\",\n  \"data\": {\n    \"result\": \"failure\",\n    \"message\": \"dummy secret string is empty\"\n  },\n  \"unset\": false\n}",
  | <                level: "info",
  | <            },
  | <        ],
  | >        output: [],
  | description: None,
  | link: None,
  | result: Some(
  | QualificationResult {
  | status: Failure,
  | title: Some(
  | "test:qualificationDummySecretStringIsTodd",
  | ),
  | link: None,
  | sub_checks: [
  | QualificationSubCheck {
  | description: "dummy secret string is empty",
  | status: Failure,
  | },
  | ],
  | },
  | ),
  | qualification_name: "test:qualificationDummySecretStringIsTodd",
  | finalized: true,
  | },
  | ]
  |  
  |  
  | Location: lib/dal/tests/integration_test/qualifications.rs:104

```

fnichol:
bors retry

PR #3866, opened 2024-05-28
Only enqueue Destroy actions when transitioning from not to_delete to to_delete

jhelwig:
When setting to_delete on a component, we only want to enqueue Destroy actions if the component is not already in the to_delete state, so we don't automatically end up with multiple Destroy actions queued for the same component.


fnichol (approved):
Nice!

fnichol:
bors merge

fnichol:
bors retry

PR #3870, opened 2024-05-28
Fix: dont confetti unless on head

jobelenus:
Confetti was firing on a changeset when deleting a component that was added in the changeset.


jobelenus:
bors merge

jobelenus:
bors merge

PR #3872, opened 2024-05-28
commit more often to reduce the chance that the background job races …

britmyerss:
…our blocking_commit

PR #3873, opened 2024-05-28
Restore test execution with func run and func run log

nickgerace:
## Description

This PR restores test executions when authoring funcs via func run and func run log. Now, test execution results are stored and test executions states will resolve to "success" when... uh, successful. We no longer need to track an execution key because we have the `FuncRunId` for the given test.

As a result of these changes, test execution integration tests now wait for both the func run to be successful and for the logs to be finalized.

Finally, action funcs can now be tested again. The real world implications of the execution cannot be controlled (like all test executions), but we will not mutate the `/root/resource` tree when testing action funcs.

## Secondary Changes

- Allow for persistent code modification when testing funcs because `save_func` will be called anyway when modifying code
  - We want to ensure the code that's actually in the buffer in the editor is sent down rather than blindly using what is retrieved from the current snapshot
- Fix code viewer flashing by not refetching after every `save_func` when editing attribute funcs (note: this may cause other issues depending on what we need to refresh, but we should not be refreshing every time)
- Move before funcs logic into `FuncRunner` as it can now be made private
- Move test in before funcs integration test module into parent secret module
- Make action kind collision error clearer when dealing with action funcs when authoring
- Consolidate duplicate "read" for `FuncRunDb`
- Remove unused `LogLine` `WsEvent` from frontend
- Fix missing blocking commits detection in tests


nickgerace:
Investigation concluded: the intermittent test failures are caused by an existing bug on `main` and not this PR. We believe it is due to issues that https://github.com/systeminit/si/pull/3872 is trying to solve. Once those are solves, this PR can be merged.

nickgerace:
bors merge

PR #3875, opened 2024-05-29
fix(dal): give dvu 10000 second interval in tests

zacharyhamm:
To prevent DVU debouncer from racing with blocking_commit jobs in tests, increase tick interval in tests to 10000 seconds, so it will never run (unless an individual test takes several hours (unless an individual test takes several hours). This is a stop gap until we implement a way to wait on the completion of the debounced dvus.


zacharyhamm:
bors try

fnichol (approved):
Not a bad solution and less lines of code than I expected!

PR #3881, opened 2024-05-29
Fix action associations when refresh mutating action funcs post failure

nickgerace:
## Description

This commit fixes action associations during `save_func` by hardening the association logic. The fix not only aims to polish action association mutation logic as a whole, but fixes a user-level bug where mutating a refresh action func after a refresh action has failed causes the system to go into an unrecoverable state. There is likely another bug lurking there related to conflicts, updating, rebasing, etc., but this commit, at least, avoids the very bad times.

Additionally, this commit adds a developer document related to using LocalStack for local development with "AWS Credential".


nickgerace:
bors merge

PR #3883, opened 2024-05-29
Remove resource payload if action function is successfull

paulocsanz:
We were not removing the payload after a successfull delete or a successfull refresh for a resource that has been removed in the real world (aws console for example).


paulocsanz:
bors try

paulocsanz:
bors merge

PR #3885, opened 2024-05-29
fix(dal,web): react to change sets applying to head/current change set

fnichol:
Fixes BUG-257


keeb:
bors merge

keeb:
verified this functionality works as expected

PR #3886, opened 2024-05-30
Bug-287 - Stop duplicating create actions on upgrade

britmyerss:
Only enqueue actions for the upgraded component that were queued for the previous version. Extend a test to make sure we don't break this going forward!


britmyerss:
bors try

PR #3888, opened 2024-05-30
fix: refresh asset pinia store with new variant id

sprutton1:
The asset pinia store was not updating with the newly generated schema variant id, so subsequent updates were all being done on the old version.

Closes https://linear.app/system-initiative/issue/BUG-275/cannot-upgrade-an-asset-twice


sprutton1:
bors try

sprutton1:
bors  r+

sprutton1:
bors cancel

sprutton1:
bors cancel

PR #3890, opened 2024-05-30
Use ConflictsAndUpdates struct instead of tuple ordering for detect_conflicts_and_updates return value

jhelwig:
We had created a `ConflictsAndUpdates` struct for use as a convenience in some tests. This promotes it for general use, instead of relying on having to already know what the order of the tuple elements mean with `detect_conflicts_and_updates`.


jhelwig:
bors merge

PR #3894, opened 2024-05-30
fix(web) - two bugs fixed

wendybujalski:
- DiagramEmptyState under multiplayer cursors
- Delete icon for components/frames now in top right corner instead of big overlay


wendybujalski:
bors merge

PR #3903, opened 2024-05-30
Allow new attribute funcs for output locations with dynamic funcs

nickgerace:
## Description

This PR allows for new attribute funcs to be bound to output locations with pre-existing dynamic funcs, including the identity func.

Here's why this is important: you can neither edit nor remove binding for `si:identity` since we do not allow authoring for intrinsics in the frontend. While going down the path to allowing overwriting `si:identity` usages when creating new funcs for a selected asset (schema with a default schema variant), we realized that we could also allow it for all dynamic functions. This requires existing attribute prototype arguments to be drained as well as enqueing all affected attribute values for a dependent values update job.

## Additional Changes

- Remove duplicate data in create func options for attribute funcs
- Restore and polish tests related to this functionality
- Fix doc comment related to the eventual parents for attribute prototypes
- Only fetch associations if currently editing a func so that we do not overwrite user's mutations in flight (i.e. prevent screen flashing when editing attribute funcs, but also ensuring new associations are populated properly)
- Rename `FETCH_FUNC_DETAILS` to `FETCH_FUNC` and add additional route for fetching associations
- Clarify purpose of reset attribute prototype by its name (though it is still very unclear)

### DIFF
--- a/app/web/src/components/AssetEditorTabs.vue
+++ b/app/web/src/components/AssetEditorTabs.vue
@@ -121,7 +121,7 @@ const closeTab = (slug: string) => {
 };
 
 const loadFuncDetailsReqStatus = funcStore.getRequestStatus(
-  "FETCH_FUNC_DETAILS",
+  "FETCH_FUNC",
   assetStore.urlSelectedFuncId,
 );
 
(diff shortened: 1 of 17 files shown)


nickgerace:
bors merge

PR #3905, opened 2024-05-31
fix: ensure default values are set

sprutton1:
If a an attribute does not exist, fallback to the prop default or null if one does not exist.

closes https://linear.app/system-initiative/issue/BUG-299/key-pair-default-key-type-is-rsa-but-it-does-not-select-as-default-in


sprutton1:
bors try

PR #3908, opened 2024-05-31
BUG-303: no negative sizes

jobelenus:
Never save a dimension lower than the minimum (240px). Never render a dimension lower than the minimum (240px). This ought to prevent any strangeness.


paulocsanz (approved):

vbustamante:
We probably should add those enforcements to the backend too, if they aren't there @jobelenus

jobelenus:
bors merge

PR #3909, opened 2024-05-31
feat(sdf): Don't reparent pasted children

vbustamante:
Make components whose parents were also copied keep parented to the copies of their original parents when pasted inside of frame


paulocsanz (approved):

vbustamante:
bors merge

PR #3910, opened 2024-05-31
feat: adding timeouts to all toasts, removing ack buttons where relevant

sprutton1:
Some of the toasts hangout without a timeout. This turns out to be undesirable in regular use. I enabled the timeouts for moving back to head on apply/abandon, conflicts, and 500s. I also removed the OK button as it doesn't seem relevant anymore.

closes https://linear.app/system-initiative/issue/BUG-293/toast-doesnt-get-automatically-dismissed-after-migrating-away-from


nickgerace (approved):
Love it!

PR #3911, opened 2024-05-31
Use ChangeStatus enum directly in SummaryDiagram... structs

jhelwig:
Rather than stringly-typing the change status, we now directly use the enum that we were previously stringifying.


jhelwig:
bors merge

PR #3913, opened 2024-05-31
feat: display error message and remove timeout for 500s

sprutton1:
Okay, re-removing the timeout for 500s so we count them as bugs and address them. I also added the error and the route to the toast. My styling here is not so great, so maybe @jobelenus or @wendybujalski can school me at some point.

image

### DIFF
--- a/app/web/src/components/toasts/FiveHundredError.vue
+++ b/app/web/src/components/toasts/FiveHundredError.vue
@@ -8,36 +8,38 @@
       />
       <p class="grow py-md">
         The server encountered an error and could not complete your request.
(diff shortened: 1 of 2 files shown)


adamhjk:
What do we think "What can I do?" does? :)

sprutton1:
@adamhjk 

> What do we think "What can I do?" does? :)

I think it's there to try to give the user some resolutions steps. It might be better to just say "Try again or reach out on Discord for help" or something along those lines instead of hiding it behind a button (that seemingly only sometimes works)

sprutton1:
More better? 
image

PR #3914, opened 2024-05-31
feat: only add - Copy once for copied components

sprutton1:
We only want to add "- Copy" to the end of a component once, so if you copy a copy it won't endlessly append "- Copy". Copy?

closes https://linear.app/system-initiative/issue/BUG-280/copying-a-component-which-has-been-copied-appends-an-additional-copy


sprutton1:
bors merge

PR #3915, opened 2024-06-01
Spike: Front end conflicts

jobelenus:
Retry set_component_position with most current data for components that are not in flight. No toasts for this endpoint. For most cases, the exact components that are conflicting are also in flight, which is a no-op, and we swallow the conflict.

Implemented generic retry! The bottom bar bounces!
image

You can see the conflicts and retry:
image

TADA!
image


britmyerss:
bors try

jobelenus:
bors merge

PR #3917, opened 2024-06-02
return error when trying to abandon HEAD change set

britmyerss:
<div><br/>via <a href="https://giphy.com/amandaceemedia/">Amanda Cee Media</a> on <a href="https://giphy.com/gifs/amandaceemedia-i-need-you-clingy-dont-leave-me-2AN7gVc1Az7pmdnFc3">GIPHY</a></div>

### DIFF
--- a/lib/sdf-server/src/server/service/change_set/abandon_change_set.rs
+++ b/lib/sdf-server/src/server/service/change_set/abandon_change_set.rs
@@ -28,10 +28,14 @@ pub async fn abandon_change_set(
     Json(request): Json<AbandonChangeSetRequest>,
 ) -> ChangeSetResult<()> {
     let mut ctx = builder.build_head(access_builder).await?;
-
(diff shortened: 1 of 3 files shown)


lib/sdf-server/src/server/service/change_set/abandon_change_set.rs
```diff
@@ -28,10 +28,14 @@ pub async fn abandon_change_set(
     Json(request): Json<AbandonChangeSetRequest>,
 ) -> ChangeSetResult<()> {
     let mut ctx = builder.build_head(access_builder).await?;
-
+    let maybe_head_changeset = ctx.get_workspace_default_change_set_id().await?;
+    if maybe_head_changeset == ctx.get_workspace_default_change_set_id().await? {
```
stack72, line 32: Won't this always fail?? You are setting the variable to the default changeset if and then checking if it's the default changeset

Don't you want to check against the current changeset??

PR #3919, opened 2024-06-02
feat: upgrade instead of duplicate on import

sprutton1:
Okay, this feels too easy which makes me think I am missing something obvious. Instead of looking at associated schemas, we list all schemas and match by name. If a match is found, we add a schema variant and set it as the default. We also no longer check if the funcs we are importing are marked as a builtin as otherwise we were stomping over the schema variant def func.


sprutton1:
bors try

sprutton1:
bors try

fnichol (approved):
Had a zoom review and it is looking good!

PR #3923, opened 2024-06-03
Delighter: when moving to the diagram from the customizer, load the last component you had selected

jobelenus:


jobelenus:
bors merge

PR #3932, opened 2024-06-04
Fix detached children not having all their values updated

nickgerace:
## Description

This PR fixes two bugs related to detaching children from parent frames. The first is the primary one this PR seeks to solve: detaching frames with deeply nested children can cause some of those deeply nested children to not have their values updated. The second was found and fixed when writing the solution to the first. The second bug involves components that are detached from up frames.

For both of the aforementioned bugs, we were not properly calculating impacted connections during detachment. We need to find all impacted connections for descendants and ascendants of a frame. We also need to find all impacted connections for all ascendants of a component.

## Secondary Changes

- Add an `omega_nesting` integration test that helped find the first bug
- Refactor lego test schemas to consolidate helpers to helper prevent test setup issues
- Fix misc bugs related to lego test schemas (e.g. the "six" output socket was not wired correctly for large lego(s))
- Log errors found when finding available inferred connections to a given input socket match (i.e. we no longer swallow the error)

## Misc Changes

- Add debug implementation to `SocketAttributeValuePair`
- Remove offset related comment from old rebaser architecture
- Replace unneeded mutable destination sockets variable when finding available inferred connections to a given input socket match
- Give functions some breathing room with blank lines


nickgerace:
bors merge

PR #3937, opened 2024-06-04
Use Component ChangeStatus, and NewEdge updates to determine if a socket-to-socket connection is new

jhelwig:
* If either if the Components that an edge is connecting the Output & Input sockets of is new, then the edge itself must be new as well.
* If neither of the Components is new, then we check to see if it's a new `AttributePrototypeArgument` (as determined by a `NewEdge` update where the destination is an `AttributePrototypeArgument`).


jhelwig:
bors merge

PR #3939, opened 2024-06-04
chore(sdf): remove check when cancelling an action

fnichol:
If an action fails to run to completion (i.e. func execution fails, there is a graph conflict, a transient/uncaught database error, etc), it may be left on the graph in a "running" state. Prior to this change we checked the action to see if it's dispatched or running and returned an error as this would imply that our system supports first-class interuptable cancellation of functions (which we do not yet support). This change removes the check and will consequently remove the action from the graph, allowing a user to re-enqueue the action to run again, etc.

Closes BUG-331


fnichol:
bors try

PR #3945, opened 2024-06-05
Make from_base_change_set flag exist for SummaryDiagramComponent & SummaryDiagramEdge

jhelwig:
The logic to drive this flag will come later, but the idea is that when this flag is `true` it is letting the frontend know that this is a "virtual" Component/Edge, and doesn't actually exist in the current change set.


jhelwig:
bors merge

PR #3948, opened 2024-06-05
Restore components & edges from base change set

jhelwig:
# Back end

## Include Components only in the base change set that would be actively removed by a change set when building the summary diagram

We now look at the series of updates that would be applied if we were to merge a change set into its base, and include "virtual" summary diagram component entries for any Components that would be removed and that only exist in the base change set. These virtual entries will have `from_base_change_set` set to true to signal that they don't actually exist in the snapshot for the change set.

## Allow importing a Component from another snapshot graph

This allows us to "undelete"/"restore" a Component from the base change set if we've removed it from the current change set, and change our mind.

## Restore Component->Component connections from base changeset whenever possible

Rather than creating a new API endpoint & method for attempting to restore a `Component -> Component` (`Output Socket -> Input Socket`) connection, `Component::connect` (and by extension `diagram/create_connection`) will now check the base change set to see if an equivalent connection exists there. If it does, then we import that pre-existing connection into the current change set. If it does not, then we'll create the connection as a brand new one.

This means that if you remove a connection, decide you didn’t actually want to do that, and re-draw it instead of clicking the “restore connection” button, we’ll do the right thing and restore it, so you don’t end up seeing a remove+add of the “same” connection, and should generally help reduce the potential churn in looking at diffs.

## Only try to create Component -> Component connections when both are in the current change set

Since we display "virtual" Components of some things that aren't actually in the current change set, we need to make sure that we don't try to create a connection to them as that would effectively be a dangling pointer.

### DIFF
--- a/app/web/src/api/sdf/dal/component.ts
+++ b/app/web/src/api/sdf/dal/component.ts
@@ -52,6 +52,7 @@ export interface RawComponent {
   updatedInfo: ActorAndTimestamp;
   toDelete: boolean;
   canBeUpgraded: boolean;
+  fromBaseChangeSet: boolean;
 }
 
 export type EdgeId = string;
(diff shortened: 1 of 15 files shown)


jhelwig:
bors try

wendybujalski:
bors merge

PR #3952, opened 2024-06-06
fix(dal): retry graph mutations after actions have run

fnichol:
Closes BUG-332

### DIFF
--- a/lib/dal/src/action.rs
+++ b/lib/dal/src/action.rs
@@ -3,12 +3,11 @@ use std::collections::{HashSet, VecDeque};
 use petgraph::prelude::*;
 use postgres_types::{FromSql, ToSql};
 use serde::{Deserialize, Serialize};
-use si_events::{ulid::Ulid, ActionResultState};
+use si_events::ulid::Ulid;
 use si_layer_cache::LayerDbError;
 use strum::{AsRefStr, Display, EnumDiscriminants, EnumIter, EnumString};
 use telemetry::prelude::*;
 use thiserror::Error;
-use veritech_client::{ActionRunResultSuccess, ResourceStatus};
 
 use crate::{
     action::{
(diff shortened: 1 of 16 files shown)


lib/dal/src/action.rs
```diff
@@ -674,34 +614,7 @@ impl Action {
     }
 }
 
-#[derive(Clone, Deserialize, Serialize, Debug, PartialEq, Eq)]
-#[serde(rename_all = "camelCase")]
-pub struct ActionReturn {
```
britmyerss, line 679: No one was listening for these events

lib/dal/src/job/definition/action.rs
```diff
@@ -121,99 +118,185 @@ impl TryFrom<JobInfo> for ActionJob {
...
-async fn process_failed_action(
-    ctx: &DalContext,
-    id: ActionId,
-    _error_message: String,
-) -> JobConsumerResult<()> {
-    info!(%id, "processing action failed");
+async fn process_failed_action(ctx: &DalContext, action_id: ActionId) -> JobConsumerResult<()> {
+    info!(%action_id, "processing action failed");
 
-    let component_id = Action::component_id(ctx, id)
```
britmyerss, line 185: one thing to note - this is our last will and testament if any of the retries/commits/anything fails above. This should be as slim of an update as possible, so we will only mark the Action as failed in this case (letting the user inspect what happened and determine what to do next)

lib/dal/src/job/definition/action.rs
```diff
@@ -121,99 +118,185 @@ impl TryFrom<JobInfo> for ActionJob {
...
 
-    let status = Action::run(ctx, id).await?;
+    let component_id = Action::component_id(ctx, action_id)
+        .await?
+        .ok_or(ActionError::ComponentNotFoundForAction(action_id))?;
+
+    Ok((prototype_id, component_id))
+}
+
+async fn process_and_record_execution(
```
britmyerss, line 179: this encapsulates all of the mutations that need to be made after a Func successfully returns and is therefore what we retry if there are conflicts. Given the scope of what we do after a func returns here (removing action node, updating resource, potentially removing the component from the graph) - we should monitor how often we end up retrying so we aren't hiding a false positive conflict case

lib/dal/tests/integration_test/action.rs
```diff
@@ -158,9 +158,9 @@ async fn run(ctx: &mut DalContext) {
         .await
         .expect("could not commit and update snapshot to visibility");
 
-    assert!(Action::run(ctx, action.id())
+    assert!(ActionPrototype::run(ctx, proto.id(), component.id())
```
vbustamante, line 161: What happens when you run `ActionPrototype::run` and no Action was created previously for that component? I don't see we taking that into account in this run code, but am unsure if that's something we can ignore

vbustamante: This test is failing lint because the action created above is not being used, but it looks to me like we could remove that statement altogether

britmyerss: Nothing should happen, I think we can remove that statement, running the prototype is now the interface for running an action func directly (and all of the graph mutations as a result of that run are processed elsewhere)

fnichol:
bors try

fnichol:
bors merge

PR #3956, opened 2024-06-06
fix(web): Make sure frames can't get smaller than child areas

vbustamante:
Compute frame content boxes on the store so reactivity works


vbustamante:
bors merge

PR #3958, opened 2024-06-07
Jhelwig/replace subgraph interaction with replace references

jhelwig:
# Commits
## Make Update::ReplaceSubgraph only import a single node, instead of the entire subgraph

The problem with importing the entire subgraph is that we also still generate `Update` entries for things below the root of the `Update::ReplaceSubgraph`, and that `find_in_self_or_create_using_onto` + `replace_references` ends up with the uinion of the "child" nodes from   both `self` and `onto` as the child nodes in the resulting graph.

What  we want is either only the child nodes from `onto` with no further `Update` entries for things below the root of the `Update::ReplaceSubgraph`, or for only the specific node referenced in `Update::ReplaceSubgraph` to be replaced, to maintain the pre-existing child nodes in `self`, and to let the `Update` entries for things below the root specified in `Update::ReplaceSubgraph` to handle the rest of the subgraph update/import.

### DIFF
--- a/lib/dal/src/workspace_snapshot/graph.rs
+++ b/lib/dal/src/workspace_snapshot/graph.rs
@@ -735,6 +735,12 @@ impl WorkspaceSnapshotGraph {
                 if edge_kind_updates.additions.len()
                     > (edge_kind_updates.removals.len() + (1 - existing_outgoing_edges_of_kind))
                 {
+                    warn!(
+                        "ExclusiveEdgeMismatch: Found {} pre-existing edges. Requested {} removals, {} additions.",
+                        existing_outgoing_edges_of_kind,
+                        edge_kind_updates.removals.len(),
+                        edge_kind_updates.additions.len(),
+                    );
                     // The net count of outgoing edges of this kind is >1. Consider *ALL* of the
                     // additions to be in conflict.
                     for (
--- a/lib/dal/src/workspace_snapshot/graph.rs
+++ b/lib/dal/src/workspace_snapshot/graph.rs
@@ -1606,6 +1606,36 @@ impl WorkspaceSnapshotGraph {
         algo::has_path_connecting(&self.graph, self.root_index, node, None)
     }
 
+    /// Import a single node from the `other` [`WorkspaceSnapshotGraph`] into `self`, if it does
+    /// not already exist in `self`.
+    pub fn find_in_self_or_import_node_from_other(
+        &mut self,
+        other: &WorkspaceSnapshotGraph,
+        other_node_index: NodeIndex,
+    ) -> WorkspaceSnapshotGraphResult<()> {
+        let node_weight_to_import = other.get_node_weight(other_node_index)?.clone();
+        let node_weight_id = node_weight_to_import.id();
+        let node_weight_lineage_id = node_weight_to_import.lineage_id();
+
+        if let Some(equivalent_node_index) =
+            self.find_equivalent_node(node_weight_id, node_weight_lineage_id)?
+        {
+            let equivalent_node_weight = self.get_node_weight(equivalent_node_index)?;
+            if !equivalent_node_weight
+                .vector_clock_write()
+                .is_newer_than(node_weight_to_import.vector_clock_write())
+            {
+                self.add_node(node_weight_to_import)?;
+
+                self.replace_references(equivalent_node_index)?;
+            }
+        } else {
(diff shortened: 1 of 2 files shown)


lib/dal/src/workspace_snapshot/graph.rs
```diff
@@ -1600,6 +1606,36 @@ impl WorkspaceSnapshotGraph {
         algo::has_path_connecting(&self.graph, self.root_index, node, None)
     }
 
+    /// Import a single node from the `other` [`WorkspaceSnapshotGraph`] into `self`, if it does
+    /// not already exist in `self`.
+    pub fn find_in_self_or_import_node_from_other(
```
zacharyhamm, line 1611: In the `Update::NewEdge` case, we call `find_in_self_or_create_using_onto` which calls `import_subgraph` even if the node exists in both graphs. I'm wondering if we should use this new method in the case that we find an equivalent node in other?

lib/dal/src/workspace_snapshot/update.rs
```diff
@@ -21,6 +21,8 @@ pub enum Update {
         destination: NodeInformation,
         edge_kind: EdgeWeightKindDiscriminants,
     },
+    // This is not correctly named. We really only want to replace the single node, as we also
+    // generate Update entries to handle processing the rest of the subgraph.
     ReplaceSubgraph {
```
zacharyhamm, line 26: I *believe* that postcard doesn't care about variant names, just order, so it might be safe to just rename this (I'd want to confirm that first, though)

jhelwig: It doesn't. It only cares about index/order, but I also wasn't 100% confident that *everything* was *always* going through postcard.

zacharyhamm:
This change makes sense to me, and I think we can do even more by restricting `import_subgraph` to the case where there's a NewEdge update to a node that does not exist in to_rebase. The only question I have is whether this actually fixes the conflict we saw in the action run. Theoretically it should be possible to produce an incorrect ExclusiveEdgeMismatch in the old code, right, if that was the problem?

lib/dal/src/workspace_snapshot/graph.rs
jhelwig: So, walking things through:
* `detect_confilcts_and_updates` walks the `onto` graph as its basis for detecting things.
* If we don't find any candidates for equivalency for a particular node in `onto` from the nodes in `to_rebase`, then we do nothing, and prune the walk of the `onto` graph at that point.
* When processing a container in `find_container_membership_conflicts_and_updates` (processing a/the parent of the potentially new node).
* There is an entry in `only_onto_edges` for the edge pointing to a node. (May or may not be to a new node)
    * We generate an `Update::NewEdge`
* If the node exists already in `to_rebase` (not new)
  * When processing the DFS event for the node, we continue walking the `onto` graph through this node as long as we continue to find equivalent nodes in `to_rebase`, generating `Update`s as necessary.
* If the node does _not_ already exist in `to_rebase` (it's new)
  *  When processing the DFS event for the node, we prune the `onto` graph walk as we could not find an equivalent node, and do not generate any `Update` for the node that only exists in `onto`.

I that when we `Update:NewEdge` in `perform_updates` that it should be a no-op if we find an equivalent node in`to_rebase` as far as importing/updating nodes goes, and that if there is no equivalent node in `to_rebase` then we want to import the "new" bits of the subgraph.

So, I don't think the new `find_in_self_or_import_node_from_other` is quite right there as it won't import the rest of the new subgraph (only the root of it), but I think you're correct that `find_in_self_or_create_using_onto` isn't quite right either as it still tries to import the entire sub-graph from `onto` in the case where it did find an equivalent pre-existing node in`to_rebase`.

jhelwig:
It doesn't look like it was an incorrect `ExclusiveEdgeMismatch`. The thing this PR is addressing is the potential to incorrectly generate a graph that violates exclusive edge constraints, because `ReplaceSubgraph` is not actually doing a *replace*, but it is doing "use the node from `onto`, and the outgoing edges from *both* `onto` and `to_rebase`".

jhelwig:
Looks like after changing `Update::AddEdge` to use a new `reuse_from_self_or_import_subgraph_from_onto`, that `find_in_self_or_create_using_onto` isn't used anywhere anymore.

jhelwig:
bors merge

PR #3960, opened 2024-06-07
Delighter: ability to search the debug view

jobelenus:
Due to how ... involved the layout of debug is, I went old school. Search the HTML and add a yellow highlight to what we find. Scroll to the first find.

image


jobelenus:
bors merge

PR #3962, opened 2024-06-07
fix(web) - many UI improvements and fixes!

wendybujalski:
- fixed loading states in the customize screen
- changed how assets with duplicate names are handled
- fixed destructive color styles across the whole UI
- fixed selection states for DetailsPanelMenuIcon
- changed Asset Functions attach function button to match IconButtons above
- added tooltip and hover state for upgrade button and other buttons on DiagramOutlineNode and ComponentCard
- "change set" is two words
- made a reference for semantic sizes and colors in the code base
- updated the SiSearch to match the better style Mark designed but we didn't get time for, more filters


wendybujalski:
bors merge

PR #3963, opened 2024-06-08
feat(si-generator): Add support for setting missingResource refresh settings

stack72:
If a resource has been removed directly in AWS, we need to ensure that our refresh funcs check for specific 404s that we know mean it's been deleted so we can remove the resource locally

We can now do that:

```
si-generator  refresh eks describe-cluster --input clusterName:properties.resource.payload.clusterName --missingResource "ResourceNotFoundException" --output 'cluster'
```

This will print the following check:

```
  if (child.exitCode !== 0) {
    const payload = _.get(component, 'properties.resource.payload');
    if (child.stderr.includes("ResourceNotFoundException") {
      return {
        status: "error",
        message: `Refresh error; resource not found; (ResourceNotFoundException)`,
      }
    }
```

That will remove the resource from the system. We can pass multiple `missingResource` options and each of them will be added as specific checks

### DIFF
--- a/bin/si-generator/src/render.ts
+++ b/bin/si-generator/src/render.ts
@@ -61,7 +61,10 @@ export async function fmt(ts: string): Promise<string> {
   }
 }
 
-export async function renderAsset(props: Array<Prop>, provider: RenderProvider): Promise<string> {
+export async function renderAsset(
+  props: Array<Prop>,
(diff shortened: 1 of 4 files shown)


nickgerace (approved):

nickgerace:
bors merge

PR #3964, opened 2024-06-10
Enhance `dal-test` helpers and schemas (ENG-2576)

nickgerace:
## Description

This PR aims to enhance `dal-test` helpers and test exclusive schemas. As a result of these changes, panics of any kind are banned by default outside of macro expansion helpers. Now, when authoring test exclusive schemas or authoring tests, error reporting should be clearer than before.

For the helpers, we leverage `color_eyre` via its `Result` type and `eyre!` macro in order to make error propagation better in tests via explicit `expect` statements and other failures. Inner panics are now disallowed in helpers themselves in order to give better error location reporting when running tests via `buck2` with default settings.

For the test exclusive schemas, everything has been refactored to be more explicit about how schemas are organized and where migrations are centralized. These changes reflect the shape of the schemas back in the old engine when they were in the dal. In addition to these changes, the helpers for test exclusive schemas have been polished, including some unnecessary usages of async functions and replaced panics. These helpers have been moved to the central module file and now, the only public, exported function is the `migrate` function. Finally, panics within test exclusive schema scaffolding have also been replaced with proper error propagation.

As a result of these changes, both `DalTestHelpersError` and `ChangeSetTestHelpers` have been removed alongside their result types.

## Example Test Failure

image

## Secondary Changes

- Fix swallowed errors when setting types in frame integration tests
- Split property editor test view into its own submodule
- Move auth token creation into expand helpers (moved from the regular test helpers)
- Temporarily disable `delete_frame_with_child_with_resource` test due to intermittent failures


lib/dal/tests/integration_test/frame.rs
```diff
@@ -556,14 +581,22 @@ async fn output_sockets_can_have_both(ctx: &mut DalContext) {
...
+    let even_frame = create_component_for_schema_name(ctx, "large even lego", "even")
+        .await
+        .expect("could not create component");
     let even_frame_component_id = even_frame.id();
-    let _ = even_frame
+    even_frame
         .set_type(ctx, ComponentType::ConfigurationFrameDown)
-        .await;
+        .await
+        .expect("could not set type");
```
nickgerace, line 591: Now, we no longer swallow errors when setting type in frame tests. This is one example.

PR #3968, opened 2024-06-11
Front-end otel sending to honeycomb via collector

jobelenus:
Let's see all the things!


dev/docker-compose.platform.yml
```diff
@@ -146,13 +146,14 @@ services:
 
   otelcol:
     #<<: *loki-logged-service
-    image: systeminit/otelcol:stable
+    image: systeminit/otelcol:20240610.210148.0-sha.143e17302-arm64v8
```
sprutton1, line 149: change this back to stable

jobelenus: ah of course

PR #3971, opened 2024-06-11
Add regression test for existing code gen funcs and new variants

nickgerace:
## Description

This PR adds a regression test for attaching existing code gen funcs to new schema variants. Specifically, the "attaching" step works, but when the "Secrets" input is added, sometimes we are unable to regenerate the variant due the following error: `pkg error: Can't convert identity to LeafInputLocation`.

The bug appears to no longer exist on main, but the test should still be valuable in catching a future regression.

## Secondary Changes

- Rename `find_by_name` to `find_id_by_name` when looking for funcs by name

### DIFF
--- a/lib/dal/src/func.rs
+++ b/lib/dal/src/func.rs
@@ -313,7 +313,7 @@ impl Func {
         Ok(Self::assemble(func_node_weight, &inner))
     }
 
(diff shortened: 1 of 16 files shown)


nickgerace:
bors merge

PR #3973, opened 2024-06-12
fix: list variant ids should only return default ids

sprutton1:
The `SchemaVariant::list_ids` func was doing just that, listing all of the variant ids, when we really want the latest default. This was causing inconsistency in api returns as we would often smash results into a hashmap and the default could have been stepped on in the process.

### DIFF
--- a/lib/dal/src/func/runner.rs
+++ b/lib/dal/src/func/runner.rs
@@ -1138,7 +1138,7 @@ impl FuncRunner {
             .value;
 
         let mut auth_funcs = vec![];
-        for secret_defining_sv_id in SchemaVariant::list_ids(ctx).await? {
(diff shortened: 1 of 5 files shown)


sprutton1:
bors try

zacharyhamm:
There should only ever be one default id. So maybe all of these should be changed to just get the default and error if there is none?

zacharyhamm:
Oh nvm. I see how this works. This is for all SVs in the workspace. Nevermind!

nickgerace (approved):

zacharyhamm (approved):
🚢 🇮🇹 
We may need to list all ids, not just defaults, at some other time, but that's another castle

sprutton1:
bors merge

PR #3978, opened 2024-06-13
fix: web otel adjustments

sprutton1:
Minor adjustments to web otel setup, adding health check endpoint for otel container


sprutton1:
bors try

PR #3981, opened 2024-06-14
Fix: when abandoning a changeset, if you've selected an asset or...

jobelenus:
module, it could be gone in the new changeset, so navigate to base authoring url!


jobelenus:
bors merge

PR #3983, opened 2024-06-14
Ensure user has access to requested Workspace & requested Change Set is part of that Workspace

jhelwig:


fnichol (approved):
As the fix is in `DalContextBuilder.build()` which is called in every SDF handler requiring a `DalContext` (let's be honest--that's virtually every handler), this helps to ensure that the check is run everywhere it needs to. Nice!

fnichol:
bors merge

lib/dal/src/context.rs
```diff
@@ -1054,6 +1056,28 @@ impl DalContextBuilder {
             no_auto_migrate_snapshots: self.no_auto_migrate_snapshots,
         };
 
+        if ctx.history_actor() != &HistoryActor::SystemInit {
```
adamhjk, line 1059: Are we worried about forging the history actor?

PR #3986, opened 2024-06-15
Frames fixes, features, and more precision when enqueuing DVU

britmyerss:
This PR does the following: 

**Add precision to which values need to rerun when upserting/orphaning components by caching the state of inferred connections and only rerunning what has been deleted/added/changed**
- When changes are made to a component tree, we now correctly capture the initial, in-between (in the case of upsert), and final state of the tree in terms of all of the various inferred connections. 
- Based on the deltas of state as we make changes to the tree, we calculate which input sockets have new/different/changed inferred connections to output sockets and enqueue them in DVU (forcing them to rerun)
- For example, with a deep-ish nested tree, if you orphan a frame in the middle, the tree is split into two, and we need to walk both trees to find what needs to be rerun, as the orphaned frame might have had children that were taking inputs from an ancestor above the direct parent/child connection that was severed
- Got schooled/saved by @nickgerace on some rust

**Fix bug with nested propagation through up frames**
- We were incorrectly failing to proceed down to an up-frame's children if we're looking for which input sockets a given down frame is driving 
- For example, if a Down Frame -contains-> Up Frame -contains-> Component and the Component is taking a value from the Down Frame, when you change that value on the Down Frame, we were never finding the Component and you'd get left with stale data

**Enable up frames to take inputs from parent down frames**
- When discussing with @keeb - this is highly desired and comes up a lot as you rarely have an Up Frame without a Down Frame in its ancestry
- This enables, for example, a VPC Up Frame inside of a Region/Credential to infer the region/credential values, where previously we were not looking and forced the user to draw edges 
- This follows the same general intent as before - if there's ambiguity (for example, the VPC Up Frame is in a Region Down Frame, and also has a Region Component inside of it) - we don't infer any connection and force the user to resolve by drawing an edge

**Adds some telemetry that was useful while figuring this out** 
- connecting child/parent spans in Pinga/DVU so we see more of a connected waterfall in Honeycomb/Grafana
- Adding workspace and changeset ids to DVU jobs (to make it easier to query by when debugging specific issues)

**Tests!**
- Add a test for the new functionality for up frames
- Add a test to increase confidence that we're handling nested frames correctly on orphaning

<div><br/>via <a href="https://giphy.com/grantkolton/">grantkoltoons</a> on <a href="https://giphy.com/gifs/2d-get-back-up-again-i-knocked-down-6iy6THGgvykxN0IA65">GIPHY</a></div>

### DIFF
--- a/lib/dal-test/src/helpers.rs
+++ b/lib/dal-test/src/helpers.rs
@@ -4,8 +4,8 @@ use color_eyre::eyre::eyre;
 use color_eyre::Result;
 use dal::key_pair::KeyPairPk;
 use dal::{
-    AttributeValue, Component, ComponentId, DalContext, InputSocket, KeyPair, OutputSocket, Schema,
-    SchemaVariant, SchemaVariantId, User, UserPk,
+    AttributeValue, Component, ComponentId, ComponentType, DalContext, InputSocket, KeyPair,
+    OutputSocket, Schema, SchemaVariant, SchemaVariantId, User, UserPk,
 };
 use names::{Generator, Name};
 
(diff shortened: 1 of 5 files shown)


nickgerace:
bors try

nickgerace:
bors merge

PR #3988, opened 2024-06-17
ENG-2572: VectorClockIds are a tuple. Editing changesets are gone

zacharyhamm:
This change mitigates the unbounded growth of vector clocks by using the change set id + the actor id as the vector clock id. Ephemeral "editing change sets" are gone, although system services get a randomly generated actor id for the duration of their DalContext. It also prunes more vector clocks, now including write clocks, on rebase, ensuring our vector clock entries remain small.

WorkspaceSnapshot method signatures have been refactored to take the vector clock id directly, instead of assuming the change set and vector clock id are the same.

All snapshots for change sets that are not applied will be migrated on SDF boot, so deploying this to prod may cause some downtime.


zacharyhamm:
bors try

stack72:
bors try

zacharyhamm:
bors try

PR #3989, opened 2024-06-17
Show all funcs for all versions of an asset and do not clone funcs on regeneration

nickgerace:
## Description

This PR contains a bunch of changes related to authoring schemas (assets), schema variants (asset versions) and funcs involved when authoring them. Primarily, this commit is focused on showing all funcs for all versions of an asset to ensure the user has all the data they need for authoring.

When you regenerate an asset when there are components on the diagram using it, a new version of the asset is created (a new schema variant). As a result, the default schema variant changes for the asset (schema). If you were to edit the output location or change anything that would affect the bindings the asset, upon regeneration, you could lose access
to the func.

_Here's what we do to solve that problem, among others that are downstream to that problem:_

- List ALL funcs in existence when asked, rather than just those that belong to assets (i.e. allow users to attach funcs that are free floating and aren't attached to any assets)
- List all funcs for all versions of an asset in the "ASSET FUNCTIONS" section (i.e. if a component is using an older version of an asset and that older version is using a func that is no longer in use, you can now access that func)
- Do not clone funcs on regeneration since the user should be explicit about changing the func to something else

### DIFF
--- a/app/web/src/store/asset.store.ts
+++ b/app/web/src/store/asset.store.ts
@@ -375,23 +375,14 @@ export const useAssetStore = () => {
             changeSetsStore.creatingChangeSet = true;
 
           this.detachmentWarnings = [];
-
           const asset = this.assetsById[assetId];
-          return new ApiRequest<
-            {
-              taskId: string;
-            },
-            AssetSaveRequest
-          >({
+
+          return new ApiRequest<null>({
             method: "post",
             url: "/variant/update_variant",
-            keyRequestStatusBy: assetId,
             params: {
               ...visibility,
               ..._.omit(asset, ["hasComponents", "createdAt", "updatedAt"]),
-            },
-            onSuccess: (response) => {
-              this.executeAssetTaskId = response.taskId;
             },
           });
         },
--- a/lib/dal/src/func.rs
+++ b/lib/dal/src/func.rs
@@ -524,7 +524,6 @@ impl Func {
     pub async fn list(ctx: &DalContext) -> FuncResult<Vec<Self>> {
         let workspace_snapshot = ctx.workspace_snapshot()?;
 
-        let mut funcs = vec![];
         let func_category_id = workspace_snapshot
             .get_category_node_or_err(None, CategoryNodeKind::Func)
             .await?;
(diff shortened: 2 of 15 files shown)


app/web/src/store/asset.store.ts
```diff
@@ -375,24 +375,15 @@ export const useAssetStore = () => {
...
-
           const asset = this.assetsById[assetId];
-          return new ApiRequest<
-            {
-              taskId: string;
-            },
-            AssetSaveRequest
-          >({
+
+          return new ApiRequest<null>({
```
nickgerace, line 380: Handled by the WsEvent and we were not using the returned data anyway.

lib/dal/src/func.rs
```diff
@@ -536,23 +535,50 @@ impl Func {
...
-            func_content_hash.push(node_weight.content_hash());
+            func_content_hashes.push(node_weight.content_hash());
             func_node_weights.push(node_weight);
         }
 
+        Self::list_inner(ctx, func_node_weights, func_content_hashes).await
+    }
+
+    /// List all [`Funcs`](Func) corresponding to the provided [`FuncIds`](Func).
+    pub async fn list_from_ids(ctx: &DalContext, func_ids: &[FuncId]) -> FuncResult<Vec<Self>> {
```
nickgerace, line 553: Break this part out for optimization so that other functions from other structs that collect `FuncIds` can use it.

lib/dal/src/func/summary.rs
```diff
@@ -32,49 +32,35 @@ pub struct FuncSummary {
...
 
 impl FuncSummary {
-    /// Returns the [summaries](FuncSummary) for all [`Funcs`](Func) in the [`ChangeSet`](crate::ChangeSet).
+    /// Returns the [summaries](FuncSummary) for all [`Funcs`](Func) in the workspace.
+    ///
+    /// This includes [`Funcs`](Func) that are "free floating" and not used by any [`Schema`] at present.
     pub async fn list(ctx: &DalContext) -> FuncSummaryResult<Vec<Self>> {
-        Self::list_inner(ctx, None).await
+        let funcs = Func::list(ctx).await?;
+        Ok(Self::from_funcs(funcs.as_slice()))
```
nickgerace, line 40: Now, we collect all funcs in the workspace in the frontend rather than those are used by at least one schema.

lib/dal/src/schema.rs
```diff
@@ -415,4 +450,27 @@ impl Schema {
...
+        Ok(func_ids)
+    }
+
+    /// Collect all [`Funcs`](crate::Func) corresponding to the provided [`SchemaId`](Schema).
+    pub async fn all_funcs(ctx: &DalContext, id: SchemaId) -> SchemaResult<Vec<Func>> {
+        let func_id_set = Self::all_func_ids(ctx, id).await?;
+        let func_ids = Vec::from_iter(func_id_set.into_iter());
+        let funcs = Func::list_from_ids(ctx, func_ids.as_slice()).await?;
+        Ok(funcs)
+    }
```
nickgerace, line 475: This allows us to get all funcs for all versions of an asset. This allows us to show all funcs for an asset in the "ASSET FUNCTIONS" section of the frontend.

nickgerace:
bors merge

PR #3990, opened 2024-06-18
feat: health check pool on Veritech startup

sprutton1:
This runs a quick loop of whatever instance type is selected for Cyclone when starting Veritech. The idea is to catch faulty jail implementations at startup so we can report the service as unhealthy if it can't run functions


sprutton1:
bors try

sprutton1:
bors try

sprutton1:
bors try

sprutton1:
bors merge

PR #3992, opened 2024-06-18
feat: add toolbox for ops helper work

sprutton1:
This creates a little helper container we can use to ease the burden of troubleshooting issues in tools/prod. This is the skeleton for the functionality plus a script for ssh'ing into instances in AWS. You can invoke it via `./components/toolbox/awsi.sh ssm -p [aws_profile] -r [aws_region]`. It will give you list of available instances you can shell into and then open up a session with whatever box you select.

### DIFF
--- a/component/toolbox/BUCK
+++ b/component/toolbox/BUCK
@@ -1,0 +1,49 @@
+load(
+    "@prelude-si//:macros.bzl",
+    "docker_image",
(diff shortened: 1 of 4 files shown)


sprutton1:
bors merge

sprutton1:
bors merge

PR #3993, opened 2024-06-18
fix(dal): pick correct ordering node in ReplaceSubgraph

zacharyhamm:
If an ordered node's children are replaced, and the RemoveEdge operations come before the ReplaceSubgraph update for the ordering node, the ordering node's vector clock write will have been incremented *during the update*, which leads us to pick the wrong node, with the result that the ordering node now has incorrect children in its order property. This fixes that by not incrementing the vector clock write for the ordering node when processing a RemoveEdge update (and *only* when processing a RemoveEdge update).

Discovery of this bug came from the VectorClockId tuple work, but this bug exists on main.

### DIFF
--- a/lib/dal/src/workspace_snapshot/graph.rs
+++ b/lib/dal/src/workspace_snapshot/graph.rs
@@ -501,7 +501,7 @@ impl WorkspaceSnapshotGraph {
 
                 let mut new_container_ordering_node_weight =
                     previous_container_ordering_node_weight.clone();
-                new_container_ordering_node_weight.push_to_order(vector_clock_id, element_id)?;
+                new_container_ordering_node_weight.push_to_order(vector_clock_id, element_id);
                 self.add_node(NodeWeight::Ordering(new_container_ordering_node_weight))?;
                 self.replace_references(container_ordering_node_index)?;
             }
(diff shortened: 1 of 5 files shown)


zacharyhamm:
bors try

lib/dal/src/workspace_snapshot/graph.rs
```diff
@@ -2128,11 +2150,15 @@ impl WorkspaceSnapshotGraph {
...
-                    self.remove_edge(
+                    self.remove_edge_inner(
                         to_rebase_change_set,
                         updated_source,
                         destination,
                         *edge_kind,
+                        // Updating the vector clocks here may cause us to pick
+                        // the to_rebase node incorrectly in ReplaceSubgraph, if
+                        // ReplaceSubgraph comes after this RemoveEdge
+                        false,
```
zacharyhamm, line 2161: changing this to true will cause the new test to fail

zacharyhamm: It should be safe to not update the ordering node vector clock here, since if we get a RemoveEdge update for an ordered container, the ordering node's node hash will have changed, and thus we should always get a ReplaceSubgraph update for that ordering node (unless we produce conflicts instead)

PR #3994, opened 2024-06-18
Authoring Contribute & Update

jobelenus:
No feature flag needed, since both flags are truthy, and when they are false—nothing about the status quo changes!

Notables:
- added iconography for both actions
- reorganized some layout to support new behaviors
- rebuilt the asset router & store for selected assets & funcs (removed a ton of confusing watchers throughout different components)... **_this means the URLs of the authoring pages are different now_**!
- can now select multiple assets with shift key
- assorted fixes to stores, tab group, et al


app/web/src/components/AssetEditorTabs.vue
```diff
@@ -94,43 +93,41 @@ watch(
       // updating tab list
       currentTabs.value = tabs;
     }
+
+    if (funcStore.selectedFuncId)
+      tabGroupRef.value?.selectTab(funcStore.selectedFuncId);
+
+    // still dont know what is racing and removing the querystring from the URL
```
jobelenus, line 100: @wendybujalski this is what I've done for the moment...

app/web/src/components/FuncEditor/FuncList.vue
```diff
@@ -1,17 +1,14 @@
...
       enableDefaultHoverClasses
       enableGroupToggle
       alwaysShowArrow
       indentationSize="none"
       leftBorderSize="none"
-      :defaultOpen="
-        (defaultOpen || (index === 0 && firstOpen)) &&
-        !(funcsForKind(kind).length <= 0)
-      "
+      defaultOpen
```
wendybujalski, line 11: Do we not want func types which have zero children to be collapsed by default anymore?

app/web/src/components/AssetEditorTabs.vue
wendybujalski: spin-shrug

wendybujalski (approved):

app/web/src/components/FuncEditor/FuncList.vue
jobelenus: So.. thats less the issue than `defaultOpen` can't really be dynamic... because its the default... if the prop changes to `true` it doesn't open the children... which is sad ;(

jobelenus:
bors merge

app/web/src/components/FuncEditor/FuncList.vue
wendybujalski: Ah I see! Well there's probably a better way to do this behavior anyway.

PR #3995, opened 2024-06-19
fix(dal): Only create attribute prototype arguments when configuring the binding, wipe existing attribute prototype arguments when resetting the attribute prototype for schema variants

britmyerss:
In the short term, when the user detaches a func, if the func was driving a socket which was previously taking it's input from another prop (as defined in the schema variant definition), they'll need to regenerate the asset to get this functionality back. If the func was driving a prop, it'll be manually editable in the component or they can choose/create a new attribute func to feed it. 

As the half-configured attribute prototype bag was driving some of the UI, I needed to also map the bag to the existing func args available, otherwise they never showed up as we no longer had a placeholder AttributePrototypeArgument that was only waiting for a path.


stack72:
bors merge

PR #3998, opened 2024-06-19
fix(dal, web): Ensure attribute functions can't set incorrect output locations

stack72:
Without this, people can set root or root/secrets - which is the dangerous props people can set

We now made it:

/root/resource_value/*
/root/domain/*
/root/si/color


stack72:
bors merge

PR #4000, opened 2024-06-20
fix(dal, sdf): Ensure we can save changes to variant details without regenerating

stack72:


stack72:
bors merge

app/web/src/store/asset.store.ts
```diff
@@ -455,6 +455,7 @@ export const useAssetStore = () => {
           return new ApiRequest<null>({
             method: "post",
             url: "/variant/update_variant",
+            keyRequestStatusBy: assetId,
```
nickgerace, line 458: Yep! This is it @stack72

PR #4001, opened 2024-06-20
fix(web): Remove Asset.Store SchemaVariantSaved Event

stack72:
We are continually reloading the page when this WsEvent fires. The variant autosaves every few seconds so we don't need to reload everything every few seconds


stack72:
bors merge

PR #4002, opened 2024-06-20
fix(web): Remove the disabled state of the regenerate asset button

stack72:
We need to be able to regenerate the asset at any time


stack72:
bors merge

PR #4003, opened 2024-06-20
fix(web): Reset Attribute Func Attachment between attribute funcs

stack72:
We need to reset the output location when we have pop the modal for attaching new attribute funcs


stack72:
bors merge

PR #4007, opened 2024-06-20
feat: add workspace token ULID

sprutton1:
This allows the auth-api to generate new ulid on create -- the unique workspace token. The idea here is that we want some workspace-specific value to exist that can act as a protected value. It can then be used to integrate with things like AWS assume-role to identity a workspace. We should not use the existing workspace ID for this as it lives in the URL, so is trivial to leak. 

Caveats:
1. This does not populate the existing workspaces in the auth-api with new tokens. Prisma does not provide a way to do this via migrations for ULIDs (that I could find)
2. This does not does populate the existing workspaces in the SI database either. However, if a workspace is retrieved from the auth-api and it contains a token, the workspace record will be updated

So, if we write a script to update the auth-api with new tokens, the SI workspaces will eventually become consistent. Or we could just update those, too. 

Follow on work is to update the auth-api database and then  decide how to thread this token value through to SecretDefinitions.

### DIFF
--- a/app/web/src/api/sdf/dal/workspace.ts
+++ b/app/web/src/api/sdf/dal/workspace.ts
@@ -3,4 +3,5 @@ export interface Workspace {
   name: string;
   created_at: IsoDateString;
   updated_at: IsoDateString;
+  token: string;
 }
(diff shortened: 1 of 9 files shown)


sprutton1:
bors try

PR #4017, opened 2024-06-21
feat: implicitly pass workspace token to all secrets

sprutton1:
When we go gather the before funcs we momentarily decrypt them before re-encrypting to pass them to Veritech. Now, we will inject the workspace token as an arg into the before func before re-encrpyting. This means:

* workspace token will be available on all secrets as `secret.WorkspaceToken`
* it will be correctly redacted as it is just another arg on the before_func

This method is probably not ideal for when we want to add more values, but I think that there is a pretty large conversation around the most right way to do this that we can save for a later date.

The AWS Cred will need to remove the WorkspaceId and use this instead. Here's it working:

image

Something about the test setup meant that I needed to manually set the token before the functions that would use it check for it. Given that the token is currently coming from the auth-api, this fairly represents the real world, but smells funny to me.

### DIFF
--- a/lib/dal-test/src/test_exclusive_schemas/fallout.rs
+++ b/lib/dal-test/src/test_exclusive_schemas/fallout.rs
@@ -232,8 +232,9 @@ fn action_funcs() -> BuiltinsResult<(FuncSpec, FuncSpec)> {
     // Add the action create func.
     let code = "async function main() {
         const authCheck = requestStorage.getItem('dummySecretString');
-        if (authCheck) {
-            if (authCheck === 'todd') {
+        const workspaceToken = requestStorage.getItem('workspaceToken');
+        if (authCheck && workspaceToken) {
+            if (authCheck === 'todd' && workspaceToken === 'token') {
                 return {
                     status: 'ok',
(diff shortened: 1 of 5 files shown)


sprutton1:
bors try

lib/dal-test/src/test_exclusive_schemas/fallout.rs
```diff
@@ -232,13 +232,14 @@ fn action_funcs() -> BuiltinsResult<(FuncSpec, FuncSpec)> {
...
-        if (authCheck) {
-            if (authCheck === 'todd') {
+        const workspaceToken = requestStorage.getItem('workspaceToken');
+        if (authCheck && workspaceToken) {
+            if (authCheck === 'todd' && workspaceToken === 'token') {
                 return {
                     status: 'ok',
                     payload: {
                         'poop': true
-                    }
```
nickgerace, line 241: @sprutton1 do we need this still or is GitHub messing with me?

nickgerace (approved):

sprutton1:
bors try

PR #4018, opened 2024-06-21
fix(dal): calc prototypes and values correctly on component upgrade

zacharyhamm:
Walk the attribute value tree of the new component and find the values in the old component that match in the new, setting them when appropriate. Also detect if component specific prototypes were configured on the old component, and attempt to configure them here.

Closes BUG-405


zacharyhamm:
bors try

PR #4019, opened 2024-06-22
fix: ignore stdin in execa

sprutton1:
This stops execa from waiting for stdin, so we no longer just hang forever if someone writes a function that does something cool like `cat` with no input.

This also moves the execaOptions out of siExec because letting people configure this inside of functions seems rife for mischief.


sprutton1:
bors merge

PR #4020, opened 2024-06-22
chore: firecracker scripts as bytes instead of from repo

sprutton1:
This moves the firecracker scripts into the firecracker library so we no longer go up to github to get them. This means you can make changes to them in a branch without having to muck with having to change which branch they are pulled from in the repo. Also some refactoring as I learn more about how to handle errors correctly.


sprutton1:
bors try

sprutton1:
bors try

PR #4023, opened 2024-06-22
fix(web, sdf, dal): Return created variant details from create_variant

stack72:
When on HEAD and creating a new variant, we were trying to load the variant on HEAD. This returns the correct information in the WsEvent to update the store and waiting for the new changeset URL to be set before trying to load the variant


stack72:
bors merge

PR #4027, opened 2024-06-22
Ensure we upsert leaf funcs to the correct attribute value for existing components

britmyerss:
When we author a new code gen or qualification, we find all existing components created with that schema variant and add the func to them. When doing this, we were incorrectly mapping new code gen functions to the qualification attribute value, now we find the correct attribute value for either the code gen or qualification attribute value and create the new attribute value for the new func there!

Also added a test!

Note: This is broken on upgrade until we fix BUG-405, meaning if you have a component on the diagram, regenerate the schema variant, add a new qualification or code gen function to the new schema variant, when you upgrade the component, the new qualification/code gen func won't be attached. Zack is fixing this in 405

<div><br/>via <a href="https://giphy.com/abcnetwork/">ABC Network</a> on <a href="https://giphy.com/gifs/thegoldbergs-season-5-episode-1-xUNd9YJwF6ifDUnqNi">GIPHY</a></div>

### DIFF
--- a/lib/dal/src/schema/variant.rs
+++ b/lib/dal/src/schema/variant.rs
@@ -1116,13 +1116,18 @@ impl SchemaVariant {
                             .await?;
 
                     // Before returning the new prototype, we need to ensure all existing components use the new
-                    // qualification.
+                    // leaf prop (either qualification or code gen).
                     let mut new_attribute_value_ids = Vec::new();
                     for component_id in Self::list_component_ids(ctx, schema_variant_id).await? {
                         let parent_attribute_value_id =
-                            Component::find_qualification_map_attribute_value_id(ctx, component_id)
-                                .await
-                                .map_err(Box::new)?;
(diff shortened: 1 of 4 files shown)


lib/dal/src/schema/variant.rs
```diff
@@ -1116,13 +1116,18 @@ impl SchemaVariant {
                             .await?;
 
                     // Before returning the new prototype, we need to ensure all existing components use the new
-                    // qualification.
+                    // leaf prop (either qualification or code gen).
                     let mut new_attribute_value_ids = Vec::new();
                     for component_id in Self::list_component_ids(ctx, schema_variant_id).await? {
                         let parent_attribute_value_id =
-                            Component::find_qualification_map_attribute_value_id(ctx, component_id)
```
britmyerss, line 1123: the bug!

nickgerace: Love it!

PR #4034, opened 2024-06-24
Clone or New Asset: provide an initial schema name in a modal

jobelenus:
This is in anticipation of making the schema name immutable after creation.

### DIFF
--- a/lib/dal/src/schema/variant/authoring.rs
+++ b/lib/dal/src/schema/variant/authoring.rs
@@ -20,9 +20,8 @@ use crate::pkg::import::import_only_new_funcs;
 use crate::pkg::{import_pkg_from_pkg, PkgError};
 use crate::schema::variant::{SchemaVariantJson, SchemaVariantMetadataJson};
 use crate::{
-    generate_unique_id, pkg, ComponentType, DalContext, Func, FuncBackendKind,
-    FuncBackendResponseType, FuncError, FuncId, Schema, SchemaError, SchemaVariant,
-    SchemaVariantError, SchemaVariantId,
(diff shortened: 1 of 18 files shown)


lib/dal/src/schema/variant/authoring.rs
```diff
@@ -350,10 +348,10 @@ impl VariantAuthoringClient {
             )
             .await?;
 
-        let schema_spec = metadata.to_spec(new_variant_spec)?;
+        let schema_spec = metadata.to_schema_spec(new_variant_spec)?;
         //TODO @stack72 - figure out how we get the current user in this!
         let pkg_spec = PkgSpec::builder()
-            .name(&metadata.name)
+            .name(&metadata.schema_name)
```
jobelenus, line 354: this is the value from line 332 `name`, right? just trying to follow 😅

lib/sdf-server/src/server/service/variant/create_variant.rs
```diff
@@ -38,9 +37,15 @@ pub async fn create_variant(
 ) -> SchemaVariantResult<impl IntoResponse> {
     let mut ctx = builder.build(request_ctx.build(request.visibility)).await?;
 
+    if Schema::is_name_taken(&ctx, &request.name).await? {
```
jobelenus, line 40: ✌️

lib/dal/src/schema/variant/authoring.rs
vbustamante: Yup, it gets packaged into the metadata object

vbustamante:
bors merge

PR #4039, opened 2024-06-25
bug(web): fix connection annotations

adamhjk:
We weren't parsing connection annotations correctly (we had all the logic, we just weren't calling it.)

This fixes that.


adamhjk:
bors merge

PR #4045, opened 2024-06-26
fix(dal): ensure that we only export the default schema variant

stack72:
While authoring, you could create 10s of versions of a variant. Our current export code would include them all in the exported module.

We only want to export the default schema variant as that's the most "stable". In the future this will be changed to be the latest locked schema variant


stack72:
bors try

PR #4050, opened 2024-06-26
Ensure secret field order is correct

nickgerace:
## Description

This PR ensures the secret field order is correct by using ordered props. Assets may need to be regenerated from other changes (e.g. props can only be ordered on recent main) in order to see this effect.


nickgerace:
bors merge

PR #4051, opened 2024-06-27
fix(web) - BUG-412 and ENG-2585

wendybujalski:
Now you can delete Arrays or Maps which are children of Arrays or Maps.

All empty props that are controlled by a parent now hide and a number of hidden props is shown in the parent.


vbustamante (approved):
🚀

wendybujalski:
bors merge

PR #4059, opened 2024-06-28
feat: allow pulling pg cert from a remote url

sprutton1:
This allows `pg-data-nats` to grab root certs from a remote instead of as a file on disk or a base64 string. The file will be written to a temp file and then read as normal. This is specifically so the module-index can load a new root RDS cert from a url because the base64 encoded latest version of that cert is too big to fit in SecretsManager. I kind of like this approach better anyhow.

You should be able to pass `--pg-cert-url` or set `SI_MODULE_INDEX__PG__CERTIFICATE_URL` to the url. Note that these certs are public and therefore do not need to be encoded as secrets.


sprutton1:
bors try

sprutton1:
bors try

PR #4062, opened 2024-06-28
Typescript chill-out, Qualification display fix, contribution always on

adamhjk:
* When showing the output of a qualification, we wrap it in a <pre> so that you can see newlines be rendered.
* We want to always show the contribute button, because otherwise we can never see it.
* Typescript is now configured to be much more relaxed, and to have an 'any' type for `_` while we figure out how to get the types for lodash loaded up.


stack72:
bors merge

PR #4065, opened 2024-07-01
Chainsaw for Funcs Part 1 - Func Authoring Parity with New Routes

britmyerss:
**Changes in this PR**

**New front end objects for Function Authoring**
- `FuncSummary` includes details about the function, the `func_id`, `FuncKind`, `display_name`, `name`, all function arguments, and all bindings for that Function
- `FuncBinding` - is the evolution of `FuncAssociations` and represents the intersection of the Func metadata and all of the required information for that Function to be configured and used by a Schema Variant (or Component)
- `FuncCode` is the actual code body and all the correct type information for type checking when authoring

**New Routes for Function Authoring**
All routes for func authoring follow the same pattern of `api/v2/workspaces/{workspace_id}/change-sets/{change_set_id}/funcs`

Reorganized the payloads for the routes to deduplicate how the front end is getting data, following the pattern we adopted for Func Arguments and Attribute Bindings

Abolished the `save_func` hammer and created separate CRUD routes for creating, modifying and deleting FuncBindings for Auth, Qualifications, Actions, and Code Gen Funcs. 

**`FuncBindings` in the Dal**
This enum encapsulates the logic for creating, updating, and deleting the various prototypes and related configuration metadata for how functions relate to Schema Variants/Components. 

I duplicated the existing tests to exercise the new code, but this is an area that we will definitely expand integration tests as the immutability for schema variants and functions come into play. 

This is set up to support the ability to add bindings for intrinsics (including setting default values) via the UI, but that work will be a follow on. 

** New WS Events** 
This is still very WIP as I work with @jobelenus on the necessary shape for these

For now, I've left all of the existing code/routes in tact so this is effectively toggled, but once the front end consumes these new routes I'll be able to remove what's been duplicated. 

Huge shoutout to @nickgerace for the work on porting authoring to the new engine - I was able to largely use the existing FuncAuthoringClient for the function specific authoring work! 

<div><br/>via <a href="https://giphy.com/family-guy/">Family Guy</a> on <a href="https://giphy.com/gifs/family-guy-family-guy-peter-chainsaw-eMRmnt6MIhfcJsMexH">GIPHY</a></div>


lib/dal/src/func/authoring.rs
```diff
@@ -466,6 +584,43 @@ impl FuncAuthoringClient {
         Ok(())
     }
 
+    #[instrument(level = "info", name = "func.authoring.save_code", skip(ctx))]
+    /// Save only the code for the given [`FuncId`]
+    pub async fn save_code(
```
vbustamante, line 589: I wonder if separate save code from save metadata is the right approach here - right now the way our system works you can get a conflict if you run those two requests in parallel, which can be easily done via the UI

Those kinds of races caused some problems for asset code recently

lib/dal/src/socket/output.rs
```diff
@@ -495,6 +495,8 @@ impl From<OutputSocket> for frontend_types::OutputSocket {
         Self {
             id: value.id.into(),
             name: value.name,
+            //default to false, but figure out how to do this better
+            eligible_to_recieve_data: false,
```
vbustamante, line 499: typo on "recieve" 😅

lib/sdf-server/src/server/service/v2/func/create_func.rs
```diff
@@ -0,0 +1,265 @@
...
+        if dal::func::is_intrinsic(name)
+            || ["si:resourcePayloadToValue", "si:normalizeToArray"].contains(&name)
+        {
+            return Err(FuncAPIError::FuncNameReserved(name.into()));
+        }
+    }
+
+    let force_change_set_id = ChangeSet::force_new(&mut ctx).await?;
+
+    let created_func_response = match request.kind {
```
vbustamante, line 66: This is more of an architectural suggestion, but i see that every branch that returns something on this match ends up running a variation of:
```
                CreateFuncResponse {
                    summary: func.into_frontend_type(&ctx).await?,
                    code: get_code_response(&ctx, created_func.id).await?,
                }
```

which makes me think that it would make senes for the match to return a Func (or even just the func id) instead of that type

lib/sdf-server/src/server/service/v2/func.rs
```diff
@@ -0,0 +1,170 @@
...
+            "/:func_id/create_argument",
+            post(argument::create_argument::create_func_argument),
+        )
+        .route(
+            "/:func_id/:func_argument_id/delete",
+            post(argument::delete_argument::delete_func_argument),
+        )
+}
+
+pub async fn get_code_response(ctx: &DalContext, func_id: FuncId) -> FuncAPIResult<FuncCode> {
```
vbustamante, line 150: I'm not sure I get 100% why this is necessary, I think a comment over here would help

vbustamante (changes requested):
Great test coverage, and it looks like it's the right change - apart from splitting the func saves

lib/dal/src/socket/output.rs
britmyerss: "I before E except after C" --- EXCEPT FOR SCIENCE!!

lib/dal/src/func/authoring.rs
britmyerss: yeah we'll feel our way through this - once you can no longer conflict with yourself it'll be less painful

vbustamante (approved):
LGTM

vbustamante:
bors merge

britmyerss:
bors retry

britmyerss:
bors retry

PR #4067, opened 2024-07-01
Allow secrets to depend on secrets with before func ordering

nickgerace:
## Description

This PR allows secrets to depend on secrets by ensuring before funcs are ordered. The ordering is determined by the underlying dependency graph based on the target component's ancestry, regardless of component type (e.g. "frame" or "component").

This PR also changes the asset builder to only allow one output socket for secret-defining assets as well as allows more than one secret prop (needed for secrets dependent on secrets).

## Secondary Changes

- Add ability to find a source component by a given input socket match where the input socket has an arity of one
- Make secret errors related to widgets more clear in func runner
- Find auth funcs for a given secret child prop id based on output socket rather than secret definition props (related to the asset builder change)
- Add ability to list default secret defining schema variants by id
- Add ability to find the output socket for a secret defining schema variant
- Change `AttributeValue::get_by_id` to `AttributeValue::get_by_id_or_error` since the function will generate an error if the attribute value does not exist
- Change `Prop::direct_child_prop_ids` to `Prop::direct_child_prop_ids_unordered` since there is an ordered function that is likely preferable for most use cases

### DIFF
--- a/lib/dal/src/component.rs
+++ b/lib/dal/src/component.rs
@@ -171,10 +171,18 @@ pub enum ComponentError {
     Serde(#[from] serde_json::Error),
     #[error("standard model error: {0}")]
     StandardModel(#[from] StandardModelError),
+    #[error("too many explicit connection sources ({0:?}) for component ({1}) and input socket ({2}) with an arity of one")]
+    TooManyExplicitConnectionSources(Vec<ComponentId>, ComponentId, InputSocketId),
+    #[error(
+        "too many inferred connections ({0:?}) for input socket match ({1:?}) with an arity of one"
+    )]
+    TooManyInferredConnections(Vec<OutputSocketMatch>, InputSocketMatch),
     #[error("transactions error: {0}")]
     Transactions(#[from] TransactionsError),
     #[error("try lock error: {0}")]
     TryLock(#[from] TryLockError),
+    #[error("unexpected explicit source ({0}) and inferred source ({1}) for input socket match ({2:?}) with an arity of one")]
+    UnexpectedExplicitAndInferredSources(ComponentId, ComponentId, InputSocketMatch),
     #[error("value source for known prop attribute value {0} is not a prop id")]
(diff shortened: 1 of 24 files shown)


lib/dal/src/func/runner.rs
```diff
@@ -1036,59 +1059,12 @@ impl FuncRunner {
...
-                    Err(other_err) => return Err(other_err)?,
-                }
-            }
-
-            if let Some(value) = maybe_value {
-                let key = Secret::key_from_value_in_attribute_value(value)?;
-                funcs_and_secrets.push((key, auth_funcs))
-            }
-        }
+        let ordered_before_funcs_with_secret_keys =
```
nickgerace, line 1062: This is where the primary changes for this PR begin.

lib/dal/src/component.rs
```diff
@@ -3575,6 +3598,69 @@ impl Component {
...
     }
+
+    /// Finds the source [`Component`] of any [`ComponentType`] for a given [`InputSocketMatch`] where the
+    /// [`InputSocket`] has an [arity](SocketArity) of [one](SocketArity::One).
+    #[instrument(
+        name = "component.source_component_for_arity_one_input_socket_match",
+        level = "debug",
+        skip_all
+    )]
+    pub async fn source_component_for_arity_one_input_socket_match(
```
nickgerace, line 3609: @britmyerss this function is where I used your inferred connections code

lib/dal/src/func/runner.rs
```diff
@@ -1114,7 +1090,127 @@ impl FuncRunner {
             }
         }
 
-        Ok(results)
+        Ok(before_functions)
+    }
+
+    /// This _private_ method generates a flattened graph of _raw_ before functions with corresponding
```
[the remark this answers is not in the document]
nickgerace: Just means `Func` and not `BeforeFunction`.

nickgerace:
bors merge

nickgerace:
bors merge

PR #4074, opened 2024-07-02
chore: temporary rollback of labeler

sprutton1:
This labeler update includes a breaking change that needs to be staged out in a specific way. Rolling this back so I can make that change.

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -1,7 +1,7 @@
 name: labeler
 
 on:
-  - pull_request
+  - pull_request_target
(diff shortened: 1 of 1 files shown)


.github/workflows/labeler.yml
```diff
@@ -10,6 +10,6 @@ jobs:
       pull-requests: write
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/labeler@v5
+      - uses: actions/labeler@v4
```
stack72, line 13: 🤦

PR #4083, opened 2024-07-03
Fix leaf func detachment and action func attachment

nickgerace:
## Description

This PR ensures detaching leaf funcs from schema variants also detaches them from components using the schema variant (unless overridden). This works by removing map entry attribute values that use the prototype being deleted (i.e. detaching logic).

This PR also fixes action func attachment showing the wrong kind by fixing the conversion from the events type to the real type.

### DIFF
--- a/lib/dal/src/action/prototype.rs
+++ b/lib/dal/src/action/prototype.rs
@@ -92,12 +92,13 @@ impl From<ActionKind> for si_events::ActionKind {
         }
     }
 }
+
 impl From<si_events::ActionKind> for ActionKind {
     fn from(value: si_events::ActionKind) -> Self {
         match value {
             si_events::ActionKind::Create => ActionKind::Create,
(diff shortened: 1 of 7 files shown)


nickgerace:
Taking into drafts because the second bug has shown up in a different form than expected.

lib/dal/src/action/prototype.rs
```diff
@@ -92,12 +92,13 @@ impl From<ActionKind> for si_events::ActionKind {
         }
     }
 }
+
 impl From<si_events::ActionKind> for ActionKind {
     fn from(value: si_events::ActionKind) -> Self {
         match value {
             si_events::ActionKind::Create => ActionKind::Create,
             si_events::ActionKind::Destroy => ActionKind::Destroy,
-            si_events::ActionKind::Manual => ActionKind::Refresh,
```
nickgerace, line 100: pain

PR #4106, opened 2024-07-06
fix(web): Redirect to unlocked function correctly

vbustamante:


app/web/src/components/FuncEditor/FuncDetails.vue
```diff
@@ -250,6 +251,7 @@ import {
   VormInput,
 } from "@si/vue-lib/design-system";
 import clsx from "clsx";
+import { ca } from "date-fns/locale";
```
vbustamante, line 254:
```suggestion

```

PR #4109, opened 2024-07-08
fix(sdf): Update install_module to allow multiple schema packages

stack72:
This will allow us to catch installing old development modules that have multiple schemas

### DIFF
--- a/lib/sdf-server/src/server/service/module/install_module.rs
+++ b/lib/sdf-server/src/server/service/module/install_module.rs
@@ -101,13 +101,22 @@ async fn install_module_inner(
     let pkg_data = module_index_client.download_module(request.id).await?;
 
     let pkg = SiPkg::load_from_bytes(pkg_data)?;
+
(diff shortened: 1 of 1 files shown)


lib/sdf-server/src/server/service/module/install_module.rs
```diff
@@ -101,13 +101,22 @@ async fn install_module_inner(
...
     let pkg = SiPkg::load_from_bytes(pkg_data)?;
+
+    let (schema_id, past_module_hashes) = if pkg.schemas()?.len() > 1 {
+        (None, None)
+    } else {
+        (
+            module_details.schema_id().map(Into::into),
+            module_details.past_hashes,
+        )
+    };
```
nickgerace, line 112: Should we log when this happens with a warning or debug log?

stack72: Nah, it's fine for now - this will only be used in a dev workflow

PR #4124, opened 2024-07-10
Show that assets can be updated

nickgerace:
## Description

This PR shows that an asset can be updated. This functionality is achieved by asking the module index what the latest module is for the corresponding local hash.

Why not use the `SchemaId` instead of the local hash? Existing SI deployments may not have the `SchemaId` in their local modules since the module-to-schema enforcement and corresponding table columns were only put in place recently.

One consideration for the future: we may want the "can I update?" logic to be moved into a different route in order to not hold up listing variants. For now, the determination logic is wrapper in an infallible wrapper in case something goes wrong, but that does not prevent poor performance.

## Secondary Changes
- Ensure `app/web` no longer mutates the `canUpdate` field for `SchemaVariants`
- Add `README` for `module-index` with instructions for local development
- Remove `SchemaVariantListEntry` as it is no longer needed

## Rendered `module-index` Documentation

[Click here!](https://github.com/systeminit/si/blob/nick/eng-2593/bin/module-index/README.md)

### DIFF
--- a/app/web/src/api/sdf/dal/schema.ts
+++ b/app/web/src/api/sdf/dal/schema.ts
@@ -66,6 +66,9 @@ export interface SchemaVariant {
   outputSockets: OutputSocket[];
   props: Prop[];
   canCreateNewComponents: boolean;
+
+  canContribute: boolean;
+  canUpdate: boolean;
 }
 
 export const outputSocketsAndPropsFor = (schemaVariant: SchemaVariant) => {
--- a/lib/module-index-client/src/types.rs
+++ b/lib/module-index-client/src/types.rs
@@ -14,8 +14,6 @@ pub enum IndexClientError {
     Request(#[from] reqwest::Error),
     #[error("Serialization error: {0}")]
     Serialization(serde_json::Error),
-    #[error("Upload error: {0}")]
-    Upload(String),
     #[error("Url parse error: {0}")]
     UrlParse(#[from] url::ParseError),
 }
@@ -74,4 +72,41 @@ pub struct ExtraMetadata {
     pub version: String,
     pub schemas: Vec<String>,
     pub funcs: Vec<FuncMetadata>,
+}
+
+#[derive(Deserialize, Serialize, Debug)]
+#[serde(rename_all = "camelCase")]
+pub struct ListLatestModulesRequest {
+    pub hashes: Vec<String>,
+}
+
+#[derive(Deserialize, Serialize, Debug)]
+#[serde(rename_all = "camelCase")]
+pub struct ListLatestModulesResponse {
+    pub modules: Vec<LatestModuleResponse>,
+}
+
+/// This struct is nearly the same as the [`ModuleDetailsResponse`], but it does not include `past_hashes` since the
+/// data is unneeded and requires additional query logic.
+#[derive(Deserialize, Serialize, Debug)]
+#[serde(rename_all = "camelCase")]
+pub struct LatestModuleResponse {
+    pub id: String,
+    pub name: String,
+    pub description: Option<String>,
+    pub owner_user_id: String,
+    pub owner_display_name: Option<String>,
+    pub metadata: serde_json::Value,
+    pub latest_hash: String,
+    pub latest_hash_created_at: DateTime<Utc>,
+    pub created_at: DateTime<Utc>,
+    pub schema_id: Option<String>,
+}
+
+impl LatestModuleResponse {
+    pub fn schema_id(&self) -> Option<Ulid> {
+        self.schema_id
+            .as_deref()
+            .and_then(|schema_id| Ulid::from_string(schema_id).ok())
+    }
 }
--- a/lib/module-index-server/src/routes/list_latest_modules_route.rs
+++ b/lib/module-index-server/src/routes/list_latest_modules_route.rs
@@ -1,0 +1,61 @@
+use axum::{
+    response::{IntoResponse, Response},
+    Json,
+};
+use hyper::StatusCode;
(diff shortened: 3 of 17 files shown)


app/web/src/api/sdf/dal/schema.ts
```diff
@@ -66,6 +66,8 @@ export interface SchemaVariant {
   outputSockets: OutputSocket[];
   props: Prop[];
   canCreateNewComponents: boolean;
+
+  canUpdate: boolean;
```
nickgerace, line 70: This has been moved to the variant itself and is no longer mutated in `app/web`.

lib/module-index-client/src/types.rs
```diff
@@ -75,3 +73,40 @@ pub struct ExtraMetadata {
...
+    pub name: String,
+    pub description: Option<String>,
+    pub owner_user_id: String,
+    pub owner_display_name: Option<String>,
+    pub metadata: serde_json::Value,
+    pub latest_hash: String,
+    pub latest_hash_created_at: DateTime<Utc>,
+    pub created_at: DateTime<Utc>,
+    pub schema_id: Option<String>,
+}
```
nickgerace, line 104: We probably do not need all this data, but the query can easily access it. We can likely reduce the amount of data sent over the wire once we have the update mechanism in place.

lib/sdf-server/src/server/service/v2/variant.rs
```diff
@@ -100,7 +162,11 @@ pub async fn get_variant(
 
     let schema_variant = SchemaVariant::get_by_id_or_error(&ctx, schema_variant_id).await?;
     let schema_id = SchemaVariant::schema_id_for_schema_variant_id(&ctx, schema_variant_id).await?;
-    let schema_variant = schema_variant.into_frontend_type(&ctx, schema_id).await?;
+    let mut schema_variant = schema_variant.into_frontend_type(&ctx, schema_id).await?;
+
+    if let Err(err) = determine_can_update_one(&ctx, &mut schema_variant, &raw_access_token).await {
```
nickgerace, line 167: We don't use this route ever (and normally, I'd delete dead code), but I covered it off anyway since there's active work here. 
cc @jobelenus @vbustamante @wendybujalski

lib/sdf-server/src/server/service/v2/variant.rs
```diff
@@ -122,32 +188,148 @@ pub async fn get_variant(
...
+                    "skipping since there's incomplete data for determining if schema variant can be updated"
+                );
+            }
+        }
     }
+
+    debug!(
+        "function 'determine_can_update_many' took: {:?}",
+        start.elapsed()
+    );
```
nickgerace, line 277: ~9ms-25ms RTT locally

lib/sdf-server/src/server/service/v2/variant.rs
```diff
@@ -122,32 +188,148 @@ pub async fn get_variant(
...
 
-        ApiError::new(status_code, self).into_response()
+    // Determine if we can update based on the latest modules.
+    for schema_variant in schema_variants.values_mut() {
+        let schema_id: SchemaId = schema_variant.schema_id.into();
+        match (latest_hashes.get(&schema_id), local_hashes.get(&schema_id)) {
+            (Some(latest_hash), Some(local_hash)) => {
+                if latest_hash != local_hash {
+                    schema_variant.can_update = true;
+                }
```
nickgerace, line 259: All of that, for this.

lib/sdf-server/src/server/service/v2/variant.rs
```diff
@@ -72,6 +128,11 @@ pub async fn list_schema_variants(
         }
     }
 
+    if let Err(err) = determine_can_update_many(&ctx, &mut schema_variants, &raw_access_token).await
+    {
+        error!(?err, "could not perform 'determine_can_update_many'");
+    }
```
nickgerace, line 134: Wrapped in an infallible wrapper for now because determining `canUpdate` should not error out the route... yet. I suspect this logic moves into its own route in the near future anyway, so let's not cause any problems.

lib/module-index-server/src/routes/list_latest_modules_route.rs
```diff
@@ -0,0 +1,63 @@
...
+    DbConnection(txn): DbConnection,
+    Json(request): Json<ListLatestModulesRequest>,
+) -> Result<Json<ListLatestModulesResponse>, ListModulesError> {
+    let raw_modules = si_module::Entity::find()
+        .from_raw_sql(Statement::from_sql_and_values(
+            DbBackend::Postgres,
+            LIST_LATEST_MODULES_QUERY,
+            [request.hashes.into()],
+        ))
+        .all(&txn)
```
nickgerace, line 54: We need to remove the ORM.

lib/sdf-server/src/server/service/v2/variant.rs
```diff
@@ -122,32 +188,148 @@ pub async fn get_variant(
...
-    fn into_response(self) -> Response {
-        let status_code = match &self {
-            Self::Transactions(dal::TransactionsError::BadWorkspaceAndChangeSet) => {
-                StatusCode::FORBIDDEN
+#[instrument(
+    name = "sdf.v2.variant.determine_can_update_many",
+    level = "debug",
+    skip_all
+)]
+async fn determine_can_update_many(
```
nickgerace, line 197: Cannot go in the `dal` since `module-index-client` should only be accessible by `sdf`. Didn't want to make a new module since the `v2` stuff is brand new and things are settling.

nickgerace:
Moving back to drafts for another test.

vbustamante (approved):
Great!

PR #4136, opened 2024-07-11
feat(auth, web) - revamped auth portal for saas launch

wendybujalski:
- Legal section now has a button to jump to the bottom
- Removed Tutorial Section from Auth Portal
- Fixed email verification system
- Removed additional questions from onboarding
- Dashboard is now called "Workspaces"
- First time users who have their email verified will be pushed directly into their default workspace on account completion
- First time modal with placeholder video - NEED TO REPLACE WITH ACTUAL TUTORIAL VIDEO
- Button on first time modal dismisses it
- App navbar changes - removed link button, added discord and docs links, manage workspaces menu option, theme and profile in profile dropdown


jobelenus (approved):
We may want to talk to Scott about a good way to protect those auth-api URLs that app/web is using... it's not ideal that they're open to the world...

PR #4146, opened 2024-07-13
Indicate what modules can be installed and updated (ENG-2593)

nickgerace:
## Description

This PR provides the functionality to be able to know what modules can be installed and updated. In the last PR (#4124), we were able to indicate what assets could be updated, but it came with two caveats:

1. The logic for communicating with the module index was wrapped inside the route to list all user facing schema variants. While the logic was wrapped in an infallible wrapper, poor performance could still occur.
2. It did not convey what modules could be installed.

This PR moves and refactors the existing logic into a new route. In addition, providing specific hashes is no longer required for the request.

## Disclaimer

This commit causes a feature regression. Now, assets do not indicate that they can be updated. This is a temporary regression since the new route provided in this PR should be used instead. Speaking of which, the new route is untested in practice due to the module index being presently unavailable in automated, Rust-based tests.

## Secondary Changes

- Move logic for listing user facing schema variants into the dal (can be tested in integration tests and used by multiple routes)
- Add regression test for listing user facing schema variants
- Split module index types into its own crate (re-exported by the client crate for users of the client)
- Add module syncing types to frontend types crate (using the new module index types crate as well)
- Split v2 variant route business logic into sub-modules
- Fix layer cache crate cargo warning for specifying dev dependencies


nickgerace:
Moving to drafts to fix the build error.

PR #4153, opened 2024-07-15
fix: ignore erroneous 404s in cypress tests

sprutton1:
We're getting erroneuous `404`s when these tests are running that seem to have to do with the redirect and hosting on S3. These 404s do not appear when using the app in practice. I'm ignoring them here because in every case we go on to validate actual behavior after we visit.


sprutton1:
Valid run: https://github.com/systeminit/si/actions/runs/9946056125

PR #4161, opened 2024-07-16
feat: add auth-api to buck2 builds

sprutton1:
Compress the dockerfile into something somewhat readable, even if I don't get everything it is doing. It builds! Made it so buck2 can build that auth-api so we get containers automatically instead of arbitrarily sending them to ECR when we deploy. Also updates the deploy action to work with this scheme so things are finally normalized. There's some jank here, but I think unraveling it any further is just an exercise in diminishing returns.

### DIFF
--- a/bin/auth-api/Dockerfile
+++ b/bin/auth-api/Dockerfile
@@ -1,118 +1,35 @@
 ARG NODE_VERSION=18.16.0
-# ARG PNPM_VERSION=8.1.1
 ARG PACKAGE_PATH=@si/auth-api
 
 FROM node:$NODE_VERSION-slim AS base
-# Check https://github.com/nodejs/docker-node/tree/b4117f9333da4138b03a546ec926ef50a31506c3#nodealpine to understand why libc6-compat might be needed.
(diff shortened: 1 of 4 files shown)


bin/auth-api/Dockerfile
```diff
@@ -1,118 +1,35 @@
...
-
-RUN --mount=type=cache,id=pnpm-store,target=/root/.pnpm-store\
-  # ↑ By caching the content-addressable store we stop downloading the same packages again and again
-  PRISMA_SKIP_POSTINSTALL_GENERATE=true\
-  pnpm install --filter . --filter @si/auth-api... --frozen-lockfile\
-  --unsafe-perm\
-  # ↑ Docker runs pnpm as root and then pnpm won't run package scripts unless we pass this arg
-  | grep -v "cross-device link not permitted\|Falling back to copying packages from store"
-
-# ↑ This inverted grep match is because using Docker's 'cache' mount type
```
nickgerace, line 34: YES! GET RID OF IT ALL!!!

PR #4162, opened 2024-07-16
Move module sync to dal, fix its tracing and add integration test

nickgerace:
## Description

This PR moves module syncing to the dal. In addition, it fixes tracing for module syncing and includes more logging. As a result of module syncing moving to the dal, there's a new integration test too.

### DIFF
--- a/lib/dal/src/module.rs
+++ b/lib/dal/src/module.rs
@@ -1,15 +1,17 @@
-use chrono::{DateTime, Utc};
-use std::collections::HashMap;
+use std::collections::hash_map::Entry;
+use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
 
+use chrono::{DateTime, Utc};
 use serde::{Deserialize, Serialize};
 use si_events::ulid::Ulid;
 use si_events::ContentHash;
-use thiserror::Error;
-use tokio::sync::TryLockError;
-
+use si_frontend_types as frontend_types;
 use si_layer_cache::LayerDbError;
 use telemetry::prelude::*;
+use thiserror::Error;
+use tokio::sync::TryLockError;
+use tokio::time::Instant;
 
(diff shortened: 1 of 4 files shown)


lib/dal/src/module.rs
```diff
@@ -402,4 +410,106 @@ impl Module {
...
+    /// [`Modules`](Module) can be upgraded and installed.
+    #[instrument(
+        name = "module.sync"
+        skip_all,
+        level = "debug",
+    )]
+    pub async fn sync(
+        ctx: &DalContext,
+        latest_modules: Vec<frontend_types::LatestModule>,
+    ) -> ModuleResult<frontend_types::SyncedModules> {
```
nickgerace, line 425: This should be picked up now. Considered adding `latest_modules` to the fields, but the object could be pretty big since it is _all_ latest modules.
cc @stack72 @jobelenus

lib/dal/src/module.rs
```diff
@@ -402,4 +410,106 @@ impl Module {
...
+        debug!("working with {} latest modules", latest_modules.len());
```
nickgerace, line 426: ...so I decided to just log the length.

lib/sdf-server/src/server/service/v2/module/sync.rs
```diff
@@ -30,119 +26,24 @@ pub async fn sync(
...
-                    schema_id,
-                    existing.latest_hash,
-                    latest_module.latest_hash,
-                ));
-            }
-            Entry::Vacant(entry) => {
-                entry.insert(latest_module.to_owned());
-            }
-        }
+    let synced_modules = Module::sync(&ctx, latest_modules.modules).await?;
```
nickgerace, line 38: Now, we collect the `latest_modules` first, drop our client out of scope and then `sync`.

PR #4168, opened 2024-07-17
feat: replace stop.sh with pure rust implementation

sprutton1:
This is a one to one replacement of the `stop.sh` script with a pure Rust implementation. Yay! The functionality here should be identical, but now it should be easier for us to manage the behaviors and failure scenarios. I also think this will make it a bit easier to write tests for the firecracker bits. Honestly the hardest part here was making the new crates play nice in buck2 land :smile: . There are two more scripts that need replaced. This was the easy one.


flake.nix
```diff
@@ -141,6 +143,17 @@
...
+          configurePhase =
+            ''
+            export LIBCLANG_PATH="${pkgs.llvmPackages.libclang.lib}/lib"
+            export BINDGEN_EXTRA_CLANG_ARGS="$(< ${pkgs.stdenv.cc}/nix-support/libc-crt1-cflags) \
+              $(< ${pkgs.stdenv.cc}/nix-support/libc-cflags) \
+              $(< ${pkgs.stdenv.cc}/nix-support/cc-cflags) \
+              $(< ${pkgs.stdenv.cc}/nix-support/libcxx-cxxflags) \
+              $NIX_CFLAGS_COMPILE"
+            export OUT=${placeholder "out"}
+            echo $OUT
```
nickgerace, line 155: you need a hug

nickgerace (approved):
Excellent work!

flake.nix
fnichol, line 155: What does this do?

fnichol (approved):
I'm not 100% sure on the fixups but if things are working, then good with me! Nice work!

flake.nix
sprutton1: Ah, that was me trying to understand some things. I'll circle around and clean it up.

PR #4175, opened 2024-07-19
fix: cfg out linux-only firecracker dependencies

sprutton1:
Fixes the flake and hide linux-only dependencies behind comp targets. This is kind of a hokey way to do this but I think the most correct way is likely a refactor of the instance itself, which is a pretty huge undertaking.

### DIFF
--- a/lib/si-firecracker/BUCK
+++ b/lib/si-firecracker/BUCK
@@ -4,14 +4,18 @@ rust_library(
     name = "si-firecracker",
     deps = [
         "//lib/cyclone-core:cyclone-core",
-        "//third-party/rust:devicemapper",
-        "//third-party/rust:krata-loopdev",
         "//third-party/rust:remain",
         "//third-party/rust:nix",
         "//third-party/rust:thiserror",
         "//third-party/rust:tokio",
         "//third-party/rust:tracing",
-    ],
+    ] + select({
+        "DEFAULT": [],
+        "config//os:linux": [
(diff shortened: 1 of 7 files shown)


lib/si-firecracker/BUCK
```diff
@@ -4,14 +4,18 @@ rust_library(
...
         "//lib/cyclone-core:cyclone-core",
-        "//third-party/rust:devicemapper",
-        "//third-party/rust:krata-loopdev",
         "//third-party/rust:remain",
         "//third-party/rust:nix",
         "//third-party/rust:thiserror",
         "//third-party/rust:tokio",
         "//third-party/rust:tracing",
-    ],
+    ] + select({
```
nickgerace, line 12: You are nixing!

lib/si-firecracker/src/firecracker.rs
```diff
@@ -61,6 +62,8 @@ impl FirecrackerJail {
     }
 
     pub async fn clean(id: u32) -> Result<()> {
```
nickgerace, line 64: My gut says to cfg out the whole function... wait isn't the file cfg'd out?

sprutton1: If I cfg out the func or the file, I I think need to slap cfgs all over the place in `local_uds.rs` where this gets called. This was the lightest touch I could figure out, but I'm open to cleaner approaches.

PR #4191, opened 2024-07-22
fix(dal): check for components that can be deleted at the end of the action job

britmyerss:
The way we were doing this before resulted in some races if multiple delete actions were running in parallel, each wouldn't know about the other and so we'd never end up removing the parent frames.  This moves the check to after the func has run and we've removed the component that needed to be deleted, so now there might be races to delete the frames but that's ok, because we want them deleted.  If there's a conflict when deleting those parent frames, we won't retry (but we'll warn) but also won't fail the job as a whole. 

<div><br/>via <a href="https://giphy.com/nickrewind/">NickRewind</a> on <a href="https://giphy.com/gifs/splat-nicksplat-rocket-power-3o6nV8mjvp0Ys8j4iI">GIPHY</a></div>

### DIFF
--- a/lib/dal/src/component.rs
+++ b/lib/dal/src/component.rs
@@ -1443,6 +1443,17 @@ impl Component {
         Ok(components)
     }
 
+    pub async fn list_to_be_deleted(ctx: &DalContext) -> ComponentResult<Vec<ComponentId>> {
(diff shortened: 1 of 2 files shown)


stack72:
Was a conflict on a deletion of a tp level frame before this PR - now it's smooth sailing!

PR #4199, opened 2024-07-23
Feat: improve diagram node selection

jobelenus:
Always have the selection box on top of all other nodes.

For example, if a frame gets trapped behind another frame, even when you select it from the diagram outliner, you have no indication where it is.

The solve here is to have the selection outline boxes drawn on a canvas layer above all other components.


jobelenus:
image

PR #4222, opened 2024-07-24
feat(sdf) - create workspace permissions system

wendybujalski:
A permissions system for creating new workspaces.
Each instance of SDF can be configured in one of three ways -
- Closed - Only the first registered user can create new workspaces
- Open - Any registered user can create new workspaces
- Allowlist - Only registered users with an email on the Allowlist can create new workspaces. The allowlist can be specific email addresses, email domains to allow all addresses from (such as @systeminit.com), or a mix of both.

### DIFF
--- a/bin/sdf/src/args.rs
+++ b/bin/sdf/src/args.rs
@@ -2,7 +2,10 @@ use std::path::PathBuf;
 
 use clap::{builder::EnumValueParser, builder::PossibleValuesParser, ArgAction, Parser};
 
-use sdf_server::{Config, ConfigError, ConfigFile, FeatureFlag, MigrationMode, StandardConfigFile};
+use sdf_server::{
+    server::{WorkspacePermissions, WorkspacePermissionsMode},
+    Config, ConfigError, ConfigFile, FeatureFlag, MigrationMode, StandardConfigFile,
+};
 use si_std::SensitiveString;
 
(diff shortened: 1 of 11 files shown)


bin/sdf/src/args.rs
```diff
@@ -172,7 +175,7 @@ pub(crate) struct Args {
     #[arg(long, env = "SI_MODULE_INDEX_URL")]
     pub(crate) module_index_url: Option<String>,
 
-    /// The base URL for the module-index API server
+    /// Allow for Posthog feature flags in SDF
```
vbustamante, line 178: Ooops, thanks for catching that 😄

bin/sdf/src/args.rs
```diff
@@ -181,6 +184,19 @@ pub(crate) struct Args {
...
+    pub(crate) create_workspace_permissions: Option<String>,
+
+    /// Create Workspace Permissions Allowlist
+    #[arg(
+        long,
+        env = "SI_CREATE_WORKSPACE_ALLOWLIST",
+        value_delimiter = ',',
+        rename_all = "snake_case"
+    )]
+    pub(crate) create_workspace_allowlist: Vec<WorkspacePermissions>,
```
vbustamante, line 199: this looks correct

vbustamante (approved):

PR #4235, opened 2024-07-26
fix: handle the tutorial popup in cypress

sprutton1:
More like cy-mess, amirite?


sprutton1:
working deploy from this branch 

https://github.com/systeminit/si/actions/runs/10114792429/job/27974449599

PR #4259, opened 2024-07-31
Add configurable timeout to function execution in `lang-js`

nickgerace:
## Description

This commit adds timeouts to `lang-js` function execution as well as the ability to configure them. By default, functions can run for a maximum of 30 minutes. This is tunable by a command-line argument, which can be used by cyclone instances. Both cyclone and veritech have tunables for the lang server's function timeout.

In addition to these changes, dead example code has been removed and the docs for `lang-js` have been revitalized, and the test suite timeout has been increased for `lang-js` unit tests. Not only that, but we now `process.exit(0)` upon successful `lang-js` runs, which you can read the comment in the source code to learn more about.

For testing, a new unit test has been added that uses a 1 second timeout in the `lang-js` test suite.

## New Docs

[See the new, rendered `lang-js` docs here](https://github.com/systeminit/si/blob/nick/eng-2618/bin/lang-js/README.md).

## Testing

This PR originally included a `dal` integration test that purposefully hit the timeout (and `lang-js` was hardcoded to timeout after ~10 seconds). Since threading through request timeout settings is not in scope for this work, the test was removed once it passed consistently.


bin/lang-js/src/function.ts
```diff
@@ -91,12 +91,17 @@ export interface OutputLine {
...
 }
 
+// This is the default timeout, configured in minutes.
+const defaultTimeout = 60; // TODO(nick): make this configurable when deploying SI
+
 export async function executeFunction(kind: FunctionKind, request: Request) {
   // Run Before Functions
   const ctx = ctxFromRequest(request);
 
+  const timeout = request.timeout ?? defaultTimeout;
```
nickgerace, line 101: Right now, this is configurable by request. We want that for two reasons: testing and tunable timeouts for ourselves. This PR does not cover deployment-level configuration and that will be covered in a second pass.

bin/lang-js/src/function.ts
```diff
@@ -169,10 +177,17 @@ export async function executeFunction(kind: FunctionKind, request: Request) {
   console.log(JSON.stringify(result));
 }
 
+function timer(minutes: number): Promise<never> {
+  return new Promise((_, reject) => {
+    setTimeout(() => reject(new Error(`function timed out after ${minutes} minutes`)), minutes * 60 * 1000);
```
nickgerace, line 182: `setTimeout` relies on milliseconds. You can provide fractional values for the `minutes` value, if desired.

bin/lang-js/src/function.ts
```diff
@@ -199,8 +214,15 @@ export async function executor<F extends Func | ActionRunFunc, Result>(
...
-  const result = await execute(vm, ctx, func, code);
-  debug({ result });
-
-  return result;
+  try {
+    console.log(`executing ${ctx.executionId} with timeout of ${timeout} minutes`);
+    const result = await Promise.race([
+      execute(vm, ctx, func, code),
+      timer(timeout),
+    ]);
```
nickgerace, line 222: Now, we see which promise comes back first.

bin/lang-js/tests/executeFunction.spec.ts
```diff
@@ -27,6 +27,7 @@ interface FuncScenario {
     arg: Record<string, any>
   }[];
   result?: any;
+  timeout?: number;
```
nickgerace, line 30: New optional field for tests to indicate that they will timeout and what we should set the timer to.

bin/lang-js/tests/executeFunction.spec.ts
```diff
@@ -220,18 +232,30 @@ describe("executeFunction", () => {
...
-        before,
-      });
-
-      const parsedLog = JSON.parse(lastLog);
-
-      expect(parsedLog).toMatchObject(scenario.result ?? {
-        protocol: "result",
-        status: "success",
-      });
+      if (scenario.timeout) {
```
nickgerace, line 235: If the test payload provides a timeout, we will expect it.

bin/lang-js/tests/executeFunction.spec.ts
```diff
@@ -220,18 +232,30 @@ describe("executeFunction", () => {
...
+        try {
+          await executeFunction(scenario.kind, {
+            ...ctx,
+            ...funcObj,
+            timeout: scenario.timeout,
+            before,
+          });
+          fail("expected function to hit timeout, but no error was thrown");
+        } catch (error) {
+          expect(error.message).toBe(`Error: function timed out after ${scenario.timeout} minutes`);
```
nickgerace, line 245: The test must error with this exact message.

bin/lang-js/src/function.ts
```diff
@@ -91,12 +91,25 @@ export interface OutputLine {
...
 
+  const timeout = request.timeout ?? defaultTimeout;
+  console.log(JSON.stringify({
+    protocol: "output",
+    executionId: ctx.executionId,
+    stream: "output",
+    level: "debug",
+    group: "log",
+    message: `Timeout in minutes: ${timeout}`,
+  }));
```
nickgerace, line 109: @johnrwatson @sprutton1 @vbustamante new log for timeout detection at the "debug" level

nickgerace:
Moving back to drafts to add full configure-ability.

nickgerace:
Now with configuration threading from `vertiech` down to `lang-js`.

PR #4318, opened 2024-08-09
Add cyclone client timeout and make func execution timeouts configurable

nickgerace:
## Description

This PR adds a timeout wrapper when executing functions via the cyclone client in veritech. In addition, the default timeout value can be overriden as well as the recently added timeout value for cyclone server's lang server process.

As a result of this commit and recent commits, the following values are now configurable:

- cyclone client timeout
  - dynamically configurable via veritech args and/or config file
- cyclone server timeout
  - statically configurable with cyclone args and/or config file
- lang-js function timeout
  - statically configurable with cyclone args and/or config file
  - dynamically configurable with lang-js args, but they are also statically configured due to how cyclone is invoked via firecracker (see the rootfs builder script for more information)

This PR also renames some functions for clarity, fixes some doc comments and does some cleanup.

### DIFF
--- a/bin/veritech/src/args.rs
+++ b/bin/veritech/src/args.rs
@@ -90,6 +90,10 @@ pub(crate) struct Args {
     /// Veritech decryption key file location [example: /run/veritech/veritech.key]
     #[arg(long)]
     pub(crate) decryption_key: Option<PathBuf>,
+
+    /// Execution timeout when communicating with a cyclone instance executing a function, in seconds
+    #[arg(long)]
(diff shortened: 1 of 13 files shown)


vbustamante (approved):

nickgerace:
All tests pass locally before we posted this PR... going to see if we missed something due to the merge queue failure.

nickgerace:
Fixed. We needed to change the dal test config since we slightly changed how the defaults worked.

PR #4325, opened 2024-08-12
feat(tracing): Getting the host as part of the otel span

stack72:


fnichol (approved):
Makes a lot of sense, I wonder why we didn't have this in the first place!

PR #4329, opened 2024-08-13
Move SummaryDiagramComponent struct to si_frontend_types & use `into_frontend_type` fn

jobelenus:
1. Move `SummaryDiagramComponent` struct to `si_frontend_types` & use `into_frontend_type` fn.
2. Stop firing event from `Component::new`, replace with events in create, paste, and upgrade.
3. Use SummaryDiagramEdge in `WsEvent::connection_created` (now `connection_upserted`) & `connection_deleted`.
4. Ghosted edge fixes in get_diagram
5. Ghosted component & edges fixes in delete_component
6. Ghosted edges fixed in delete_connection
7. Copying and pasting components with edges between them sends correct WsEvents
8. Upgrading components sends correct WsEvents.

### DIFF
--- a/lib/dal/src/component.rs
+++ b/lib/dal/src/component.rs
@@ -7,9 +7,9 @@ use petgraph::Direction::Outgoing;
 use serde::{Deserialize, Serialize};
 use si_pkg::KeyOrIndex;
 use socket::{ComponentInputSocket, ComponentOutputSocket};
-use std::collections::{HashMap, VecDeque};
+use std::collections::{hash_map, HashMap, HashSet, VecDeque};
 use std::hash::Hash;
-use std::num::ParseFloatError;
+use std::num::{ParseFloatError, ParseIntError};
 use std::sync::Arc;
 use telemetry::prelude::*;
 use thiserror::Error;
@@ -29,8 +29,9 @@ use crate::attribute::value::{
     AttributeValueError, ChildAttributeValuePair, DependentValueGraph, ValueIsFor,
 };
 use crate::change_set::ChangeSetError;
+use crate::change_status::ChangeStatus;
 use crate::code_view::CodeViewError;
-use crate::diagram::{SummaryDiagramComponent, SummaryDiagramInferredEdge};
+use crate::diagram::{SummaryDiagramEdge, SummaryDiagramInferredEdge};
 use crate::history_event::HistoryEventMetadata;
 use crate::layer_db_types::{ComponentContent, ComponentContentV1};
 use crate::prop::{PropError, PropPath};
@@ -49,6 +50,10 @@ use crate::workspace_snapshot::WorkspaceSnapshotError;
 use crate::{AttributePrototypeId, SocketArity};
 use frame::{Frame, FrameError};
 use resource::ResourceData;
+use si_frontend_types::{
+    DiagramSocket, DiagramSocketDirection, DiagramSocketNodeSide, GridPoint, Size2D,
+    SummaryDiagramComponent,
+};
 
 use crate::{
     implement_add_edge_to, pk, AttributePrototype, AttributeValue, AttributeValueId, ChangeSetId,
@@ -154,8 +159,10 @@ pub enum ComponentError {
     OutputSocketNotFoundForComponentId(OutputSocketId, ComponentId),
     #[error("output socket {0} has more than one attribute value")]
     OutputSocketTooManyAttributeValues(OutputSocketId),
-    #[error("parse float error: {0}")]
+    #[error(transparent)]
     ParseFloat(#[from] ParseFloatError),
+    #[error(transparent)]
+    ParseInt(#[from] ParseIntError),
     #[error("prop error: {0}")]
     Prop(#[from] PropError),
     #[error("found prop id ({0}) that is not a prop")]
@@ -564,11 +571,6 @@ impl Component {
                 .await
                 .map_err(|err| ComponentError::Action(Box::new(err)))?;
         }
-
-        WsEvent::component_created(ctx, component.id())
-            .await?
-            .publish_on_commit(ctx)
-            .await?;
 
         Ok(component)
     }
@@ -2359,11 +2361,6 @@ impl Component {
 
         ctx.workspace_snapshot()?.remove_node_by_id(id).await?;
 
-        WsEvent::component_deleted(ctx, id)
-            .await?
-            .publish_on_commit(ctx)
-            .await?;
-
         Ok(())
     }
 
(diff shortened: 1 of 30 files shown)


lib/dal/src/diagram.rs
```diff
@@ -296,6 +130,25 @@ impl SummaryDiagramEdge {
...
     }
+
+    pub fn assemble_outgoing(
+        outgoing_connection: OutgoingConnection,
+        from_component: &Component,
+        to_component: &Component,
+        change_status: ChangeStatus,
+    ) -> ComponentResult<Self> {
+        Ok(SummaryDiagramEdge {
+            from_component_id: outgoing_connection.from_component_id,
```
jobelenus, line 141: Are these from & to "correct"? I don't have to flip them?

lib/dal/src/component.rs
```diff
@@ -3952,13 +3972,156 @@ impl Component {
...
+    ) -> ComponentResult<HashMap<ComponentId, bool>> {
+        let mut components = HashMap::new();
+        let base_change_set_ctx = ctx.clone_with_base().await?;
+        for component_id in component_ids {
+            let maybe_component =
+                Component::try_get_by_id(&base_change_set_ctx, component_id).await?;
+            match maybe_component {
+                Some(_) => components.insert(component_id, true),
+                _ => components.insert(component_id, false),
+            };
```
nickgerace, line 3988: I would do the following:
```rust
components.insert(component_id, maybe_component.is_some());
```

lib/dal/src/component.rs
```diff
@@ -3952,13 +3972,156 @@ impl Component {
...
+                entry.insert(socket_value.to_owned());
+
+                socket_value
+            }
+            hash_map::Entry::Occupied(entry) => entry.get().to_owned(),
+        };
+        let schema = SchemaVariant::schema_for_schema_variant_id(ctx, schema_variant.id()).await?;
+
+        let default_schema_variant =
+            SchemaVariant::get_default_for_schema(ctx, schema.id()).await?;
```
nickgerace, line 4049: Are we always wanting to walk back to the default even though we already have a variant?

nickgerace (approved):

lib/dal/src/component.rs
jobelenus: We seem to use it for `can_be_upgraded`... I copy pasta'd this code.. just moved it from one place to another.

lib/dal/src/diagram.rs
britmyerss: yeah those look right to me! you can also just pull the to and from off the outgoing_connection instead of also passing them in

lib/sdf-server/src/server/service/diagram/create_connection.rs
```diff
@@ -67,16 +69,22 @@ pub async fn create_connection(
...
-        request.from_socket_id,
-        request.to_socket_id,
-    )
-    .await?
-    .publish_on_commit(&ctx)
-    .await?;
+    let from_component = Component::get_by_id(&ctx, request.from_component_id).await?;
+    let to_component = Component::get_by_id(&ctx, request.to_component_id).await?;
+    for incoming_connection in to_component.incoming_connections(&ctx).await? {
+        if incoming_connection.to_input_socket_id == request.to_socket_id {
```
jhelwig, line 75: Shouldn't this make sure that all of the `to_socket_id`, `from_component_id`, and `from_socket_id` match? As it currently is, it looks like it'll generate the `WsEvent` for every edge going to the same `InputSocket` as the original request, regardless of which `Component` or `OutputSocket` is on the source side.

lib/sdf-server/src/server/service/diagram/delete_component.rs
```diff
@@ -47,16 +54,71 @@ pub async fn delete_components(
         .await?;
         components.insert(component_id, component_still_exists);
 
+        let exists_on_head = components_existing_on_head
+            .get(&component_id)
+            .unwrap_or(&false);
```
jhelwig, line 59: This feels really awkward. It's really unusual to see a reference to a bool. It looks like here is the only place that's using the output of `Component::exists_on_head`. Rather than returning a `HashMap<ComponentId, bool>` it would probably be more ergonomic to have it return a `HashSet<ComponentId>` of only the `ComponentId`s that do exist on `HEAD`. If it did that, then this check becomes `let exists_on_head = components_existing_on_head.contains(&component_id);` with no need for an extra unwrapping or defaults.

lib/sdf-server/src/server/service/diagram/delete_connection.rs
```diff
@@ -43,15 +45,68 @@ pub async fn delete_connection(
...
     let input_socket = InputSocket::get_by_id(&ctx, request.to_socket_id).await?;
 
+    let base_change_set_ctx = ctx.clone_with_base().await?;
+
+    let base_from_component = Component::get_by_id(&base_change_set_ctx, request.from_component_id)
+        .await
+        .ok();
+    let base_to_component = Component::get_by_id(&base_change_set_ctx, request.to_component_id)
+        .await
+        .ok();
```
jhelwig, line 62: These should use `try_get_by_id` instead, since it returns a `ComponentResult<Option<Self>>`. While the frontend can't do anything about any errors that happen, we shouldn't hide them from our monitoring.

lib/dal/src/component.rs
```diff
@@ -3148,13 +3183,153 @@ impl Component {
...
+        let base_change_set_ctx = ctx.clone_with_base().await?;
+        for component_id in component_ids {
+            let maybe_component =
+                Component::try_get_by_id(&base_change_set_ctx, component_id).await?;
+            components.insert(component_id, maybe_component.is_some());
+        }
+        Ok(components)
+    }
+
+    pub async fn into_frontend_type(
```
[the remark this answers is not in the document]
jobelenus: I'm following Fletcher's example here for the `into_frontend_type` fns. He had them on the DAL object. But, like you said, from/into works either way.

lib/dal/src/workspace_snapshot.rs
```diff
@@ -1642,105 +1642,65 @@ impl WorkspaceSnapshot {
...
-
-            // * Get incoming edges
-            for incoming_connection in base_change_set_component
+            .map_err(Box::new)?;
+        #[derive(Hash, Clone, PartialEq, Eq)]
+        struct UniqueEdge {
+            to_component_id: ComponentId,
+            from_component_id: ComponentId,
+            from_socket_id: OutputSocketId,
+        }
```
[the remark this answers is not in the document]
britmyerss: @jhelwig my thought was that the input socket ids might change (for example if the component was upgraded) - but yeah your example is a good one that I hadn't thought of... we should probably handle it

jobelenus:
Realized what is happening here. The edges are trying to draw to the upgraded component which has been removed/replaced. But the replacement isn't drawn on the diagram, so its looking for rendered socket positions which don't exist.

> OK, there is only one weird thing happening. When upgrading a component the edge goes over the wire correctly. But it isn't visible until you move either component the edge is connecting. WTF?!

jobelenus:
@zacharyhamm the only unexpected thing that happened is, when upgrading a component, the edges between the upgraded component returned by `Diagram::assemble` are "Added" rather than "Unmodified", which is _technically_ true because they connect on a different socket ID on the upgraded component. Before I rebased your changes in they were "Unmodified".

PR #4331, opened 2024-08-14
Cleanup and encapsulate the logic for finding inferred connections

britmyerss:
This PR includes bug fixes and cleanup as the functionality of Frames has evolved since it was first ported to the new engine and there was lingering traversal bugs depending on if you were searching for Outgoing or Incoming inferred connections. 

There is a new module InferredConnectionGraph which encapsulates the logic around finding Inferred Connections (incoming and outgoing), moving this logic out of the Component module.  

All of the existing tests apply and I added a new one to account for the fix to ensure that multiple Input Sockets for an Up Frame don't connect to the same Output Socket.

<div><br/>via <a href="https://giphy.com/kiszkiloszki/">Kiszkiloszki</a> on <a href="https://giphy.com/gifs/art-loop-hand-bgT142FWjRwV4hWsvX">GIPHY</a></div>

### DIFF
--- a/lib/dal/src/attribute/value.rs
+++ b/lib/dal/src/attribute/value.rs
@@ -57,7 +57,7 @@ pub use dependent_value_graph::DependentValueGraph;
 
 use crate::attribute::prototype::AttributePrototypeError;
 use crate::change_set::ChangeSetError;
-use crate::component::InputSocketMatch;
+use crate::component::socket::ComponentInputSocket;
 use crate::func::argument::{FuncArgument, FuncArgumentError};
 use crate::func::intrinsics::IntrinsicFunc;
 use crate::func::runner::{FuncRunner, FuncRunnerError};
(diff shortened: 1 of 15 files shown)


lib/dal/src/attribute/value.rs
```diff
@@ -719,40 +719,38 @@ impl AttributeValue {
...
-                        input_socket_match.component_id,
-                        output_match.component_id,
+                        component_input_socket.component_id,
+                        component_output_socket.component_id,
                     )
                     .await
                     .map_err(Box::new)?
                     {
-                        return Ok(vec![]);
+                        // XXX: We need to properly handle the difference between "there is
```
nickgerace, line 743: Is this check no longer necessary?

lib/dal/src/component/inferred_connection_graph.rs
```diff
@@ -0,0 +1,610 @@
...
+//! the [`ComponentType`], and the [`SocketArity`], in addition to where a [`Component`] exists
+//! relative to its parents, children, siblings, and all components in the 'tree'.
+//!
+//! A [`InferredConnectionGraph`] can be assembled by providing one or many [`ComponentId`]s, building
+//! one or many trees that where the [`Component`] exists, then interrogating each [`Component`]'s [`InputSocket`]s
+//! to find which [`OutputSocket`]s they can be connected to.
+//!
+//! As [`OutputSocket`]s do not care which [`InputSocket`]s are pulling data from them, we
+//! build this mapping only from the perspective of the [`InputSocket`] and use that mapping to hydrate both
+//! the Incoming and Outgoing Inferred Connections for a given [`ComponentId`]
```
nickgerace, line 13: Nice.

lib/dal/src/attribute/value.rs
britmyerss: no it is, I thought about encapsulating that, but we need to return the connections even if data shouldn't flow (because we show them on the UI)

PR #4356, opened 2024-08-19
Implement USR2 signal handler

johnrwatson:
Implements USR2 signal interpreter on a new tokio thread which can then be used by the applications to run custom business logic like:
- Go into maintenance mode/take yourself out of the working pool
- Change application behaviour on an individual node basis
- Debug out current state of running applications/like a statuscheck

Initially, we can use this signal handler to move a node out of the working pool/for maintenance mode to prevent write corruption while updating the application binaries.

With this PR, it just logs out a message after the message has been caught successfully

<hr/>

chore: Fixes minor variable-naming issue within rebaser which was inconsistent based on the other binaries.

### DIFF
--- a/bin/cyclone/src/main.rs
+++ b/bin/cyclone/src/main.rs
@@ -32,6 +32,9 @@ async fn main() -> Result<()> {
         telemetry_application::init(config, &task_tracker, shutdown_token.clone())?
     };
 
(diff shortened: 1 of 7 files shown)


johnrwatson:
Superseeding with new branch

PR #4357, opened 2024-08-19
NodeWeights for all!

jhelwig:


lib/dal/examples/rebase/main.rs
```diff
@@ -5,7 +5,7 @@ use si_layer_cache::db::serialize;
 
 use dal::{
     workspace_snapshot::graph::{correct_transforms::correct_transforms, RebaseBatch},
-    WorkspaceSnapshotGraphV2,
+    WorkspaceSnapshotGraphVCurrent,
```
nickgerace, line 8: Painful naming but I can't think of better 🥴

lib/dal/src/action/prototype.rs
```diff
@@ -50,7 +55,7 @@ pub enum ActionPrototypeError {
     #[error("schema variant error: {0}")]
     SchemaVariant(#[from] SchemaVariantError),
     #[error("schema variant not found for prototype: {0}")]
-    SchemaVariantFoundForPrototype(ActionPrototypeId),
+    SchemaVariantNotFoundForPrototype(ActionPrototypeId),
```
nickgerace, line 58: That must have been confusing

lib/dal/src/layer_db_types/content_types.rs
```diff
@@ -435,48 +451,66 @@ pub struct SchemaContentV1 {
...
-                    finalized_once: v1.finalized_once,
-                    is_builtin: v1.is_builtin,
-                    is_locked: true,
-                })
-            }
-            later => later,
-        };
+        let mut working_content = self;
+        loop {
+            working_content = match working_content {
```
nickgerace, line 466: Good strategy for this

lib/dal/src/layer_db_types/content_types.rs
```diff
@@ -435,48 +451,66 @@ pub struct SchemaContentV1 {
...
+            };
+        }
 
         // extract latest
-        let latest = match at_least_v2 {
-            SchemaVariantContent::V2(v2) => v2,
+        let latest = match working_content {
+            SchemaVariantContent::V3(v3) => v3,
             _ => unreachable!(),
         };
```
nickgerace, line 515: Feels like we could refactor to remove the blank (😱) unreachable. This is nit though.

lib/dal/src/workspace_snapshot/graph/detect_updates.rs
```diff
@@ -13,7 +13,7 @@ use crate::{
     EdgeWeight, EdgeWeightKind, EdgeWeightKindDiscriminants,
 };
 
-use super::WorkspaceSnapshotGraphV2;
+use super::WorkspaceSnapshotGraphV3;
```
nickgerace, line 16: Maybe we want the type aliased name used everywhere (i.e. `WorkspaceSnapshotGraphVCurrent`)?

lib/dal/src/workspace_snapshot/graph/v3.rs
```diff
@@ -0,0 +1,1565 @@
...
+        source: Option<Ulid>,
+        kind: CategoryNodeKind,
+    ) -> WorkspaceSnapshotGraphResult<Option<(Ulid, NodeIndex)>> {
+        let source_index = match source {
+            Some(provided_source) => self.get_node_index_by_id(provided_source)?,
+            None => self.root_index,
+        };
+
+        // TODO(nick): ensure that two target category nodes of the same kind don't exist for the
+        // same source node.
```
nickgerace, line 321: Well... maybe someday.

lib/dal/src/workspace_snapshot/migrator/v3.rs
nickgerace, line 1: Looks good to me.

nickgerace (approved):

lib/dal/src/workspace_snapshot/graph/detect_updates.rs
jhelwig: I kind of left this one since I'm not sure whether detect updates may end up living as a thing that's part of the version of the graph format. (So `workspace_snapshot/graph/v57/detect_updates.rs`.) The idea was that anything written against a specific version of the graph would not use the `VLatest` type alias.

Ideally, that `VLatest` alias wouldn't be necessary at all, but we have a ton of code living in places outside of `WorkspaceSnapshot` & `WorkspaceSnapshotGraph` that have intimate knowledge of the internal workings of `WorkspaceSnapshotGraph`. I'd like to keep us moving in a direction of pushing all of that code down into the versioned `WorkspaceSnapshotGraphV...` with higher-level interfaces exposed out through `WorkspaceSnapshot`. So, basically the idea we started with at the very beginning of the graph work, but that we (as a group) weren't great on keeping up on.

nickgerace: Gotcha, that sounds good. I can see why the alias is a good stopgap for now then.

lib/dal/src/layer_db_types/content_types.rs
zacharyhamm: it may make sense to return an error here, although we know this is actually unreachable, so I'm ok with it.

lib/dal/src/workspace_snapshot/node_weight/traits.rs
```diff
@@ -23,10 +27,162 @@ pub type CorrectTransformsResult<T> = Result<T, CorrectTransformsError>;
...
-        _workspace_snapshot_graph: &WorkspaceSnapshotGraphV2,
+        _workspace_snapshot_graph: &WorkspaceSnapshotGraphV3,
         updates: Vec<Update>,
         _from_different_change_set: bool,
     ) -> CorrectTransformsResult<Vec<Update>> {
         Ok(updates)
     }
 }
+
+// XXX: Should this actually require Sized? rustc is very unhappy about CorrectExclusiveOutgoingEdge without it.
```
zacharyhamm, line 38: I believe the higher-ranked trait bound on CorrectExclusiveOutgoingEdge has an implicit `Sized` bound, and we could maybe change it to `?Sized` to get rid of the `Sized` requirement here

lib/dal/src/layer_db_types/content_types.rs
zacharyhamm: On a second thought: we should implement a proc macro at some point to handle this gradual upgrade process so that we can always be sure it's unreachable.

lib/dal/src/workspace_snapshot/node_weight/traits.rs
jhelwig: I don't know that anything really needs to change. I was just surprised by the error I was getting without the explicit trait bound here.

PR #4369, opened 2024-08-19
ENG-2639: collapsed frames

jobelenus:
- Use diagram outliner to collapse frames
- Use context menu to collapse frames
- Add caret icon to frame upper left to collapse frames
- Store positions, sizes, and collapsed state in browser local storage
- Incoming connections terminate at the center of the frame
- Display children stats for collapsed frames in the bottom right corner
- Removing resize handles on collapsed frame
- Collapsed frames can be moved
- Enforce boundary box on collapsed child within a parent
- Send stats to posthog


wendybujalski (approved):
😎👍

PR #4382, opened 2024-08-20
feat: add maintenance timeout to toast as 60s

johnrwatson:
Adds a basic timeout to the maintenance toast so it doesn't need to be dismissed by the user. We don't need a reload here either as the service should just kick back in and resolve the session again on the next API call.

<hr/>

NB: Also fixed a minor typo in `sed s/maintanance/maintenance/g` file naming


johnrwatson:
Accidentally rebased in Nick's commits, will re-do

PR #4385, opened 2024-08-21
feat(module-index, sdf, dal): Add ability to understand if contribution made

stack72:
When we make a contribution, we now want to store the schema_Variant_id and the schema_variant_version so that we remove the contribute button if the user has already sent the contribution

This makes it a nicer user experience! In the future, we may want to mark assets that have been contributed by the user and that have not made their way back to the system

https://github.com/user-attachments/assets/3967fb74-0e71-4993-8a28-ecc1efb4e553


zacharyhamm (approved):
looks straightforward to me!

PR #4386, opened 2024-08-21
Allow components with secrets to be copy/pasted

jkeiser:
With @nickgerace

Right now, if you try to copy/paste a component with a credential hooked up, we fail to copy and toss an error.

### How it works now

This is because when we copy/paste the attributes to the new component, we currently:

1. Extract the attribute value as JSON and set the new attribute value to it (which removes reactive links to secret nodes, input/output sockets, and props).
2. Set the name and clear the resource_value.
3. Go back and fix up the prototypes to point to the correct nodes.

However, when there is a secret, we don't get that far, because when we try to set a secret to a JSON value it immediately throws an error trying to get the secret to compute dependent values.

### This PR

This PR straight up clones the attribute prototypes and skips the JSON setting, and does the name setting and resource clearing at the end. This should get rid of dependent values churn during copy/paste as well as make the process simpler.

While trying to untangle the logic, I built a helper to iterate child attribute pairs (which I've seen done several other places, so we should be able to consolidate).

### Roads not taken

An alternate path not explored was to share code with the "upgrade component to new schema variant" code, which also clones a value and carries data from old to new. It's a very different path, however, and it's unclear how much code sharing there really is to be had besides a superficial structure similarity.

NOTE: I will squash this down before pushing; right now the commits are self-contained enough that some people might prefer to look at the commits, but they don't matter long term.

### DIFF
--- a/lib/dal/src/attribute/value.rs
+++ b/lib/dal/src/attribute/value.rs
@@ -127,6 +127,12 @@ pub enum AttributeValueError {
     ChangeSet(#[from] ChangeSetError),
     #[error("component error: {0}")]
     Component(#[from] Box<ComponentError>),
+    #[error("duplicate key or index {key_or_index} for attribute values {child1} and {child2}")]
+    DuplicateKeyOrIndex {
+        key_or_index: KeyOrIndex,
+        child1: AttributeValueId,
+        child2: AttributeValueId,
+    },
     #[error("empty attribute prototype arguments for group name: {0}")]
     EmptyAttributePrototypeArgumentsForGroup(String),
     #[error("func error: {0}")]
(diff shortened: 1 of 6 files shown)


lib/si-pkg/src/spec/attribute_value.rs
```diff
@@ -34,6 +35,14 @@ impl fmt::Display for KeyOrIndex {
         Ok(())
     }
 }
+impl Hash for KeyOrIndex {
```
zacharyhamm, line 38: surprised you can't derive a hash for this type

lib/dal/src/attribute/value.rs
```diff
@@ -2211,6 +2251,67 @@ impl AttributeValue {
...
+            let key_or_index = match &key {
+                Some(key) => KeyOrIndex::Key(key.clone()),
+                None => KeyOrIndex::Index(index as i64),
+            };
+            if let Some(old_index) = pair_index.get(&key_or_index) {
+                return Err(AttributeValueError::DuplicateKeyOrIndex {
+                    key_or_index,
+                    // It is impossible at this point to have anything except FirstOnly
+                    // in the vec.
+                    child1: pairs[*old_index].first().unwrap(),
```
zacharyhamm, line 2276: Although logically impossible to panic here, this should be written to be panic free, just in case:

`pairs.get(*old_index).and_then(|pair| pair.first())`. You can just make the child1 type on this error an `Option` to prevent having to `ok_or` or `unwrap`

zacharyhamm (changes requested):
this looks great! Although I did not build and test it, it makes sense. The "pairs" concept was a little confusing at first but makes sense. Might be good to explain the usage more clearly in the comments.

One thing we want to avoid is the call to unwrap and the direct vec indexing with square brackets. Both of those can panic.

lib/dal/src/attribute/value.rs
zacharyhamm: (our linter will not allow you to ship this with an unwrap, which is likely why it's failing)

lib/si-pkg/src/spec/attribute_value.rs
zacharyhamm: try adding #[derive(Hash)] and see if it works :)

jkeiser: I don't know why I thought I couldn't ... it works fine!

lib/dal/src/attribute/value.rs
jkeiser: Do we mind the use of `unreachable!()` or will that trigger lints too?

zacharyhamm: unreachable! won't trigger lints, no, and is fine if you know the spot is really unreachable. But if you can produce an error instead, that's always better

jkeiser: I think I found a way to avoid the lint; as long as I keep first_children around I can just reach inside that to get the old ID.

zacharyhamm: the direct indexing is also a panic case. if the index is not there, it panics

zacharyhamm (approved):

lib/dal/src/attribute/value.rs
jkeiser: Fixed as suggested by stuffing an existing index in there if it's None, and adding a comment noting that it should never happen.

jkeiser: I just now grokked the putting an Option in the error. I'm ok either way if you'd prefer that!

nickgerace (approved):

PR #4392, opened 2024-08-21
fix(dal): send fewer status updates in dvu

zacharyhamm:
Streaming update messages for every attribute value produced a ton of data, putting pressure on both nats and on our frontend which had to store and process thousands of update messages for large workspaces. Instead just add a start and finish message for each individual component touched in the dependent values update job.


zacharyhamm:
Want to test this a little more before merging

PR #4399, opened 2024-08-22
feat(pinga): consume work from a Jetstream work queue

fnichol:
This change refactors the Pinga server to consume its work (i.e. job requests) via a NATS Jetstream work queue. Leaving outstanding work on an external queue means that a Pinga instance won't lose pending requests when it's shut down or restarted.

The refactoring uses our `lib/naxum` crate to consume the requests from a Jetstream Stream and otherwise keeps the existing logic largely in place. Importantly:

- Processing a job request will almost unconditionally be successfully consumed. This means that if a job itself fails, it will be considered an "acked" message and a successful job run. This matches the prior behavior and can be changed in the future. The implication is that NATS currently will not help to re-deliver a job that has already been run (even if it failed).
- Jobs that have failed will no longer be tracked via the `JobFailure` DAL object and related database table. It appears as though there is no consumer of this data and it's very likely that the `JobFailure` concept came about before we improved our tracing/logging/telemetry. If this is required again, it can be revived.
- A "blocking" job will still use a core NATS reply subject as a mechanism to reply to the caller of the job. Core NATS seems to be the appropriate medium in this case as there *should* already be a subscriber to that inbox and if the caller goes away, then core NATS will simply drop the reply as no one is listening. A reply subject is added the job request via a NATS message header as the request is published on a NATS Jetstream Stream and there isn't a user-facing API to provide such a reply subject (which is by design).

In addition to using `lib/naxum` in general the following structural changes have been made:

- Remove the old graceful shutdown mechanics and rely on `CancellationToken` and `TaskTracker` types instead (these work better with naxum anyway).
- Both a Pinga server and any producer of job requests (that would be a DAL `Context` that is committing with pending jobs) will attempt to create the `PINGA_JOBS` Jetstream Stream instance, ensuring that the cold boot order of services is not important.

There is still further work to bring Pinga to a behavioral parity with prior version and still more work to address several known issues:

- The span-per-request facility isn't fully threaded through and isn't inherently OpenTelemetry-aware.
- It's unclear if some of the instrumented spans are being dropped or grouped differently without this consistent request span.
- On a graceful shutdown, any jobs in flight will simply be cancelled, whereas we would prefer the in flight jobs be given a window of time to complete before finishing a shutdown.
- Not all concurrency metrics are present. These are better delivered as middleware that can wrap around the inner service, but this hasn't been written up yet (note that this middleware can be used across Pinga, Veritech, and Rebeaser services).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1345,6 +1345,7 @@ dependencies = [
  "once_cell",
  "paste",
  "petgraph",
+ "pinga-core",
  "postcard",
  "postgres-types",
  "pretty_assertions_sorted",
@@ -3790,6 +3791,13 @@ dependencies = [
 ]
 
 [[package]]
+name = "pinga-core"
+version = "0.1.0"
+dependencies = [
+ "si-data-nats",
+]
+
+[[package]]
 name = "pinga-server"
 version = "0.1.0"
 dependencies = [
@@ -3798,6 +3806,8 @@ dependencies = [
  "derive_builder",
  "futures",
  "nats-subscriber",
+ "naxum",
+ "pinga-core",
  "remain",
  "serde",
  "serde_json",
@@ -3814,6 +3824,7 @@ dependencies = [
  "tokio",
  "tokio-stream",
  "tokio-util",
+ "tower",
  "ulid",
  "veritech-client",
 ]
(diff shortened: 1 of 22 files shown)


nickgerace (approved):

nickgerace:
ENG-2651

PR #4403, opened 2024-08-22
chore(deployment): Post start and stop to discord

stack72:
This will post start and finish of production deployments to discord

### DIFF
--- a/.github/workflows/deploy-stack.yml
+++ b/.github/workflows/deploy-stack.yml
@@ -1,9 +1,10 @@
 name: Deploy Services
 
-run-name: Deploy of ${{ inputs.service }} to ${{ inputs.environment }} by @${{ github.actor }}
+run-name:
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-stack.yml
```diff
@@ -76,12 +77,27 @@ jobs:
             echo 'services=["${{ inputs.service }}"]' >> "$GITHUB_OUTPUT"
           fi
 
+  notify-start:
+    name: Start Deployment Notification
+    runs-on: ubuntu-latest
+    if: github.event.inputs.environment == 'production'
```
sprutton1, line 83: I _think_ you need to use `${{ inputs.environment }}` since this uses both workflow call and dispatch. I seem to remember getting caught up on this at one point.

stack72:
Closing this out as this will be handled in the new deployment stream work

PR #4425, opened 2024-08-24
feat(dal): graph goes vroom

zacharyhamm:
Profiling of slow rebases with lots of updates revealed that most of the time was being spent in replace_references, both calculating the merkle tree hash and cloning node weights in copy_node_index_by_id. This "copy on write" logic no longer makes sense to me with the operational transforms. And in any case, we are still copy on write (from the read only graph to the working copy). This PR removes the "copy on write" logic which no longer makes sense. Instead we mutate the graph in place, and keep a list of touched nodes. Using that list of touched nodes, we recalculate the merkle tree hash for the graph once, in `cleanup_and_merkle_tree_hash`. The result is a massive speed up in graph operations. Even huge rebases come in under a second, and usually under 300ms. Small rebase operations stay very fast, typically under 50ms.

### DIFF
--- a/clippy.toml
+++ b/clippy.toml
@@ -3,4 +3,8 @@ disallowed-methods = [
     { path = "std::env::vars", reason = "should not directly access environment variables within library crates, favor configuration injection, passing parameters, etc." },
     { path = "std::env::var_os", reason = "should not directly access environment variables within library crates, favor configuration injection, passing parameters, etc." },
     { path = "std::env::vars_os", reason = "should not directly access environment variables within library crates, favor configuration injection, passing parameters, etc." },
+    { path = "dal::workspace_snapshot::WorkspaceSnapshot::write_working_copy_to_disk", reason = "The snapshot should only be written to disk when debugging"},
+    { path = "dal::workspace_snapshot::WorkspaceSnapshot::write_readonly_graph_to_disk", reason = "The snapshot should only be written to disk when debugging"},
+    { path = "dal::workspace_snapshot::WorkspaceSnapshot::tiny_dot_to_file", reason = "The snapshot should only be written to disk when debugging"},
+    { path = "dal::workspace_snapshot::graph::RebaseBatch::write_to_disk", reason = "Rebase batches should only be written to disk when debugging"},
 ]
--- a/lib/dal/src/workspace_snapshot.rs
+++ b/lib/dal/src/workspace_snapshot.rs
@@ -355,18 +355,22 @@ impl WorkspaceSnapshot {
         let self_clone = self.clone();
         let updates = slow_rt::spawn(async move {
             let mut working_copy = self_clone.working_copy_mut().await;
-            working_copy.cleanup();
+            working_copy.cleanup_and_merkle_tree_hash()?;
 
-            self_clone.read_only_graph.detect_updates(&working_copy)
+            Ok::<Vec<Update>, WorkspaceSnapshotGraphError>(
+                self_clone.read_only_graph.detect_updates(&working_copy),
+            )
         })?
-        .await?;
+        .await??;
 
         Ok((!updates.is_empty()).then_some(RebaseBatch::new(updates)))
     }
 
     /// Calculates the set of updates made to `updated_snapshot` against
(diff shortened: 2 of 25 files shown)


lib/dal/src/workspace_snapshot/graph.rs
```diff
@@ -1590,6 +1525,36 @@ impl WorkspaceSnapshotGraphV2 {
         Ok(())
     }
 
+    /// Does a dfs post-order walk of the tree, recalculating the merkle tree
+    /// hash for any nodes we have touched while working on the graph. Should be more
+    /// efficient than recalculating the entire merkle tree hash, since we will
+    /// only walk "up" the paths that have been touched and thus need to be
+    /// recalculated.
```
jhelwig, line 1532: From the doc comment, it sounds like the intent was that only a subset of the graph would be walked, but this will walk the entire graph and only recalculate on nodes from the `touched_node_indices` up to the root. Is the doc comment inaccurate, or was the implementation intended to be different?

zacharyhamm: I'll update the comment. It walks the whole graph but it only updates the merkle tree hash for the paths that have children in the touched node list. The walk is fast, the merkle tree hash part is slow

clippy.toml
```diff
@@ -3,4 +3,8 @@ disallowed-methods = [
     { path = "std::env::vars", reason = "should not directly access environment variables within library crates, favor configuration injection, passing parameters, etc." },
     { path = "std::env::var_os", reason = "should not directly access environment variables within library crates, favor configuration injection, passing parameters, etc." },
     { path = "std::env::vars_os", reason = "should not directly access environment variables within library crates, favor configuration injection, passing parameters, etc." },
+    { path = "dal::workspace_snapshot::WorkspaceSnapshot::write_working_copy_to_disk", reason = "The snapshot should only be written to disk when debugging"},
+    { path = "dal::workspace_snapshot::WorkspaceSnapshot::write_readonly_graph_to_disk", reason = "The snapshot should only be written to disk when debugging"},
+    { path = "dal::workspace_snapshot::WorkspaceSnapshot::tiny_dot_to_file", reason = "The snapshot should only be written to disk when debugging"},
+    { path = "dal::workspace_snapshot::graph::RebaseBatch::write_to_disk", reason = "Rebase batches should only be written to disk when debugging"},
```
nickgerace, line 9: Thank you

lib/dal/src/workspace_snapshot.rs
```diff
@@ -355,18 +355,22 @@ impl WorkspaceSnapshot {
...
-            working_copy.cleanup();
+            working_copy.cleanup_and_merkle_tree_hash()?;
 
-            self_clone.read_only_graph.detect_updates(&working_copy)
+            Ok::<Vec<Update>, WorkspaceSnapshotGraphError>(
+                self_clone.read_only_graph.detect_updates(&working_copy),
+            )
         })?
-        .await?;
+        .await??;
```
nickgerace, line 364: Took me a few times to read but got it. You're directly returning the updates because now the cleanup method is fallible and you need to be explicit about the return type.

lib/dal/src/workspace_snapshot/graph.rs
```diff
@@ -181,6 +198,15 @@ impl WorkspaceSnapshotGraphV2 {
...
     }
 
+    /// Add a node to the list of touched nodes, so that it is taken into
+    /// account when recalculating the merkle tree hash for this graph. If a
+    /// node weight is modified, or if a an outgoing edge is added or removed
+    /// to/from this node, you must touch the node, or the merkel tree hash will
+    /// not be updated correctly.
+    pub fn touch_node(&mut self, node_index: NodeIndex) {
+        self.touched_node_indices.insert(node_index);
+    }
```
nickgerace, line 208: Clever

nickgerace (approved):

lib/dal/src/workspace_snapshot.rs
zacharyhamm: yeah the async closure needs to get a return type somewhere so that `?` works

PR #4428, opened 2024-08-26
Fix error updating secrets

jkeiser:
Right now, if you update a secret in-place and then quickly click on a component that uses it (after rebase but before DVU), an error toast pops up saying it couldn't find the secret.

This is because when we get the property values for a component for editing, we try to look up the secret by matching the key stored in the dependent value (cached in the node) against the keys in all Secret node. Since the secret key has since been updated, the secret can't be found. It eventually gets better, but it happens enough to drive John Watson crazy.

This change gets us past the error by sending back null (telling the frontend it's not set). Going forward, we'll talk through how to handle situations where dependent values aren't up to date with @wendybujalski and @jobelenus -- there is a general problem to solve there, as @jhelwig points out. And we should consider whether we should include more information (such as the secret id) in the JSON value for secrets, to reduce how often this will affect people. (Even if we did that, the problem would persist when the secret is *deleted*.)

This only affects the property editor value calculation.

Fixes BUG-481.


jobelenus:
@jkeiser Thanks for the call out here! The only remaining thing on my mind here is, perhaps instead of `debug!` when we hit this null case—do we want to add an `error!` so that we can measure in honeycomb how often this exact problem is effecting people? That might help us prioritize when/how much energy to put into a better resolution here?

PR #4430, opened 2024-08-26
chore: refactor pool_noodle loop to use cancellation token and config

sprutton1:
Post the Veritech Jetstream changes, this refactors PoolNoodle to run over a single loop that works based on tasks in a queue. The veritech cancellation token is threaded through to we shutdown and clean up when veritech is shut down. We now only spin up as many tasks as there is max_concurrency to help manage back pressure (it's currently set to an arbitrarily high number).

This also does some quality of life changes, a config struct, better defaults, and more metrics to track things like the number of jobs waiting for instances when we're under enough pressure.

Here's an undersized pool fighting for its life.
image

### DIFF
--- a/lib/si-pool-noodle/src/pool_noodle.rs
+++ b/lib/si-pool-noodle/src/pool_noodle.rs
@@ -1,19 +1,18 @@
+use crate::lifeguard::LifeGuard;
+use crate::task::{PoolNoodleTask, PoolNoodleTaskType};
 use crossbeam_queue::ArrayQueue;
-use std::fmt;
 use std::fmt::Display;
-use std::fmt::Formatter;
 use std::result;
 use std::sync::Arc;
 use telemetry_utils::metric;
-use tokio::time::sleep;
-use tokio::time::Duration;
+use tokio::time::{self, sleep};
 use tokio_util::sync::CancellationToken;
-use tracing::{debug, error, info, warn};
+
+use tokio::time::Duration;
+use tracing::{debug, info, warn};
 
 use crate::errors::PoolNoodleError;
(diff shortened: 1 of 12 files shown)


lib/si-pool-noodle/src/pool_noodle.rs
```diff
@@ -76,233 +107,185 @@ where
...
-                            match PoolNoodleInner::spawn(id, &me.spec).await {
-                                Ok(instance) => {
-                                    debug!("PoolNoodle: instance started: {}", id);
-                                    Self::push_to_ready(me.clone(), instance);
+                tokio::select! {
+                    _ = me.shutdown_token.cancelled() => {
+                        debug!("main loop received cancellation");
+                        break
+                    }
+                    Ok(permit) = semaphore.acquire() => {
```
nickgerace, line 146: We discussed potential failure modes here and @sprutton1 is going to confirm if this works as intended

lib/si-pool-noodle/src/pool_noodle.rs
```diff
@@ -76,233 +107,185 @@ where
...
-                        }
-                        Err(e) => {
-                            warn!("PoolNoodle: failed to clean instance: {}", id);
-                            warn!("{:?}", e);
-                            Self::push_to_clean(me.clone(), id);
-                        }
+            // drain the pool of cleanup tasks and drop all prepare tasks
+            while !me.work_queue.is_empty() {
+                let me = me.clone();
+                match me.work_queue.pop() {
```
nickgerace, line 180: We discussed potentially simplifying this to `while let Some`

nickgerace (approved):

PR #4433, opened 2024-08-26
Further destroy the existence of "nilId" (ENG-2469)

nickgerace:
## Description

This PR furthers the destruction of `nilId` throughout the codebase. It does this by migrating as many dal-based identifiers from the `pk!` macro to the `id!` macro as possible. The macros are extremely similar, but the `id!` macro does not include the ability to create `nilIds` easily.

All remaining locations using the old `pk!` macro have been marked with a `TODO` comment indicating what's to come.

## Reminder

This macro (`id!`) works for primary keys, just as the old macro (`pk!`) worked for identifiers in the past.

## Additional Changes

In addition to these changes both the `node.rs` and `job_failure.rs` modules have been removed due to lack of use. The job failures table will still be migrated in order to not interrupt existing SI deployments, but it is effectively dead.

You may also notice that `AttributePrototypeSource` had its `EnumIter` and `EnumString` derived traits removed. This is because they rely on the identifiers to implement or derive `Default`, which the new macro does not provide an implementation for (intentionally).


nickgerace:
Local testing succeeded including actions, authoring, creating and attaching new funcs, etc.

PR #4436, opened 2024-08-27
Break up WorkspaceSnapshotGraph & Migrator into versioned module namespacing

jhelwig:
Since we're planning on moving to a graph V3 in the near future, this helps create a structure to keep things more organized than having every version ever in the same file.


nickgerace (approved):

lib/dal/src/workspace_snapshot/migrator.rs
```diff
@@ -135,106 +139,29 @@ impl SnapshotGraphMigrator {
...
+                }
+                WorkspaceSnapshotGraph::V1(inner_graph) => {
+                    working_graph = WorkspaceSnapshotGraph::V2(migrate_v1_to_v2(inner_graph)?);
+                }
+                WorkspaceSnapshotGraph::V2(_) => {
+                    // Nothing to do, this is the newest version,
+                    break;
                 }
             }
         }
```
fnichol, line 162: Yeah, I like this!

lib/dal/src/workspace_snapshot/migrator/v2.rs
```diff
@@ -0,0 +1,92 @@
...
+        content_address::ContentAddress,
+        graph::{deprecated::DeprecatedWorkspaceSnapshotGraphV1, LineageId},
+        node_weight::NodeWeight,
+    },
+    EdgeWeight, WorkspaceSnapshotGraphV2,
+};
+
+use super::SnapshotGraphMigratorResult;
+
+pub fn migrate_v1_to_v2(
```
fnichol, line 16: In a future change we might want to consider wrapping an `#[instrument]` around this one in order to have some "upgrade cost" info. Kind of nice that it's in one place

fnichol (approved):
Yeah, I'm personally really liking the look of this, nicely done!

PR #4438, opened 2024-08-27
Convert veritech to use naxum and jetstream (ENG-2652)

nickgerace:
## Description

This PR converts veritech to use both naxum and jetstream, like #4399 did for pinga. By using both technologies, we now have durable queues for function requests and easier flexibility for running multiple veritech instances in production systems. The "kill execution" functionality for veritech still uses core NATS due to its need for fan-out-style communication, but it also uses naxum for its handler, extractor, etc. benefits.

## Additional Changes

We've also removed access to reconciliation functions at the veritech level. Sadly, this commit does not remove reconciliation functions from layers "above" veritech (e.g. the dal) nor does it remove reconciliation function from layers "below" veritech (e.g. cyclone and lang-js). This is because this commit is already large enough and aims to keep the bulk of changes within veritech itself.

This commit also renames "cancel execution" to "kill execution" to unify language across the codebase. As a result of this change, we needed to alias `FuncRunState::Cancelled` to the new enum variant `FuncRunState::Killed`. You'll change telemetry-related changes to this renaming as well.

Veritech shutdown handle has been removed in favor of shutdown tokens, much like pinga and our other services. Remaining pinga shutdown handle code has been remove from dal-test as well.

The `FuncDispatchContext` now requires `WorkspaceId` and `ChangeSetId` to be added. This was necessary to satisfy naxum-specific subject naming conventions. It may also open the door to more precise tracing at the veritech level.

## Disclaimer

Tracing and telemetry may have drifted at the veritech level, but should have remained in tact. This goal of this commit is keep veritech functionally the same with follow-up tunings for tracing, usability, etc.


nickgerace:
Merging. All tests pass. Func killing works with the admin dashboard. Actions run after composition with secrets, frames and generic fields. Qualifications pass. Authoring works. All mentioned features work interwoven with one another.

nickgerace:
Pain.
image

PR #4439, opened 2024-08-27
feat: Return 422 on user code errors

vbustamante:
To ensure bad user code is not reported as server errors, we return 422 Unprocessable Content when that's the reason for an API call failure.

To enable this, I had to make the language server errors static and parseable into a rust enum, that has an extra UserCodeException variant containing a string so easily support JS errors on the frontend.

This PR also cleans up code error messages in the frontend

image

### DIFF
--- a/lib/cyclone-core/src/progress.rs
+++ b/lib/cyclone-core/src/progress.rs
@@ -1,7 +1,5 @@
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
-
-/// The error kind for function result failures when the error occurs within the veritech server.
-const FUNCTION_RESULT_FAILURE_ERROR_KIND_VERITECH_SERVER: &str = "veritechServer";
+use strum::Display;
 
 /// A line of output, streamed from an executing function.
 ///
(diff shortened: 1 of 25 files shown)


lib/cyclone-core/src/progress.rs
```diff
@@ -146,9 +144,20 @@ impl FunctionResultFailure {
...
+    InvalidReturnType,
+    ReconciliationFieldWrongType,
+    UserCodeException(String),
+    VeritechServer,
+}
+
 #[derive(Debug, Deserialize, Eq, PartialEq, Serialize, Clone)]
 pub struct FunctionResultFailureError {
-    pub kind: String,
+    pub kind: FunctionResultFailureErrorKind,
```
vbustamante, line 160: This was the scariest part to add, since it's carried all the way from cyclone back to the DAL basically. From what I've tested it does not break any protocols but I'd appreciate extra eyes

nickgerace: Looks good to me. You also don't have to worry about #4438 because we only dropped `Reconciliation` funcs from veritech (still "above" in dal, still "below" in cyclone and lang-js).

lib/veritech-server/src/server.rs
```diff
@@ -92,6 +94,26 @@ pub enum ServerError {
...
+                timestamp(),
+            ),
+            _ => FunctionResultFailure::new_for_veritech_server_error(
+                execution_id,
+                self.to_string(),
+                timestamp(),
+            ),
+        })
+    }
+}
```
nickgerace, line 115: As a heads up, `Cancelled` becomes `Killed` in #4438.

nickgerace (approved):

lib/veritech-server/src/server.rs
vbustamante: Oh ok, I'll wait for that to merge before merging this one

PR #4444, opened 2024-08-28
feat(naxum): add ability to limit incoming work to a naxum app

fnichol:
This change adds an alternative naxum app setup `naxum::serve_with_incoming_limit` which takes an optional number. When set, the app will limit the number of incoming messages being handled at the same time.


fnichol:
References: ENG-2655

nickgerace (commented):
great!

PR #4446, opened 2024-08-28
chore(pinga,veritech): normalize server config & setups

fnichol:
This change harmonizes the server configuration and server setups between Pinga and Veritech. The following changes are made:

- Veritech takes an `--instance-id` option on the CLI and a random ID is generated if not provided (similar to other services).
- Veritech takes a `--concurrency` option to override the default concurrency limit.
- The concurrency option is `--concurrency` in both Veritech & Pinga.
- The config file value `concurrency_limit` is honored in both Veritech & Pinga.
- Default for concurrency limits are now in each respective server's `config` module and is used to render the value indirectly through its `Config` type.
- Veritech's execution timeout option is updated to `--cyclone-client-execution-timeout-secs` to reflect the numerial value.
- Veritech's execution timeout is a duration and is computed via its `Config` type.
- Veritech's `main.rs` is updated to be follow the same pattern as Pinga and Rebaser.
- Veritech's binary sets a graceful shutdown timeout, currently set to 6 hours (which likely needs adjustment but was inifinte prior to this change).
- Veritech's server setup now uses a `from_config` function, similar to Pinga and Rebaser.

### DIFF
--- a/lib/veritech-server/src/server.rs
+++ b/lib/veritech-server/src/server.rs
@@ -1,87 +1,53 @@
-use futures::{join, Future, StreamExt};
-use naxum::handler::Handler as _;
-use naxum::middleware::ack::AckLayer;
-use naxum::middleware::trace::{DefaultMakeSpan, DefaultOnRequest, TraceLayer};
-use naxum::ServiceBuilder;
-use naxum::ServiceExt as _;
+use std::{
+    collections::HashMap,
+    convert::Infallible,
+    fmt,
+    future::{Future, IntoFuture as _},
+    io,
+    sync::Arc,
+    time::Duration,
+};
 
-use si_crypto::{VeritechDecryptionKey, VeritechDecryptionKeyError};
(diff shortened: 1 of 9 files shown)


lib/veritech-server/src/server.rs
```diff
@@ -102,12 +68,50 @@ impl fmt::Debug for Server {
 }
 
 impl Server {
-    #[instrument(name = "veritech.init.cyclone.uds", level = "info", skip_all)]
-    pub async fn for_cyclone_uds(config: Config, token: CancellationToken) -> ServerResult<Self> {
+    #[instrument(name = "veritech.init.from_config", level = "info", skip_all)]
+    pub async fn from_config(config: Config, token: CancellationToken) -> ServerResult<Self> {
```
nickgerace, line 72: Ah this is welcome

nickgerace (commented):
also great!

PR #4449, opened 2024-08-28
V to drag

umag:
A small quality of life improvement for drag-to-scroll keybinding


umag:
After some tests, maybe it should be another key, like B, cause v interferes a bit with paste

keeb:
Hi @umag - sorry for the delay in getting this reviewed. Thank you very much for the contribution! 

The ultimate decision here is to implement a feature with customizable keybinds as opposed to expanding the existing key mappings.

PR #4453, opened 2024-08-28
Allow ordered task groups for graceful shutdown

nickgerace:


lib/si-service/src/shutdown.rs
```diff
@@ -42,36 +42,44 @@ impl ShutdownError {
...
+        let mut count = 0;
+        for (tracker, token) in trackers_with_tokens {
+            count += 1;
+            info!("performing graceful shutdown for task group {count}");
             tracker.close();
             token.cancel();
+            tracker.wait().await;
         }
-    }
+    };
```
nickgerace, line 77: The PR and commit descriptions get into this further, but using a manger struct could result in better tracing when draining (e.g. "we are now draining the telemetry group").

For now, let's use the raw types as we are refactoring frequently here, at present.

bin/sdf/src/main.rs
```diff
@@ -210,8 +197,20 @@ async fn async_main() -> Result<()> {
...
+        ] {
+            count += 1;
+            info!("performing graceful shutdown for task group {count}");
+            tracker.close();
+            token.cancel();
+            tracker.wait().await;
+        }
+
+        // TODO(nick): we need to handle telemetry shutdown properly as well.
         telemetry_shutdown.wait().await?;
```
nickgerace, line 214: We'll be tackling sdf soon to get it in-line with the other services.

fnichol (approved):
Nicely done, we are on our way to correct, clean, and controlled process shutdown!

PR #4456, opened 2024-08-29
Fix otel missing span for errors

jobelenus:


lib/vue-lib/src/pinia/pinia_api_tools.ts
```diff
@@ -462,6 +461,8 @@ export const initPiniaApiToolkitPlugin = (config: { api: AxiosInstance }) => {
           completed.resolve({
             error: err,
           });
+          span.setAttributes({ "http.status_code": err.response.status });
+          span.end();
```
fnichol, line 465: Ah, was ending the span a necessary part of the solution? I guess that makes a lot of sense, there isn't something like a Rust `Drop` trait to fall back on.

fnichol (approved):
Is it weird that I'm excited for the return of front-end errors?

lib/vue-lib/src/pinia/pinia_api_tools.ts
jobelenus: Yep

PR #4459, opened 2024-08-29
feat(dal): dvu component concurrency limit

zacharyhamm:
Adds a column to the workspaces table that allows us to control how many components can be updated by the DependentValuesUpdate job in one run. When the limit is reached, the job will write the current set of DVU roots back to the graph so the dependency graph can be reconstructed in the state it was when the job finished. Currently every workspace will get a hefty default limit of 256 components, but this is configurable by updating the column in the workspaces table, which the Admin panel will make possible in a future change.

Also reduces the dvu interval for the rebaser to 800ms, so that dvu jobs are enqueued more frequently.

This adds a database migration, and a new node weight, so cannot be rolled back easily (however no graph migration will be required)

### DIFF
--- a/lib/rebaser-server/src/config.rs
+++ b/lib/rebaser-server/src/config.rs
@@ -66,7 +66,7 @@ pub struct Config {
     #[builder(default = "random_instance_id()")]
     instance_id: String,
 
-    #[builder(default = "5000")]
+    #[builder(default = "800")]
     dvu_interval_millis: u64,
 }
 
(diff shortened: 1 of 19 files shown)


lib/rebaser-server/src/config.rs
```diff
@@ -66,7 +66,7 @@ pub struct Config {
     #[builder(default = "random_instance_id()")]
     instance_id: String,
 
-    #[builder(default = "5000")]
+    #[builder(default = "800")]
```
zacharyhamm, line 69: @fnichol can you think of any issues with the leader election stuff that changing this timer might cause @jkeiser

PR #4462, opened 2024-08-30
fix: Clone object children of arrays and maps correctly

vbustamante:
Previously, when upgrading a component and reaching an attribute value that didn't have a correspondent on the new variant yet, we assumed it was a map or array scalar item and created it, which resulted in us not creating attribute values for fields of objects inside arrays and maps. This resulted in unusable array entries on the prop editor. 

Now, when cloning objects that are inside arrays and maps, we forcefully create all attributes.

This unearthed another problem, which was that we were using the prop id + the key to determine equivalency between avs of new and old components, meaning in the case of fields of objects within arrays (which don't have keys or indices themselves), we'd copy the value from random items with the same prop id. This is resolved by matching on the full av path instead of just the key or index

### DIFF
--- a/lib/dal/src/component.rs
+++ b/lib/dal/src/component.rs
@@ -601,7 +601,7 @@ impl Component {
         let self_schema_variant_id = Component::schema_variant_id(ctx, self.id).await?;
         let mut dvu_roots = vec![];
 
-        // Gather up a bunch of data about the current schema variant
+        // Gather a bunch of data about the current schema variant
         let mut new_input_sockets = HashMap::new();
         for input_socket_id in
             InputSocket::list_ids_for_schema_variant(ctx, self_schema_variant_id).await?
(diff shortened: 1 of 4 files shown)


lib/dal/src/component.rs
```diff
@@ -643,202 +644,228 @@ impl Component {
...
+                            // we explicitly set the value on the component to the old
+                            // value it used to have, as if the user had explicitly set it
+                            // themselves. (You could argue they basically implicitly set
+                            // the value of root/si/type when they created the child
+                            // components.)
+                            if prop_path == ["root", "si", "type"] {
+                                let old_value = Prop::default_value(ctx, old_prop_id).await?;
+                                let new_value = Prop::default_value(ctx, new_prop_id).await?;
+                                if old_value != new_value {
+                                    error!("root/si/type different");
```
zacharyhamm, line 762: is this an error?

zacharyhamm (approved):
makes sense, and great find!

lib/dal/src/component.rs
vbustamante: no, it was an error!

PR #4465, opened 2024-08-30
feat(si-service): refactor `shutdown::graceful` to a builder pattern

fnichol:
This change updates the common binary program `graceful()` function into a function that returns an internal builder capable of the same graceful shutdown logic as before. This refactoring will allow for future extensions (i.e. additional optional shutdown scenarios) as well as making the setup a little more ergonomic.


zacharyhamm:
love it!

I was wondering if we wanted to look at refactoring a bunch of our code to use https://elastio.github.io/bon/ which looks really ergonomic

PR #4468, opened 2024-08-30
Post-deploy smoketest tracer bullet

jobelenus:
- Adds the si-smoketest package with an sdf client to allow writing of API tests
- Adds an endpoint that allows direct login to the auth portal, so that we can get an API token to run API test
- Adds a github action to run these tests


vbustamante:
Needs to me merged so we can deploy the auth API and test the actions

PR #4475, opened 2024-09-03
fix(*): upgrade cacache-rs to get panic fixes

zacharyhamm:
Unwraps in cacache-rs are eliminated in a recent PR, which should solve some panics we get (mostly in CI).

For sync-deps, I had to pin serde_json since the latest requires a more recent rust than we use.


vbustamante (approved):
YES

PR #4482, opened 2024-09-04
feat(sdf-server,dal,web): Download workspace snapshots in Admin panel

zacharyhamm:
Adds a workspace and changeset picker, with the ability to download a snapshot to disk directly from SDF.

### DIFF
--- a/app/web/package.json
+++ b/app/web/package.json
@@ -112,6 +112,7 @@
     "@types/node": "^18.15.11",
     "@types/tinycolor2": "^1.4.3",
     "@types/validator": "^13.7.2",
+    "@types/wicg-file-system-access": "^2023.10.5",
     "@vitejs/plugin-vue": "^5.0.3",
     "cypress": "^13.8.1",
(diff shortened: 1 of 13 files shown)


frantzmiccoli:
Is this supposed to be anywhere in the version deployed on https://app.systeminit.com/ because I can't find any way to download an export.

stack72:
> Is this supposed to be anywhere in the version deployed on https://app.systeminit.com/ because I can't find any way to download an export.

Hi @frantzmiccoli 

This is not functionality we expose to the user - this is for troubleshootin

frantzmiccoli:
@stack72 thanks for the heads up. So once on a system, there is a vendor lock-in, correct?

stack72:
> @stack72 thanks for the heads up. So once on a system, there is a vendor lock-in, correct?

No, there is a way for users to export their own workspace. System Initiative doesn't have any layers of abstraction for assets soot resources - we are 1:1 with the cloud APIs themselves so you can easily stop using it and still manage your cloud resources

PR #4490, opened 2024-09-05
feat: API test structural changes

johnrwatson:
Various changes to the libraries + call pattern in the test deno application. 

Adds a structured JSON output based on the execution of the tests itself. 

Allows the users to also specify individual tests which is helpful if you're developing a new one you don't have to run the whole collection all the time.

Test report looks like this for now:
```
Test Report:
[
  {
    "test_name": "create_and_delete_component",
    "start_time": "2024-09-05T11:42:30.081Z",
    "finish_time": "2024-09-05T11:42:32.858Z",
    "test_duration": "2777ms",
    "test_result": "success",
    "message": "",
    "test_execution_sequence": 1,
    "uuid": "5d39beb5-1fe1-4592-87aa-fa1146185b87"
  },
  {
    "test_name": "get_head_changeset",
    "start_time": "2024-09-05T11:42:32.858Z",
    "finish_time": "2024-09-05T11:42:33.077Z",
    "test_duration": "219ms",
    "test_result": "success",
    "message": "",
    "test_execution_sequence": 2,
    "uuid": "31662fbc-0893-4f23-87ad-0b20a8108fbe"
  }
]
```

### DIFF
--- a/bin/si-smoketest/binary_exeuction_lib.ts
+++ b/bin/si-smoketest/binary_exeuction_lib.ts
@@ -1,0 +1,18 @@
+// This lib can be used for binary args management, i.e.
+// parsing inputs
+// setting default variables
+// setting verbosity/logging levels
+// interrogating the environment to figure out exactly how to execute the tests
+
+export function parseArgs(args: string[]) {
+    if (args.length < 2) {
+      throw new Error("Expected at least 2 args: workspaceId, userEmail, and optionally userPassword and test names");
+    }
+  
+    const workspaceId = args[0];
+    const userId = args[1];
+    const password = args[2] || undefined; // Password is optional
+    const testsToRun = args.slice(3); // Skip the first three arguments
+  
+    return { workspaceId, userId, password, testsToRun };
+}
--- a/bin/si-smoketest/main.ts
+++ b/bin/si-smoketest/main.ts
@@ -1,90 +1,61 @@
 // main.ts
 import assert from "node:assert";
 import { SdfApiClient } from "./sdf_api_client.ts";
+import { createDefaultTestReportEntry, printTestReport, TestReportEntry } from "./test_execution_lib.ts";
+import { parseArgs } from "./binary_exeuction_lib.ts"; 
 
 if (import.meta.main) {
-  assert(Deno.env.get("SDF_API_URL")?.length > 0, "Expected SDF_API_URL env var");
-  assert(Deno.env.get("AUTH_API_URL")?.length > 0, "Expected AUTH_API_URL env var");
-
-  assert([2, 3].includes(Deno.args.length), "Expected args: workspaceId, userEmail and userPassword (optional)");
-  const workspaceId = Deno.args[0];
-  const userId = Deno.args[1];
-  const password = Deno.args[2];
-
+  const { workspaceId, userId, password, testsToRun } = parseArgs(Deno.args);
+  
+  // Init the SDF Module
   const sdfApiClient = await SdfApiClient.init(workspaceId, userId, password);
 
-  // Run tests
-  let testStart = new Date();
-  await get_head_changeset(sdfApiClient);
-  console.log(`get_head_changeset OK - ${(new Date() - testStart)}ms`);
+  // Dynamically load test files from the ./tests directory
+  const testFiles: { [key: string]: string } = {};
+  for await (const dirEntry of Deno.readDir("./tests")) {
+    if (dirEntry.isFile && dirEntry.name.endsWith(".ts")) {
+      const testName = dirEntry.name.replace(".ts", "");
+      testFiles[testName] = `./tests/${dirEntry.name}`;
+    }
+  }
 
-  testStart = new Date();
-  await create_and_delete_component(sdfApiClient);
-  console.log(`create_and_delete_component OK - ${(new Date() - testStart)}ms`);
+  // If no tests are specified, run all tests by default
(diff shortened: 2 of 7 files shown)


bin/si-smoketest/binary_exeuction_lib.ts
```diff
@@ -0,0 +1,18 @@
+// This lib can be used for binary args management, i.e.
```
vbustamante, line 1: I feel this code could still love inside our main code, since it won't be used anywhere else and is very coupled to test startup, but admittedly it's a nitpick

bin/si-smoketest/main.ts
```diff
@@ -1,90 +1,61 @@
 // main.ts
 import assert from "node:assert";
 import { SdfApiClient } from "./sdf_api_client.ts";
+import { createDefaultTestReportEntry, printTestReport, TestReportEntry } from "./test_execution_lib.ts";
+import { parseArgs } from "./binary_exeuction_lib.ts"; 
 
 if (import.meta.main) {
-  assert(Deno.env.get("SDF_API_URL")?.length > 0, "Expected SDF_API_URL env var");
-  assert(Deno.env.get("AUTH_API_URL")?.length > 0, "Expected AUTH_API_URL env var");
```
vbustamante, line 7: Was the intent to remove this checks? I can see them anywhere else

johnrwatson: Ah good shout, this was indeed a mistake during pulling the arg parsing out, will fix

bin/si-smoketest/binary_exeuction_lib.ts
johnrwatson: It could, but I think that as we develop out the other options for execution this is going to get a bit distracting from the intent of main.ts. 

In my opinion, main.ts should really be focused on driving the tests through the test_execution library, rather than handling all the cases that aren't even valid, like some of these:

i.e. 
- you specify a ramp test but don't give it a profile to ramp against
- you specify two types of test execution at the same time
- you specify an output folder that can't exist or be written to by the executing process
- etc etc

Handling all these cases will be hundreds of lines of .ts I'd imagine, so I thought it would be worth segregating off early to save us a refactor.

Happy to move it back in line in main if you wish, your call 🙂

bin/si-smoketest/main.ts
johnrwatson: Amended, sorry about that! I did have the feeling I missed something, see:
`checkEnvironmentVariables(Deno.env.toObject());`

bin/si-smoketest/tests/get_head_changeset.ts
```diff
@@ -0,0 +1,10 @@
+import assert from "node:assert";
+import { SdfApiClient } from "../sdf_api_client.ts";
+
+export default async function get_head_changeset(sdfApiClient: SdfApiClient) {
+  const data = await sdfApiClient.listOpenChangeSets();
```
jobelenus, line 5: I don't necessarily love the separation of the URL and payload from the test (e.g. inherently coupled data living in two different places)... but I can get over that 😇

vbustamante: This is more of a smoke test for the smoke tester, it's made to make sure the simplest get request can execute, but other places may nee to list open changesets - this test can got away but the helper in the client would stay

bin/si-smoketest/binary_exeuction_lib.ts
vbustamante: I'm hyper aware of codebase fragmentation because this usually turns into the user who will debug the code later having to traverse 3 different files to find out a pretty simple algorithm that is only used in one place on the other side - I'd rather have the hundreds of lines in hand when deciding how to break it apart, you know?

That being sad, our priority is to get this running and this works perfectly fine so I won't block it at all, let's get this train running!

vbustamante (approved):

bin/si-smoketest/tests/get_head_changeset.ts
jobelenus: @johnrwatson and I brainstormed on this little bit:
```
interface API_REQUEST {
    path: string,
    method: string,
    headers: Record<string, string>
}

const APIS<string, API_REQUEST> = {
    open_changeset: { path: '/change_set/list_open_change_sets', method: 'GET', headers: { ... } },
    schema_variants: { path: '/v2/workspaces/${workspaceId}/change-sets/${changeSetId}/schema-variants', method: 'GET'},
}

await sdfApiClient.call(APIS.open_changeset, payload)
```

johnrwatson: Dreamy, let me implement

vbustamante: there's already a fetch method on the client

vbustamante: We either implement the methods for each endpoint or this interface and update fetch, otherwise there's like 2 abstractions for the same thing

vbustamante: And since the methods are already there (and this PR is blocking me merging my test a bit 😅 ) i'd suggest we leave it as it is

vbustamante: (I'm merging this just to get the code locally but this can be talked about/implement in a folloup PR, I'm don't wanna kill the conversation by merging this pr)

PR #4503, opened 2024-09-05
Feat/run api tests in ci

johnrwatson:
Run the API tests in CI.

Adds them to the cron workflow every 15 minutes + splits them to run on individual nodes to drive them through as quickly as possible.


vbustamante (approved):

nickgerace (commented):
Thanks folks!

PR #4505, opened 2024-09-06
feat: create a compute executor that can run CPU-intensive tasks

fnichol:
This change adds an `si_runtime::compute-executor` function which builds and returns a `DedicatedExecutor` tuned for running CPU-intensive tasks.

There are a few tunings applied to the underlying Tokio runtime:

- The Tokio I/O types such as net, process, signal, etc. are not enabled
- A reduced thread priority is set when the Tokio runtime spawns threads, allowing another primary/main reactor a chance to have some preferential priority

  In addition, the `si_runtime::build_runtime` is renamed to `si_runtime::main_tokio_runtime` with better thread names by default.

### DIFF
--- a/lib/si-runtime-rs/src/lib.rs
+++ b/lib/si-runtime-rs/src/lib.rs
@@ -1,19 +1,62 @@
 //! Common Tokio runtime related behavior.
 
-use tokio::runtime::Runtime;
+use std::{
+    sync::atomic::{AtomicUsize, Ordering},
+    time::Duration,
+};
 
-pub const RT_DEFAULT_THREAD_STACK_SIZE: usize = 2 * 1024 * 1024 * 3;
-pub const RT_DEFAULT_BLOCKING_POOL_SIZE: usize = 2048; // this is 4x the tokio default of 512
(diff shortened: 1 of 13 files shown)


lib/si-runtime-rs/src/lib.rs
```diff
@@ -1,19 +1,62 @@
...
+use std::{
+    sync::atomic::{AtomicUsize, Ordering},
+    time::Duration,
+};
+
+use tokio::runtime::{Builder, Runtime};
+use tokio_dedicated_executor::{DedicatedExecutor, DedicatedExecutorInitializeError};
+
+pub const DEFAULT_TOKIO_RT_THREAD_STACK_SIZE: usize = 2 * 1024 * 1024 * 3;
+pub const DEFAULT_TOKIO_RT_BLOCKING_POOL_SIZE: usize = 2048; // this is 4x the tokio default of 512
```
zacharyhamm, line 12: I 4x'ed this because spawn_blocking was working poorly on macOS, I don't know if we need it to be so large (and we should use fewer blocking pool threads with this executor)

PR #4509, opened 2024-09-06
build(deps): upgrade to NATS server 2.10.20

fnichol:
Notably in the release notes:

> Fix regression in KV CAS operations on R=1 replicas introduced in
> v2.10.19

See: https://github.com/nats-io/nats-server/releases/tag/v2.10.20
References: nats-io/nats-server#5841


nickgerace (commented):
Thanks!

PR #4527, opened 2024-09-09
chore(ci): Add cron to deploy once a day

stack72:
Deploy at 1600 UTC each weekday


nickgerace (approved):

sprutton1:
Do we need to be deploying tools prod on a cron? Or was this supposed to be for prod?

PR #4529, opened 2024-09-09
fix: fail pipeline on tests, correct concurrency naming

sprutton1:
This fixes the concurrency of the api-tests in the pipeline so they don't inadvertently cancel the cypress tests. Tests should also exit 1 if they contain a `failure`.


.github/workflows/deploy-stack.yml
```diff
@@ -105,7 +105,7 @@ jobs:
     # We want to ensure that in-progress cron runs against tools-prod
     # are canceled when we do a deploy so they don't fail erroneously
     concurrency:
-      group: e2e-${{ inputs.environment }}
+      group: api-test-${{ inputs.environment }}
```
johnrwatson, line 108: Ooops , good catch

PR #4533, opened 2024-09-09
feat(dal): use corrections to protect socket arity

zacharyhamm:
If a new connection update is detected for input socket with an Arity of one, issue RemoveEdge updates for any existing connections to ensure we don't accidentally merge multiple inputs to a 1-arity socket.

### DIFF
--- a/lib/dal/src/workspace_snapshot/node_weight/content_node_weight.rs
+++ b/lib/dal/src/workspace_snapshot/node_weight/content_node_weight.rs
@@ -1,16 +1,24 @@
+use std::collections::HashMap;
+
+use petgraph::prelude::*;
 use serde::{Deserialize, Serialize};
 use si_events::merkle_tree_hash::MerkleTreeHash;
 use si_events::{ulid::Ulid, ContentHash};
 
-use crate::workspace_snapshot::content_address::ContentAddressDiscriminants;
-use crate::workspace_snapshot::graph::deprecated::v1::DeprecatedContentNodeWeightV1;
-use crate::workspace_snapshot::{
-    content_address::ContentAddress,
-    graph::LineageId,
-    node_weight::traits::CorrectTransforms,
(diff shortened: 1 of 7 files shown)


lib/dal/src/workspace_snapshot/node_weight/content_node_weight.rs
```diff
@@ -177,4 +185,110 @@ impl From<DeprecatedContentNodeWeightV1> for ContentNodeWeight {
...
+
+        if let Some(self_idx) = graph.get_node_index_by_id_opt(self.id()) {
+            if let Some(NodeWeight::InputSocket(InputSocketNodeWeight::V1(input_socket_inner))) =
+                graph
+                    .edges_directed(self_idx, Incoming)
+                    .filter(|edge_ref| {
+                        EdgeWeightKindDiscriminants::Prototype == edge_ref.weight().kind().into()
+                    })
+                    .filter_map(|edge_ref| graph.get_node_weight_opt(edge_ref.source()))
+                    .next()
```
jhelwig, line 240: It's definitely adding on to the method call chaining going on, but if you use the "get node weight of kind" & `SiVersionedNodeWeight` trait here, you won't have to hard-code what version the inner is.

Something like:
```suggestion
            if let Ok(input_socket_inner) =
                graph
                    .edges_directed(self_idx, Incoming)
                    .filter(|edge_ref| {
                        EdgeWeightKindDiscriminants::Prototype == edge_ref.weight().kind().into()
                    })
                    .filter_map(|edge_ref| graph.get_node_weight_opt(edge_ref.source()))
                    .next()
                    .get_input_socket_node_weight()
                    .map(|nw| node_weight::traits::SiVersionedNodeWeight::inner(nw))
```

zacharyhamm: ah yes, i forgot about our new traits

PR #4535, opened 2024-09-09
Annihilate reconciliation func remnants (again)

nickgerace:
## Description

This PR removes remnants of reconciliation funcs in the codebase... again. Since they are unused and since their re-introduction would likely take a different shape, we are destroying what remains. We loved you!

This is a second attempt (#4523 and #4454) and now works with existing graphs due to deserializing enums that contained reconciliation funcs.


lib/dal/src/func/backend.rs
```diff
@@ -88,6 +87,7 @@ pub enum FuncBackendKind {
     JsAttribute,
     JsAuthentication,
     Json,
+    // NOTE(nick): this has been deprecated. Not adding serde deprecated tag in case it affects the type.
     JsReconciliation,
```
zacharyhamm, line 91: We should also probably remove the `#[remain::sorted]`, since we actually only want to add to the *end* of this enum (and the one below), and maybe add a note that changing this enum will change a type that goes into the content store and so should be done with care

lib/si-events-rs/src/func_run.rs
```diff
@@ -66,6 +66,7 @@ pub enum FuncBackendKind {
     JsAttribute,
     JsAuthentication,
     Json,
+    // NOTE(nick): this has been deprecated. Not adding serde deprecated tag in case it affects the type.
     JsReconciliation,
```
zacharyhamm, line 70: It looks like these are postcard serialized as well: https://github.com//systeminit/si/blob/nick/eng-2653/lib/si-layer-cache/src/db/func_run_log.rs#L119

So let's remove the remain sorted here and also add a note about postcard footguns

lib/si-pkg/src/spec/func.rs
```diff
@@ -70,6 +70,7 @@ pub enum FuncSpecBackendKind {
     JsAttribute,
     JsAuthentication,
     Json,
+    // NOTE(nick): this has been deprecated. Not adding serde deprecated tag in case it affects the type.
```
zacharyhamm, line 73: These are *not* postcard serialized and myay even be safe to remove since we likely don't have any in the module store, but the safest is to leave them for now

zacharyhamm (commented):
I tested against a workspace from main and was able to deserialize function objects so I think we're good. I made some suggestions about adding some warnings and removing `#[remain::sorted]`, which we should never use for postcard serialized enums since the order is important and we'll need to break sort order if we expand the enums

zacharyhamm:
Actually, I'm seeing a deserialization issue in the layer cache updates, going to track it down.

lib/si-events-rs/src/func_run.rs
nickgerace: Gotcha. Will take the same action as the above.

lib/si-pkg/src/spec/func.rs
nickgerace: Yeah I kept these for the conversion. Might change the comment to be more clear.

nickgerace:
Made the changes that we discussed.

PR #4553, opened 2024-09-11
chore(sdf): restructure sdf binary & library to resemble other services

fnichol:
This is...a large churn refactoring, but very long overdue.

The goal with this change is to bring our HTTP API service (aka "SDF") in line with the structure of our other service components. The impetus for this work was after a production incident where our new "graceful shutdown" logic was being short circuited when SDF was running in the `migrateAndQuit` mode (written literally years ago).

The following are notable changes:

- Internalize all `sdf_server::Server` setup into a `from_config` and `from_services` constructor.
- Remove all former shutdown logic and replace it with `CancellationToken`s and `TaskTracker`s.
- Centralize axum app creation for use between the library and integration test code.
- Fix web socket shutdown logic, in the "updates" and "crdt" sessions. Now on graceful shutdown, SDF will issue a close frame with an `AWAY` code and all active web socket sessions will close.
- Centralize server initialization code for use in the "just migrate and don't run the server" scenario.
- Create a `Migrator` type which can either be created via a `from_config` constructor or returned from an `sdf_server::Server` via a `migrator()` method. This type is used to run all database/builtin/data migrations.
- Move all modules under `sdf_server::server` into the root module directly. Frankly, this was my doing way back when there was a `client` module under a `lib/sdf` crate. There is no need for the additional nesting and makes it appear as though those modules are somehow more special or internal than similar code in our other server crates.


lib/sdf-server/src/app.rs
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    // TODO(fnichol): This really smells, we should not be flipping the app behavior differently if
+    // we're in a testing scenario. What is the configurable differences necessary to drive the app
+    // under test? *That* is how the app should be configured--no `for_tests` bool.
+    //
+    // Until that is refactored though...this constructor remains.
+    #[allow(clippy::too_many_arguments)]
+    #[inline]
+    pub fn from_services_for_tests(
```
nickgerace, line 57: Frankly, I think it's just a wrapper and I think we can dismantle it later. Of course, there may be a config change or two between prod and testing, but I think this is mainly old patterns.

nickgerace (commented):
Awesome PR!!

PR #4554, opened 2024-09-12
fix: Only allow keypair access from the original workspaceId

vbustamante:
Only allow keypairs to be accessed from their original workspace, to avoid secret leaks. If a secret is unusable when gathering before funcs, the respective before func will be skipped as if the secret was empty.

Also adds UI that shows unusable secrets clearly:

image

image

image

### DIFF
--- a/lib/dal/src/key_pair.rs
+++ b/lib/dal/src/key_pair.rs
@@ -11,8 +11,8 @@ use thiserror::Error;
 use crate::{
     pk,
     serde_impls::{base64_bytes_serde, nonce_serde},
-    standard_model_accessor_ro, DalContext, HistoryEvent, HistoryEventError, Timestamp,
-    TransactionsError, Workspace, WorkspaceError, WorkspacePk,
+    standard_model_accessor_ro, DalContext, HistoryEvent, HistoryEventError, TenancyError,
+    Timestamp, TransactionsError, Workspace, WorkspaceError, WorkspacePk,
 };
 
 mod key_pair_box_public_key_serde;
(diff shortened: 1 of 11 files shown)


lib/dal/src/secret.rs
```diff
@@ -319,6 +319,21 @@ impl Secret {
...
 
+    /// Returns if the secret can be decrypted in this workspace
+    pub async fn can_be_decrypted(&self, ctx: &DalContext) -> SecretResult<bool> {
+        let key = self.encrypted_secret_key;
+
+        let encrypted_secret = EncryptedSecret::get_by_key(ctx, key)
+            .await?
+            .ok_or(SecretError::EncryptedSecretNotFound(key))?;
+
+        match encrypted_secret.key_pair(ctx).await {
```
nickgerace, line 330: Good idea

lib/dal/src/key_pair.rs
```diff
@@ -111,6 +115,11 @@ impl KeyPair {
         let row = ctx.txns().await?.pg().query_one(GET_BY_PK, &[&pk]).await?;
         let json: serde_json::Value = row.try_get("object")?;
         let key_pair_row: KeyPairRow = serde_json::from_value(json)?;
+
+        if key_pair_row.workspace_pk != ctx.tenancy().workspace_pk()? {
+            return Err(KeyPairError::UnauthorizedKeyAccess);
+        }
```
nickgerace, line 121: Yep!

nickgerace (approved):

PR #4556, opened 2024-09-12
feat: add_action api test and fix up readme

MegaWatt01:
Added action_list API Endpoint to client and created a test checking an action is added to the action_list when a component is created on a change set. 

Updated Readme for API testing tool. 

<div></div>

### DIFF
--- a/bin/si-api-test/tests/add_action.ts
+++ b/bin/si-api-test/tests/add_action.ts
@@ -1,0 +1,81 @@
+import assert from "node:assert";
+import { SdfApiClient } from "../sdf_api_client.ts";
+import { runWithTemporaryChangeset } from "../test_helpers.ts";
+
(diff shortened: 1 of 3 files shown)


bin/si-api-test/tests/add_action.ts
```diff
@@ -0,0 +1,81 @@
...
+  };
+  const createComponentResp = await sdfApiClient.call({
+    route: "create_component",
+    body: createComponentPayload,
+  });
+
+  const newComponentId = createComponentResp?.componentId;
+  assert(newComponentId, "Expected to get a component id after creation");
+  console.log(newComponentId);
+  console.log(changeSetId);
```
sprutton1, line 56: nit: we probably don't need these, the test output is pretty chatty as it is.

sprutton1 (approved):
Thanks for contribution! Looks good minus the one nit below.

PR #4562, opened 2024-09-12
Do not merge: chunking idea for metadata

nickgerace:
This commit is an idea for chunking data between SI services for billing events due to an unbounded field size. This idea introduces complications and does not get us closer to good data warehouse design. Thus, PR #4564 is preferred over this.


nickgerace:
This PR was purely posted for history's sake. This was an idea tossed around, but is not a good one. See #4561 for more information.

PR #4569, opened 2024-09-13
feat: move events away from base64 encooded, add timestamp

johnrwatson:
We were base64 encoding before forklifting, this wasn't required but was caused by confusion in the rust docs for the AWS client lib. Also when the data landed in S3 we can't analyse it.

Also we had no idea which order the events landed in, so I've added a timestamp to each event.


jkeiser (approved):
Looks right through and through, thanks! So awesome to see the JSON go all the way through!

PR #4570, opened 2024-09-13
fix(dal): ensure DVU finishes after function failure

zacharyhamm:
The condition for finishing the DVU after function failure was broken. This should ensure we always send StatusFinished events even if an execution fails.

Should close BUG-513


jhelwig:
What was it about the previous logic that wasn't correct? It was trying to protect against reporting "done" when there are no independent IDs, but there are still functions running (for exmple, if we've gotten to the point of having started running all of the leaf nodes in the graph, and there is nothing else left in the dependency graph).

PR #4571, opened 2024-09-13
fix: kill execution sender test

zacharyhamm:
The kill execution sender test runs a function that sleeps for 30 seconds, but this might not be enough time in CI for the test to run, given how many services it has to interact with. Up this to 240 seconds, which hopefully should make this test less flaky.


nickgerace (commented):
Good idea!

PR #4603, opened 2024-09-16
Properly update SchemaVariant nodes

jhelwig:


lib/dal/src/workspace_snapshot/node_weight/input_socket_node_weight/v1.rs
```diff
@@ -155,7 +155,7 @@ impl SiNodeWeight for InputSocketNodeWeightV1 {
     fn node_hash(&self) -> ContentHash {
         let mut content_hasher = ContentHash::hasher();
         content_hasher.update(self.arity.to_string().as_bytes());
-        content_hasher.update(self.content_address.to_string().as_bytes());
+        content_hasher.update(self.content_address.content_hash().to_string().as_bytes());
```
jhelwig, line 158: Turns out that `self.content_address.to_string()` is just `SchemaVariant`, and doesn't actually include any of the hash. 🙃

lib/dal/src/workspace_snapshot/node_weight/func_node_weight.rs
```diff
@@ -100,7 +100,7 @@ impl FuncNodeWeight {
 
     pub fn node_hash(&self) -> ContentHash {
         ContentHash::from(&serde_json::json![{
-            "content_address": self.content_address,
+            "content_hash": self.content_address.content_hash(),
```
jhelwig, line 103: These are _technically_ fine, since `serde_json` will basically output a form of the debug output of the enum, but it's such an easy thing to forget that it's special handling from `serde_json` that it felt like we should avoid using it directly for the `node_hash` in general.

jhelwig:
The `node_hash` fixes are going in as part of a different PR, and the rest of the work on this is happening in a different PR.

PR #4606, opened 2024-09-16
feat: nack veritech function run failures

sprutton1:
This allows Veritech to `nack` function run requests that fail prior to the actual function execution, such as if the cyclone pool is starved or the message fails to deserialize. On `nack`, the message will be put back into Jetstream, allowing an Veritech to pick it up (the same one or a different one) and try again. This should help cover off on pool starvation scenarios, giving us a chance the retry failures and let the pool refill. We will retry messages up to five times before giving up.

In the future, we should consider a [dead letter queue](https://docs.nats.io/using-nats/developer/develop_jetstream/consumers#dead-letter-queues-type-functionality) so we can determine what to do with these when they completely fail out.

This PR also reduces the execution pool timeout to 2 minutes, assuming that if we can't serve the `get` within that window another Veritech might be able to. I also added a timeout to the pool healthcheck to catch cases where we might hang on startup.

For @fnichol and @nickgerace , I made a light attempt at refactoring this to remove the code duplication, but found wrestling with the types made it a fairly large effort and thought it would be better served in a PR that does the refactor only so we aren't also contending with behavior changes while trying to get it right.

### DIFF
--- a/lib/veritech-server/src/handlers.rs
+++ b/lib/veritech-server/src/handlers.rs
@@ -120,25 +120,25 @@ pub async fn process_request(
             let request: ActionRunRequest =
                 serde_json::from_slice(&msg.payload).map_err(HandlerError::RequestDerializing)?;
             info!(execution_kind = %NATS_ACTION_RUN_DEFAULT_SUBJECT_SUFFIX, execution_id = %request.execution_id, "validated request and about to execute");
-            action_run_request_task(state, request, reply_subject).await
+            action_run_request_task(state, request, reply_subject).await?
         }
         (Some(NATS_RESOLVER_FUNCTION_DEFAULT_SUBJECT_SUFFIX), None) => {
             let request: ResolverFunctionRequest =
                 serde_json::from_slice(&msg.payload).map_err(HandlerError::RequestDerializing)?;
             info!(execution_kind = %NATS_RESOLVER_FUNCTION_DEFAULT_SUBJECT_SUFFIX, execution_id = %request.execution_id, "validated request and about to execute");
-            resolver_function_request_task(state, request, reply_subject).await
+            resolver_function_request_task(state, request, reply_subject).await?
         }
         (Some(NATS_SCHEMA_VARIANT_DEFINITION_DEFAULT_SUBJECT_SUFFIX), None) => {
             let request: SchemaVariantDefinitionRequest =
                 serde_json::from_slice(&msg.payload).map_err(HandlerError::RequestDerializing)?;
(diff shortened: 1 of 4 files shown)


lib/veritech-server/src/server.rs
```diff
@@ -288,6 +289,7 @@ impl Server {
         async_nats::jetstream::consumer::pull::Config {
             durable_name: Some(CONSUMER_NAME.to_owned()),
             filter_subject: subject::incoming(subject_prefix).to_string(),
+            max_deliver: CONSUMER_MAX_DELIVERY,
```
sprutton1, line 292: @fnichol I messed around with including `backoff` here, but it causes it to retry if the first timer expires before we ack/nack, causing quintuplicate function runs. If I understand correctly, we aren't acking until after execution, so we would be racing the timeout.

lib/veritech-server/src/handlers.rs
```diff
@@ -150,10 +150,8 @@ async fn action_run_request_task(
     state: AppState,
     cyclone_request: ActionRunRequest,
     reply_mailbox: Subject,
-) {
```
nickgerace, line 153: These were infallible wrappers. If you want to fail, I'd delete all four wrappers and just call the inner function directly.

lib/veritech-server/src/handlers.rs
```diff
@@ -621,34 +652,50 @@ async fn validation_request(
...
+    let result = tokio::select! {
         _ = tokio::time::sleep(timeout) => {
             error!("hit timeout for communicating with cyclone server");
             kill_sender_remove_blocking(&state.kill_senders, execution_id).await?;
             Err(HandlerError::CycloneTimeout(
                 timeout,
             ))
         },
-        _ = kill_receiver => {
+        Ok(_) = kill_receiver => {
```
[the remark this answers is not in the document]
nickgerace: Ugh! That's annoying. Sorry about that. I missed that error scenario when @vbustamante and I first did this. Glad you found it though. Big hug.

lib/veritech-server/src/server.rs
nickgerace: This sounds right to me. I think right sizing the delivery max is a good plan B.

nickgerace (approved):

lib/veritech-server/src/server.rs
fnichol: Ah, I think the reason for that first timer expiring comes down to the `act_wait` value, which is 30 seconds by default. We fire `AckKind::Progress` messages by default every (30 -1) seconds but only *after* we start processing the message. So that time to first `AckKind::Progress` message is `time_after_pull_waiting_for_instance + 30` seconds.

fnichol (approved):
This is looking great!

PR #4620, opened 2024-09-17
Fail on apply if DVU roots are not empty

nickgerace:
## Description

This PR ensures that we fail on applying a change set in `sdf` if DVU roots are not empty. A toast will be fired if this happens, which should be rare and require no additional user action.

### DIFF
--- a/app/web/src/components/toasts/RetryApply.vue
+++ b/app/web/src/components/toasts/RetryApply.vue
@@ -1,0 +1,43 @@
+<template>
+  <div>
+    <div
+      class="flex flex-row gap-sm items-center dark:bg-black bg-white max-h-[75vh] overflow-y-auto max-w-lg"
+    >
+      <Icon
+        name="alert-circle"
+        class="text-warning-600 content-center ml-md"
+        size="lg"
+      />
+      <p class="grow py-md max-w-lg">
+        Error when attemping to apply changes: found dependent values that need
+        to be processed. Please retry applying your changes! No further action
+        should be required.
+      </p>
+    </div>
+    <div class="flex flex-row gap-sm items-center mt-xs">
+      <VButton
+        class="grow text-action-300 dark:hover:text-white hover:text-black hover:bg-action-400 hover:underline"
+        label="Close"
+        tone="empty"
+        variant="solid"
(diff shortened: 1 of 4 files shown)


lib/sdf-server/src/service/change_set.rs
```diff
@@ -73,11 +75,16 @@ pub type ChangeSetResult<T> = std::result::Result<T, ChangeSetError>;
...
         let (status, error_message) = match self {
-            ChangeSetError::ChangeSetNotFound => (StatusCode::NOT_FOUND, self.to_string()),
-            ChangeSetError::DalChangeSetApply(_) => (StatusCode::CONFLICT, self.to_string()),
             ChangeSetError::ActionAlreadyEnqueued(_) => {
                 (StatusCode::NOT_MODIFIED, self.to_string())
             }
+            ChangeSetError::ChangeSetNotFound => (StatusCode::NOT_FOUND, self.to_string()),
+            ChangeSetError::DalChangeSetApply(_) => (StatusCode::CONFLICT, self.to_string()),
+            ChangeSetError::DvuRootsNotEmpty(_) => (
+                StatusCode::PRECONDITION_REQUIRED,
```
johnrwatson, line 84: super nice

nickgerace: Thank you Mr. Watson 🫡

app/web/src/components/toasts/RetryApply.vue
nickgerace, line 1: This is a direct fork of `FiveHundredError`. For this PR, I did not want to bundle both errors into one Vue component, but I could see a future where we have a `GenericError` toast with an optional title (and if one isn't provided, you get the error title that `FiveHundredError` currently has).

cc @wendybujalski

wendybujalski: Yeah, unifying into a GenericError component will definitely be good to do down the line. But this works for now!

wendybujalski (approved):
👍

app/web/src/components/toasts/RetryApply.vue
nickgerace: Thank you @wendybujalski!

PR #4624, opened 2024-09-18
Brit/bug 524 wrong asset name after changing regenerating applying to

britmyerss:


lib/dal/src/component.rs
```diff
@@ -576,6 +576,15 @@ impl Component {
 
         component.set_name(ctx, &name).await?;
 
+        //set a component specific prototype for the root/si/type as we don't want it to ever change other than manually
```
britmyerss, line 579: this might be controversial - but we were already trying to set a component specific prototype for the root/si/type prop on auto-upgrade, so this covered off on a number of edge cases I was hitting to just default it when the component was created

jhelwig: I think it makes sense that we might want certain props to always be considered as "overridden" as soon as a `Component` is created. I think that longer term I'd like to see it as an annotation on the prop itself, so we don't need to hard-code special casing individual props, and we'd be able to have it be more generalized.

stack72:
Tested this locally and it works well! thank you :)

PR #4629, opened 2024-09-19
Ensure values are updated correctly when changing a component's type.

britmyerss:
This PR adds an ability to `update_type_by_id` which, when called, and the component type is changing to or from an up or down frame, ensures we update any relevant attribute values that have different inputs given the new type. 

I also fixed up some of the tests that are manually setting the type on creation to use a helper, and added comments to differentiate when we should `update_type_by_id` vs. `set_type_by_id` which does not update necessary values, nor does it ensure that the new component type won't cause other problems.

### DIFF
--- a/lib/dal/src/component.rs
+++ b/lib/dal/src/component.rs
@@ -100,6 +100,8 @@ pub enum ComponentError {
     ChangeSet(#[from] ChangeSetError),
     #[error("code view error: {0}")]
     CodeView(#[from] CodeViewError),
+    #[error("component has children, cannot change to component type")]
+    ComponentHasChildren,
     #[error("component {0} has more than one value for the {1} prop")]
     ComponentHasTooManyValues(ComponentId, PropId),
     #[error("component {0} has an unexpected schema variant id")]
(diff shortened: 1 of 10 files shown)


lib/dal/src/component.rs
```diff
@@ -1619,12 +1623,81 @@ impl Component {
 
         Ok(serde_json::from_value(type_value)?)
     }
+    /// Sets the [`AttributeValue`] for root/si/type to the given [`ComponentType`]
+    /// NOTE: This does NOT ensure that this change is valid, nor does it account for
+    /// needing to update other attribute values in cases where the new type is an up or
+    /// down frame
+    pub async fn set_type_by_id(
```
jkeiser, line 1630: It's not clear from names that set and update do different things--if this is sort of an "unsafe/unchecked" method, it's probably worth giving one of them a worse name so we can read some sort of difference at callsites. `set_type_by_id_unchecked` vs. `set_type_by_id` maybe, so people using autocomplete hesitate before calling the unchecked one? I think you could call this a nit, but it confused me quickly seeing the two different calls in the PR right away :)

(Could be we have a convention for what "update" means I'm not aware of though!

PR #4636, opened 2024-09-19
Switch to new asset details panel paradigm

nickgerace:
## Description

This PR switches to the new asset details paradigm, which involves intrinsic func editing. To start, it will use the unset func and the identity func to set bindings. This PR includes several UI changes to support this includng a caution lines setting, converging on one design for the dropdowns, filtering for the dropdowns, checkboxes for the dropdown, and sorting configurable props and output sockets by name.

## Secondary Changes

This PR also ensures that we retain func bindings when regenerating assets and switches to the new asset details panel paradigm, which replaces `setValueFrom` and it enables intrinsic func authoring for all.

The bug was caused by relying on the identity func when merging prototypes for both prop and socket output locations between two schema variants. Now, conditional logic based on the identity func has been removed because identity func bindings now work similarly to custom attribute func bindings: both involve editing the asset, but not the asset's code.

As a result of these changes, `setValueFrom` will now "no-op". The documentation and inlay hints have been updated accordingly. This does not affect existing assets as they will go through the import and export module code rather than through asset func execution. In addition, the identity-based bindings will appear in the asset details panel. However, the next time that the asset func is regenerated, `setValueFrom` will "no-op" and only the bindings in the asset details panel will be respected.

## Misc Changes

This commit also fixes tracing naming for the `regenerate_variant` function, sorting when working with the asset details panel and has some general asset details panel polish.

In addition to these changes, we also ensure that graph cycles aren't 500 errors when editing attribute bindings.

### DIFF
--- a/app/docs/src/reference/typescript/asset_builder/classes/MapKeyFuncBuilder.md
+++ b/app/docs/src/reference/typescript/asset_builder/classes/MapKeyFuncBuilder.md
@@ -110,8 +110,14 @@ this
 
 > **setValueFrom**(`valueFrom`): `this`
 
-DEPRECATED: Set the value of this socket using a ValueFromBuilder.
-The recommended way to do this is to attach an attribute function.
+DEPRECATED: this method no longer does anything. It will be ignored
+when executing the asset function. Please use the asset editing
+interface to perform equivalent functionality.
+
+If the entry is new, you will need to regenerate the asset first!
+
+In the past, this was used to set the value of this entry using a
+ValueFromBuilder.
 
 #### Parameters
 
--- a/app/web/src/api/sdf/dal/schema.ts
+++ b/app/web/src/api/sdf/dal/schema.ts
@@ -97,7 +97,9 @@ export const outputSocketsAndPropsFor = (
   return opts;
 };
 
-const groupedPropsFor = (schemaVariant: SchemaVariant): GroupedOptions => {
+export const groupedPropsFor = (
+  schemaVariant: SchemaVariant,
+): GroupedOptions => {
   const rootPropOptions = schemaVariant.props
     .filter(
       (p) =>
@@ -188,14 +190,24 @@ const groupedPropsFor = (schemaVariant: SchemaVariant): GroupedOptions => {
 export const inputSocketsAndPropsFor = (
   schemaVariant: SchemaVariant,
 ): GroupedOptions => {
-  const inputSocketOptions = schemaVariant.inputSockets.map((socket) => ({
+  const opts = groupedPropsFor(schemaVariant);
+  opts["Input Sockets"] = rawInputSocketsFor(schemaVariant);
+  return opts;
+};
+
+export const inputSocketsFor = (
+  schemaVariant: SchemaVariant,
+): GroupedOptions => {
+  return {
+    "Input Sockets": rawInputSocketsFor(schemaVariant),
+  };
+};
+
+const rawInputSocketsFor = (schemaVariant: SchemaVariant) => {
+  return schemaVariant.inputSockets.map((socket) => ({
     label: socket.name,
     value: `s_${socket.id}`,
   }));
-
-  const opts = groupedPropsFor(schemaVariant);
-  opts["Input Sockets"] = inputSocketOptions;
-  return opts;
 };
 
 export const findSchemaVariantForPropOrSocketId = (
--- a/app/web/src/components/AssetDetailIntrinsicInput.vue
+++ b/app/web/src/components/AssetDetailIntrinsicInput.vue
@@ -10,6 +10,7 @@
       display.funcId === identityFuncId ? 'input-socket' : 'circle-slash'
     "
     :disabled="isLocked || display.funcId === unsetFuncId"
+    :showCautionLines="display.funcId === unsetFuncId"
     iconRightRotate="down"
     :nullLabel="display.funcId === unsetFuncId ? 'Unset' : 'not set'"
     type="dropdown-optgroup"
@@ -17,25 +18,28 @@
   >
     <template #rightOfInput>
       <DropdownMenu ref="contextMenuRef" :forceAbove="false">
-        <DropdownMenuItem header :checkable="false" class="uppercase"
-          >Change Configuration</DropdownMenuItem
-        >
         <DropdownMenuItem
-          :checked="display.funcId === unsetFuncId"
-          @select="() => emit('changeToUnset', display)"
-          >Unset Binding</DropdownMenuItem
-        >
-        <DropdownMenuItem
-          :checked="display.value?.startsWith('p_')"
-          label="Bind to Input Socket"
-          :submenuItems="socketSubmenu"
-          @select="() => emit('changeToIdentity', display, null)"
+          label="Change Configuration"
+          header
+          class="uppercase"
         />
         <DropdownMenuItem
-          :checked="display.value?.startsWith('s_')"
+          label="Unset"
+          checkable
+          :checked="selectedFilter === 'unset'"
+          @select="selectFilter('unset')"
+        />
+        <DropdownMenuItem
+          label="Bind to Input Socket"
+          checkable
+          :checked="selectedFilter === 'inputSocket'"
+          @select="selectFilter('inputSocket')"
+        />
+        <DropdownMenuItem
           label="Bind to Prop"
-          :submenuItems="propSubmenu"
-          @select="() => emit('changeToIdentity', display, null)"
+          checkable
+          :checked="selectedFilter === 'prop'"
+          @select="selectFilter('prop')"
         />
       </DropdownMenu>
       <DetailsPanelMenuIcon
@@ -56,7 +60,6 @@ import {
   VormInput,
   DropdownMenu,
   DropdownMenuItem,
-  DropdownMenuItemObjectDef,
 } from "@si/vue-lib/design-system";
 import {
   FuncKind,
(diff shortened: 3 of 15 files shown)


app/web/src/components/AssetDetailsPanel.vue
```diff
@@ -146,7 +146,7 @@
           @focus="focus"
         />
       </Stack>
-      <template v-if="ffStore.SHOW_INTRINSIC_EDITING">
+      <template>
```
nickgerace, line 149: @wendybujalski @jobelenus I believe blank template is useless here, right?

lib/dal/src/schema/variant/authoring.rs
```diff
@@ -302,49 +300,52 @@ impl VariantAuthoringClient {
         }
     }
 
-    #[allow(clippy::too_many_arguments)]
-    #[instrument(name = "variant.authoring.update_variant", level = "info", skip_all)]
+    #[instrument(
+        name = "variant.authoring.regenerate_variant",
```
nickgerace, line 304: This was subtle, but would have been a pain later.
cc @stack72 @johnrwatson @sprutton1

lib/dal/tests/integration_test/schema/variant/authoring/regenerate.rs
```diff
@@ -79,3 +88,348 @@ async fn regenerate_variant(ctx: &mut DalContext) {
     // ensure the func is attached
     assert!(funcs_for_default.into_iter().any(|func| func.id == func_id));
 }
+
+#[test]
+async fn retain_bindings(ctx: &mut DalContext) {
+    let name = "Toto Wolff";
+    let description = None;
+    let link = None;
+    let category = "Mercedes AMG Petronas";
```
nickgerace, line 97: https://www.youtube.com/watch?v=HNvBdyocYFo

app/web/src/components/AssetDetailsPanel.vue
nickgerace: Well... that answers that.
Screenshot 2024-09-24 at 5 04 22 PM

lib/dal/src/schema/variant/authoring.rs
sprutton1: Nice catch

nickgerace:
We are going to discuss how to do this and when's the best time.

jobelenus (approved):

app/docs/src/reference/typescript/asset_builder/classes/MapKeyFuncBuilder.md
```diff
@@ -110,8 +110,14 @@ this
...
-DEPRECATED: Set the value of this socket using a ValueFromBuilder.
-The recommended way to do this is to attach an attribute function.
+DEPRECATED: this method no longer does anything. It will be ignored
+when executing the asset function. Please use the asset editing
+interface to perform equivalent functionality.
+
+If the entry is new, you will need to regenerate the asset first!
+
+In the past, this was used to set the value of this entry using a
+ValueFromBuilder.
```
nickgerace, line 120: You'll see this new deprecation notice in both the asset editor inlay hints and the documentation on the docs site now.

app/web/src/api/sdf/dal/schema.ts
```diff
@@ -97,7 +97,9 @@ export const outputSocketsAndPropsFor = (
   return opts;
 };
 
-const groupedPropsFor = (schemaVariant: SchemaVariant): GroupedOptions => {
+export const groupedPropsFor = (
```
nickgerace, line 100: Now that we have filtering, we needed to break out these functions. They will use one another to reduce potential side effects though, so this should go unnoticed by all other users of `inputSocketsAndPropsFor`.

app/web/src/components/AssetDetailIntrinsicInput.vue
```diff
@@ -10,32 +10,36 @@
       display.funcId === identityFuncId ? 'input-socket' : 'circle-slash'
     "
     :disabled="isLocked || display.funcId === unsetFuncId"
+    :showCautionLines="display.funcId === unsetFuncId"
```
nickgerace, line 13: Didn't bundle this in with `disabled` for `VormInput` since it has users everywhere. I don't love this, but at least it is explicit and doesn't happen as a consequence of another property changing inside `VormInput`.

app/web/src/components/AssetDetailIntrinsicInput.vue
```diff
@@ -157,13 +125,53 @@ const unsetFuncId = computed(() => {
...
+    return "unset";
+  } else if ("socketName" in props.data) {
+    // NOTE(nick): fallback to the input data if the display is empty. We need this in case a "emit" blows the
+    // component away. This could be cleaner to avoid having multiple branches.
+    return "inputSocket";
+  } else if ("path" in props.data) {
+    // NOTE(nick): fallback to the input data if the display is empty. We need this in case a "emit" blows the
+    // component away. This could be cleaner to avoid having multiple branches.
+    return "prop";
+  }
```
nickgerace, line 143: So... the initial filter will be incorrect if an "emit" changed the filter from Unset to either Input Socket or Prop. This happens because it is based on the _displayed_ value, which will be "not set" in both the case of an Input Socket or a Prop. Therefore, we need to fallback and check the incoming data. This could be cleaner, but it works.

app/web/src/components/AssetDetailIntrinsicInput.vue
```diff
@@ -157,13 +125,53 @@ const unsetFuncId = computed(() => {
...
+  const variant = assetStore.variantFromListById[props.schemaVariantId];
+  if (!variant) return {};
+
+  if (selectedFilter.value === "prop") {
+    return groupedPropsFor(variant);
+  } else if (selectedFilter.value === "inputSocket") {
+    return inputSocketsFor(variant);
+  }
+  return {};
+});
```
nickgerace, line 170: This is the filter in action.

app/web/src/components/AssetDetailIntrinsicInput.vue
```diff
@@ -157,13 +125,53 @@ const unsetFuncId = computed(() => {
...
+  if (props.isLocked) return;
+
+  selectedFilter.value = item;
+
+  if (item === "unset") {
+    emit("changeToUnset", display.value);
+  } else {
+    emit("changeToIdentity", display.value, null);
+  }
+};
```
nickgerace, line 157: Wrapper around the old select to handle both the filter changing and the emit side effect.

app/web/src/components/AssetDetailsPanel.vue
```diff
@@ -146,47 +146,47 @@
           @focus="focus"
         />
       </Stack>
-      <template v-if="ffStore.SHOW_INTRINSIC_EDITING">
-        <Stack class="p-xs" spacing="none">
-          <span class="uppercase font-bold py-3"
-            >CONFIGURE DATA PROPAGATION</span
+      <Stack class="p-xs" spacing="none">
```
nickgerace, line 149: Big diff, but basically, the flag is gone.

app/web/src/components/AssetDetailsPanel.vue
```diff
@@ -146,47 +146,47 @@
...
+              :data="config"
+              @change="updateOutputSocketIntrinsics"
+              @changeToUnset="changeToUnset"
+              @changeToIdentity="changeToIdentity"
+            />
+          </li>
+        </ul>
+        <p v-else class="text-xs pb-4 pt-2">
+          No output sockets exist for asset.
+        </p>
```
nickgerace, line 172: New text to let the user know why the space is empty.

app/web/src/components/AssetDetailsPanel.vue
```diff
@@ -146,47 +146,47 @@
...
+              :schemaVariantId="schemaVariantId"
+              :isLocked="editingAsset.isLocked"
+              :data="prop"
+              @change="updatePropIntrinsics"
+              @changeToUnset="changeToUnset"
+              @changeToIdentity="changeToIdentity"
+            />
+          </li>
+        </ul>
+        <p v-else class="text-xs pb-4 pt-2">No props exist for asset.</p>
```
nickgerace, line 188: New text to let the user know why the space is empty.

app/web/src/components/AssetDetailsPanel.vue
```diff
@@ -383,6 +381,7 @@ const _configurableProps = computed(() => {
     };
     config.push(d);
   });
+  config.sort((a, b) => a.name.localeCompare(b.name));
```
nickgerace, line 384: Now the left hand options should be more stable in appearance.

wendybujalski (approved):

app/web/src/components/AssetDetailsPanel.vue
```diff
@@ -433,6 +432,7 @@ const _outputSocketIntrinsics = computed(() => {
 
       bindings.push(d);
     });
+  bindings.sort((a, b) => a.socketName.localeCompare(b.socketName));
```
nickgerace, line 435: Now the left hand options should be more stable in appearance.

app/web/src/store/feature_flags.store.ts
```diff
@@ -10,7 +10,6 @@ const FLAG_MAPPING = {
   FIRST_TIME_TUTORIAL_MODAL: "first_time_tutorial_modal",
   DEV_SLICE_REBASING: "dev-slice-rebasing",
   ADMIN_PANEL_ACCESS: "si_admin_panel_access",
-  SHOW_INTRINSIC_EDITING: "asset_detail_intrinsic_editing",
```
nickgerace, line 13: Death of the flag.

lib/vue-lib/src/design-system/forms/VormInput.vue
```diff
@@ -69,6 +69,12 @@ you can pass in options as props too */
             : null
         "
         class="vorm-input__input-wrap"
+        :class="
+          clsx(
+            'vorm-input__input-wrap',
+           showCautionLines ? themeClasses('bg-caution-lines-light', 'bg-caution-lines-dark') : ''
+          )
+        "
```
nickgerace, line 77: Implementation of the caution lines, solely used by "unset" in the intrinsic editor now.

PR #4638, opened 2024-09-19
feat: rearchitect rebaser to support more scale

fnichol:
A rather large change which changes how a Rebaser server consumes
requests for change sets and when it performs related "dependent values
update" (aka "DVU") runs.

General NATS Jetstream Architecture
-----------------------------------

A Rebaser server uses 2 NATS Jetstream streams to track its work:

- `REBASER_REQUESTS`: as before all requests are consumed from this
  stream and can be thought of as a work queue. In this implementation,
  the stream is a limits-based stream, as the Jetstream consumer model
  is different. As before a request for workspace `$wk_id` and change
  set `$cs_id` are published on `rebaser.requests.$wk_id.$cs_id`.
- `REBASER_TASKS`: this is a classic Jetstream work queue stream, but
  where there is only *one* message per subject and where each message
  represents an exclusive task for a single Rebaser to spin up and run.
  In this case, there are only "process" tasks in which a Rebaser will
  consume the enqueued requests for a change set and run DVUs in serial
  (that is, only one at a time and one after the other as needed). When
  a request message is sent to the `REBASER_REQUESTS` stream, another
  body-less message is sent to this tasks stream. These messages are
  published on `rebaser.tasks.$wk_id.$cs_id.process`.

Tasks Stream
------------

On the tasks stream, a single Jetstream consumer is setup to share work
accross multiple Rebasers as clients. When a Rebaser starts to process a
task message it continuously sends `AckKind::Progress` messages to keep
the message from being ack'd or redelivered. If the task encounters an
error, it can return a `Result::Err(_)` from its Naxum handler which
will trigger an `AckKind::Nack(None)` message, causing an immediate
redlivery of the message to another Rebaser.

If a task needs to be interrupted due to a graceful shutdown of a Rebaser
server, the handler it will return an `Error::Interupted(_)` error which
ensures that the message is nack'd and will be redelivered. If a task
can be cleanly completed, the the handler will return a `Result::Ok(_)`
which triggers an `AckKind::Ack`, causing the message to be deleted from
the tasks stream and so will *not* be ran again.

Requests Stream
---------------

When a "process" task is running, a dedicated NATS consumer is created
to exclusively process all requests for a change set in serial (that is,
one at a time). This consumer is known as an "ordered consumer" and is
push-based (rather than the default pull-based consumers). An ordered
consumer is much lighter weight and ephemeral as far as a NATS cluster
is concerned and thus should reduce the stress on NATS when many change
sets are created/active over a short period of time.

This ordered consumer is set up with a timeout that detects when no
message has entered the subject (or no message has been pulled into the
Naxum app) within a period. When this "quiescence" period is seen, this
triggers a specific graceful shutdown of the "process" task where its
exit state is to return `Result::Ok(_)` and ack the task message. In
this way, change sets which become inactive (that is, no Rebaser request
message) are spun down to conserve resources and allow the Rebaser to
focus only on "active" change sets.

Using an ordered consumer means that we can no longer use a message
`ack` to delete a processed messages (this is a trick of a work queue
stream). Therefore when a request has been successfully processed, the
message is deleted from the stream using its sequence number. A new
Naxum middleware called `PostProcess` provides us a way to handle this
delete on an `OnSuccess` callback. In the `OnFailure` case we simply
don't delete the message which means the next message is still the first
and only message to process.

Another tracked Tokio task is running alongside this consuming requests
task, called the `SerialDvuTask`. It is waiting for a
`tokio::sync::Notify` to fire which is triggered by the
request-consuming Tokio task. If, during the run of a DVU, another
request is processed that requires another DVU run, then the `Notify`
will be re-enabled. That way when the `SerialDvuTask` loop comes back to
check, the `Notify` will be set thus trigger "yet one more" DVU run.

Other Structural Changes
------------------------

Among other changes, some of note are:

- The publisher of a Rebaser request send it directly to the
  `REBASER_REQUESTS` stream. Due to its current RPC (i.e.
  request-then-await-response) communication pattern, setting up the
  Rebaser in this way makes it operate more like Pinga and Veritech.
  Future work is likely going to bring back the concept of "activities"
  that flow for interested parties to follow, however prior to this
  commit, the only users of the activity stream were callers to the
  Rebaser and the Rebaser server itself.
- A first class client for the Rebaser is provided in
  `lib/rebaser-client` which abstracts the NATS communication and the
  on-the-wire request formatting details, ideally allowing us to evolve
  the Rebaser's messaging internally where possible.
- The start of an API messaging framework is provided largely in the
  `lib/rebaser-core` crate in the `api_types` module. Several Rust
  traits are provided that when implemented gives the user a versioned
  message that can be upgraded and understands its serialization format
  independently of versioning. This should allow us to evolve the
  Rebasers request and response messages with both the client and server
  able to detect unsupported message versions without having to
  deserialize the message payload itself. Big props to @jhelwig's work
  on version graph snapshots for the inspiration.
- Rebaser request and response messages are transmitted with several
  metadata header in the NATS message including:
  - `X-CONTENT-TYPE`: describes the serialization format in a content
    type/MIME header compatible string. Current values are
    `application/json` and `application/cbor` with `application/cbor`
    the current default.
  - `X-MESSAGE-TYPE`: describes the message type, typically the primary
    Rust name. Current values are `EnqueueUpdatesRequest` and
    `EnqueueUpdatesResponse`.
  - `X-MESSAGE-VERSION`: a positive integer, where a greater number
    value implies a newer version. It is likely that code that would
    understand version 5 of a message may be able to understand version
    4, 3, 2, 1 but not 6 as an example.
  - `NATS-MSG-ID`: this is a standard NATS header which can be used for
    message de-duplication and is populated with a random
    at-request-time `Ulid` value, managed via a `RequestId` type
    (required for all API data types).


lib/dal-test/src/helpers/change_set.rs
```diff
@@ -178,7 +178,32 @@ impl ChangeSetTestHelpers {
...
+                if !ctx_clone
+                    .workspace_snapshot()?
+                    .has_dependent_value_roots()
+                    .await?
+                {
+                    break;
+                }
+                tokio::time::sleep(Duration::from_millis(25)).await;
+            }
+        }
```
nickgerace, line 205: Normally I'd be weary about the potential infinite loop, but this is a test helper, so looks great to me!

lib/dal/src/context.rs
```diff
@@ -476,30 +526,39 @@ impl DalContext {
...
-            });
+        let maybe_rebase = match self.write_current_rebase_batch().await? {
+            Some(updates_address) => DelayedRebaseWithReply::WithUpdates {
+                rebaser: self.rebaser(),
+                workspace_pk: self.workspace_pk()?,
+                change_set_id: self.change_set_id(),
+                updates_address,
+            },
+            None => DelayedRebaseWithReply::NoUpdates,
+        };
```
nickgerace, line 537: Nice! I like this structure.

lib/naxum/src/middleware/post_process/service.rs
```diff
@@ -0,0 +1,81 @@
...
+                "NATS Jetstream message from parts should succeed, this is a bug!; error={:?}",
+                err
+            ),
+        };
+
+        let info = Arc::new(Info::from(
+            // TODO(fnichol): the middleware here is infallible, but this call could, in theory
+            // error. There's probably a better alternative here...
+            message.info().expect("failed to parse message info"),
+        ));
```
nickgerace, line 67: These look like they'd be hit by logic errors caught by local dev and tests. I'm okay with panics here.

lib/rebaser-client/src/lib.rs
```diff
@@ -0,0 +1,272 @@
...
+            id,
+            workspace_id,
+            change_set_id,
+            updates_address,
+            from_change_set_id,
+        });
+
+        // Cut down on the amount of `String` allocations dealing with ids
+        let mut wid_buf = [0; WorkspacePk::ID_LEN];
+        let mut csid_buf = [0; ChangeSetId::ID_LEN];
```
nickgerace, line 167: Smart!

lib/rebaser-client/src/lib.rs
```diff
@@ -0,0 +1,272 @@
...
+            self.context.metadata().subject_prefix(),
+            workspace_id.array_to_str(&mut wid_buf),
+            change_set_id.array_to_str(&mut csid_buf),
+        );
+
+        // There is one more optional future here which is confirmation from the NATS server that
+        // our publish was acked. However, the task stream will drop new messages that are
+        // duplicates and this returns an error on the "ack future". Instead, we'll keep this as
+        // fire and forget.
+        self.context.publish(tasks_subject, vec![].into()).await?;
```
nickgerace, line 199: Looks good to me.

lib/rebaser-core/src/api_types.rs
```diff
@@ -0,0 +1,483 @@
...
+    }
+}
+
+// TODO(fnichol): Guess what? We should consolidate these `id!` macros! Here's the current catch:
+// most of our "event" ids require serializing for over the wire transmission but don't always need
+// to write into the Postgres database. A lot of our `id!` macros out of dal (and now si-events)
+// get coupled to requiring Postgres marshalling so we just tack that on. For the moment this id
+// type has no Postgres requirements and so wil purposfully *omit* the `postgres_types::FromSql`
+// trait implementations. It think there's a way to split these concners up, but it requires a
+// couple more small crates and I'm just not down to do that right this moment. Cheers!
```
nickgerace, line 59: Yeah, we need to tackle this at some point. They're bigger than they need to be and we have a lot of duplicated `id!` usage for semantically the same types. I'd borderline stuff it all in an `si-id` crate. We can talk about it later, though. Good call out!

lib/rebaser-core/src/api_types.rs
```diff
@@ -0,0 +1,483 @@
...
+        impl TryFrom<CoolRequestV3> for CoolRequestV4 {
+            type Error = UpgradeError;
+
+            fn try_from(value: CoolRequestV3) -> Result<Self, Self::Error> {
+                Ok(Self {
+                    id: value.id,
+                    name: value.name,
+                    organization: None,
+                })
+            }
```
nickgerace, line 369: This is a really clever idea for upgrade testing before we even use or need it. Well done

lib/rebaser-core/src/nats.rs
```diff
@@ -0,0 +1,129 @@
...
+
+    let stream = context
+        .get_or_create_stream(async_nats::jetstream::stream::Config {
+            name: nats_stream_name(prefix, NATS_REBASER_TASKS_STREAM_NAME),
+            description: Some("Rebaser tasks".to_owned()),
+            retention: async_nats::jetstream::stream::RetentionPolicy::WorkQueue,
+            discard: async_nats::jetstream::stream::DiscardPolicy::New,
+            max_messages_per_subject: 1,
+            discard_new_per_subject: true,
+            allow_direct: true,
```
nickgerace, line 29: This tracks with what I've seen with how the system behaves live: one "lock" per change set until the rebaser yields it or the change set is no longer in use (or timed out, etc.).

lib/rebaser-server/src/change_set_processor_task.rs
```diff
@@ -0,0 +1,420 @@
...
+// Await either a graceful shutdown signal from the task or an inactive request stream trigger.
+async fn graceful_shutdown_signal(
+    task_token: CancellationToken,
+    quiescence_token: CancellationToken,
+) {
+    tokio::select! {
+        _ = task_token.cancelled() => {}
+        _ = quiescence_token.cancelled() => {}
+    }
+}
```
nickgerace, line 211: Happy that this is its own function too. Good idea

lib/rebaser-server/src/change_set_processor_task.rs
```diff
@@ -0,0 +1,420 @@
...
+        let AppState {
+            workspace_id,
+            change_set_id,
+            nats,
+            ctx_builder,
+            run_notify,
+        } = state;
+        let mut ctx = ctx_builder
+            .build_for_change_set_as_system(workspace_id.into(), change_set_id.into())
+            .await?;
```
nickgerace, line 297: Part of me worries about how this API could be misused, but I can't think of a better way right now. We did need this though rather than making somewhat BS contexts, so thank you

lib/rebaser-server/src/change_set_processor_task.rs
```diff
@@ -0,0 +1,420 @@
...
+            // one more dvu
+            if ctx
+                .workspace_snapshot()?
+                .has_dependent_value_roots()
+                .await?
+            {
+                run_notify.notify_one();
+            }
+
+            if let Some(workspace) = Workspace::get_by_pk(&ctx, &workspace_id.into()).await? {
```
nickgerace, line 325: I feel like we should fail if this is empty, unless you're accounting for workspace "deletion". Even then, I suspect it would be a soft delete and this should always return `Some`.

lib/rebaser-server/src/change_set_processor_task.rs
```diff
@@ -0,0 +1,420 @@
...
+                        let new_snapshot_id = ctx
+                            .write_snapshot()
+                            .await?
+                            .ok_or(dal::WorkspaceSnapshotError::WorkspaceSnapshotNotWritten)?;
+                        // Manually update the pointer to the new address/id that reflects the new
+                        // Action states.
+                        change_set.update_pointer(&ctx, new_snapshot_id).await?;
+                        // No need to send the request over to the rebaser as we are the rebaser.
+                        ctx.commit_no_rebase().await?;
+                    }
```
nickgerace, line 341: Yep, this tracks with how I understand how/when actions should run. Looks good.

lib/rebaser-server/src/change_set_processor_task.rs
```diff
@@ -0,0 +1,420 @@
...
+                .await
+                .map_err(Error::PublishReply)?;
+        }
+
+        // TODO(fnichol): hrm, is this *really* true that we've written to the change set. I mean,
+        // yes but until a dvu has finished this is an incomplete view?
+        let mut event = WsEvent::change_set_written(&ctx, change_set_id.into()).await?;
+        event.set_workspace_pk(workspace_id.into());
+        event.set_change_set_id(Some(change_set_id.into()));
+        event.publish_immediately(&ctx).await?;
```
nickgerace, line 371: Smells like a "fuck it we ball, reload that UI" trigger. I'm okay with it for now, but you're right... not necessarily going to be true. Good to call out!

lib/rebaser-server/src/handlers.rs
```diff
@@ -0,0 +1,297 @@
...
+    }
+}
+
+fn rebaser_requests_per_change_set_consumer_config(
+    nats: &NatsClient,
+    subject_prefix: Option<&str>,
+    workspace: &ParsedWorkspaceId<'_>,
+    change_set: &ParsedChangeSetId<'_>,
+) -> push::OrderedConfig {
+    push::OrderedConfig {
```
nickgerace, line 283: I saw this yesterday and am a bit surprised it comes from `push` and not `pull`... just an interesting note lol

lib/rebaser-server/src/lib.rs
```diff
@@ -11,7 +11,7 @@
     dead_code,
     improper_ctypes,
     missing_debug_implementations,
-    missing_docs,
+    // missing_docs, // TODO(fnichol): re-enable
```
nickgerace, line 14: I don't blame you

lib/rebaser-server/src/middleware.rs
```diff
@@ -0,0 +1,153 @@
...
+    #[status_code = 406]
+    #[body = "unsupported message type"]
+    pub struct UnsupportedMessageTypeError;
+}
+
+define_rejection! {
+    #[status_code = 406]
+    #[body = "unsupported message version"]
+    pub struct UnsupportedMessageVersionError;
+}
```
nickgerace, line 105: These are clever. Nice.

lib/rebaser-server/src/rebase.rs
```diff
@@ -90,24 +74,21 @@ pub async fn perform_rebase(
...
-            message.payload.rebase_batch_address,
-        ))?;
+        .ok_or(RebaseError::MissingRebaseBatch(request.updates_address))?;
 
     debug!(
-        "to_rebase_address: {}, rebase_batch_address: {}",
-        to_rebase_workspace_snapshot_address, message.payload.rebase_batch_address
+        to_rebase_workspace_snapshot_address = %to_rebase_workspace_snapshot_address,
+        updates_address = %request.updates_address,
     );
```
nickgerace, line 84: Thank you! These fields will be easier to index in Jaegar.

nickgerace (approved):
Looks great to me! I like the naxum-spawned-by-naxum work and it's easy to follow. Clever use of the tasks stream basically being held locks to change sets.

lib/naxum/src/middleware/post_process/service.rs
fnichol: I *may* be able to deal with this differently

lib/rebaser-core/src/api_types.rs
fnichol: Yep, I almost made such a crate!

lib/rebaser-server/src/change_set_processor_task.rs
fnichol: Me too!

lib/rebaser-server/src/change_set_processor_task.rs
fnichol: I had noticed that in many places we were using a `ctx.build_default()` (which had incorrectly set tenancy and visibility), then making other calls to mutate the `DalContext` into the right state. But yes, setting up this "god object" is becoming a little nerve-wracking.

lib/rebaser-server/src/change_set_processor_task.rs
fnichol: Yeah I wonder--this code was mostly lifted off prior versions of the Rebaser so I didn't question what it was doing all that much. I wonder what @zacharyhamm or @jhelwig might think?

lib/rebaser-server/src/handlers.rs
fnichol: Yes, we were suggested that a push consumer would be okay to use here (maybe slightly preferred in our use case). My understanding is that under the covers it uses a core NATS subject, fires all messages that you're subscribed to there and set the client up to subscribe to the single core NATS subject (so old-school NATS firehose style). Kind of neat!

lib/rebaser-server/src/lib.rs
fnichol: Oh yeah, I did fix this up again 😄

lib/rebaser-server/src/middleware.rs
fnichol: A port over from tricks in the Axum code base. Yes, I like these macros!

nickgerace (approved):
Re-approved!

PR #4647, opened 2024-09-20
feat: Amend status codes in SDF server

johnrwatson:
Adds catches for a lot of missing/improve status 500 throws within sdf.

I've tried to leave a few groups at 500 still:
- DAL errors
- Postgres/third party connection errors
- Other things that genuinely should throw, like your workspace is broken

I'm still to do Session and Qualification, I'll follow up on those.

### DIFF
--- a/lib/sdf-server/src/service/change_set.rs
+++ b/lib/sdf-server/src/service/change_set.rs
@@ -75,8 +75,14 @@ pub type ChangeSetResult<T> = std::result::Result<T, ChangeSetError>;
 impl IntoResponse for ChangeSetError {
     fn into_response(self) -> Response {
         let (status, error_message) = match self {
-            ChangeSetError::ActionAlreadyEnqueued(_) => {
+            ChangeSetError::ActionAlreadyEnqueued(_) | ChangeSetError::ActionPrototype(_) => {
                 (StatusCode::NOT_MODIFIED, self.to_string())
+            }
+            ChangeSetError::SchemaVariant(_) | ChangeSetError::Schema(_) => {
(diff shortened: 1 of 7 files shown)


lib/sdf-server/src/service/change_set.rs
```diff
@@ -75,9 +75,15 @@ pub type ChangeSetResult<T> = std::result::Result<T, ChangeSetError>;
 impl IntoResponse for ChangeSetError {
     fn into_response(self) -> Response {
         let (status, error_message) = match self {
-            ChangeSetError::ActionAlreadyEnqueued(_) => {
+            ChangeSetError::ActionAlreadyEnqueued(_) | ChangeSetError::ActionPrototype(_) => {
```
stack72, line 78: what error does this catch?

lib/sdf-server/src/service/component.rs
```diff
@@ -118,16 +118,31 @@ impl IntoResponse for ComponentError {
             | ComponentError::NotFound(_) => (StatusCode::NOT_FOUND, self.to_string()),
             ComponentError::PropertyEditor(err) => match err {
                 PropertyEditorError::ComponentNotFound
+                | PropertyEditorError::PropertyEditorValueNotFoundByPropId(_)
                 | PropertyEditorError::SchemaVariantNotFound(_) => {
                     (StatusCode::NOT_FOUND, err.to_string())
                 }
-
+                PropertyEditorError::AttributePrototype(_)
```
stack72, line 125: Are we sure this is a LOOP_DETECTED?

johnrwatson: It's bad request, you've misread the rust I think. Loop detected is the next match arm

PR #4650, opened 2024-09-20
fix(auth-portal): Adding the updated ToS

stack72:


vbustamante:
image
@stack72 the auth portal also needs to have the current version updated in LegalAgreementPage

PR #4671, opened 2024-09-23
fix(auth-api): Switch the new create workflow to be behind a feature flag

stack72:
This was behind a check on email - we want it to be behind a check for feature flag specificially

### DIFF
--- a/bin/auth-api/src/services/users.service.ts
+++ b/bin/auth-api/src/services/users.service.ts
@@ -9,6 +9,7 @@ import {
   SAAS_WORKSPACE_URL,
 } from "./workspaces.service";
 import { tracker } from "../lib/tracker";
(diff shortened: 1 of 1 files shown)


bin/auth-api/src/services/users.service.ts
```diff
@@ -149,7 +150,11 @@ export async function createOrUpdateUserFromAuth0Details(
       lastName: user.lastName,
     });
 
-    if (user.email.includes("@systeminit.com")) {
+    const saasReleaseEnabled = await posthog.isFeatureEnabled(
+      "auth_portal_saas_release",
+      user.id,
+    );
+    if (saasReleaseEnabled) {
```
vbustamante, line 157: Looks good!

vbustamante (approved):

PR #4684, opened 2024-09-24
feat: add test for applying with 10 open change sets

sprutton1:
All of these test workspaces now have the test asset installed.

1. ensure the HEAD has 0 components by deleting them all and applying
2. create 10 change sets and create one of the test components
3. `pop` one of the change sets off and apply it
4. walk all of the remaining change sets, ensure they now have two components  on the graph, and abandon them

### DIFF
--- a/bin/si-api-test/tests/create_and_apply_across_change_sets.ts
+++ b/bin/si-api-test/tests/create_and_apply_across_change_sets.ts
@@ -1,0 +1,268 @@
+import assert from "node:assert";
+import { SdfApiClient } from "../sdf_api_client.ts";
+
+export default async function create_and_and_apply_across_change_sets(
+  sdfApiClient: SdfApiClient,
+) {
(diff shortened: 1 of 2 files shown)


bin/si-api-test/tests/create_and_apply_across_change_sets.ts
johnrwatson, line 1: If we merge this it'll run on all CI runs automatically and likely slow the API tests run down. 

Might be worth putting it in a different folder or add a variant for a single execution in this tests folder to increase the coverage

sprutton1: I tuned the test down to 10 change sets and made sure it cleans up after itself. The runs looks to be sub-minute now. That cool? https://github.com/systeminit/si/actions/runs/11018151975

johnrwatson: Perfect!

PR #4690, opened 2024-09-24
fix: Don't override function data from own websocket events

vbustamante:


jobelenus (approved):
I think this is a great solution for this particular issue where each field is isolated (e.g. changing field A won't change field B). :shipit:

PR #4692, opened 2024-09-24
fix(auth-api): Ensure that we create a unique lago subscription

stack72:


bin/auth-api/src/lib/lago.ts
```diff
@@ -27,7 +27,7 @@ export async function createCustomer(
 
 export async function createTrialSubscription(userPk: string) {
   const plan_code = "launch_trial";
-  const external_id = `1234_${plan_code}`;
+  const external_id = `${userPk}_${plan_code}`;
```
johnrwatson, line 30: <3

PR #4696, opened 2024-09-25
feat(docs): Adding an explanation for how to authenticate to AWS from SI

stack72:


app/docs/src/explanation/aws-authentication.md
```diff
@@ -0,0 +1,60 @@
...
+                "StringEquals": {
+                    "sts:ExternalId": "YOUR WORKSPACE TOKEN"
+                }
+            }
+        }
+    ]
+}
+```
+
+Tahe the arn of the role that you created and set that in the `AssumeRole` field
```
sprutton1, line 57:
```suggestion
Take the arn of the role that you created and set that in the `AssumeRole` field
```

PR #4701, opened 2024-09-25
fix(dal): Warn instead of error during DVU when functions fail

britmyerss:
Turns it down to a warn event instead of an error


sprutton1 (approved):

zacharyhamm (approved):
💯 
What we *really* need to do is surface the errors to the user

PR #4717, opened 2024-09-27
Upsert new inferred edges on creation

nickgerace:
## Description

This PR informs the frontend of new, inferred edges on component creation. This happens when creating a component and directly placing it into a frame.

This PR also handles an edge case for mutated, inferred connections when detaching children of frames.

Example scenario: there's a grandparent up-frame, a parent down-frame and a child component or up-frame. If the parent frame has an input socket with an arity of one and both the child and grandparent have an output socket with a matching annotation, then the edge will be mutated. The parent will now have a different value at its input socket because the source side of the edge has changed.


nickgerace:
Reviewed with @britmyerss and added a new commit for an edge case. Ran through a few extra scenarios and we should be good!

PR #4729, opened 2024-10-01
Fix incorrect error strings for malformed inputs

nickgerace:
## Description

This PR fixes incorrect error strings for malformed inputs when creating or editing attribute func bindings. It does this by using a typed error payload with every permutation of a malformed input.


lib/dal/src/func/binding/attribute.rs
```diff
@@ -98,37 +126,87 @@ impl AttributeBinding {
...
+                        input_socket_id,
+                        static_argument_value,
+                    ),
+                ))
             }
-
-            (Some(prop_id), None) => AttributeFuncDestination::Prop(prop_id.into()),
-            _ => {
-                return Err(FuncBindingError::MalformedInput(
-                    "cannot set more than one output location".to_owned(),
```
nickgerace, line 127: Similar issue here. Yes, the "output" location string is correct, but it also bundles the "no option provided" case with the "multiple options provided" case.

lib/dal/src/func/binding/attribute_argument.rs
```diff
@@ -123,26 +126,4 @@ impl AttributeArgumentBinding {
             attribute_func_input_location,
         })
     }
-    pub fn assemble_attribute_input_location(
```
nickgerace, line 126: Moved this to live with its friends. The `AttributeArgumentBinding` struct is relatively self-contained domain-wise after this change.

PR #4731, opened 2024-10-01
feat: Garbage collect unused schema variants on demand

vbustamante:


zacharyhamm:
I haven't read through this but I'm curious: is the case where a variant is unused in one change set but used in another covered (deleting it would then break the components in the other change set on replay from head)

vbustamante:
@zacharyhamm's observation helped me figure out a better way to determine which variants need to be deleted

PR #4734, opened 2024-10-02
fix: Don't send websocket events with component's temp id

vbustamante:
Previously, on component upgrade, we were reattaching parents before the component creation had been finalized, and its id was a temp one instead of the one of the component it was going to replace. Since the parent_upsert functions send websocket events implicitly, even though the correct connection was made, the frontend would get the temp id intead of the real component id, making the graph wrong


vbustamante:
fixes BUG-620

PR #4740, opened 2024-10-04
feat: add custom spicedb container image

sprutton1:
We're planning to use SpiceDb as our authz solution. This adds it to the tilt stack. I set the container to build when starting tilt so we can add changes to `schema.zed` to validate changes as we go without having to push container images along the way. We will still build and ship images in CI.

I also reorganized the tilt file a bit.

### DIFF
--- a/dev/Tiltfile
+++ b/dev/Tiltfile
@@ -10,13 +10,10 @@ cfg = config.parse()
 
 groups = {
     "platform": [
-        "nats",
-        "otelcol",
         "db",
+        "nats",
         "postgres",
-        "db-test",
-        "postgres-test",
-        "jaeger",
+        "spicedb",
     ],
     "backend": [
         "pinga",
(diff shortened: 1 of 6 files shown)


dev/Tiltfile
```diff
@@ -30,14 +27,17 @@ groups = {
...
     ],
     "telemetry": [
-        "prometheus",
-        "promtail",
         "grafana",
+        "jaeger",
         "loki",
+        "otelcol",
+        "prometheus",
+        "promtail",
```
fnichol, line 40: ahhhhhh lexical sorting, I'm at ease

dev/docker-compose.platform.yml
```diff
@@ -174,3 +174,14 @@ services:
     image: localstack/localstack
     ports:
       - "4566:4566"
+
+  spicedb:
+    #<<: *loki-logged-service
+    build: ../component/spicedb
+    environment:
+      - "SPICEDB_LOG_FORMAT=console"
+      - "SPICEDB_GRPC_PRESHARED_KEY=hobgoblin"
```
fnichol, line 183: It's started!

fnichol (approved):
A very nice start for us to build upon!

PR #4758, opened 2024-10-09
CorrectTransforms should always try to consult the NodeWeight from the Updates first

jhelwig:
If there is a NodeWeight in the set of updates, we should try consult that version when performing CorrectTransforms, as it will be more up to date & relevant than the version already in the graph.


jhelwig:
Unnecessary as all of the `CorrectTransforms` implementations already account for needing to consult the version from the updates to get the latest information.

PR #4764, opened 2024-10-10
Install module upgrades as editing assets

nickgerace:
## Description

This PR installs module upgrades as editing assets in order to not clobber your local changes to a module's asset. Now, you'll be able to work with the editing asset from upstream before applying your change set. Alongside the editing asset will be its unlocked asset func so that regenerations are possible.

This commit also blocks upgrading modules when you having an editing asset for the module available. If there's an older module that does not know its core `SchemaId`, then we will install anyway, which is how things work on main today. However, those modules should not exist anymore since the `SchemaId` is now required for new modules.

The import code has been polished and modernized a bit with unused variables and code being dropped wholesale.

### DIFF
--- a/lib/dal/src/pkg/import.rs
+++ b/lib/dal/src/pkg/import.rs
@@ -30,7 +30,7 @@ use crate::{
     prop::PropPath,
     schema::variant::leaves::{LeafInputLocation, LeafKind},
     DalContext, EdgeWeightKind, Func, FuncId, InputSocket, OutputSocket, OutputSocketId, Prop,
-    PropId, PropKind, Schema, SchemaId, SchemaVariant, SchemaVariantId,
+    PropId, PropKind, Schema, SchemaVariant, SchemaVariantId,
 };
 use crate::{AttributePrototype, AttributePrototypeId};
 
@@ -196,7 +196,7 @@ async fn import_change_set(
             metadata.name(),
         );
 
-        let (_, schema_variant_ids) = import_schema(
+        let schema_variant_ids = import_schema(
             ctx,
             schema_spec,
             installed_module.clone(),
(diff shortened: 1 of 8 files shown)


lib/dal/src/schema/variant.rs
```diff
@@ -1084,6 +1084,27 @@ impl SchemaVariant {
         Ok(Func::get_by_id_or_error(ctx, asset_func_id).await?)
     }
 
+    /// This method unlocks the asset [`Func`] without creating a copy.
+    ///
+    /// **Warning:** this is a somewhat dangerous as we should normally create a copy of an asset
```
zacharyhamm, line 1089: Why is it dangerous?

lib/dal/src/pkg/import.rs
```diff
@@ -453,125 +453,178 @@ async fn import_schema(
...
+        // (again, the one we are currently working with). This is because we need to ensure that
+        // the user's local changes to the assets are not automatically clobbered when installing
+        // updated assets from the module index. Not only that, but its corresponding asset func
+        // needs to be unlocked so that the user can regenerate and make changes at will, so we
+        // will do that too.
+        //
+        // If the schema did not already exist, we can safely set the default schema variant and we
+        // don't care about whether or not it becomes locked.
+        let mut can_lock = true;
+        if index == default_variant_spec_index {
```
zacharyhamm, line 575: What happens if there is already an unlocked variant in this change set for this schema?

zacharyhamm (commented):
No immediate red flags to me. But I don't think I've completely wrapped my head around variant locking

lib/dal/src/pkg/import.rs
nickgerace: This is stopped at the `install_module` route level. Do you think we should be checking that in here instead?

lib/dal/src/schema/variant.rs
nickgerace: Anytime we unlock, we want to copy the object being unlocked. This specific issue that the PR tackles is a "special case".

PR #4776, opened 2024-10-11
fix for tokio blocking pool size on macos

wendybujalski:
this should prevent the severe slowdown Wendy has been seeing on her M1 macbook


zacharyhamm:
Stupid kill execution test

PR #4782, opened 2024-10-11
Mock audit logs for frontend

nickgerace:
## Description

This PR adds mocked audit logs for the frontend. Currently they are based in the `dal`, though it is likely that they will move to their own crate.

It's not in this PR, but there is a difference between a "frontend type" audit log and the audit log that will be ultimately be stored in the database. The purpose of this commit is to unlock frontend creativity by providing generated audit logs. It is not meant to reflect the structure that will ultimately be used by backend services.

### DIFF
--- a/lib/sdf-server/src/service/v2/audit_log.rs
+++ b/lib/sdf-server/src/service/v2/audit_log.rs
@@ -1,0 +1,38 @@
+use axum::{
+    response::{IntoResponse, Response},
+    routing::get,
+    Router,
+};
+use thiserror::Error;
+
+use crate::{service::ApiError, AppState};
+
+pub mod list_audit_logs;
+
+#[remain::sorted]
+#[derive(Debug, Error)]
+pub enum AuditLogError {
+    #[error("audit log error: {0}")]
+    AuditLog(#[from] dal::audit_log::AuditLogError),
+    #[error("transactions error: {0}")]
+    Transactions(#[from] dal::TransactionsError),
+}
+
+pub type AuditLogResult<T> = Result<T, AuditLogError>;
+
+impl IntoResponse for AuditLogError {
+    fn into_response(self) -> Response {
+        let err_string = self.to_string();
+
+        #[allow(clippy::match_single_binding)]
+        let (status_code, maybe_message) = match self {
+            _ => (ApiError::DEFAULT_ERROR_STATUS_CODE, None),
+        };
+
+        ApiError::new(status_code, maybe_message.unwrap_or(err_string)).into_response()
+    }
+}
+
+pub fn v2_routes() -> Router<AppState> {
+    Router::new().route("/", get(list_audit_logs::list_audit_logs))
+}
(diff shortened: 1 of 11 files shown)


lib/si-frontend-types-rs/src/audit_log.rs
```diff
@@ -0,0 +1,21 @@
+use serde::{Deserialize, Serialize};
+use si_events::{
+    audit_log::{AuditLogKind, AuditLogService},
+    Actor,
+};
+
+#[derive(Clone, Debug, Deserialize, Serialize, Eq, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct AuditLog {
```
nickgerace, line 9: This the "frontend type" of the audit log. Note that this is explicitly not the backend representation, which should represent the data model of the audit log in the database. That is not included in this PR since this PR is aimed at preparing the frontend for mocked data.

lib/si-events-rs/src/audit_log.rs
```diff
@@ -0,0 +1,20 @@
...
+
+#[remain::sorted]
+#[derive(Clone, Debug, Deserialize, Serialize, Eq, PartialEq)]
+pub enum AuditLogKind {
+    CreateComponent,
+    DeleteComponent,
+    PerformedRebase,
+    RanDependentValuesUpdate,
+    UpdatePropertyEditorValue,
+}
```
nickgerace, line 20: These are in `si-events` for now, but we will see where they ultimately end up.

lib/sdf-server/src/service/v2/audit_log/list_audit_logs.rs
```diff
@@ -0,0 +1,25 @@
...
+    AccessBuilder(access_builder): AccessBuilder,
+    PosthogClient(_posthog_client): PosthogClient,
+    OriginalUri(_original_uri): OriginalUri,
+    Path((_workspace_pk, change_set_id)): Path<(WorkspacePk, ChangeSetId)>,
+) -> AuditLogResult<Json<Vec<frontend_types::AuditLog>>> {
+    let ctx = builder
+        .build(access_builder.build(change_set_id.into()))
+        .await?;
+
+    let audit_logs = dal::audit_log::generate(&ctx).await?;
```
nickgerace, line 22: Just return the generated data from a crate outside of `sdf`. Note that this route does not account for pagination semantics that we will soon want. That will come in a later PR.

lib/sdf-server/src/service/v2/audit_log.rs
```diff
@@ -0,0 +1,38 @@
...
+pub type AuditLogResult<T> = Result<T, AuditLogError>;
+
+impl IntoResponse for AuditLogError {
+    fn into_response(self) -> Response {
+        let err_string = self.to_string();
+
+        #[allow(clippy::match_single_binding)]
+        let (status_code, maybe_message) = match self {
+            _ => (ApiError::DEFAULT_ERROR_STATUS_CODE, None),
+        };
```
nickgerace, line 30: This will get expanded over time, so no need to collapse this yet.

wendybujalski (approved):

PR #4784, opened 2024-10-13
fix: spawn one cleanup task per distinct disk cache

zacharyhamm:
Currently, every 3600 seconds, 10s of thousands of disk cache tasks will attempt to evict stale files, exhausting CPU. Let's stop the disk cache eviction feature from spawning one long running eviction task per layer cache event. Instead spawn just one per distinct disk cache (as defined by the write path).


zacharyhamm:
@sprutton1 just dropping a note about this one

PR #4787, opened 2024-10-14
fix: add no-op mode and ensure item cleanup for disk cache

sprutton1:
* Allows the disk cache cleanup to be run in no-op for validation
* Adds a config setting for adjusting the disk cache cleanup mode
* Re-enable disk cache cleanup
* Track and report on the disk cache cleanups
* Ensure we are removing both the index and the content when we remove from the disk cache

```
2024-10-14T15:25:13.975559Z  INFO ThreadId(57) si_layer_cache::disk_cache: Disk cache cleanup mode no-op removed 0 out of 0 expired items
2024-10-14T15:26:13.984540Z  INFO ThreadId(50) si_layer_cache::disk_cache: Disk cache cleanup mode no-op removed 0 out of 368 expired items

2024-10-14T16:12:05.830532Z  INFO ThreadId(64) si_layer_cache::disk_cache: Disk cache cleanup mode remove removed 0 out of 0 expired items
2024-10-14T16:12:15.869308Z  INFO ThreadId(30) si_layer_cache::disk_cache: Disk cache cleanup mode remove removed 133 out of 133 expired items
2024-10-14T16:12:25.876136Z  INFO ThreadId(60) si_layer_cache::disk_cache: Disk cache cleanup mode remove removed 0 out of 0 expired items
```

### DIFF
--- a/component/init/configs/service.toml
+++ b/component/init/configs/service.toml
@@ -59,3 +59,4 @@ seconds_to_idle = "$SI_CACHE_EVICTION_SECONDS_TO_IDLE"
 
 [layer_db_config.disk_cache_config]
 path = "$SI_LAYER_CACHE_DISK_PATH"
+cleanup_mode = "$SI_LAYER_CACHE_CLEANUP_MODE"
--- a/lib/si-layer-cache/src/disk_cache.rs
+++ b/lib/si-layer-cache/src/disk_cache.rs
@@ -1,3 +1,4 @@
+use std::fmt;
 use std::sync::Arc;
 
 use std::path::PathBuf;
@@ -16,6 +17,7 @@ const DEFAULT_CHECK_CACHE_TTL_SECONDS: u64 = 60 * 60; // check every hour
 
 #[derive(Clone, Debug)]
 pub struct DiskCache {
+    cleanup_mode: DiskCacheCleanupMode,
     ttl: Duration,
     ttl_check_interval: Duration,
     write_path: Arc<PathBuf>,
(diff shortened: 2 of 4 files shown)


component/init/configs/service.toml
```diff
@@ -59,3 +59,4 @@ seconds_to_idle = "$SI_CACHE_EVICTION_SECONDS_TO_IDLE"
 
 [layer_db_config.disk_cache_config]
 path = "$SI_LAYER_CACHE_DISK_PATH"
+cleanup_mode = "$SI_LAYER_CACHE_CLEANUP_MODE"
```
zacharyhamm, line 62: nice

lib/si-layer-cache/src/disk_cache.rs
```diff
@@ -75,27 +81,42 @@ impl DiskCache {
...
 
     async fn cleanup(&self) {
+        let mut would_remove = 0;
+        let mut removed = 0;
+        let now = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .expect("unable to get the current time, what does this mean? How could this happen?")
+            .as_millis();
+
         for md in cacache::list_sync(self.write_path.as_ref()).flatten() {
```
zacharyhamm, line 91: So, the way this is written, it will occupy an entire tokio worker thread until it has finished.

add a `tokio::task::yield_now().await` here in each iteration to yield execution in between each removal iteration.

zacharyhamm (commented):
Makes sense to me. One note about the actual deletion task, which will occupy an entire worker thread

zacharyhamm (approved):

PR #4791, opened 2024-10-14
Management functions for import

jobelenus:
image

### DIFF
--- a/lib/dal/src/func/binding/management.rs
+++ b/lib/dal/src/func/binding/management.rs
@@ -1,0 +1,106 @@
+use serde::{Deserialize, Serialize};
+use telemetry::prelude::*;
+
+use crate::{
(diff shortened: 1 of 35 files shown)


lib/dal/src/func/binding/management.rs
```diff
@@ -0,0 +1,108 @@
...
+#[derive(Deserialize, Serialize, Debug, Clone, PartialEq, Eq, Hash)]
+pub struct ManagementBinding {
+    // unique ids
+    pub schema_variant_id: SchemaVariantId,
+    pub management_prototype_id: ManagementPrototypeId,
+    pub func_id: FuncId,
+}
+
+impl ManagementBinding {
+    /// Creates an [`ActionPrototype`] with the specified [`ActionKind`] for a given [`SchemaVariantId`]
```
britmyerss, line 20: Probably want to update code comments :)

PR #4801, opened 2024-10-15
Compartmentalize required graph implementation knowledge

jhelwig:
The idea is that that we should move as much of logic that is directly interacting with `NodeWeight` & edges/`EdgeWeight` as close to the versioned `WorkspaceSnapshotGraph` as possible, restricting things "outside" of `WorkspaceSnapshot` to only using interfaces exposed throug `Ext` traits on `WorkspaceSnapshot` that do not allow direct manipulation of the graph.

By limiting things that are directly interacting with the graph structure to things on `WorkspaceSnapshot` (or ideally `WorkspaceSnapshotGraph`), we can hopefully limit the amount of duplicated (and diverging) implementations of various logic.


zacharyhamm:
@jhelwig I believe you're going to run into issues with the dal-macros since `main` is now on 1.81.0 and it looks like this branch is still on the old flake

zacharyhamm:
> @jhelwig I believe you're going to run into issues with the dal-macros since `main` is now on 1.82.0 and it looks like this branch is still on the old flake

Nevermind it all passed on rebase

lib/dal/src/socket/input.rs
```diff
@@ -114,10 +109,10 @@ pub struct InputSocket {
 
 impl InputSocket {
     pub async fn get_by_id(ctx: &DalContext, id: InputSocketId) -> InputSocketResult<Self> {
-        let workspace_snapshot = ctx.workspace_snapshot()?;
-        let node_weight = workspace_snapshot.get_node_weight_by_id(id).await?;
-
-        Self::get_from_node_weight(ctx, &node_weight).await
+        ctx.workspace_snapshot()?
+            .get_input_socket(ctx, id)
```
fnichol, line 113: This is *phenomenal*!

lib/dal/src/workspace_snapshot/graph/v4/socket/input.rs
```diff
@@ -0,0 +1,177 @@
...
+
+impl InputSocketExt for WorkspaceSnapshotGraphV4 {
+    fn new_input_socket(
+        &mut self,
+        schema_variant_id: SchemaVariantId,
+        input_socket_id: InputSocketId,
+        lineage_id: LineageId,
+        arity: SocketArity,
+        content_hash: ContentHash,
+    ) -> WorkspaceSnapshotGraphResult<InputSocketNodeWeight> {
```
zacharyhamm, line 25: I wonder if these traits should have an associated error type, or their own error type. Adding InputSocketError to the WorkspaceSnapshotGraphError seems like we're coupling the graph to the things on it even more closely. Not blocking at all, and something we can look at later.

jhelwig:
Yeah, that PR only touched the test output files, and the overlapping diff hunks in my PR are identical, so there shouldn't be any conflicts.

lib/dal/src/workspace_snapshot.rs
```diff
@@ -176,16 +190,19 @@ impl WorkspaceSnapshotError {
...
+/// ## Internals
+///
+/// The concurrency types used here to give us interior mutability in the tokio run time are *not*
+/// sufficient to prevent data races when operating on the same graph on different threads, since
+/// our graph operations are not "atomic" and the graph *WILL* end up being read from different
+/// threads while a write operation is still in progress if it is shared between threads for
+/// modification. For example after a node is added but *before* the edges necessary to place that
+/// node in the right spot in the graph have been added. We need a more general solution here, but
+/// for now an example of synchronization when accessing a snapshot across threads can be found in
+/// [`crate::job::definition::DependentValuesUpdate`].
```
fnichol, line 205: Good idea calling this out

zacharyhamm (approved):
Makes sense to me! Would love to decouple the extension traits from direct dependency on the DalContext as well.

fnichol (approved):
This is looking like a really nice approach that helps put the impl details closer to the versioned structs that would dictate the details (while not forcing endless duplication)

lib/dal/src/workspace_snapshot.rs
zacharyhamm: _whispers_ that was me (we had a bug in DVU until we did proper synchronization)

lib/dal/src/workspace_snapshot/graph/v4/socket/input.rs
jhelwig: Yeah, it really is reinforcing my thinking that we really need to rework how we have errors set up.

jhelwig:
> Makes sense to me! Would love to decouple the extension traits from direct dependency on the DalContext as well.

Same. The idea was that I would try to move as much as possible into the `Ext` trait(s) for `WorkspaceSnapshotGraph`, and as soon as I needed access to a `DalContext`, that bit would go in the `Ext` trait(s) for `WorkspaceSnapshot`, and would only do as much as was necessary with the `DalContext` there. 

The biggest problems that I see in the way of decoupling even further are needing access to the content store/layerdb (`DalContext`), and having more meaningful errors without exploding the number of `WorkspaceSnapshotError` & `WorkspaceSnapshotGraphError` variants (or having dependencies on the other error types throughout the `dal`).

At some point I'd like to explore a more flexible error type/variant setup than the somewhat rigid one provided by `thiserror`. Or at least, see if there would be a more flexible way to use `thiserror` that doesn't result in a variant count explosion (which is kind of what we have now).

PR #4802, opened 2024-10-15
Replace billing lambda with dynamic view

jkeiser:
To reduce the number of moving parts in billing that can fail, and to keep data fresher and more up to date, this creates a dynamic SQL view that replaces what is currently a lambda populating a table.

The view uses the SQL LEAD() function to replicate the lambda's primary job of getting the next event's timestamp so we have a full date range in the current event.

The lambda and its schedule have been deleted. 

The cost is performance (cached queries take 8 seconds instead of 3 right now). There are plenty of opportunities to fix that, including indexing and materialized views, none of which we need to pursue right now.

NOTE: this PR also reorganizes the queries to expose the resource counts of each workspace we used to calculate max_resource_count for each hour. This helped with diagnostics but is also designed to let us (later) segment invoices by workspace.

Fixes ENG-2741.


zacharyhamm (approved):

component/lambda/sql/billing.sql
```diff
@@ -3,15 +3,6 @@ from data catalog
 database 'si-prod-data'
 iam_role 'arn:aws:iam::944151301248:role/si-prod-redshift,arn:aws:iam::798856999590:role/si-prod-glue-data-bucket';
```
johnrwatson, line 4: chore: might be worth pushing this to an env var, it's got our account IDs in it

johnrwatson (approved):

component/lambda/sql/billing.sql
jkeiser: Oh geez! Thanks!

britmyerss (approved):

PR #4803, opened 2024-10-15
feat(ENG-2745): actions can set resourceId

zacharyhamm:
Actions can now return a resourceId which will set the /root/si/resourceId prop (if it exists on the component). This enables creates that set resourceId instead of the payload, allowing the refresh to set the resource payload.


jobelenus (approved):
I assume that somewhere along the line, after this bit, "component updated" WsEvent fires??

zacharyhamm:
Actions should trigger update events of some kind

PR #4807, opened 2024-10-15
fix(auth-portal): Remove ability to self service change role

stack72:
Disabling this feature as it will be replaced with ReBAC


britmyerss (approved):

DtmVip (commented):
Dddy

PR #4811, opened 2024-10-16
feat: add permissions lib and builder for relations and permissions

sprutton1:
Adds a lib for defining concrete permission and relation types and a builder for interacting with them. The spicedb client itself must be passed in.


fnichol (approved):
Nice!

PR #4815, opened 2024-10-16
Add second cron for running e2e tests against tools production, fix deployment so we cancel running e2e tests if a deploy is kicked off

britmyerss:
<div><br/>via <a href="https://giphy.com/paramountplus/">Paramount+</a> on <a href="https://giphy.com/gifs/paramountplus-season-3-paramount-icarly-wIAzgC4sqiGaeKMhr2">GIPHY</a></div>

### DIFF
--- a/.github/workflows/deploy-stack.yml
+++ b/.github/workflows/deploy-stack.yml
@@ -28,14 +28,15 @@ on:
         description: "version to deploy"
         default: "stable"
 
-# We want to ensure that in-progress cron runs against prod
-# are canceled when we do a deploy so they don't fail erroneously.
+# We only one one deploy happening at a time per environment, so if one is
+# in progress, we'll wait for it to finish before starting the
+# next one.
+# Note that this will also wait for E2E tests to finish, as per the crons workflows
 concurrency:
-  group: e2e-${{ inputs.environment }}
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy-stack.yml
```diff
@@ -32,10 +32,9 @@ on:
 # are canceled when we do a deploy so they don't fail erroneously.
 concurrency:
   group: e2e-${{ inputs.environment }}
-  cancel-in-progress: false
```
sprutton1, line 35: I don't know that we want this to be true as it will also cancel in-progress deploys, which may not be safe if we're in the middle of a migration or something like that.

.github/workflows/tools-crons.yml
```diff
@@ -0,0 +1,22 @@
+name: Tools Crons
+
+on:
+  schedule:
+    - cron: "*/15 * * * *" # Runs every 15 minutes
+
+# We will wait until these tests finish before starting a deploy
+concurrency:
+  group: deploy-tools-prod
```
sprutton1, line 9:
```suggestion
  group: deploy-tools
```

sprutton1 (changes requested):

sprutton1 (approved):

PR #4816, opened 2024-10-16
Feat: migrate current diagram to use views transparently 

vbustamante:
Makes component geometries now be stored in separate nodes bound to a default view, that's created on graph migration. This allows to system to retain its functionality while we work on implementing the usage of multiple views


lib/dal/src/workspace_snapshot/graph/v4/tests.rs
```diff
@@ -3,7 +3,7 @@ use std::collections::HashMap;
 use si_events::{ulid::Ulid, ContentHash};
 
 use crate::{
-    workspace_snapshot::{graph::WorkspaceSnapshotGraphV3, node_weight::NodeWeight},
+    workspace_snapshot::{graph::WorkspaceSnapshotGraphV4, node_weight::NodeWeight},
```
jhelwig, line 6: Do you think it would make sense to move these tests to a `graph/current/...`, and update them to use `WorkspaceSnapshotGraphVCurrent`?

There will always be things that are specific to whatever changes that were made for that graph version (which should probably live in `graph/v4/`), but a lot (if not all) of the tests in this area feel more suited for "graphs (in general) should behave like ...".

lib/dal/src/workspace_snapshot/node_weight/component_node_weight.rs
```diff
@@ -120,6 +122,110 @@ impl ComponentNodeWeight {
...
         ]
     }
+
+    pub(crate) async fn try_upgrade_and_create_external_geometry(
+        ctx: &DalContext,
+        v4_graph: &mut WorkspaceSnapshotGraphV4,
+        default_view_idx: NodeIndex,
+        component_node_weight: &ComponentNodeWeight,
+    ) -> NodeWeightResult<()> {
+        dbg!("upgrade one");
```
jhelwig, line 132: Probably should be removed.

lib/dal/src/workspace_snapshot/node_weight/component_node_weight.rs
```diff
@@ -120,6 +122,110 @@ impl ComponentNodeWeight {
...
+
+        let component_content: ComponentContent = cas
+            .try_read_as(&component_node_weight.content_hash())
+            .await?
+            .ok_or_else(|| NodeWeightError::MissingContentFromStore(component_node_weight.id))?;
+
+        // When migrating from graph v3 to v4 all components should be on v1
+        let ComponentContent::V1(content) = component_content.clone() else {
+            return Ok(());
+        };
```
jhelwig, line 149: Rather than silently pretending that we were able to upgrade this `Component` we should probably return an error. Or at the very least, log an `error!(...)` if we encounter a component content that isn't the correct version.

lib/dal/src/workspace_snapshot/node_weight/geometry_node_weight/v1.rs
```diff
@@ -0,0 +1,48 @@
...
+    }
+}
+
+impl CorrectTransforms for GeometryNodeWeightV1 {}
+
+impl CorrectExclusiveOutgoingEdge for GeometryNodeWeightV1 {
+    fn exclusive_outgoing_edges(&self) -> &[EdgeWeightKindDiscriminants] {
+        &[EdgeWeightKindDiscriminants::Represents]
+    }
+}
```
jhelwig, line 48: Should we also enforce that a `Geometry` can only be used in a single `View`?

jhelwig (changes requested):
No major comments. Just a few questions/suggestions, and some minor "please change".

lib/dal/src/workspace_snapshot/node_weight/geometry_node_weight/v1.rs
vbustamante: @jhelwig The edges flows from the view to the geometry, to avoid the need to create the positions category. Is there a way to have an exclusive incoming edge?

jhelwig: `ComponentNodeWeight` has an example of doing this for the `FrameContains` incoming edge on a `Component`: https://github.com/systeminit/si/blob/32013b25ab5d2848c4df4c5ddacf1df206074f7b/lib/dal/src/workspace_snapshot/node_weight/component_node_weight.rs#L253

lib/dal/src/workspace_snapshot/graph/v4.rs
```diff
@@ -0,0 +1,1711 @@
...
+
+        self.graph
+            .node_weight_mut(current_idx)
+            .ok_or(WorkspaceSnapshotGraphError::NodeWeightNotFound)?
+            .set_id_and_lineage(new_id, new_lineage_id);
+
+        self.add_node_finalize(new_id, new_lineage_id, current_idx)?;
+
+        Ok(())
+    }
```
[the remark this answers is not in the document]
vbustamante: it does not warn anything, but i removed it anyway

lib/dal/src/workspace_snapshot/graph/v4.rs
```diff
@@ -0,0 +1,1711 @@
...
+        f.debug_struct("WorkspaceSnapshotGraph")
+            .field("root_index", &self.root_index)
+            .field("node_index_by_id", &self.node_index_by_id)
+            .field("graph", &self.graph)
+            .finish()
+    }
+}
+
+impl WorkspaceSnapshotGraphV4 {
+    pub async fn new(ctx: &DalContext) -> WorkspaceSnapshotGraphResult<Self> {
```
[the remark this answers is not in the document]
vbustamante: Yeah this is functional but I don't really love the solution. Unit tests are great so I made my best to keep them working

lib/dal/src/workspace_snapshot/graph/v4.rs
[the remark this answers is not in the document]
jhelwig: Looks like I was thinking of a clippy lint. With editor integrations, it's not always immediately clear whether it's coming from rustc, clippy, or something else. https://rust-lang.github.io/rust-clippy/rust-1.82.0/index.html#/unused_async

jhelwig (dismissed):

jhelwig (approved):

PR #4818, opened 2024-10-17
fix: show funcs for unlocked variants even if not bound to default variant

zacharyhamm:


britmyerss (approved):

lib/sdf-server/src/service/v2/func/list_funcs.rs
```diff
@@ -49,12 +49,13 @@ async fn treat_single_function(
 
     // check if it is to be filtered away
     let mut schema_default_map: HashMap<SchemaId, SchemaVariantId> = HashMap::new();
+    let mut unlocked_variants = HashSet::new();
 
     // If func is unlocked or intrinsic, we always use it,
     // otherwise we return funcs that are bound to default variants
-    // OR not bound anything
+    // OR not bound anything, OR editing variants
```
vbustamante, line 56:
```suggestion
    // OR not bound to anything, OR to editing variants
```
My typo, but let's just go head and fix it

vbustamante (approved):
Makes sense

lib/sdf-server/src/service/v2/func/list_funcs.rs
zacharyhamm: too late! in the merge queue :)

PR #4820, opened 2024-10-17
feat: allow editing user roles in auth portal

sprutton1:
This adds the Approver role and allows owners with the `CHANGE_USER_ROLE` feature flag to edit roles.

As the `Owner` of a workspace, I can edit roles and delete users.

image

image

As a non-`Owner`, I cannot.
image


stack72 (dismissed):

app/auth-portal/src/components/MemberListItem.vue
```diff
@@ -0,0 +1,120 @@
...
+      <ErrorMessage :requestStatus="changeMembershipReq" />
+      <VormInput
+        v-model="changeWorkspaceRoleType"
+        :options="changeWorkspaceRoleTypeDropdownOptions"
+        :disabled="
+          !featureFlagsStore.CHANGE_USER_ROLE ||
+          memUser.role === 'OWNER' ||
+          !isWorkspaceOwner
+        "
+        label="Role"
```
vbustamante, line 22:
```suggestion
        noLabel
```

Since this is an input in a named column, it'd be better to just hide the label on the form field itself

vbustamante (changes requested):
Looks good! Just a little UI nitpick, that should fix alignment on each rox

vbustamante (approved):

PR #4821, opened 2024-10-17
feat: sync user relations from auth-api

sprutton1:
* set `spicedb_client` as an optional part of the SDF state, set to `None` by default (to ensure we can deploy safely to environments that lack spicedb)
* when users are deleted from a workspace or when user roles are changed (in the auth-api), we will sync the `Approvers` in spicedb so we have an up-to-date list of users with that role
* add `Owner` as a relation as they should be the only user allowed to modify the `Approvers` list
* on `auth-connect` or `auth-reconnect`, if the user is the original creator of the workspace (and therefore the `Owner`), we will set them as such if they haven't been already
  * we could also just backfill this relationship using a migration? We _should_ have this data available in the auth-api

### DIFF
--- a/component/otelcol/config.yaml
+++ b/component/otelcol/config.yaml
@@ -21,72 +21,182 @@ processors:
       - context: spanevent
         statements:
           # js-browser document-load
-          - set(span.attributes["webjs.connectEnd_ms"], (time_unix_nano - span.start_time_unix_nano) / 1000000) where name == "connectEnd" and instrumentation_scope.name == "@opentelemetry/instrumentation-document-load"
-          - set(span.attributes["webjs.connectStart_ms"], (time_unix_nano - span.start_time_unix_nano) / 1000000) where name == "connectStart" and instrumentation_scope.name == "@opentelemetry/instrumentation-document-load"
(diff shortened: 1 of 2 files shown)


component/spicedb/schema.zed
```diff
@@ -2,5 +2,6 @@
 
   definition workspace {
       relation approver: user
-      permission approve = approver
+      relation owner: user
+      permission approve = approver+owner
```
fnichol, line 6: 🤯 woah! (haha, makes a lot of sense)

lib/permissions/src/schema/schema.zed
```diff
@@ -2,5 +2,6 @@ definition user {}
 
 definition workspace {
     relation approver: user
+    relation owner: user
     permission approve = approver
```
fnichol, line 6: Did you want `permission approve = approver+owner` as well, like in `component/spicedb/schema.zed`?

lib/sdf-server/src/config.rs
```diff
@@ -456,6 +470,7 @@ fn buck2_development(config: &mut ConfigFile) -> Result<()> {
         Some(postgres_cert.clone().try_into()?);
     config.pkgs_path = pkgs_path;
     config.layer_db_config.pg_pool_config.dbname = "si_layer_db".to_string();
+    config.spicedb_config.enabled = true;
```
fnichol, line 473: Yay!

fnichol (dismissed):
Looks good!

fnichol (approved):

PR #4823, opened 2024-10-18
fix(sdf): Make component not found to be 404 not 500

stack72:
Screenshot 2024-10-18 at 11 06 46


zacharyhamm (approved):
I believe this is happening because we are fetching data about components for the applied change set *after* apply

PR #4824, opened 2024-10-18
chore(nix): Adding the Azure CLI to the Nix Flake

stack72:


johnrwatson (dismissed):

johnrwatson (approved):
I think this should be fine

PR #4829, opened 2024-10-18
Add custom extractor for filtering audit logs

nickgerace:
## Description

This PR adds a custom extractor for filtering audit logs. Generally, it handles array notated query string parameters, which is likely the first time we needed to do so in `sdf`.

### DIFF
--- a/lib/sdf-server/src/extract.rs
+++ b/lib/sdf-server/src/extract.rs
@@ -271,6 +271,36 @@ async fn tenancy_from_claim(
     Ok(Tenancy(dal::Tenancy::new(claim.workspace_pk)))
 }
 
+/// Use instead of [`axum::extract::Query`] when the query contains array params using "[]"
+/// notation.
+///
+/// Inspiration : https://dev.to/pongsakornsemsuwan/rust-axum-extracting-query-param-of-vec-4pdm
+pub struct QueryWithVecParams<T>(pub T);
+
+#[async_trait]
+impl<T> FromRequestParts<AppState> for QueryWithVecParams<T>
+where
+    T: serde::de::DeserializeOwned,
+{
+    type Rejection = (StatusCode, Json<serde_json::Value>);
+
+    async fn from_request_parts(
+        parts: &mut Parts,
+        _state: &AppState,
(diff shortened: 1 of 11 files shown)


lib/si-data-spicedb/tests/integration_test/mod.rs
```diff
@@ -147,11 +147,7 @@ async fn check_permissions() {
         .await
         .expect("failed to create a relation");
 
-    let perms = Permission::new(
-        workspace_object.clone(),
-        "approve",
-        user_object.clone(),
-    );
+    let perms = Permission::new(workspace_object.clone(), "approve", user_object.clone());
     let bad_perms = Permission::new(workspace_object, "approve", user_object2);
```
nickgerace, line 151: `cargo fmt`

lib/sdf-server/src/extract.rs
```diff
@@ -271,6 +271,32 @@ async fn tenancy_from_claim(
...
+
+    async fn from_request_parts(
+        parts: &mut Parts,
+        _state: &AppState,
+    ) -> Result<Self, Self::Rejection> {
+        let query = parts
+            .uri
+            .query()
+            .ok_or(not_found_error("no query string found in uri"))?;
+        let deserialized = serde_qs::from_str(query).map_err(|err| internal_error(err))?;
```
zacharyhamm, line 295: Isn't this a 422?

zacharyhamm: I mean: shouldn't this produce a 422 and not a 500? If we fail to deserialize it's because the input was invalid

zacharyhamm: Same with the not found. That's invalid input, not a resource not found

nickgerace: Ah fair point. Will fix

zacharyhamm: (or maybe 400: Bad Request)

zacharyhamm (dismissed):

zacharyhamm (approved):

zacharyhamm (approved):

PR #4830, opened 2024-10-18
Always sort object fields in their schema-defined order

jkeiser:
Right now, object fields on a component are rearranged sometimes (for example, when setting properties, switching an object property from "set from socket" to "set from manual," and when updating a component to a new asset version. This has several effects:

* This causes the component property editors to show properties in different orders at different times (and for different components from the same asset), making it hard to work with.
* The orders also don't tend to match what was specified by the asset author, which makes it hard for authors to group related fields or put important fields first (hurting the experience for authors and users alike).
* This also causes BUG-633, where copy/pasting a component can cause values to copy into the wrong fields. This is because the copy/paste code assumes the order of the original and copy (for objects) are the same when it copies the values over.

This PR changes `get_child_av_ids_in_order()` so that it essentially ignores the ordering in the node, using the ordering of the props instead.

Fixes BUG-214 and BUG-633.

### Alternate Fixes

Fixing the places where we rearrange fields would be a good idea, but by itself insufficient: in the wild, we have plenty of components with the wrong order. (The two places we currently rearrange fields--that I know of--are `AttributeValue::update(<object prop>, json!({}))` and component upgrade.)

We could both fix rearrangement *and* fix property ordering with a minor version migration. I didn't pursue this because the risk of a migration is nonzero (and if anything isn't quite right we'd have to make a *third* version to fix it). However, we still have the option and can use most of this code if we decide to do it in a followup. The tests will likewise remain important.

### DIFF
--- a/lib/dal/src/attribute/value.rs
+++ b/lib/dal/src/attribute/value.rs
@@ -134,6 +134,8 @@ pub enum AttributeValueError {
     },
     #[error("empty attribute prototype arguments for group name: {0}")]
     EmptyAttributePrototypeArgumentsForGroup(String),
+    #[error("object field is not a child prop of the object prop: {0}")]
+    FieldNotChildOfObject(AttributeValueId),
     #[error("func error: {0}")]
     Func(#[from] FuncError),
     #[error("func argument error: {0}")]
@@ -166,6 +168,10 @@ pub enum AttributeValueError {
     MissingPropEdge(AttributeValueId),
     #[error("missing prototype for attribute value {0}")]
     MissingPrototype(AttributeValueId),
+    #[error(
+        "found multiple child attribute values ({0} and {1}, at minimum) for the same prop: {2}"
+    )]
+    MultipleAttributeValuesSameProp(AttributeValueId, AttributeValueId, PropId),
     #[error("found multiple props ({0} and {1}, at minimum) for attribute value: {2}")]
     MultiplePropsFound(PropId, PropId, AttributeValueId),
     #[error("no component prototype found for attribute value: {0}")]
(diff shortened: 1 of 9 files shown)


lib/dal/src/attribute/value.rs
```diff
@@ -2397,30 +2419,66 @@ impl AttributeValue {
...
+                        return Err(AttributeValueError::Unreachable);
+                    }
+
+                    Ok(child_ids_in_prop_order)
+                }
             }
         } else {
-            // Leaves don't have ordering nodes
-            Ok(vec![])
+            // TODO can this ever happen? Seems like AttributeValueError::MissingPropEdge()
```
zacharyhamm, line 2478: Probably not

lib/dal/src/attribute/value.rs
```diff
@@ -2397,30 +2419,66 @@ impl AttributeValue {
...
+                // attribute values in prop order, not attribute
+                // value order, so that we always return them in the
+                // same order (the order the props were created for
+                // the schema variant), not the order they were set
+                // on the attribute value
+                //
+                // TODO doing this on read papers over the fact that we're storing them in an
+                // order we do not prefer. We should really just ensure it's impossible to
+                // create a node with these out of sync, or remove the ordering node from
+                // object props altogether.
```
zacharyhamm, line 2442: do you remove remove the ordering node from object attributes?

zacharyhamm (approved):
clippy is going to ding your tests for a few nits, but this makes sense to me

lib/dal/src/attribute/value.rs
jkeiser: From object AttributeValues, yeah.

lib/dal/src/attribute/value.rs
jkeiser: I can update this to just return the error--it would make things simpler for sure.

jkeiser: (I'll make a quick followup PR.)

zacharyhamm (dismissed):

zacharyhamm (approved):

PR #4835, opened 2024-10-20
Add edges between dependent components in the actions dependencies graph for both frames and edges.

britmyerss:
Also restore right click context menu for edges.

<div><br/>via <a href="https://giphy.com/travisband/">Travis</a> on <a href="https://giphy.com/gifs/travisband-travis-fran-healy-guys-wait-for-me-Xm3b41ZkDH5nOYueh6">GIPHY</a></div>

### DIFF
--- a/app/web/src/components/ModelingDiagram/ModelingDiagram.vue
+++ b/app/web/src/components/ModelingDiagram/ModelingDiagram.vue
@@ -3068,7 +3068,7 @@ const allElementsByKey = computed(() =>
       ...Object.values(componentsStore.nodesById),
       ...groups.value,
       ...sockets.value, // protected to not run on drag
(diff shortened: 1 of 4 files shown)


jhelwig (approved):
👍🏻 This looks like a good step in the right direction.

Not something to change in this PR, but something we'll probably want to address sooner rather than later: Reading through things again while looking at the changes here, I think we're missing handling another scenario in any of the iterations of this we've had so far. If we have something like `Component A (Action 1) -> Component B -> Component C -> Component D -> Component E (Action 2)` (where `->` is `OutputSocket connects to InputSocket`), then I think we'll consider `Action 1`, and `Action 2` as independent. I believe that we'd want `Action 2` to depend on `Action 1`.

PR #4840, opened 2024-10-21
fix(auth-api): Protection of Owner based API endpoints

stack72:


bin/auth-api/src/routes/workspace.routes.ts
```diff
@@ -65,6 +68,18 @@ async function extractOwnWorkspaceIdParam(ctx: CustomRouteContext) {
   return workspace;
 }
 
+async function isOwnerspaceOwner(ctx: CustomRouteContext) {
```
sprutton1, line 71:
```suggestion
async function isWorkspaceOwner(ctx: CustomRouteContext) {
```

stack72: d'oh!

sprutton1 (approved):

PR #4845, opened 2024-10-21
Orphan all children before removing old frame version on schema version upgrade

jhelwig:
While attempting to (re)connect the children of a frame to the newly upgraded version, we try to find the `OutputSocket` `AttributeValue`s from the old version, but this has already been removed. Rather than creating specialized (re)connect logic for the schema version upgrade scenario, we can have the existing logic treat it as a brand new parent/child relationship by orphaning the children before we remove the old version of the parent frame.

### DIFF
--- a/lib/dal/src/component.rs
+++ b/lib/dal/src/component.rs
@@ -3092,6 +3092,11 @@ impl Component {
         // ========================================
         // Delete original component
         // ========================================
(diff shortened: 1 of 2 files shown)


stack72 (approved):
This makes total sense that we need to orphan the components from the old frame!

nickgerace (approved):

PR #4852, opened 2024-10-22
Add pending events architecture

nickgerace:
## Introduction

This PR adds the pending events architecture with a corresponding `EventSessionId`.

## Description

The pending events architecture is another step towards removing all "do this and wait for a reply" logic in the engine. Specifically, it tackles events that are dependent on a successful rebase. The idea is that a new NATS JetStream stream will have "pending events" queue and dequeue rapidly, with `sdf` and `pinga` being the primary publishers and the `rebaser` being the primary consumer. Once a successful rebase happens, the `rebaser` will drain the pending events for that specific rebase and publish them on the relevant stream (e.g. the billing events stream).

Why not use `RebaseBatchAddress` to identify uniqueness in events when performing a rebase? Technically, it is possible to reproduce the same `RebaseBatchAddress` for a rebase. Therefore, we will ensure the `rebaser` has a unique `EventSessionId` to know which events to drain.

## Additional Changes

This PR also splits out the "api-types" subsystem in "rebaser-core" into a new crate called "naxum-api-types". The new crate contains stricter lints and documentation. It can be used for versioning messages sent using "naxum".

On top of these changes, the "service" field of an audit log has been removed. This may be useful in the future, but not all services are self aware of their instance type outside of telemetry. Not only that, but the data is likely unimportant. Therefore, it has been removed. The frontend audit dashboard will still work since the "service" field is unused there for the same reasons.

## Context

This PR is a continuation of the initial effort from #4814.


fnichol (approved):
Zoom peer reviewed

PR #4854, opened 2024-10-22
Track server-level tasks in the rebaser

nickgerace:
## Description

This PR adds the ability to track server-level tasks in the rebaser. This is useful for spawning tasks in the change set processor that are allowed to outlive the lifetime of the processor task itself.

## Context

This PR is a continuation of the initial effort from https://github.com/systeminit/si/pull/4814.


fnichol (approved):
Zoom peer reviewed

PR #4858, opened 2024-10-23
Victor/eng 2696 migrate current diagram to view

vbustamante:
Let's try this again without a possible loop in the changeset listing


lib/dal/src/workspace_snapshot/migrator.rs
```diff
@@ -76,78 +85,68 @@ impl SnapshotGraphMigrator {
...
+        info!("Migrating {} snapshot(s)", open_change_sets.len(),);
 
-        loop {
-            let change_sets_to_migrate = change_set_graph.independent_ids();
-            if change_sets_to_migrate.is_empty() {
-                break;
+        for change_set in open_change_sets {
+            let mut change_set = ChangeSet::find(ctx, change_set.id)
+                .await?
+                .ok_or(ChangeSetError::ChangeSetNotFound(change_set.id))?;
```
vbustamante, line 93: This should prevent any infinite loops

zacharyhamm (approved):

PR #4861, opened 2024-10-23
feat: add metrics to layer cache hits

sprutton1:
Remove the span collection for layer cache hits, it spams way too hard to be useful. Add individual metric calls for layer cache hits in different places.

I suspect that the memory hits will be too much here, too, but we can pull them out if that's the case.

image


zacharyhamm (approved):
Let's start with this. I'd love to be able to see the difference between the different kinds of cache but we can see how to hack that together with macros if we think this doesn't tell us enough

PR #4862, opened 2024-10-23
Update lambdas whenever `main` updates

jkeiser:
This makes it so that we can deploy a new version of a lambda by updating `main`. To use it, you replace your lambda with the contents of `github_lambda_loader.py`, and it will:

1. Pulls the `si_*.py` modules from [`components/lambda/functions` on main](https://github.com/systeminit/si/tree/main/component/lambda/functions)
2. Pulls `<lambda function name>.py` from the same place (e.g. `upload-billing-hours.py`). This means different lambdas will automatically pick up the right script.
3. Imports `lambda_handler` from `<lambda function name>.py` (AWS will run this function).

We're already doing this for the `workspace_delegations_population` lambda, with a miniature version of this; this just lets us write modules.

### DIFF
--- a/component/lambda/functions/github_lambda_loader.py
+++ b/component/lambda/functions/github_lambda_loader.py
@@ -1,0 +1,33 @@
+# From https://github.com/systeminit/si/blob/main/component/lambda/functions/github_lambda_loader.py
+import importlib, urllib.request, urllib.parse, hashlib, pathlib, os, sys, tempfile, logging
+
+logger = logging.getLogger()
+logger.setLevel("INFO")
+
+# The name of the file containing the lambda handler (defaults to <lambda_function_name>.py)
+SI_IMPORT_LAMBDA_HANDLER = os.environ.get('SI_IMPORT_LAMBDA_HANDLER', f"{os.environ['AWS_LAMBDA_FUNCTION_NAME']}.py")
+# Where to download the Python files from (e.g. https://raw.githubusercontent.com/systeminit/si/refs/heads/main/component/lambda/functions/)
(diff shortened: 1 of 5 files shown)


component/lambda/functions/github_lambda_loader.py
```diff
@@ -0,0 +1,30 @@
+import importlib, urllib.request, urllib.parse, hashlib, pathlib, os, sys, tempfile, logging
+
+logger = logging.getLogger()
+logger.setLevel("INFO")
+
+# The name of the file containing the lambda handler (defaults to <lambda_function_name>.py)
+SI_IMPORT_LAMBDA_HANDLER = os.getenv('SI_IMPORT_LAMBDA_HANDLER') or f"{os.environ['AWS_LAMBDA_FUNCTION_NAME']}.py"
+# Where to download the Python files from
+SI_GITHUB_FUNCTIONS_URL = os.getenv('SI_GITHUB_FUNCTIONS_URL') or "https://raw.githubusercontent.com/systeminit/si/refs/heads/main/component/lambda/functions/"
```
johnrwatson, line 9: nitpick: I think I'd prefer these not to have a default and just bomb out if not set

johnrwatson (approved):
It'd be great to move the existing lambda to do the same behaviour/identical driving method, but this is great

component/lambda/functions/github_lambda_loader.py
jkeiser: For SI_GITHUB_FUNCTIONS_URL I think that makes a lot of sense. For SI_IMPORT_LAMBDA_HANDLER I kind of liked enforcing a convention where you name the script the same as the lambda--I just added the env var in case someone ever wanted to do something weird. (Same with the list of libraries)

johnrwatson (approved):

PR #4865, opened 2024-10-23
chore: Don't retry migrations if they fail, raise timeout to 20 minutes

vbustamante:


component/toolbox/scripts/supporting-funcs/ssm-funcs.sh
```diff
@@ -55,6 +55,11 @@ start_and_track_ssm_session() {
       --instance-id "$instance_id" \
       | jq -r '.StandardOutputContent')
     echo "$output" >"$results_directory/$instance_id.json"
+  elif [ "$status" == "InProgress" ]; then
+    output="{\"instance_id\": \"$instance_id\", \"status\": \"error\", \"message\": \"Caller timeout out after waiting\"}"
+    echo "$output" >"$results_directory/$instance_id.json"
+    echo "The github action has timed out, but the migration may still be running. Check the ssm logs on aws"
```
sprutton1, line 61: nit: this technically could happen to scripts besides the one that runs migrations, so we shouldn't be too specific here

sprutton1 (approved):

PR #4871, opened 2024-10-24
feat: Create update and list endpoints for views

vbustamante:

### DIFF
--- a/lib/sdf-server/src/service/diagram.rs
+++ b/lib/sdf-server/src/service/diagram.rs
@@ -20,9 +20,8 @@ use telemetry::prelude::*;
 use thiserror::Error;
 use tokio::task::JoinError;
 
-use crate::AppState;
-
 use super::ApiError;
+use crate::AppState;
 
 pub mod create_component;
 pub mod create_connection;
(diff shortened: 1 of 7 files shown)


lib/sdf-server/src/service/diagram.rs
```diff
@@ -171,4 +177,7 @@ pub fn routes() -> Router<AppState> {
         .route("/get_diagram", get(get_diagram::get_diagram))
         .route("/list_schemas", get(list_schemas::list_schemas))
         .route("/dvu_roots", get(dvu_roots::dvu_roots))
+        .route("/list_views", get(list_views))
+        .route("/create_view", post(create_view))
+        .route("/update_view", post(update_view))
```
jobelenus, line 182: I'd really love to have these under the v2 namespace... is that terribly annoying?

jobelenus (changes requested):
1 change for the URL structure... `ViewView` is hilarious 😆

lib/sdf-server/src/service/diagram.rs
vbustamante: @jobelenus 

It is 😄 
But yeah we should do that eventually. The only thing 100% don't want is to have it partially migrated. I'll just migrate everything on diagram to v2. 

I'll have to change the stores too, let me know if there's any chance we have a conflict

vbustamante: ended up just migrating the views to v2 on a new controller

jobelenus (approved):

PR #4874, opened 2024-10-25
feat: replace memory and disk cache with foyer

sprutton1:
This replaces both `moka` and `cacache` with [foyer](https://github.com/foyer-rs/foyer/tree/main). The primary motivation for this replacement is to ensure that we have proper eviction for disk entries that does not pin the CPU so much that the service locks up. Foyer comes with a wealth of features and tuneables that gives us many interesting options for the future. What's implemented here:

* a cache with a memory capacity represented as percentage of the total
* cache ordering based on the size of the items in the cache
* a disk cache with a capacity of a fixed, but configurable, size
* rate limiting on items entering the disk cache to allow for management of blocking writes
* configuration around the number of flushing threads and  buffer pool size to fine tune writing to disk
* a FIFO eviction pattern on items in the disk cache
* automatic recovery of existing caches on disk
* removal of the disk persistence tasks as we can now rely on foyer to do this for us. This should help a lot with thread starvation issues as we spawn a lot of persistence tasks. Ideally, we can move this work out here entirely now and rely on a separate service to do this long-term persistence

This should extensively tested, ideally in a "real" environment. Running load tests locally show performance that is at least as good, if not better, than what is currently on `main`

There's room to grow here. Some things that are obvious after doing this work:

* func runs and func run logs seem out of place here and seem to mostly use the relational database to to their work. Maybe we should move them out?
* foyer is so tuneable that it might take us a while to find the best setup for our needs
* having six! caches per service makes it difficult to build sane configurations around. We should move the individual caches so that they share a single hybrid cache and can still talk to postgres when necessary. This should make it significantly easier for us to tune things. We can also do some interesting ordering based on the different key types

The PR is broken into commits to make review more manageable. A few of the signatures changed, so a large number of files were touched.


sprutton1:
try

lib/dal/src/attribute/prototype.rs
```diff
@@ -148,16 +148,12 @@ impl AttributePrototype {
...
-            .layer_db()
-            .cas()
-            .write(
-                Arc::new(AttributePrototypeContent::V1(content.clone()).into()),
-                None,
-                ctx.events_tenancy(),
-                ctx.events_actor(),
-            )
-            .await?;
+        let (hash, _) = ctx.layer_db().cas().write(
```
zacharyhamm, line 151: Ahh, so all these methods are sync now huh? interesting

lib/pinga-server/src/config.rs
```diff
@@ -184,7 +184,7 @@ fn default_concurrency_limit() -> usize {
 }
 
 fn default_layer_db_config() -> LayerDbConfig {
-    LayerDbConfig::default_for_service("pinga")
+    LayerDbConfig::default()
```
zacharyhamm, line 187: What changed here? No service-specific defaults?

lib/si-layer-cache/src/db.rs
```diff
@@ -39,6 +37,8 @@ pub mod rebase_batch;
 pub mod serialize;
 pub mod workspace_snapshot;
 
+const GIGABYTES: usize = 1024 * 1024 * 1024;
```
zacharyhamm, line 40: image

lib/si-layer-cache/src/db.rs
```diff
@@ -118,65 +102,89 @@ where
...
 
-        let snapshot_cache: LayerCache<Arc<WorkspaceSnapshotValue>> = LayerCache::new(
+        let snapshot_cache: Arc<LayerCache<Arc<WorkspaceSnapshotValue>>> = LayerCache::new(
             workspace_snapshot::CACHE_NAME,
             pg_pool.clone(),
-            memory_cache_config.clone(),
-            disk_cache_config.clone(),
+            cache_config
+                .clone()
+                .with_memory_percentage(0.50)
```
zacharyhamm, line 180: A good solution before working out how to use a single cache for everything

lib/si-layer-cache/src/db/workspace_snapshot.rs
```diff
@@ -22,22 +22,22 @@ pub struct WorkspaceSnapshotDb<V>
 where
     V: Serialize + DeserializeOwned + Clone + Send + Sync + 'static,
 {
-    pub cache: LayerCache<Arc<V>>,
+    pub cache: Arc<LayerCache<Arc<V>>>,
```
zacharyhamm, line 25: Curious what changed here to require this

lib/si-layer-cache/src/hybrid_cache.rs
```diff
@@ -0,0 +1,174 @@
...
+                            self.insert(key.into(), deserialized.clone());
+                            Some(deserialized)
+                        }
+                        Err(e) => {
+                            error!(
+                        "Failed to deserialize stored bytes from memory cache for key ({:?}): {}",
+                        key,
+                        e
+                    );
+                            self.remove(key);
```
zacharyhamm, line 85: Formatting here seems strange...

lib/si-layer-cache/src/hybrid_cache.rs
```diff
@@ -0,0 +1,174 @@
...
+            .await
+            .map_err(|e| LayerDbError::Foyer(e.into()))?;
+
+        Ok(Self { cache })
+    }
+
+    pub async fn get(&self, key: &str) -> Option<V> {
+        match self.cache.obtain(key.into()).await {
+            Ok(Some(entry)) => match entry.value() {
+                // todo: bad clone here
```
zacharyhamm, line 69: I believe all our inner types here are Arcs? So the clone is cheap

lib/pinga-server/src/config.rs
sprutton1: Before we were feeding the service name to the disk cache path, but we were also generating a random path in that default like `default-cache-pinga-aj34s`, so passing the name was only useful if you needed to look at the specific disk cache of one of the services locally.

lib/si-layer-cache/src/hybrid_cache.rs
sprutton1: cargo fmt disagrees :D

lib/si-layer-cache/tests/integration_test/db/func_execution.rs
zacharyhamm, line 1: Was this just unused?

sprutton1: Yeah, I think it was the prototype of func runs before those landed

zacharyhamm (approved):
This seems like a great improvement overall. When the disk cache is full it just keeps going. I think we should probably build in some memory overhead to the memory layer so we don't OOM (i'm trying to exhaust the ram on my machines but they... have a lot of ram)

zacharyhamm:
We'll need the disk cache portion to be more configurable. If the available size is smaller than the configured size, the system stops writing rebase batches

lib/si-layer-cache/src/db.rs
fnichol: Agreed!

lib/si-layer-cache/src/db/rebase_batch.rs
```diff
@@ -102,7 +102,7 @@ where
         let mut tried = 0;
         let read_wait = Instant::now();
         while tried < MAX_TRIES {
-            if let Some(v) = self.cache.memory_cache().get(&key).await {
+            if let Some(v) = self.cache.cache().get(&key).await {
```
fnichol, line 105: My first thought was that Foyer appears to check in memory (and early return) and then fall back to check on disk, and so our retry loop may no longer apply. However, does the `self.cache.get(..).await` call below fall back to querying postgres currently? If so, then I understand 😄

lib/si-layer-cache/src/db/workspace_snapshot.rs
```diff
@@ -50,7 +50,7 @@ where
         let key = WorkspaceSnapshotAddress::new(&postcard_value);
         let cache_key: Arc<str> = key.to_string().into();
 
-        self.cache.insert(cache_key.clone(), value_clone).await;
+        self.cache.insert(cache_key.clone(), value_clone);
```
fnichol, line 53: It's interesting that the enqueued disk write is handled out-of-band of the `insert` method call, making it non-async. Kind of clever

lib/si-layer-cache/src/db/workspace_snapshot.rs
```diff
@@ -102,7 +102,7 @@ where
         let mut tried = 0;
         let read_wait = Instant::now();
         while tried < MAX_TRIES {
-            if let Some(v) = self.cache.memory_cache().get(&key).await {
+            if let Some(v) = self.cache.cache().get(&key).await {
```
fnichol, line 105: Same question as above with respect to the while/retry loop

lib/si-layer-cache/src/hybrid_cache.rs
fnichol: Yeah, I think that's the case in our specific impl, that way a 30mb snapshot stays as a single 30mb hunk of heap memory and not `n` copies over time.

lib/si-layer-cache/src/hybrid_cache.rs
```diff
@@ -0,0 +1,174 @@
...
+pub struct CacheConfig {
+    disk_capacity: usize,
+    disk_path: PathBuf,
+    disk_admission_rate_limit: usize,
+    disk_buffer_size: usize,
+    disk_buffer_flushers: usize,
+    memory: u64,
+    memory_percentage: f32,
+    name: String,
+}
```
fnichol, line 144: Love that we own our own config here for managing that over time!

lib/si-layer-cache/src/hybrid_cache.rs
```diff
@@ -0,0 +1,174 @@
...
+    pub fn with_path_join(mut self, path: impl AsRef<Path>) -> Self {
+        self.disk_path = self.disk_path.join(path);
+        self
+    }
+}
+
+impl Default for CacheConfig {
+    fn default() -> Self {
+        let sys = sysinfo::System::new_all();
+        let path = tempfile::TempDir::with_prefix_in("default-cache-", "/tmp")
```
fnichol, line 169: I think I *finally* understand why I have so many `/tmp/default-cache-*` directories! This was already the case in main, but I couldn't remember if those dirs were from SI code. Hold on a second while I delete some stuff... 😄

lib/si-layer-cache/src/hybrid_cache.rs
```diff
@@ -0,0 +1,174 @@
...
+            .expect("unable to create tmp dir for layerdb")
```
fnichol, line 170: I *think* we're okay with this potential panic as we're constructing configs and even their caches early at boot time for our services (i.e. no delayed time bomb). Alternatively you may be able to do something like `Tempdir::with_prefix(...).unwrap_or("/tmp/default-cache-failed-tmpdir")...` (likely with some Rust type changes to thread the `PathBuf` type through. The downside is that path may not be unique and empty leading to other problems later on. 

I guess I'm now wishing for a `TryDefault` trait now!

lib/si-layer-cache/src/layer_cache.rs
```diff
@@ -18,66 +21,61 @@ pub struct LayerCache<V>
...
+        tracker.spawn(lc.clone().shutdown_handler(token.clone()));
+        Ok(lc)
     }
 
-    async fn spawn_disk_cache_write_vec(&self, key: Arc<str>, value: Vec<u8>) -> LayerDbResult<()> {
-        self.disk_cache().insert(key, value).await?;
+    async fn shutdown_handler(self: Arc<Self>, token: CancellationToken) -> LayerDbResult<()> {
+        token.cancelled().await;
+        debug!("shutting down layer cache {}", self.name);
+        self.compute_executor.shutdown();
```
fnichol, line 61: You should be able to skip shutting down the compute executor. I had to double check the impl, but calling `shutdown` will trigger the shutdown of all clones of the executor so there could be a shutdown race between this code and other other code that is using the same cloned executor. I was looking for where in layer-cache we were shutting this down, then didn't find it which made me check.

lib/si-layer-cache/src/layer_cache.rs
```diff
@@ -18,66 +21,61 @@ pub struct LayerCache<V>
...
+        self.cache().close().await?;
```
fnichol, line 62: I'm assuming this is where we await the flushing of all pending disk writes?

lib/dal/src/context.rs
```diff
@@ -524,10 +524,12 @@ impl DalContext {
         let events_actor = self.events_actor();
 
         let rebase_batch_address = slow_rt::spawn(async move {
```
fnichol, line 526: Not important for this PR, but I'm reminded I wanted to move the `slow_rt::spawn` calls to use the `self.compute_executor.spawn()` calls instead.

fnichol: Some follow on work, no doubt (that would need independent testing confirmation)

fnichol (changes requested):
This is looking great! I only had a couple of questions and maybe one request to drop the compute executor shutdown. Otherwise, we're ready to rock!

fnichol:
> We'll need the disk cache portion to be more configurable. If the available size is smaller than the configured size, the system stops writing rebase batches

That's a good catch!

lib/si-layer-cache/src/db/rebase_batch.rs
sprutton1: This one calls directly into the cache, so no postgres. I _think_ that we're waiting for the rebase to come across nats and be populated into memory here, hence the loop.

lib/si-layer-cache/src/db/workspace_snapshot.rs
sprutton1: It's kind of what we did, too, with the persister, but we exposed more of that detail up front than they do. The disk cache is optional technically, so it makes sense that this is this way

lib/si-layer-cache/src/hybrid_cache.rs
sprutton1: Yeah, my thinking is that since this is on boot, we likely want to fail here if we can't create the directories correctly as we probably will panic somewhere very shortly after this

lib/si-layer-cache/src/layer_cache.rs
sprutton1: I'll remove it. Should shutdown calls be dedup'd and no-op if it's called more than once? Otherwise you kind of need to know that this has been cloned and shutdown elsewhere

lib/si-layer-cache/src/layer_cache.rs
sprutton1: Yep, flushes and reclamation wait here. I'll add a comment

zacharyhamm (approved):

lib/si-layer-cache/src/db/rebase_batch.rs
fnichol: Right, that does make sense!

fnichol (dismissed):
Look good!

zacharyhamm (approved):

PR #4877, opened 2024-10-25
audit logs JSON display and other improvements

wendybujalski:
- Each event row can now be expanded to see the raw JSON data for each event
- A button appears on the header area to collapse all rows when any rows are open
- Each timestamp cell now has a tooltip to display the raw timestamp data
- Each change set cell now has a tooltip to display the change set id
- Also fixed some typos in the auth-api README
- Posthog tracking each time a set of logs are loaded (including when the user first navigates to the Audit page)

### DIFF
--- a/app/web/src/components/AuditLogCell.vue
+++ b/app/web/src/components/AuditLogCell.vue
@@ -1,0 +1,73 @@
+<template>
+  <td
+    v-tooltip="tooltip"
+    align="center"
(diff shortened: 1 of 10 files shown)


stack72 (dismissed):

nickgerace (approved):
Group reviewed

nickgerace (approved):

PR #4885, opened 2024-10-29
AI Asset Generator

jkeiser:
This adds a button to the asset schema code editor that lets you generate your asset schema function with AI.

https://github.com/user-attachments/assets/9cfd344e-2188-45e5-9855-cededb8ed23e

* Generation is imperfect, but passable. This turn is about hooking it up end to end so we can iterate more rapidly.
* AI generation is in the new `asset_sprayer` crate, which `sdf` calls out to to generate the code. `dal` does not depend on this.
* Locked behind the `ai-generator` feature flag, which is restricted to SI internal only
* Use `SI_OPENAI_API_KEY` when launching `buck2` to pass your API key
* Use `SI_ASSET_SPRAYER_PROMPTS_DIR=<path to your si dir>/lib/asset-sprayer/prompts` when launching `buck2` to have it load prompts from disk for faster prompt dev cycle
* Deployable using SI config files with `[openai] api_key=...` in config file

### DIFF
--- a/app/web/src/components/Ai/GenerateAwsAssetSchemaPanel.vue
+++ b/app/web/src/components/Ai/GenerateAwsAssetSchemaPanel.vue
@@ -1,0 +1,73 @@
+<template>
+  <div class="flex flex-row gap-xs items-end p-xs justify-end w-full">
+    <VormInput
+      v-model="awsCommand.cli"
+      label="CLI"
+      type="text"
+      disabled
+      :maxLength="3"
+      @enterPressed="generateAwsAssetSchema"
+    />
+    <VormInput
+      v-model="awsCommand.command"
+      label="Command"
+      type="text"
+      :disabled="generateRequestStatus.isPending"
+      @enterPressed="generateAwsAssetSchema"
+    />
+    <VormInput
+      v-model="awsCommand.subcommand"
+      label="Subcommand"
+      type="text"
+      :disabled="generateRequestStatus.isPending"
+      @enterPressed="generateAwsAssetSchema"
+    />
+    <VButton
+      v-tooltip="'Generate Schema From AWS Command'"
+      class="mb-[2px]"
+      :loading="generateRequestStatus.isPending"
+      loadingIcon="sparkles"
+      loadingText="Generating ..."
+      :requestStatus="generateRequestStatus"
+      label="Generate"
+      size="sm"
+      @click="generateAwsAssetSchema"
+    />
+  </div>
+</template>
+
+<script lang="ts" setup>
+import { reactive, PropType } from "vue";
+import { VormInput, VButton } from "@si/vue-lib/design-system";
+import { useAssetStore } from "@/store/asset.store";
+import { SchemaVariant } from "@/api/sdf/dal/schema";
+
+const assetStore = useAssetStore();
+
+const props = defineProps({
+  /** Asset we're generating for */
+  asset: { type: Object as PropType<SchemaVariant>, required: true },
+});
+
+/** The AWS command entered in the form */
+const awsCommand = reactive({
+  cli: "aws",
+  command: "sqs",
+  subcommand: "create-queue",
+});
+
+/** Status of the  */
+const generateRequestStatus = assetStore.getRequestStatus(
+  "GENERATE_AWS_ASSET_SCHEMA",
+  props.asset.schemaVariantId,
+);
+
+/** Actually trigger schema generation */
+const generateAwsAssetSchema = async () => {
+  await assetStore.GENERATE_AWS_ASSET_SCHEMA(
+    props.asset.schemaVariantId,
+    awsCommand.command,
+    awsCommand.subcommand,
+  );
+};
+</script>
--- a/app/web/src/components/AssetEditorHeader.vue
+++ b/app/web/src/components/AssetEditorHeader.vue
@@ -1,69 +1,109 @@
 <template>
-  <div class="p-xs flex flex-row gap-xs items-center">
-    <div class="flex flex-col min-w-0 flex-grow">
-      <TruncateWithTooltip
-        :showTooltip="showTooltip"
-        class="text-2xl font-bold pb-2xs flex flex-row items-center gap-xs"
-      >
-        <div
-          :class="
-            clsx(
-              'flex flex-row items-center gap-xs flex-none max-w-full',
-              selectedFunc && [
-                'cursor-pointer hover:underline',
-                themeClasses('text-action-500', 'text-action-300'),
-              ],
-            )
-          "
-          @click="onClick"
-        >
-          <NodeSkeleton :color="selectedAsset.color" size="mini" />
-          <TruncateWithTooltip ref="truncateRef1" hasParentTruncateWithTooltip>
-            {{ schemaVariantDisplayName(selectedAsset) }}
-          </TruncateWithTooltip>
-        </div>
+  <div class="flex flex-col flex-grow">
+    <!-- Main asset header -->
+    <div class="p-xs flex flex-row gap-xs items-center">
+      <div class="flex flex-col min-w-0 flex-grow">
         <TruncateWithTooltip
-          v-if="selectedFunc"
-          ref="truncateRef2"
-          hasParentTruncateWithTooltip
+          :showTooltip="showTooltip"
+          class="text-2xl font-bold pb-2xs flex flex-row items-center gap-xs"
         >
-          / {{ selectedFunc.kind }} / {{ selectedFunc.name }}
+          <div
+            :class="
+              clsx(
+                'flex flex-row items-center gap-xs flex-none max-w-full',
+                selectedFunc && [
+                  'cursor-pointer hover:underline',
+                  themeClasses('text-action-500', 'text-action-300'),
+                ],
+              )
+            "
+            @click="onClick"
+          >
+            <NodeSkeleton :color="selectedAsset.color" size="mini" />
+            <TruncateWithTooltip
+              ref="truncateRef1"
+              hasParentTruncateWithTooltip
+            >
+              {{ schemaVariantDisplayName(selectedAsset) }}
+            </TruncateWithTooltip>
+          </div>
+          <TruncateWithTooltip
+            v-if="selectedFunc"
+            ref="truncateRef2"
+            hasParentTruncateWithTooltip
+          >
+            / {{ selectedFunc.kind }} / {{ selectedFunc.name }}
+          </TruncateWithTooltip>
         </TruncateWithTooltip>
-      </TruncateWithTooltip>
-      <div
-        class="text-xs italic flex flex-row flex-wrap gap-x-lg text-neutral-600 dark:text-neutral-200 items-center"
-      >
-        <div>
-          <span class="font-bold">Asset Created At: </span>
-          <Timestamp :date="selectedAsset.created_at" size="long" />
+        <div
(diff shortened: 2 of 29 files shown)


app/web/src/components/AiButton.vue
```diff
@@ -0,0 +1,61 @@
...
+import { VormInput, IconButton, VButton } from "@si/vue-lib/design-system";
+import { useAssetStore } from "@/store/asset.store";
+
+const assetStore = useAssetStore();
+
+/** Whether the form is open or not */
+const open = ref(false);
+const setOpen = async (value: boolean) => {
+  open.value = value;
+}
```
nickgerace, line 45: I don't believe this has to be async, but I could be wrong

app/web/src/components/AssetEditorHeader.vue
```diff
@@ -46,6 +46,7 @@
       class="flex-none"
       :color="selectedAsset.color"
     />
+    <AiButton v-if="featureFlagsStore.AI_GENERATOR && selectedAsset && !selectedFunc" :selectedAsset="selectedAsset" />
```
nickgerace, line 49: I love the component layout here in the frontend

app/web/src/store/asset.store.ts
```diff
@@ -565,6 +565,25 @@ export const useAssetStore = (forceChangeSetId?: ChangeSetId) => {
...
+          return new ApiRequest<SchemaVariant>({
+            url: API_PREFIX.concat([id, "generate_aws_asset_schema"]),
+            params: { command, subcommand },
+            keyRequestStatusBy: id,
+            // onSuccess: () => {
+            //   if (code) {
+            //     const f = funcStore.funcCodeById[schemaVariant.assetFuncId];
+            //     if (f) f.code = code;
+            //   }
+            // }
```
nickgerace, line 584: Do we want to keep this for the future?

app/web/src/store/feature_flags.store.ts
```diff
@@ -13,6 +13,7 @@ const FLAG_MAPPING = {
   ON_DEMAND_ASSETS: "on_demand_assets",
   MANAGEMENT_FUNCTIONS: "management-functions",
   AUDIT_PAGE: "audit-page",
+  AI_GENERATOR: "ai-generator",
```
nickgerace, line 16: Flag makes sense!

bin/sdf/src/args.rs
```diff
@@ -202,6 +202,28 @@ pub(crate) struct Args {
     #[arg(long, env = "SI_AUTH_API_URL")]
     pub(crate) auth_api_url: Option<String>,
 
+    /// Override for the openai API key
+    #[arg(long, env = "SI_OPENAI_API_KEY")]
+    // TODO This should be a SensitiveString but if it is, it gets passed as "..."
+    // pub(crate) openai_api_key: Option<SensitiveString>,
+    pub(crate) openai_api_key: Option<String>,
```
nickgerace, line 209: Yes... good call out. I have a few ideas we can employ here, but this is non-blocking

lib/asset-sprayer/prompts/aws/asset_schema.yaml
nickgerace, line 1: Very clever

lib/asset-sprayer/src/config.rs
```diff
@@ -0,0 +1,45 @@
...
+
+impl SIOpenAIConfig {
+    pub fn into_openai_config_opt(self) -> Option<OpenAIConfig> {
+        if self.api_key.is_none()
+            && self.api_base.is_none()
+            && self.org_id.is_none()
+            && self.project_id.is_none()
+        {
+            return None;
+        }
```
nickgerace, line 28: I like this. Consume yourself to make the third party crate config.

lib/asset-sprayer/src/lib.rs
```diff
@@ -0,0 +1,118 @@
+//! This create provides centralized logic for working with the billing events NATS Jetstream stream.
```
nickgerace, line 1: We should reword this since it is leftover from another crate

lib/asset-sprayer/src/prompts.rs
```diff
@@ -0,0 +1,147 @@
...
+        match self {
+            Prompt::AssetSchema => "aws/asset_schema.yaml",
+        }
+    }
+
+    fn yaml_embedded(&self) -> &'static str {
+        match self {
+            Prompt::AssetSchema => include_str!("../prompts/aws/asset_schema.yaml"),
+        }
+    }
```
nickgerace, line 31: I get nervous when not using path joining features from the standard library, but I think it's safe to say these path separators in the strings will be fine.

lib/asset-sprayer/src/prompts.rs
```diff
@@ -0,0 +1,147 @@
...
+            ChatCompletionRequestMessage::System(message) => {
+                if let ChatCompletionRequestSystemMessageContent::Text(text) = &mut message.content
+                {
+                    *text = Self::replace_prompt_text(text.clone(), replace, prompt).await?;
+                }
+            }
+            _ => (),
+        }
+        Ok(message)
+    }
```
nickgerace, line 98: Is mutating the message passed in not a desirable option?

lib/asset-sprayer/src/prompts.rs
```diff
@@ -0,0 +1,147 @@
...
+        }
+
+        // Copy the remainder of the text
+        result.push_str(text);
+
+        Ok(result)
+    }
+
+    async fn get(url: &str) -> reqwest::Result<String> {
+        println!("Fetching: {}", url);
```
nickgerace, line 140: We should formally log this

lib/dal/src/schema/variant.rs
```diff
@@ -1057,6 +1057,11 @@ impl SchemaVariant {
         self.asset_func_id
     }
 
+    pub fn asset_func_id_or_error(&self) -> SchemaVariantResult<FuncId> {
+        self.asset_func_id
+            .ok_or(SchemaVariantError::MissingAssetFuncId(self.id))
+    }
```
nickgerace, line 1063: Nice!

lib/sdf-server/src/extract.rs
```diff
@@ -147,6 +148,25 @@ impl FromRequestParts<AppState> for HandlerContext {
...
+
+    async fn from_request_parts(
+        _parts: &mut Parts,
+        state: &AppState,
+    ) -> Result<Self, Self::Rejection> {
+        let asset_sprayer = state
+            .asset_sprayer()
+            .ok_or(not_found_error("openai not configured"))?;
+        // TODO why do we clone? Does this keep the pool around at all?
+        Ok(Self(asset_sprayer.clone()))
```
nickgerace, line 166: Because it is taken as a reference from the app state:
```rust
pub fn asset_sprayer(&self) -> Option<&AssetSprayer> {
    self.asset_sprayer.as_ref()
}
```

lib/sdf-server/src/service/v2/variant/generate_aws_asset_schema.rs
```diff
@@ -0,0 +1,94 @@
...
+        .build(access_builder.build(change_set_id.into()))
+        .await?;
+
+    // Generate the code
+    let code = asset_sprayer
+        .aws_asset_schema(&aws_command.command, &aws_command.subcommand)
+        .await?;
+
+    // Update the function
+    let force_change_set_id = ChangeSet::force_new(&mut ctx).await?;
```
nickgerace, line 47: This won't be an issue with your PR, but usually you see `force_change_set_id` at the top just in case `ctx` is used immediately. This isn't a blocker for me to move it higher... though as I'm saying this I'm realizing that footgun is big in many of our routes... not an issue for this PR though!

lib/sdf-server/src/service/v2/variant/generate_aws_asset_schema.rs
```diff
@@ -0,0 +1,94 @@
...
+        &variant.schema_name,
+        variant.display_name.clone(),
+        variant.category.clone(),
+        variant.description.clone(),
+        variant.link.clone(),
+        variant.color.clone(),
+        variant.component_type.into(),
+        Some(code),
+    )
+    .await?;
```
nickgerace, line 65: Love the minimal route. Is it possible to get the asset sprayer into an integration test? I suspect the required token makes that difficult, though I wonder if we can test the library without generating real code.

nickgerace (approved):
Excellent work! Nailed it.

app/web/src/components/AiButton.vue
jkeiser: Damn straight

bin/sdf/src/args.rs
jkeiser: Yeah, I'm quite sure I was just using it wrong and there's a secret incantation I'm missing!

lib/asset-sprayer/src/prompts.rs
jkeiser: oh lord, I'll fix this one before checkin

lib/sdf-server/src/extract.rs
jkeiser: oh, this would be a cargo culted comment. removing

lib/sdf-server/src/service/v2/variant/generate_aws_asset_schema.rs
jkeiser: Will do, I'm modifying other stuff here anyway.

stack72 (approved):

stack72 (dismissed):

stack72 (approved):

app/web/src/components/Ai/GenerateAwsAssetSchemaPanel.vue
```diff
@@ -0,0 +1,75 @@
+<template>
+  <div class="flex flex-row gap-xs place-items-end">
+    <VormInput
+      ref="cli"
```
wendybujalski, line 4: These refs do not appear to be in use - was this just pattern matching?

app/web/src/components/Ai/GenerateAwsAssetSchemaPanel.vue
```diff
@@ -0,0 +1,75 @@
...
+      type="text"
+      :disabled="generateRequestStatus.isPending"
+      @enterPressed="generateAwsAssetSchema"
+    />
+    <VButton
+      :loading="generateRequestStatus.isPending"
+      loadingIcon="sparkles"
+      loadingText="Generating ..."
+      :requestStatus="generateRequestStatus"
+      tooltip="Generate Schema From AWS Command"
```
wendybujalski, line 33: VButton doesn't have a tooltip prop, so to make this tooltip functional I'll change it to the v-tooltip Vue directive.

wendybujalski (commented):
Looks good! I've got a few UI tweaks and improvements that I'll push shortly and then I think this feature is ready to merge. We can keep it behind a feature flag until we're sure it's ready for users to try.

wendybujalski (dismissed):

app/web/src/components/Ai/GenerateAwsAssetSchemaPanel.vue
jkeiser: Yep! I'll try and remove them and if it compiles, win!

lib/asset-sprayer/src/prompts.rs
jkeiser: In some places it was better to return a new thing, and some places &mut was better, but since they were all the same function stack I just wanted to pick a lane :)

wendybujalski (approved):

PR #4886, opened 2024-10-29
Replace mocked audit log data with real data

nickgerace:
## Description

This PR replaces mocked audit log data with real data. It moves `AuditLog` to "V2", but all active streams are "beta" streams, so backwards compatibility should only matter for the lifespan of the future flag and the "beta" streams.

This PR also reduces the events down to just creating a component, updating a property editor value and deleting a component. Other event kinds have been removed and `AuditLog` versions now exist in their own submodules.

In addition to these changes, the frontend has adapted to the new audit log shape. However, the filters and dropdowns are broken, which we will fix as we determine what data should look like and how to interact with it.

The audit log page now has a loader for reloading the set of audit logs and contains a few changes to the headers. Unused fields have been dropped for now too.


stack72 (approved):

nickgerace:
Moving to drafts due to performance issues.

lib/dal/src/audit_logging.rs
```diff
@@ -2,53 +2,53 @@
...
-    #[error("workspace not found: {0}")]
-    WorkspaceNotFound(WorkspacePk),
+    #[error("user error: {0}")]
+    User(#[from] Box<UserError>),
+    #[error("user not found for id: {0}")]
+    UserNotFound(UserPk),
 }
 
-pub type AuditLoggingResult<T> = Result<T, AuditLoggingError>;
+type Result<T> = std::result::Result<T, AuditLoggingError>;
```
fnichol, line 51: Do you like this pattern? I think I am starting to like this myself. It's easy to work with within modules and it almost never needs to leak outside. And if it does, you can always pub-export an `AuditLoggingResult` type alias. Anyway, cool!

lib/dal/src/audit_logging.rs
```diff
@@ -166,3 +164,151 @@ pub(crate) async fn write_final_message(ctx: &DalContext) -> AuditLoggingResult<
...
+    let stream = stream_wrapper.stream().await?;
+    let consumer = stream
+        .create_consumer(async_nats::jetstream::consumer::pull::Config {
+            filter_subject: stream_wrapper.subject(workspace_id.into()).to_string(),
+            ..Default::default()
+        })
+        .await?;
+
+    // TODO(nick): remove hard-coded max messages value.
+    let mut messages = consumer.fetch().max_messages(200).messages().await?;
```
fnichol, line 187: Yeah, maybe one day an alternative `list_by_chunk` function that takes the size of messages and maybe a sequence number (or similar) as the cursor. I'm guessing you're working into a paging strategy...

lib/si-events-rs/src/audit_log.rs
```diff
@@ -1,53 +1,36 @@
...
+use crate::{Actor, ChangeSetId};
+
+mod v1;
+mod v2;
+
+pub type AuditLogKind = AuditLogKindV2;
+pub type AuditLogKindDiscriminants = AuditLogKindV2Discriminants;
+
+// TODO(nick): switch to something like "naxum-api-types" crate to avoid sizing issues.
+#[allow(clippy::large_enum_variant)]
```
fnichol, line 16: Well! I forgot about this scenario! I wonder if `Box`ing each variants member would not be too catastrophic for the API ergonomics. Reminds me...I bet there are some other request/reponse types which will suffer from this. And it's a little worse because these types get larger and larger over time when we are passing them around like they are pretty small. Anyway, very interesting find!

fnichol (commented):
Looking good so far!

lib/dal/src/audit_logging.rs
nickgerace: I do 😄  because I don't actually use the type externally: it is for convenience and footgun removal. I'd argue we should move towards this pattern because exporting result types feels like the wrong move "by default".

lib/dal/src/audit_logging.rs
nickgerace: Exactly. We have a "fake" paging strategy now, which is to dump a ton of logs and then simulate paging. I am also considering switching from this standard, non-durable pull consumer to an ordered push consumer that can perform this behavior. Right now, we're not so focused on the NATS consumption pieces and are focused on the frontend experience, but this will become important very soon.

lib/si-events-rs/src/audit_log.rs
nickgerace: Yeah this will be a pain. Frankly, I'm considering using that `naxum-api-types` crate I made from your Rebase Request versioning code that I ported over. I am a bit concerned about exploding enums here.

lib/dal/src/audit_logging.rs
jkeiser: I am a fan. A lot of crates do this, too!

stack72 (approved):

stack72 (approved):

PR #4891, opened 2024-10-30
Don't filter out bindings for schema variants that have components on the canvas

britmyerss:
<div><br/>via <a href="https://giphy.com/billions/">Billions</a> on <a href="https://giphy.com/gifs/billions-season-4-episode-1-showtime-550h6RQfuOuhtaDwdl">GIPHY</a></div>


nickgerace (approved):

zacharyhamm (approved):
This should work, but it means we'll touch every component on the graph when listing funcs. It might be better to implement a quick "has_components" that detects if there is a single use edge to a component for the SV

PR #4892, opened 2024-10-30
feat: add `WorkspacePermission` axum middleware & v2 change set apply

fnichol:

### DIFF
--- a/lib/sdf-server/src/service/session.rs
+++ b/lib/sdf-server/src/service/session.rs
@@ -12,7 +12,7 @@ use serde::{Deserialize, Serialize};
 use si_data_spicedb::SpiceDbError;
 use thiserror::Error;
 
-use crate::AppState;
+use crate::{middleware::WorkspacePermissionLayer, AppState};
 
 use super::ApiError;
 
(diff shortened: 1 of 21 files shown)


lib/sdf-server/src/service/session.rs
```diff
@@ -93,6 +93,8 @@ pub fn routes() -> Router<AppState> {
         .route("/load_workspaces", get(load_workspaces::load_workspaces))
         .route(
             "/refresh_workspace_members",
-            post(refresh_workspace_members::refresh_workspace_members),
+            post(refresh_workspace_members::refresh_workspace_members).layer(
```
britmyerss, line 96: Just for my own understanding - anyone who is an owner of a change set, whenever they change someone's permissions, this will kick in and here we're ensuring that only owners can refresh the workspace members, yeah?

fnichol: Ah yes, that's basically it!

stack72: that is already how the Auth API does it :)

sprutton1 (approved):

PR #4893, opened 2024-10-30
feat: create and manage other components

zacharyhamm:
Management functions can now take the components they manage as input, and they can create components, including components of other schemas. Created components are automatically "managed" by the component that creates them.


zacharyhamm:
@jobelenus and @vbustamante this is the first version of this that will expose the geometry. Do we need to rethink this before it goes live? I'm not sure how views + geometry will work

jobelenus (approved):

PR #4896, opened 2024-10-31
feat(sdf-server): Capturing the list_audit_logs events for Posthog

stack72:


wendybujalski (commented):
We are actually already firing this same Posthog event from the frontend UI, so this is not necessary.

PR #4899, opened 2024-10-31
Remove optional parameters from service.toml

jkeiser:
We don't use these in any production environment, and this will get us where we can deploy.

### DIFF
--- a/lib/asset-sprayer/src/config.rs
+++ b/lib/asset-sprayer/src/config.rs
@@ -1,20 +1,27 @@
 use async_openai::config::OpenAIConfig;
 use serde::{Deserialize, Serialize};
+use serde_with::{serde_as, NoneAsEmptyString};
 use si_std::SensitiveString;
-use std::path::PathBuf;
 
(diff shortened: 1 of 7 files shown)


lib/asset-sprayer/src/config.rs
```diff
@@ -1,20 +1,27 @@
...
 use serde::{Deserialize, Serialize};
+use serde_with::{serde_as, NoneAsEmptyString};
 use si_std::SensitiveString;
-use std::path::PathBuf;
 
 /// OpenAI configuration. Mirrors OpenAI configuration, but with Deserialize support.
+#[serde_as]
 #[derive(Clone, Debug, Default, Deserialize, Serialize)]
 pub struct SIOpenAIConfig {
+    #[serde_as(as = "NoneAsEmptyString")]
```
fnichol, line 10: Nice!

fnichol (approved):
That's a nice solution and should help when/if we need some optional settings that may or may not have values rendered out in config file templates

PR #4910, opened 2024-11-01
Solidify audit log structure in the UI

nickgerace:
## Description

This PR solidifies the audit log structure in the UI. In order to do this, there is now an `AuditLogV3`.

> [!NOTE]
> Backwards compatibility only matters for the "beta" streams and we can reset the counter once
we switch over to the production streams.

Now, we have a display name for the event, an entity type for the event and an entity name for the event.

This PR re-orders the columns and headers as well as adds selected and hover background coloration to the rows. It does not contain fixes for the filters nor are the selected and hover colors final. In those verticals, this PR is just another step towards where we want to be.

### DIFF
--- a/app/web/src/components/Workspace/WorkspaceAuditLog.vue
+++ b/app/web/src/components/Workspace/WorkspaceAuditLog.vue
@@ -22,9 +22,7 @@
               class="flex items-center gap-2xs pr-xs whitespace-nowrap flex-none"
             >
               <div>Page</div>
-              <div class="font-bold">
-                {{ currentFilters.page }} of {{ totalPages }}
-              </div>
+              <div class="font-bold">{{ currentPage }} of {{ totalPages }}</div>
             </div>
             <IconButton
               v-tooltip="
@@ -106,11 +104,13 @@
             <tr
               :class="
                 clsx(
-                  'h-md text-sm',
-                  themeClasses(
-                    'odd:bg-neutral-200 even:bg-neutral-100',
-                    'odd:bg-neutral-700 even:bg-neutral-800',
-                  ),
+                  'h-lg text-sm',
+                  rowCollapseState[Number(row.id)]
+                    ? 'bg-action-300'
+                    : themeClasses(
+                        'odd:bg-neutral-200 even:bg-neutral-100 hover:bg-action-200',
+                        'odd:bg-neutral-700 even:bg-neutral-800 hover:bg-action-200',
+                      ),
                 )
               "
             >
@@ -157,7 +157,7 @@ import {
   createColumnHelper,
 } from "@tanstack/vue-table";
 import clsx from "clsx";
-import { h, computed, ref, withDirectives, resolveDirective } from "vue";
+import { h, computed, ref } from "vue";
 import { trackEvent } from "@/utils/tracking";
 import { AuditLogDisplay, LogFilters, useLogsStore } from "@/store/logs.store";
 import { AdminUser } from "@/store/admin.store";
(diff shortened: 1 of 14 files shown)


app/web/src/components/Workspace/WorkspaceAuditLog.vue
```diff
@@ -22,9 +22,7 @@
               class="flex items-center gap-2xs pr-xs whitespace-nowrap flex-none"
             >
               <div>Page</div>
-              <div class="font-bold">
-                {{ currentFilters.page }} of {{ totalPages }}
-              </div>
+              <div class="font-bold">{{ currentPage }} of {{ totalPages }}</div>
```
nickgerace, line 25: Now you will see `Page 0 of 0` rather than `Page 1 of 0`.

app/web/src/components/Workspace/WorkspaceAuditLog.vue
```diff
@@ -106,11 +104,13 @@
...
-                    'odd:bg-neutral-200 even:bg-neutral-100',
-                    'odd:bg-neutral-700 even:bg-neutral-800',
-                  ),
+                  'h-lg text-sm',
+                  rowCollapseState[Number(row.id)]
+                    ? 'bg-action-300'
+                    : themeClasses(
+                        'odd:bg-neutral-200 even:bg-neutral-100 hover:bg-action-200',
+                        'odd:bg-neutral-700 even:bg-neutral-800 hover:bg-action-200',
+                      ),
```
nickgerace, line 113: We'll nail these colors and style in a follow-up PR, but this was aimed at just making it a little better in the meantime.

lib/dal/src/audit_logging.rs
```diff
@@ -184,50 +198,93 @@ pub async fn list(ctx: &DalContext) -> Result<Vec<FrontendAuditLog>> {
         })
         .await?;
 
-    // TODO(nick): remove hard-coded max messages value.
-    let mut messages = consumer.fetch().max_messages(200).messages().await?;
+    // TODO(nick): replace hard-coded max messages value with proper pagination.
+    let mut messages = consumer.fetch().max_messages(1000).messages().await?;
```
nickgerace, line 202: Just a reminder that we will need to stop doing the "post-process" filter and pagination we are doing now and will need to replace it with a real solution. We use a non-durable, regular pull consumer now and we may need to use an ordered push consumer (also non-durable by its nature) when we do this.

lib/dal/src/audit_logging.rs
```diff
@@ -184,50 +198,93 @@ pub async fn list(ctx: &DalContext) -> Result<Vec<FrontendAuditLog>> {
...
+                        .map_err(Box::new)?
+                        .ok_or(AuditLoggingError::ChangeSetNotFound(change_set_id.into()))?;
+                    match change_set.status {
+                        ChangeSetStatus::Abandoned
+                        | ChangeSetStatus::Failed
+                        | ChangeSetStatus::Rejected => {
+                            trace!(?change_set.status, ?change_set.id, "skipping change set for audit log assembly due to status");
+                            return Ok(None);
+                        }
+                        ChangeSetStatus::Applied
```
nickgerace, line 235: Right now, viewing events for a `Applied` change sets might be weird on HEAD since the filters are broken (known) at the moment. If you clear the filters, you'll be able to see the events, but it will be unclear that they are on HEAD.

lib/si-events-rs/src/audit_log/v3.rs
```diff
@@ -0,0 +1,14 @@
...
+#[derive(Debug, Deserialize, Serialize, Clone, PartialEq, Eq)]
+pub struct AuditLogV3 {
+    pub actor: Actor,
+    pub kind: AuditLogKindV3,
+    pub entity_name: Option<String>,
+    pub timestamp: String,
+    pub change_set_id: Option<ChangeSetId>,
+}
+
+pub type AuditLogKindV3 = super::v2::AuditLogKindV2;
```
nickgerace, line 14: Just re-use the last kind.

lib/si-frontend-types-rs/src/audit_log.rs
```diff
@@ -70,6 +75,22 @@ pub enum AuditLogDeserializedMetadata {
...
+        match discrim {
+            Discrim::CreateComponent => ("Created", "Component"),
+            Discrim::DeleteComponent => ("Deleted", "Component"),
+            Discrim::UpdatePropertyEditorValue => ("Updated property in Component", "Property"),
+            Discrim::UpdatePropertyEditorValueForSecret => {
+                ("Updated secret in Component", "Property for Secret")
+            }
+        }
+    }
+}
```
nickgerace, line 92: Right now, these aren't stored on the persisted logs themselves. We may want to change this, but this works for fast iteration on the feature before it is released.

nickgerace: @stack72 this might be useful if we want to tune the exact wording of things in the UI for a bit. This will require an SDF compile, so we could also move this to the frontend too.

lib/dal/src/audit_logging.rs
```diff
@@ -184,50 +198,93 @@ pub async fn list(ctx: &DalContext) -> Result<Vec<FrontendAuditLog>> {
...
+                timestamp: inner.timestamp,
+                change_set_id,
+                change_set_name,
+                metadata: serde_json::to_value(deserialized_metadata)?,
+            }))
+        }
+        AuditLog::V2(_) | AuditLog::V1(_) => {
+            debug!("skipping older audit logs in beta...");
+            Ok(None)
+        }
```
nickgerace, line 284: Just a reminder that for right now while the beta streams are active, you'll need to make new logs. Worrying about the conversations for logs that we will be throwing out later is likely wasted time. We'll see if we want to start converting if we want to see how logs live long term.

stack72 (approved):

PR #4911, opened 2024-11-01
Allow billing to support more than 2 subscriptions

jkeiser:
Right now we support only free trial and pay-as-you-go plans in our billing ingestion and price modeling. This allows us to support custom Enterprise plans and to move SI internal users to a new plan that doesn't generate invoices.

SQL
----

The SQL changes here primarily allow us to support more than two subscriptions, so that you can switch between multiple plans:

* The primary SQL change is to `latest_owner_subscriptions`, which now lists each subscription as a separate row instead of combining the previous 2 subscriptions into a single row per owner. Everything else is fallout from that.
* This also fixes a bug where we would report the last day's usage of the free trial against your pay-as-you-go plan. It would not have been invoiced, but it's inaccurate and would have caused larger problems as we move between two plans that *do* charge. The issue was that the end date of the free trial is a *date* and therefore 1 day off from the start date of the pay-as-you-go subscription. We want to report all the hours during that gap day to the free trial. The new verification views caught this issue.
* The `workspace_verifications` schema is now real and tests for a much more extensive set of potential errors. We will follow up with a lambda that runs these verifications so that we can alert if they begin showing us issues.
* I tested these in production by putting a new set of views alongside the old ones, so the results could be compared:
  - The new views produce identical results to the old ones for our current 2 plans.
  - There are no new errors (after running the verification views).
  - These are entirely readonly procedures and cannot affect existing data unless one replaces the existing views (which I haven't done--and won't until we move through review).

Lambdas
----------
The lambda changes make it easier to set up and run the lambdas, as well as override environment variables:

* Run `source ./setup_venv.sh` and you can run them at that point.
* Common `SiLambda` handles common environment, logging and Lago/Redshift configuration.
* The event passed to the lambda can contain environment variable overrides, allowing you to test behavior without affecting the normal runs.
* `si_logger` is removed in favor of `logging.debug/info/error`.
* logging configuration happens in `SiLambda()` and honors the `SI_LOG_LEVEL` environment variable.
* Pass environment variables in the event to override them without changing the main lambda (if you want to test a lambda with different parameters).
* New dry run env var with `SI_DRY_RUN` to make testing general lambda runs safer.

The output of the views used for billing uploads is not different from before; the lambdas will need to be adjusted to right now compared to before indeed need to be modified to *accept* more than 2 plans. However, as long as there are 2 plans they will behave the same as before.


component/lambda/functions/billing-set-prices.py
```diff
@@ -1,140 +1,139 @@
...
     LagoSubscription,
 )
 from si_types import OwnerPk
 
 
-session = boto3.Session()
-redshift = Redshift.from_env(session)
-lago = LagoApi.from_env(session)
-
 paul = OwnerPk("01GW0KXH4YJBWC7BTBAZ6ZR7EA")
```
johnrwatson, line 14: nitpick: this should be removed

component/lambda/functions/billing-set-prices.py
```diff
@@ -1,140 +1,139 @@
...
-def get_subscription(
-    external_subscription_id: ExternalSubscriptionId,
-) -> LagoSubscription:
-    try:
-        return cast(
-            LagoSubscription,
-            lago.get(f"/api/v1/subscriptions/{external_subscription_id}").json()[
-                "subscription"
-            ],
+        Prices will be set based on the user's usage during this month.
```
johnrwatson, line 25: query: Can you elaborate slightly? I don't quite follow why if the day is >5 that you would want to run this logic

component/lambda/functions/upload-billing-hours.py
```diff
@@ -1,110 +1,106 @@
...
 
-
-session = boto3.Session()
-redshift = Redshift.from_env(session)
-lago = LagoApi.from_env(session)
-
-
-paul = OwnerPk("01GW0KXH4YJBWC7BTBAZ6ZR7EA")
 billable_metric_code = "resource-hours"
 cost_per_resource_hour = 0.007
```
johnrwatson, line 11: query: if we need to add a custom plan price for a user, are we able to do that / where are the user:pricing mappings (if anywhere)

johnrwatson (approved):
Looks good to me, but testing it / having it live is the only real way of validating a change like this

component/lambda/functions/billing-set-prices.py
jkeiser: Will do! The reason is that we need to use *last* month's price if we're past the 1st of the month but last month's invoices aren't finalized yet (due to the grace period and the fact that last month's invoice will pick up any price overrides automatically until it is finalized).

component/lambda/functions/upload-billing-hours.py
jkeiser: There are none--right now the system supports (a) the pay-as-you-go plan with its 100-concurrent-resources threshold; and (b) plans where the price is set in Lago and don't have that kind of threshold. Custom enterprise plans are not so likely to have a threshold where they go to zero, and this keeps us from having to maintain a big mapping for now.

johnrwatson (approved):

PR #4915, opened 2024-11-04
feat(audit-logs): Extending the audit logs api

stack72:
Including things like RunAction and ChangeSet manipulation in the audidt logs

### DIFF
--- a/app/web/src/components/AuditLogHeader.vue
+++ b/app/web/src/components/AuditLogHeader.vue
@@ -10,7 +10,7 @@
           header.id !== 'json' &&
           'cursor-pointer hover:underline',
         header.id === 'json' && 'w-8 px-2xs',
-        themeClasses('bg-shade-0', 'bg-shade-100'),
+        themeClasses('bg-neutral-400 text-black', 'bg-shade-100'),
       )
     "
     @mousedown="startActive"
--- a/app/web/src/components/Workspace/WorkspaceAuditLog.vue
+++ b/app/web/src/components/Workspace/WorkspaceAuditLog.vue
@@ -106,10 +106,10 @@
                 clsx(
                   'h-lg text-sm',
                   rowCollapseState[Number(row.id)]
-                    ? 'bg-action-300'
+                    ? themeClasses('bg-action-200', 'bg-action-900')
                     : themeClasses(
-                        'odd:bg-neutral-200 even:bg-neutral-100 hover:bg-action-200',
-                        'odd:bg-neutral-700 even:bg-neutral-800 hover:bg-action-200',
+                        'odd:bg-neutral-200 even:bg-neutral-100 hover:border hover:border-action-500',
+                        'odd:bg-neutral-700 even:bg-neutral-800 hover:border hover:border-action-300',
                       ),
                 )
               "
@@ -190,7 +190,9 @@ const DEFAULT_FILTERS = {
   excludeSystemUser: false,
   kindFilter: [],
   serviceFilter: [],
-  changeSetFilter: [changeSetsStore.selectedChangeSetId],
+  changeSetFilter: changeSetsStore.headSelected
+    ? []
+    : [changeSetsStore.selectedChangeSetId],
   userFilter: [],
 } as LogFilters;
 const currentFilters = ref<LogFilters>({ ...DEFAULT_FILTERS });
--- a/lib/dal/src/audit_logging.rs
+++ b/lib/dal/src/audit_logging.rs
@@ -226,13 +226,12 @@ async fn assemble_for_list(
                         .map_err(Box::new)?
                         .ok_or(AuditLoggingError::ChangeSetNotFound(change_set_id.into()))?;
                     match change_set.status {
-                        ChangeSetStatus::Abandoned
-                        | ChangeSetStatus::Failed
-                        | ChangeSetStatus::Rejected => {
+                        ChangeSetStatus::Failed | ChangeSetStatus::Rejected => {
                             trace!(?change_set.status, ?change_set.id, "skipping change set for audit log assembly due to status");
                             return Ok(None);
                         }
-                        ChangeSetStatus::Applied
+                        ChangeSetStatus::Abandoned
+                        | ChangeSetStatus::Applied
                         | ChangeSetStatus::Approved
                         | ChangeSetStatus::NeedsAbandonApproval
                         | ChangeSetStatus::NeedsApproval
--- a/lib/dal/src/context.rs
+++ b/lib/dal/src/context.rs
@@ -582,6 +582,11 @@ impl DalContext {
     }
 
     pub async fn commit_no_rebase(&self) -> TransactionsResult<()> {
+        // Since we are not rebasing, we need to write the final message and flush all
+        // pending audit logs.
+        self.write_audit_log_final_message().await?;
+        self.publish_pending_audit_logs(None, None).await?;
+
         if self.blocking {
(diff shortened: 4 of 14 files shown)


app/web/src/components/AuditLogHeader.vue
```diff
@@ -10,7 +10,7 @@
           header.id !== 'json' &&
           'cursor-pointer hover:underline',
         header.id === 'json' && 'w-8 px-2xs',
-        themeClasses('bg-shade-0', 'bg-shade-100'),
+        themeClasses('bg-neutral-400 text-black', 'bg-shade-100'),
```
nickgerace, line 13: We should drop `text-black`, but this is non-blocking and I'll do in a follow-up PR.

app/web/src/components/Workspace/WorkspaceAuditLog.vue
```diff
@@ -106,10 +106,10 @@
                 clsx(
                   'h-lg text-sm',
                   rowCollapseState[Number(row.id)]
-                    ? 'bg-action-300'
+                    ? themeClasses('bg-action-200', 'bg-action-900')
```
nickgerace, line 109: This will also need the hover classes we added, but that is also non-blocking and I will do it in a follow-up PR.

app/web/src/components/Workspace/WorkspaceAuditLog.vue
```diff
@@ -106,10 +106,10 @@
...
                     : themeClasses(
-                        'odd:bg-neutral-200 even:bg-neutral-100 hover:bg-action-200',
-                        'odd:bg-neutral-700 even:bg-neutral-800 hover:bg-action-200',
+                        'odd:bg-neutral-200 even:bg-neutral-100 hover:border hover:border-action-500',
+                        'odd:bg-neutral-700 even:bg-neutral-800 hover:border hover:border-action-300',
```
nickgerace, line 112: We will break this out into a computed in a follow-up PR.

app/web/src/components/Workspace/WorkspaceAuditLog.vue
```diff
@@ -190,7 +190,9 @@ const DEFAULT_FILTERS = {
   excludeSystemUser: false,
   kindFilter: [],
   serviceFilter: [],
-  changeSetFilter: [changeSetsStore.selectedChangeSetId],
+  changeSetFilter: changeSetsStore.headSelected
+    ? []
+    : [changeSetsStore.selectedChangeSetId],
```
nickgerace, line 195: @stack72 What is this change?

lib/dal/src/audit_logging.rs
```diff
@@ -226,13 +226,12 @@ async fn assemble_for_list(
...
                     match change_set.status {
-                        ChangeSetStatus::Abandoned
-                        | ChangeSetStatus::Failed
-                        | ChangeSetStatus::Rejected => {
+                        ChangeSetStatus::Failed | ChangeSetStatus::Rejected => {
                             trace!(?change_set.status, ?change_set.id, "skipping change set for audit log assembly due to status");
                             return Ok(None);
                         }
-                        ChangeSetStatus::Applied
+                        ChangeSetStatus::Abandoned
```
nickgerace, line 233: @stack72 do you want to show all abandoned audit logs? I thought we didn't want to do that.

I suspect this is because you want to show the "I abandoned this change set" event only, but not what you did in that change set before abandoning it. Is that right?

lib/dal/src/context.rs
```diff
@@ -582,6 +582,11 @@ impl DalContext {
     }
 
     pub async fn commit_no_rebase(&self) -> TransactionsResult<()> {
+        // Since we are not rebasing, we need to write the final message and flush all
+        // pending audit logs.
+        self.write_audit_log_final_message().await?;
+        self.publish_pending_audit_logs(None, None).await?;
```
nickgerace, line 588: Perfect

lib/sdf-server/src/service/change_set/abandon_change_set.rs
```diff
@@ -41,6 +42,8 @@ pub async fn abandon_change_set(
         .await?
         .ok_or(ChangeSetError::ChangeSetNotFound)?;
 
+    let name = change_set.clone().name;
```
nickgerace, line 45: nit: we can just call `change_set.name.to_owned()` so we only clone the field. Non-blocking and I'll fix in a follow-up.

lib/sdf-server/src/service/change_set/apply_change_set.rs
```diff
@@ -85,6 +86,9 @@ pub async fn apply_change_set(
         }),
     );
 
+    ctx.write_audit_log(AuditLogKind::ApplyChangeset {}, "-".to_string())
```
nickgerace, line 89: Ah I think we'll need the backend API to make this optional after all... I'll fix that in a follow-up.

lib/si-events-rs/src/audit_log/v2.rs
```diff
@@ -50,4 +53,36 @@ pub enum AuditLogKindV2 {
         after_secret_name: Option<String>,
         after_secret_id: Option<SecretId>,
     },
+    ApplyChangeset {},
```
nickgerace, line 56: This doesn't need `{}`, but I'll change that in a follow-up.

lib/si-frontend-types-rs/src/audit_log.rs
```diff
@@ -73,6 +73,42 @@ pub enum AuditLogDeserializedMetadata {
...
+        func_name: String,
+    },
+    ActionRun {
+        prototype_id: ActionPrototypeId,
+        action_kind: ActionKind,
+        func_id: FuncId,
+        func_display_name: Option<String>,
+        func_name: String,
+        run_status: bool,
+    },
```
nickgerace, line 111: These also need the `camelCase`

app/web/src/components/Workspace/WorkspaceAuditLog.vue
stack72: This removes filtering to HEAD when on HEAD - it shows all changes across the ecosystem

lib/dal/src/audit_logging.rs
stack72: yes @britmyerss wanted to show those - believes it's the right level of detail to show

lib/si-frontend-types-rs/src/audit_log.rs
stack72: great spot!

lib/si-events-rs/src/audit_log/v2.rs
```diff
@@ -1,7 +1,10 @@
...
 use strum::Display;
 
-use crate::{Actor, AttributeValueId, ChangeSetId, ComponentId, PropId, SchemaVariantId, SecretId};
+use crate::{
+    ActionKind, ActionPrototypeId, Actor, AttributeValueId, ChangeSetId, ComponentId, FuncId,
+    PropId, SchemaVariantId, SecretId,
+};
 
 #[derive(Debug, Deserialize, Serialize, Clone, PartialEq, Eq)]
 pub struct AuditLogV2 {
```
nickgerace, line 10: Calling this out for other reviewers: we are adding variants to the end of an enum that is serialized and deserialized with `serde_json`, so this is safe for backwards compatibility. We do not use a special form of serialization here: just the standard one.

nickgerace (approved):

PR #4916, opened 2024-11-04
feat: connect components in management functions

zacharyhamm:
Create and update in management functions can now create connections to managed components.


lib/dal/src/management/mod.rs
```diff
@@ -321,20 +653,38 @@ impl<'a> ManagementOperator<'a> {
...
-                    let mut component = Component::get_by_id(self.ctx, component_id).await?;
-                    component
-                        .set_raw_geometry(self.ctx, raw_geometry.to_owned())
-                        .await?;
+
+                if let Some(remove_conns) = &update_conns.remove {
+                    for to_remove in remove_conns {
+                        self.remove_connection(component_id, to_remove).await?;
+                    }
                 }
```
nickgerace, line 679: Does it matter if you add before remove or vice versa?

zacharyhamm: Hmm that's a good question. I guess if you wanted to "reconnect" we'd want to ensure removes happened before adds

jobelenus (approved):

jobelenus (approved):

PR #4923, opened 2024-11-05
chore(module-index): Update sea-orm

stack72:
This was required to protect against a dependabot reported vulnerability https://github.com/systeminit/si/security/dependabot/325

### DIFF
--- a/lib/module-index-server/src/models/si_module.rs
+++ b/lib/module-index-server/src/models/si_module.rs
@@ -90,7 +90,7 @@ impl sea_query::ValueType for ModuleKind {
     }
 
     fn column_type() -> sea_query::ColumnType {
-        sea_query::ColumnType::String(None)
+        sea_query::ColumnType::String(StringLen::None)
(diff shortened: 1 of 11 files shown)


zacharyhamm (approved):

lib/module-index-server/src/models/si_module.rs
```diff
@@ -90,7 +90,7 @@ impl sea_query::ValueType for ModuleKind {
     }
 
     fn column_type() -> sea_query::ColumnType {
-        sea_query::ColumnType::String(None)
+        sea_query::ColumnType::String(StringLen::None)
```
stack72, line 93: This is breaking change they introduced in Sea ORM 1.x.x

zacharyhamm:
We should be safe against this vulnerability since i believe we limit content size to well below 4GB

fnichol (approved):
Yep, that's looking about right, nicely done!

PR #4924, opened 2024-11-05
Move the Audit Trail feature out of beta

nickgerace:
## Description

This PR not only moves the Audit Trail feature out of beta, but also adds some UI polish, adds dependent values update logging and formalizes the underlying audit logging architecture.

## UI Changes

- Keep hover border enabled whether or not the row has been selected

## Log Structure Changes

- Reset the `AuditLog` and `AuditLogKind` version counter back to `V1` since we are moving from the beta streams to the production streams
- Rename `display_name` to `title` since we do not want to imply that the field alone paints the full picture (the frontend uses `entity_type` in the row right after the `title` to form the phrase)
- Require the entity name

## Log Kind Changes

- Add `UpdateDependentInputSocket`, `UpdateDependentOutputSocket` and `UpdateDependentProperty` kinds
- Rename `ActionRun` to `RunAction` to meet `AuditLogKind` conventions
- Rename `ChangeSet`-related events to ensure the casing is `ChangeSet` rather than `Changeset`
- Use the change set name for the entity name for `ApplyChangeSet`
- Remove empty `structs` from kinds without any metadata fields (i.e. no empty braces)
- Use `remain::sorted` for both the backend and frontend kinds since `serde_json` serialization and deserialization allows for new kinds inserted anywhere in the enum while maintaining backwards compatibility

## Production Changes

- Create and use two NATS JetStream streams: `AUDIT_LOGS` and `PENDING_EVENTS` (services now use `audit.log.>` and `pending.event.>` prefixes)
- Drop usage of two NATS JetStream streams: `AUDIT_LOGS_BETA` and `PENDING_EVENTS_BETA` (services no longer use `beta.audit.log.>` and `beta.pending.event.>` prefixes)

## Secondary Changes

- Add new `audit_log` submodule in dependent values update that's contained in the same file with its own imports and error type
- Collect the "input" `AttributeValueIds` when preparing prototype arguments for prototype function execution
- Cache the `before_value` before spawning a prototype function execution in dependent values update
- Refactor the audit logging integration test to be dynamic to the number of audit logs generated
- Document the `AuditLog` frontend type since it will be the stable interface for the frontend and the backend interfaces are versioned


lib/dal/src/audit_logging.rs
```diff
@@ -217,7 +216,7 @@ async fn assemble_for_list(
     audit_log: AuditLog,
 ) -> Result<Option<FrontendAuditLog>> {
     match audit_log {
-        AuditLog::V3(inner) => {
+        AuditLog::V1(inner) => {
```
zacharyhamm, line 219: V3 to V1? Just surprising

lib/si-events-rs/src/audit_log/v1.rs
```diff
@@ -1,35 +1,134 @@
...
-    Eq,
-    PartialEq,
-    Hash,
-    AsRefStr,
-    Display,
-    EnumIter,
-    EnumString,
-)]
+#[derive(Debug, Deserialize, Serialize, Clone, PartialEq, Eq, Display, EnumDiscriminants)]
 pub enum AuditLogKindV1 {
```
fnichol, line 20: This looks good--and even convenient to see all the possibilities together. I wonder long term if all the embedded fields will push the size of the enum to be larger than expected. Although for this to happen, we'd have to have a variant with many more fields or a field with a large stack size type. Anyway, nothing actionable, just reflecting out loud 😄

fnichol (dismissed):
Looking good here! I don't know if I'm going to have the time to test this out locally though. The NATS stream transitions look good though so I wouldn't expect any upgrade/deployment issues

lib/si-events-rs/src/audit_log/v1.rs
nickgerace: Thank you! I think this is going to be a pain soon enough. I'm okay with it for now due to the amount of escape hatches we have available to us, but yes, you are right. It's going to get huge.

lib/dal/src/job/definition/dependent_values_update.rs
```diff
@@ -485,6 +505,12 @@ async fn execution_error_detail(
     ))
 }
 
+type PrototypeFunctionExecutionResult = (
```
zacharyhamm, line 508: Would be even better as a struct (not a blocker). I'm guilty of it too but we should definitely just make a type when we start returning a 3-tuple or bigger

zacharyhamm (approved):
🚢 🇮🇹

lib/dal/src/audit_logging.rs
nickgerace: Yes. Ideally, we would have went from `VBeta3` to `V1` or something analogous. We are wholesale dropping the old beta streams in favor of the production ones, so this is a good time to reset the counter.

lib/dal/src/job/definition/dependent_values_update.rs
nickgerace: Absolutely. I think we need to do that precisely. I can do it in my follow-up PR to fix the filters.

PR #4926, opened 2024-11-05
feat: modify schemas managed by management prototype

zacharyhamm:
Adds managed schemas to the management bindings, and allows you to edit them from the function editor. This makes it possible to author management functions that create, update, and connect other asset/variant types than the managing component's type.

### DIFF
--- a/app/web/src/components/FuncEditor/ManagementDetails.vue
+++ b/app/web/src/components/FuncEditor/ManagementDetails.vue
@@ -1,31 +1,94 @@
 <template>
-  <div class="p-xs flex flex-col gap-xs">
-    <div class="text-neutral-700 type-bold-sm dark:text-neutral-50">
+  <div class="p-xs flex flex-col gap-xs w-full">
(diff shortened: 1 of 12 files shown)


zacharyhamm:
@britmyerss the changes to `update_bindings` are the ones i was talking about earlier

app/web/src/components/FuncEditor/ManagementDetails.vue
```diff
@@ -35,6 +98,66 @@ const binding = computed(() => {
...
+    ),
+  );
+  filteredOptions.sort((a, b) => a.label.localeCompare(b.label));
+
+  return filteredOptions;
+});
+
+const addSchema = async () => {
+  if (binding.value) {
+    const updated_binding = _.clone(binding.value);
```
jobelenus, line 137: BTW, I prefer using `structuredClone` (yay ES6) rather than lodash. It is both faster and more thorough/correct.

stack72 (approved):

PR #4927, opened 2024-11-05
Use SpiceDB to facilitate Change Set Approvals end to end

britmyerss:
This PR includes the following changes: 
- When the `REBAC` feature flag is enabled, users will no longer see the consensus voting approval and will instead use the new approval flow that leverages SpiceDB for authorization
- Note: it is not recommended to have multiple users in the workspace with varying states of toggle
- Uses the `list_subjects` function of SpiceDB to find all users that have the permission to Approve a change set being applied (and extends an integration test for this) 
- Note: this should be refactored to follow a more ergonomic interface, but that will happen as a follow up.
- If a user is an approval, they'll be able to `Approve and Apply` a change set
- If a user is not an approver, they will be able to request approval and withdraw their request
- Non-approvers can re-open a change set before it has been applied, and only once it has been approved will they be able to apply
- Provides a way for approvers to see all change sets that need review, from anywhere. 

The way to test this locally is to add another user to your local workspace, then enable the feature flag for both users! 

<div><br/>via <a href="https://giphy.com/PermissionIO/">PermissionIO</a> on <a href="https://giphy.com/gifs/PermissionIO-power-bruce-almighty-ive-got-the-BtEw37CXZti8yfq3Ke">GIPHY</a></div>

### DIFF
--- a/app/web/src/api/sdf/dal/change_set.ts
+++ b/app/web/src/api/sdf/dal/change_set.ts
@@ -8,6 +8,8 @@ export enum ChangeSetStatus {
   Abandoned = "Abandoned",
   NeedsApproval = "NeedsApproval",
   NeedsAbandonApproval = "NeedsAbandonApproval",
+  Rejected = "Rejected",
+  Approved = "Approved",
 }
 
 export type ChangeSetId = string;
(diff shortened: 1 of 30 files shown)


wendybujalski (dismissed):
👍

jobelenus (dismissed):

jobelenus (approved):

PR #4929, opened 2024-11-06
Sync subscriptions from Lago when they change

jkeiser:
This checks for subscription changes in Lago and syncs them to SI when they change or new ones are added.

It also switches workspace_delegations_population.py to use the shared libraries, and moves the Auth API stuff to its own module.

### DIFF
--- a/component/lambda/functions/workspace_delegations_population.py
+++ b/component/lambda/functions/workspace_delegations_population.py
@@ -1,336 +1,215 @@
-import os
+from collections.abc import Iterable
 import json
-import boto3
-import botocore
-from pip._vendor import requests
-import botocore.session as bc
-from botocore.client import Config
+from typing import NotRequired, Optional, TypedDict, cast, overload
 import time
 from datetime import datetime
+from si_lambda import SiLambda, SiLambdaEnv
+from si_types import WorkspaceId, OwnerPk, SqlTimestamp
+from si_lago_api import ExternalSubscriptionId, LagoSubscription, LagoSubscriptionsResponse, IsoTimestamp
+import logging
+from itertools import groupby
 
-secret_name = os.environ['LAMBDA_REDSHIFT_ACCESS']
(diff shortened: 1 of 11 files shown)


johnrwatson:
Is this PR a mix of two PRS? It's got content about the AI asset sprayer too in it

component/lambda/functions/workspace_delegations_population.py
```diff
@@ -1,336 +1,212 @@
...
+            SELECT owner_pk, subscription_id, subscription_start_date, subscription_end_date, plan_code, external_id
+              FROM workspace_operations.owners
+              LEFT OUTER JOIN workspace_operations.latest_owner_subscriptions USING (owner_pk)
+        """))
+
+        for owner_pk, si_subscriptions in groupby(latest_owner_subscriptions, lambda sub: sub['owner_pk']):
+            si_subscriptions = { sub['external_id']: sub for sub in si_subscriptions }
+
+            # Get Lago subscriptions for the owner
+            lago_subscriptions = cast(LagoSubscriptionsResponse, self.lago.get(f"/api/v1/subscriptions?external_customer_id={owner_pk}&status[]=pending&status[]=active&status[]=terminated&status[]=active").json())
```
johnrwatson, line 106: I don't think this is a good idea to run every few minutes in the workspace delegations, I reckon we'll get rate limited by Lago or the lambda will timeout.

e.g. If we have 25000 customer records in Lago this one lambda will take at least 20 minutes to just iterate through this loop (assuming 50ms per call), where the maximum single lambda runtime is 15 minutes.

jkeiser: We discussed this, and I agree :) We have a little time, but I need to wait until after the next AI turn. What we planned:

1. Separate missing workspace upload vs. subscription sync into separate lambdas
2. Update record_timestamps if nothing has changed for a customer; only pick the N most out of date customers on each run

johnrwatson (approved):

PR #4930, opened 2024-11-06
chore(si-layer-cache): rework cache config to include reservations

fnichol:
This change is intended to make a running SI service with a layer cache more resilient when running under load over a long period of time. The biggest change is introducing the concept of a reserved percentage of disk and memory space that a lower level cache will *not* use.

The interesting tunables in a `CacheConfig` are:

- `memory_reserved_percent` (default `40`%) - the percentage of total system memory that will never be used by the cache. This means that if there was 10 gigabytes of RAM detected, then 4 gigabytes would never be used for the cache, but for other parts of the running service and other system processes (i.e. systemd services, SSH, logging, agents, etc.)
- `memory_usable_max_percent` (default `100`%) - the percentage of remaining system memory that will be allocated to the cache. An amount of 100% would be the amount of total RAM minus the reserved RAM (as chosen above). In the 10 gigabytes of RAM detected example, with a 40% reservation value, there would be 6 gigabytes remaining. If the `memory_usable_max_percent` value was 50%, then the configured amount of RAM for the cache would be `10 * (1 - 0.4) * 0.5 = 3 gigabytes`.
- `disk_reserved_percent` (default `5`%) - the percentage of total disk space that will never be used for the cache. The total disk space is determined by the `disk_path` provided and will correspond to the mount point. If the detected disk size was 10 gigabytes, then 0.5 gigabytes wouldn't be used by the cache with a 5% value.
- `disk_usable_max_percent` (default `100`%) - the percentage of remaining disk space that will be allocated to the cache. As in the memory version, with a 5% reservation and a 50% usable max value, the configured disk cache size would be `10 * (1 - 0.05) * 0.5 = 4.75 gigabytes`.


zacharyhamm (approved):
Nicely done

johnrwatson (approved):

PR #4931, opened 2024-11-06
feat(sdf, dal): Add audit log for a user to a workspace

stack72:


lib/sdf-server/src/service/session/auth_connect.rs
```diff
@@ -205,6 +206,9 @@ async fn find_or_create_user_and_workspace(
     // ensure workspace is associated to user
     user.associate_workspace(&ctx, *workspace.pk()).await?;
 
+    ctx.write_audit_log_on_head(AuditLogKind::Login {}, "Person".to_string())
```
stack72, line 209: I've made this show as a person logged in for now - this opens up so that we can add API, AI Agent etc later

nickgerace (approved):

PR #4932, opened 2024-11-06
Spike: Split up sdf-server

jhelwig:
`lib/axum-util` really should be `sdf-server-common`. It started out as a more generic place to put `axum` related things that were defined in `sdf-server`, but needed by both `sdf-server` and the various `sdf-server-v2-*` crates. It ended up being very specific to `sdf-server`, and the `sdf-server-v2-*` crates, though.


jhelwig:
Did not speed up compilation.

PR #4935, opened 2024-11-06
Final Audit Log UI polishing run

nickgerace:
## Description

This PR fixes the audit log headers and expands their functionality. Now, all headers are populated from live or compile-time data (other than user invitations) to ensure that the application has up-to-date headers.

The second commit moves the filtering and pagination logic into the audit log list function. More information below!

Screenshot 2024-11-06 at 11 21 48 PM

## Additional Information

This PR also introduces a new filter for the entity type. Now, you can filter by entity type and/or the raw event kind. The raw event kind is viewable by hovering over the title in the "Event" section. Change Set and User filtering now works as intended too with a streamlined filtering scheme for the System actor (thank you `#[serde(untagged)]`!).

Other changes include matching the expand button's style to that of the ellipsis on the action card and removing old "ip" header references in the audit log UI code. The max message count in the audit logging list call has been increased from 1000 to 10000 as well, though this ultimately needs to be replaced with real pagination soon.

## Timestamp Refreshing

This PR introduces timestamp refreshing and its first user is the audit logs UI. Every sixty seconds, the timestamp string will refresh. What was "17 minutes ago" will become "18 minutes ago", as an example. In the future, we may want to tie this to a clock for more precision.

## Audit Logging Integration Test Fix

The audit logging integration test now accounts for a busy CI environment and wraps the "source stream has no messages" check into a new test helper function with a timeout and interval pattern.

## Known Issue

With smaller windows, the leftmost filter dropdown can clip off the screen. This is due to anchoring the `DropdownMenu` to the `th` element. This requires a further look, but should not hinder users of medium or larger windows.

Screenshot 2024-11-06 at 10 59 08 PM

## What about the second commit?

In the second commit, we moved the filtering and pagination logic into the audit log list function. The fake pagination logic is still in place, but this moves us a step closer towards real pagination by ensuring all callers go through the same interface.

The second commit also makes the message fetch dynamic based on the number of messages on the screen. Now, users and change sets are cached while filtering audit logs.

## Why not properly paginate?

The hold-up? Reverse timestamp sort. Consuming off the stream provides a linearized view of audit logs. This is great when you know where you want to start and end on a stream, but we do not have that luxury here since we need to know how many logs fit our filtered criteria as well as other relevant information.


app/web/src/api/sdf/dal/audit_log.ts
nickgerace, line 1: These are copied from the backend to ensure that filters are always up to date for the user. If we fetched them from the backend, we would not know when to push a `WsEvent` to update them in the frontend. Thus, this is the best compromise for user experience and performance.

If you scroll down further, you'll see a comment in the backend type that calls out that you need to add the entity type and kind to the frontend too in this specific file.

app/web/src/components/Workspace/WorkspaceAuditLog.vue
```diff
@@ -173,34 +173,33 @@ import AuditLogDrawer from "../AuditLogDrawer.vue";
...
-
 const collapseAllRows = () => {
   rowCollapseState.value = new Array(PAGE_SIZE).fill(false);
 };
+
 const DEFAULT_FILTERS = {
   page: 1,
   pageSize: PAGE_SIZE,
   sortTimestampAscending: false,
-  excludeSystemUser: false,
```
nickgerace, line 194: Now we can handle the system user directly in the `userFilter` array.

lib/dal/src/audit_logging.rs
```diff
@@ -88,6 +88,13 @@ pub(crate) async fn publish_pending(
         None => (TaskTracker::new(), true),
     };
 
+    // TODO(nick): somewhere in this function (or alongside it), we need to deserialize the events and send them up to
+    // the frontend. The frontend can make its own decision to include it into its paginated and filtered view, but we
+    // will make one event that sends an array of published audit logs. For chunking, we can do a naive array list
+    // length check or something else (e.g. size check) and send multiple messages, if need be. The first move might be
+    // to add a broadcast channel were shuttle sends the message paylods in bytestream format. Then, after shuttle
+    // exits, the dal can deserialize those messages.
```
nickgerace, line 96: This is the next bit of work: getting rid of the refresh icon in the dashboard and publishing `WsEvents` for the newly published audit logs.

lib/dal/src/audit_logging.rs
```diff
@@ -238,7 +245,7 @@ pub async fn list(ctx: &DalContext) -> Result<Vec<FrontendAuditLog>> {
         .await?;
 
     // TODO(nick): replace hard-coded max messages value with proper pagination.
-    let mut messages = consumer.fetch().max_messages(1000).messages().await?;
+    let mut messages = consumer.fetch().max_messages(10000).messages().await?;
```
nickgerace, line 248: Again, calling out that this will need to be replaced soon, but I'm bumping it further for now while we test the limits.

lib/sdf-server/src/service/v2/audit_log/list_audit_logs.rs
```diff
@@ -12,16 +12,24 @@ use si_frontend_types as frontend_types;
 use super::AuditLogResult;
 use crate::extract::{AccessBuilder, HandlerContext, PosthogClient, QueryWithVecParams};
 
-#[derive(Deserialize, Serialize, Debug)]
+#[remain::sorted]
+#[derive(Deserialize, Debug)]
+pub enum UserFilter {
+    System,
+    #[serde(untagged)]
+    User(UserPk),
```
nickgerace, line 20: This is how we handle `string[]` with the "System" string or a `ChangeSetId` conversion to an enum

stack72 (approved):

PR #4937, opened 2024-11-07
Mgmt func history

zacharyhamm:
Shows a history of all management function executions for the change set
in the history tab, and allows you to see the latest execution logs via
the component's management function panel.

### DIFF
--- a/app/web/src/components/ChangesPanelHistory.vue
+++ b/app/web/src/components/ChangesPanelHistory.vue
@@ -1,37 +1,75 @@
 <template>
-  <ScrollArea
-    v-if="actionsStore.historyActions.length > 0"
-    ref="mainDivRef"
-    @click="deselectOnClickEmptySpace"
-  >
-    <!-- TODO(Wendy)- SEARCH BAR SHOULD GO HERE -->
-    <ActionsList
-      v-for="[detail, actions] in actionsStore.historyActionsGrouped"
-      :key="detail.changeSetId"
-      :actions="actions"
-      :changeSet="getChangeSet(detail)"
-      :clickAction="clickAction"
-      :selectedFuncRunIds="selectedFuncRunId ? [selectedFuncRunId] : []"
-      kind="history"
-      @history="openHistory"
-    />
+  <ScrollArea ref="mainDivRef" @click="deselectOnClickEmptySpace">
+    <TreeNode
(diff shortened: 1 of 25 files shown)


app/web/src/components/ChangesPanelHistory.vue
```diff
@@ -1,5 +1,9 @@
 <template>
-  <ScrollArea ref="mainDivRef" @click="deselectOnClickEmptySpace">
+  <div
+    v-if="featureFlagsStore.MANAGEMENT_FUNCTIONS"
```
zacharyhamm, line 3: @wendybujalski I'll pull this down to test the flag in a second. But will this hide the *entire* history if you don't have the flag on? We just want to hide the management func history

wendybujalski: Nah, I just separated the two options for flag on/off to make it simpler.

app/web/src/components/ChangesPanelHistory.vue
```diff
@@ -1,69 +1,143 @@
...
     />
-  </ScrollArea>
-  <EmptyStateCard
-    v-else
-    iconName="actions"
-    primaryText="No Actions Have Been Taken"
-    secondaryText="There is no action history to display for this change set."
-  />
+  </div>
+  <div v-else ref="actionDivRef" class="py-2xs w-full h-full">
```
wendybujalski, line 77: @zacharyhamm Look here!

zacharyhamm: ahh! i see! very clever

zacharyhamm: that was hard to parse in the diff but it's clear from the full file.

lib/veritech-core/src/lib.rs
```diff
@@ -222,7 +222,7 @@ impl VeritechRequest {
             VeritechRequest::KillExecution(kill_execution_request) => {
                 &kill_execution_request.execution_id
             }
-            VeritechRequest::Mangement(management_request) => &management_request.execution_id,
+            VeritechRequest::Management(management_request) => &management_request.execution_id,
```
vbustamante, line 225: 💯

jobelenus (approved):

PR #4938, opened 2024-11-07
chore(si-data-spicedb): support TLS-enabled SpiceDB endpoints

fnichol:
This change adds URL-detecting TLS support when connecting to SpiceDB endpoints starting with `https://`. Prior to this change the SpiceDB client would attempt a connection and then be disconnected when TLS negotiation failed. Later API calls would fail with a "transport error" message.

References: lunaetco/spicedb-client#3

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -155,9 +155,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.92"
+version = "1.0.93"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


fnichol:
In the process of debugging this issue and then developing a fix, I left us a `read-schema` example Rust program which can be run with `buck2 run lib/si-data-spicedb/examples/read-schema`. It relies on `SPICEDB_URL` and `SPICEDB_PRESHARED_KEY` as program inputs.

stack72 (approved):

PR #4942, opened 2024-11-08
fix(sdf): Remove ManagePermissionLayer on refresh_workspace_members

stack72:
This API endpoint is used more as a Webhook than it is with managing information

The AuthPortal talks to this endpoint and tells the endpoint to assemble the correct data for the wroksapce. It doesn't push that data to SDF. so if a user who doesn't have access to manage my workspace hits the endpoint, they cannot manipulate the workspace. This means we can work without this check at this time.


fnichol (approved):
Ah, the solutions to our problems are unique sometimes!

PR #4946, opened 2024-11-08
Multiple Views

jobelenus:

### DIFF
--- a/lib/sdf-server/src/service/v2/view.rs
+++ b/lib/sdf-server/src/service/v2/view.rs
@@ -81,7 +81,12 @@ impl IntoResponse for ViewError {
     fn into_response(self) -> Response {
         let (status_code, error_message) = match self {
             ViewError::NameAlreadyInUse(_) => (StatusCode::UNPROCESSABLE_ENTITY, self.to_string()),
-
+            ViewError::DalDiagram(
+                dal::diagram::DiagramError::DeletingLastGeometryForComponent(_, _),
+            )
+            | ViewError::Component(ComponentError::ComponentAlreadyInView(_, _)) => {
+                (StatusCode::FORBIDDEN, self.to_string())
+            }
             _ => (StatusCode::INTERNAL_SERVER_ERROR, self.to_string()),
         };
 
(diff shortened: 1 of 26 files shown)


lib/sdf-server/src/service/v2/view.rs
```diff
@@ -81,7 +81,12 @@ impl IntoResponse for ViewError {
     fn into_response(self) -> Response {
         let (status_code, error_message) = match self {
             ViewError::NameAlreadyInUse(_) => (StatusCode::UNPROCESSABLE_ENTITY, self.to_string()),
-
+            ViewError::DalDiagram(
+                dal::diagram::DiagramError::DeletingLastGeometryForComponent(_, _),
+            )
+            | ViewError::Component(ComponentError::ComponentAlreadyInView(_, _)) => {
+                (StatusCode::FORBIDDEN, self.to_string())
```
zacharyhamm, line 88: Is FORBIDDEN the right status code here? It feels like this is a input error

jobelenus: @zacharyhamm 400? or 422?

jobelenus: Axum uses 422 when you send it the wrong JSON, so i'd rather not overload that one

vbustamante: I went by the wikipedia definition of it:

> [403 Forbidden](https://en.wikipedia.org/wiki/HTTP_403)
The request contained valid data and was understood by the server, but the server is refusing action. This may be due to the user not having the necessary permissions for a resource or needing an account of some sort, or **attempting a prohibited action (e.g. creating a duplicate record where only one is allowed)**. This code is also typically used if the request provided authentication by answering the WWW-Authenticate header field challenge, but the server did not accept that authentication. The request should not be repeated.

vbustamante: If we want to keep 403 only for authorization levels or something like that we could change it though, but it feels like it makes sense on this case where the actions are forbidden by the application logic

jobelenus: 401 is authz

vbustamante (dismissed):

zacharyhamm (dismissed):

zacharyhamm (approved):

wendybujalski (approved):

zacharyhamm (approved):
last time? 🤞

PR #4957, opened 2024-11-12
docs(eks-how-do): Create EKS How-to

stack72:


app/docs/src/how-tos/aws-eks.md
```diff
@@ -0,0 +1,520 @@
...
+NOTE: When creating a nodegroup and/or a fargate profile, AWS requires the
+cluster to be `ACTIVE`. The create actions will fail with `warning` messages.
+You can retry these actions after a period of time.
+
+## Accessing the cluster
+
+By default, the user used to create the cluster will be the user to have access
+to it. It is not recommended to use access keys directly with System Initiative.
+The workspace uses assumeRolePolicy, so specific users will need granded access
+to the cluster.
```
johnrwatson, line 392: & use kubectl to get valid responses from the clusters apiserver

app/docs/src/how-tos/aws-eks.md
```diff
@@ -0,0 +1,520 @@
...
+
+### Generating a KubeConfig
+
+Using the AWS CLI, run the following command:
```
johnrwatson, line 396: There's nothing here to say get a set of credentials into AWS, basically it might be worth saying to the reader to setup some creds which you want to authenticate to the cluster & you can check who the user is with `aws sts get-caller-identity`

johnrwatson: I reckon a lot of users `will not` have a valid set at this stage

app/docs/src/how-tos/aws-eks.md
```diff
@@ -0,0 +1,520 @@
...
+### Update EKS Cluster Access Config
+
+![Update Cluster Config](./aws-eks/update-cluster-config.png)
+
+Update the `demo-eks-cluster` `accessConfig.authenticationMode` to be
+`API_AND_CONFIG_MAP`
+
+Right click on the component and select `Update Cluster Access Config`.
+
+Note: This change needs to take effect in the cluster before you can add access
```
johnrwatson, line 439: Tell them to put the secondary create on hold to flaunt the feature that you can do that, then have them click it on head after the primary applies correctly?

app/docs/src/how-tos/aws-eks.md
```diff
@@ -0,0 +1,520 @@
...
+
+![Create Access Entry](./aws-eks/create-access-entry.png)
+
+Add an `Access Entry` to your `demo-eks-cluster` eks cluster frame.
+
+Set the component name to be `My User`.
+
+Set the `type` to be `STANDARD`.
+
+Set the `principalArn` to be `<enter your userArn>`.
```
johnrwatson, line 452: Add a note here saying that this is an example user/user which you want to grant access to the cluster + give an example of a user arn in the correct form to help them

app/docs/src/how-tos/aws-eks.md
```diff
@@ -0,0 +1,520 @@
...
+
+Connect the `Security Group ID` output socket of
+eks-cluster-sg`component to the`Source Traffic Security Group
+ID`input socket of this`nodegroup-all-protocol-all-ports` component.
+
+### Apply your Change Set
+
+![Apply Change Set](./aws-eks/apply.png)
+
+Press `Escape` or click anywhere on the canvas background to select the
```
johnrwatson, line 373: Another utterance

app/docs/src/how-tos/aws-eks.md
```diff
@@ -0,0 +1,520 @@
...
+Click the `Apply Change Set` button to:
+
+- Create the correct access policies for your user
+
+NOTE: By adding a user to the Cluster access management, you will need to
+regenerate a new kubeconfig.
+
+### Connect to the Cluster
+
+After generating an updated Kubeconfig, you can run the command:
```
johnrwatson, line 480: Not sure you need to get a new kubeconfig, as the allocation of access should be totally within the cluster/aws. you should be able to just run can-i again to check it's now fixed

app/docs/src/how-tos/aws-eks.md
```diff
@@ -0,0 +1,520 @@
...
+
+```
+$ kubectl auth can-i get pods
+yes
+```
+
+Now you can access the details of the pods:
```
johnrwatson, line 487: The first line here isn't rendering + looks a bit confusing. Basically it doesn't show the command to run, you can just copy it out of the terminal

johnrwatson (approved):
Minor comments, see diff

app/docs/src/how-tos/aws-eks.md
stack72: I have linked to the AWS CLI docs and the configuring it

PR #4959, opened 2024-11-12
fix: use change set sharded store for management runs

zacharyhamm:
Some reactivity issues and a bug when fetching history were a consequence of using the FuncRuns pinia store, which is not namespaced to the change set. This moves the management functionality over to its own, change set specific store, which resolves these issues.

Fixes BUG-664


wendybujalski (approved):
Looks good!

wendybujalski (approved):

PR #4961, opened 2024-11-12
Lambda that runs verification scripts to check for errors in billing data

jkeiser:
This runs the workspace_verifications SQL queries in a lambda, so that we can run it periodically and create alerts to let us know when things have gone weird.

(It's been sitting in my tree and needs a-flushin'.)


johnrwatson (approved):
Nice job.

One thing that concerns me is how we are going to adjust the price retrospectively to something higher/other than 0.007 and how that rolls out effectively

jkeiser:
That's actually one of the things addressed by this pr :) it's configurable by environment variable or event property now.

PR #4963, opened 2024-11-12
fix(auth-portal): Set editor label to be collaborator

stack72:
Screenshot 2024-11-12 at 23 53 30

Screenshot 2024-11-12 at 23 56 49


fnichol (approved):
This appears to change the label without affecting the underlying value so that we shouldn't have to update any other services. Nice!

PR #4970, opened 2024-11-14
Make the audit log UI behave as a data table

nickgerace:
## Description

This PR makes the audit log UI behave as a true data table. Now, filters only modify the view of the data that has already been queried. Filters are now populated purely by the data returned and all forms of filter option gathering have been replaced accordingly.

## Primary Changes

As part of these changes, pagination has been temporarily removed in favor of a button to load more entries. The number of entries to be loaded and filters chosen are preserved for each change set as the user moves across the workspace.

The underlying query is based on the current selected change set. If you are not on HEAD, you only see audit logs for the current change set. If you are on HEAD, you see audit logs for HEAD and all change sets that have been applied or abandoned.

In addition to these changes, the refresh button has been removed and now we react to `WsEvents` with a debouncer in place for re-querying the logs.

## What about the backend?

Shuttle now has the ability to publish messages on destination streams with a given suffix subject. We need this to ensure that only individual change sets have their messages filtered. At the HEAD change set level, we will just look at all audit logs in the workspace.

A second change to the backend is to the ephemeral pull consumer in the list call. We need to use a sequence-based delivery policy to get the newest audit logs due to JetStream's LIFO stream semantics.

## QoL changes

Quality of life changes include but are not limited to the following:

- The loader for the initial load has been moved underneath the header
  row and is handled separately from other load page requests
- A message will be displayed when no rows match the filter criteria

## Additional Information

This PR is a continuation of #4965.


britmyerss (approved):

lib/sdf-server/src/extract.rs
```diff
@@ -296,36 +296,6 @@ async fn tenancy_from_claim(
     Ok(Tenancy(dal::Tenancy::new(claim.workspace_pk)))
 }
 
-/// Use instead of [`axum::extract::Query`] when the query contains array params using "[]"
-/// notation.
-///
-/// Inspiration : https://dev.to/pongsakornsemsuwan/rust-axum-extracting-query-param-of-vec-4pdm
-pub struct QueryWithVecParams<T>(pub T);
```
nickgerace, line 303: @fnichol surely we won't want this again someday 😅

PR #4971, opened 2024-11-14
Move otelcol back to platform group in Tiltfile

jhelwig:
While otelcol is used for telemetry, it is also a core service that the applications expect to be available. When otelcol is not started, we get a _large_ amount of error output from `sdf`, `veritech`, etc., as well as from the integration tests when only `buck2 run dev:platform` has been run with no other intervention.

### DIFF
--- a/dev/Tiltfile
+++ b/dev/Tiltfile
@@ -12,6 +12,7 @@ groups = {
     "platform": [
         "db",
         "nats",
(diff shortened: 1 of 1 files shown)


zacharyhamm (approved):
Makes sense to me

johnrwatson (approved):

PR #4974, opened 2024-11-14
feat: management edges

zacharyhamm:
Management edges now render on the diagram, and can be drawn from one component to another, if that component is capable of managing the other. They can also be deleted.

### DIFF
--- a/app/web/src/store/components.store.ts
+++ b/app/web/src/store/components.store.ts
@@ -188,11 +188,22 @@ export const getAssetIcon = (name: string) => {
 };
 
 const edgeFromRawEdge =
-  (isInferred: boolean) =>
+  ({
+    isInferred,
+    isManagement,
+  }: {
+    isInferred?: boolean;
+    isManagement?: boolean;
+  }) =>
   (e: RawEdge): Edge => {
     const edge = structuredClone(e) as Edge;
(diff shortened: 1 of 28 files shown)


app/web/src/store/components.store.ts
```diff
@@ -188,11 +188,22 @@ export const getAssetIcon = (name: string) => {
 };
 
 const edgeFromRawEdge =
-  (isInferred: boolean) =>
+  ({
```
jobelenus, line 191: Nice 👍

lib/dal/src/management/mod.rs
```diff
@@ -485,7 +493,11 @@ impl<'a> ManagementOperator<'a> {
             auto_geometry
         };
 
-        Ok((component.id(), geometry))
+        Ok(CreatedComponent {
+            component,
+            geometry,
```
jobelenus, line 498: Doesn't this need a viewId? (i.e. all geometries are related to a view)

jobelenus (changes requested):

lib/dal/src/management/mod.rs
zacharyhamm: This is the relative geometry within the current view, which will get transformed into the view geo when we call `set_geometry`. Everything is happening within the view passed in (right now that is only the default view)

jobelenus (dismissed):

jobelenus (approved):

PR #4975, opened 2024-11-14
feat: view nodes on diagram api

vbustamante:


zacharyhamm (commented):
Diagram changes look fine to me! Happy to do a full review when this is ready

lib/dal/src/workspace_snapshot/node_weight/geometry_node_weight/v1.rs
```diff
@@ -39,7 +47,196 @@ impl GeometryNodeWeightV1 {
     }
 }
 
-impl CorrectTransforms for GeometryNodeWeightV1 {}
+// IF this geometry is for a DO and its on a view that already has a geometry for this DO
+// replace the original geometries content for the content of the new one, and erase other updates for self
+
+// replace new node update with a replaceNode update with the new content
+impl CorrectTransforms for GeometryNodeWeightV1 {
+    fn correct_transforms(
```
vbustamante, line 55: New correct transforms (@jhelwig I'd appreciate a review on this one)

lib/dal/src/workspace_snapshot/node_weight/geometry_node_weight/v1.rs
```diff
@@ -39,7 +47,196 @@ impl GeometryNodeWeightV1 {
...
+            // Modify the update by replacing the old duplicate geometry with the new one,
+            // instead of creating the new one
+            match updates.get_mut(self_creation_idx) {
+                Some(update @ Update::NewEdge { .. }) => {
+                    let _old_update = mem::replace(update, replaced_update);
+                }
+                // If there, somehow, isn't an entry at that idx, or it is not a
+                // Update::NewEdge, then something has gone horribly wrong, as we got the
+                // idx by iterating over the updates, and looking at what kind of thing was
+                // there.
```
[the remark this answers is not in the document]
vbustamante: yes it should, fixed it locally, will push

jhelwig (dismissed):
Assuming the `/try` succeeds, 👍🏻 .

jhelwig (dismissed):

lib/dal/src/diagram/diagram_object.rs
```diff
@@ -0,0 +1,106 @@
...
+    implement_add_edge_to,
+    workspace_snapshot::node_weight::{
+        category_node_weight::CategoryNodeKind,
+        diagram_object_node_weight::{DiagramObjectKind, DiagramObjectNodeWeight},
+        traits::SiVersionedNodeWeight,
+        NodeWeight,
+    },
+    DalContext, EdgeWeightKind, EdgeWeightKindDiscriminants,
+};
+use jwt_simple::prelude::{Deserialize, Serialize};
```
zacharyhamm, line 12: This is probably not the right place to be importing Deserialize, Serialize (should come from serde)

lib/dal/src/diagram/diagram_object.rs
```diff
@@ -0,0 +1,106 @@
...
+
+    fn assemble(node_weight: DiagramObjectNodeWeight) -> Self {
+        Self {
+            id: node_weight.id(),
+            object_kind: node_weight.object_kind(),
+        }
+    }
+
+    pub async fn new_for_view(ctx: &DalContext, view_id: ViewId) -> DiagramResult<Self> {
+        println!("new for view");
```
zacharyhamm, line 51: Stray debugging

lib/dal/src/diagram/geometry.rs
```diff
@@ -54,6 +61,14 @@ impl Geometry {
         result: DiagramResult,
     );
 
+    implement_add_edge_to!(
+        source_id: GeometryId,
+        destination_id: Ulid,
```
zacharyhamm, line 66: Do diagram objects have their own id type? If so we should use that instead of ulid

vbustamante:
> image
> Create some views in one change set. Select one of those view besides default. Create a new change set, and you'll be greeted with this error

this is an interesting one. When we redirect to another changeset we're not being smart about going to the default view if the selected view is present 

It seems to be frontend only, I'll fix it

vbustamante:
@zacharyhamm (and anyone interested in testing the special case we're dealing with):

On the main branch, create some views and apply to head

Pull this branch

create two changesets, A and B

on both changesets, create a view node. It has to be the same view node inside of the same view on both changesets, since we need to make sure we're not creating duplicate geometries. Put the view nodes in different positions in each changeset

apply changeset A. changeset B should work as normal, should have a single view node, and the view node should have taken the position of the view node from changeset A

lib/dal/src/diagram/geometry.rs
vbustamante: They don't

lib/dal/src/diagram/diagram_object.rs
vbustamante: good call, fixed

lib/dal/src/diagram/diagram_object.rs
vbustamante: fixed

zacharyhamm (approved):

zacharyhamm (approved):

PR #4981, opened 2024-11-15
Added message to empty secrets panel

keeb:


app/web/src/components/SecretsPanel.vue
```diff
@@ -99,7 +99,16 @@
...
+      <p>No secret definitions found.</p>
+      <br />
+      <p>
+        To configure a secret, drag the corresponding asset onto the canvas.
+      </p>
+      <br />
+      <p class="text-sm">
+        Example: Drag the AWS Credential to the canvas, and the AWS Credential
+        Secret will show up here for configuration.
+      </p>
```
nickgerace, line 111: I think this is great! If @wendybujalski is cool with the usages of `<p>` and `<br>` then this looks good to me.

nickgerace (approved):

app/web/src/components/SecretsPanel.vue
nickgerace: I ended up approving this because we can polish this if desired and it looks good overall.

PR #4985, opened 2024-11-18
Get the change set for audit log published events

nickgerace:
This PR ensures that we get the change set when preparing audit log publish events. We do not want to rely on the reference as it will only be stored when updating the visibility.


johnrwatson (approved):

nickgerace:
Transient error...

Screenshot 2024-11-18 at 12 03 38 PM

PR #4986, opened 2024-11-18
fix(sdf): Ensure we reload secrets when dynamically installing component

stack72:
This is a bit of a hammer but there's *currently* no way to understand that a secret has been installed as being a secret is defined internally to the module schema

This will alleviate the pain for now and will allow us to turn this feature on for users


zacharyhamm (approved):
👍 I didn't realize we had that WsEvent already

PR #4988, opened 2024-11-18
Ensure API requests with ForceChangeset always flip `creatingChangeset` back off on failure

jkeiser:
This creates a new `changeSetsStore.nonHeadApiRequest()` function for API calls that force a new changeset to be created. It will flip `creatingChangeSet` on, and then flip it back off if the API fails. There are a number of instances where we flip it on, but don't flip it back off.

This should not change behavior in any instances where API requests succeed; boilerplate reduction here is entirely rote replacement (or should be--another set of eyes is welcome)

### DIFF
--- a/app/web/README.md
+++ b/app/web/README.md
@@ -41,7 +41,7 @@ This template should help get you started developing with Vue 3 and Typescript i
 
 ## IDE Setup Instructions
 ### [VSCode](https://code.visualstudio.com/) (preferred)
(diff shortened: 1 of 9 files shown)


jobelenus (approved):

stack72:
@jkeiser is this still something we want to push forward with?

jkeiser:
I will close for now and get back to it next time I'm working on something that forces a changeset. Thanks for following up!

PR #5001, opened 2024-11-19
Write to audit logs table from forklift

nickgerace:
## Description

This PR ensures that we skip publishing ws events when audit logs are published for ephemeral change sets.

### DIFF
--- a/lib/dal/src/audit_logging.rs
+++ b/lib/dal/src/audit_logging.rs
@@ -131,13 +131,37 @@ pub(crate) async fn publish_pending(
                 error!(?err, "audit logs shuttle error");
             }
 
-            match WsEvent::audit_logs_published(&ctx_clone_for_ws_event).await {
-                Ok(event) => {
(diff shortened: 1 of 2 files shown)


lib/dal/src/audit_logging.rs
```diff
@@ -131,13 +131,37 @@ pub(crate) async fn publish_pending(
...
+                                error!(?err, "error when publishing ws event for audit logs");
+                            }
+                        }
+                        Err(err) => error!(?err, "error when creating ws event for audit logs"),
                     }
                 }
-                Err(err) => error!(?err, "error when creating ws event for audit logs"),
+                Ok(None) => {
+                    trace!("skipping ws event creation for audit logs: no change set found")
+                }
```
nickgerace, line 160: This is the case we are hitting. I am fairly confident we are running into the ephemeral change set here, but will need to follow up once we switch to PG-backed audit logs for the UI.

stack72 (approved):

PR #5002, opened 2024-11-20
Write audit logs to database in forklift

nickgerace:
## Description

This PR adds the ability for forklift to write audit logs to the new audit database. By default, this functionality is disabled and can be enabled with the `--enable-audit-logs-app` flag or configuration file option for forklift.

## Changes to audit logs types

- Move all postgres-related logic to new "database" module, splitting up all functionality into sub-modules
- Refactor insert method to ensure insert query works
- Move all functionality out of the root library file
- Move metadata struct from the frontend types crate to the events crate and ensure that it is versioned because we need it for database insertions
- Make entity type optional for the metadata rather than using
  whitespaces or empty strings

## Changes to forklift

- Add optional inner audit logs app based on the new config option
- Move all apps to a new app module, including the handlers and app states
- Ensure app setup flows through a single error type which then owns app setup error types for each app
- Ensure audit logs app durable consumer uses an ack layer, but relies  on the limits-based queue semantics to keep the messages on it
- Add postgres cert for local development, including development environment detection
- Add flag and config option to enable audit logs app
- Add audit config option for the corresponding postgres config

## Misc changes

- Ensure the existing audit logging list function works from other changes in this commit
- Remove unused pub export for development environment detection for sdf

## Considerations

- Empty metadata for audit logs will be written as "JSON null" and not "PG null" for the database despite the column being nullable (this is okay, just so long as we know!)
- We have faith in naxum concurrency limits and `PgPool` configuration enough such that we can singularly insert audit logs into the database one at a time (this is subject to change)


fnichol (approved):
Looking good and ready to deploy (and be disabled by default to boot!)

PR #5006, opened 2024-11-20
fix(layer-cache): add size hinting when inserting Rust types into caches

fnichol:
This change addresses an issue when inserting a Rust type into a Foyer cache, where the memory used by that rust variable was not being accurately calculated. Prior to this change, the `size_of_val()` would dereference one level of "smart pointers" (i.e. deref an `Arc`, or `Vec`, or `String`), but no more levels. This led to an undercount of memory for that Rust value. For example, perhaps a workspace snapshot that was 30mb in size would be incorrectly calculated as 5mb.

As determining the size of an arbitrary Rust object is surprisingly difficult, we opted to use the *uncompressed* size of serialized Postcard bytes as our "size hint" (note that Postcard adds very little extra information into the stream of bytes, so could be considered as a very close approximation) and to provide this size hint to Foyer's `with_weighter()` configuration. In our caches, when we have a `MaybeDeserialized::RawBytes` variant, the size of its Bytes is the size of that value (and so can be used directly).

### DIFF
--- a/lib/si-layer-cache/src/hybrid_cache.rs
+++ b/lib/si-layer-cache/src/hybrid_cache.rs
@@ -37,7 +37,7 @@ where
     V: Serialize + Clone + Send + Sync + 'static,
 {
     RawBytes(Vec<u8>),
-    DeserializedValue(V),
+    DeserializedValue { value: V, size_hint: usize },
 }
 
 #[derive(Clone, Debug)]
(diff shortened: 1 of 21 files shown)


lib/si-layer-cache/src/hybrid_cache.rs
```diff
@@ -107,7 +107,12 @@ where
         let cache: HybridCache<Arc<str>, MaybeDeserialized<V>> = HybridCacheBuilder::new()
             .with_name(&config.name)
             .memory(memory_cache_capacity_bytes)
-            .with_weighter(|_key: &Arc<str>, value: &MaybeDeserialized<V>| size_of_val(value))
+            .with_weighter(
+                |_key: &Arc<str>, value: &MaybeDeserialized<V>| match value {
+                    MaybeDeserialized::RawBytes(bytes) => bytes.len(),
+                    MaybeDeserialized::DeserializedValue { size_hint, .. } => *size_hint,
+                },
+            )
```
zacharyhamm, line 115: Any idea what the difference ends up being between the uncompressed postcard and the actual rust types? Feels like they can't be identical.

I wonder if a future solution to get it totally right is a SizeOf trait of some kind, where we can implement a function on every type that will report its size reliably

zacharyhamm (approved):
Amazing catch here! Memory management in rust is managed so automatically that it's easy to forget about the corner cases like this. My testing didn't go beyond rebooting a running system on this branch and ensuring it still worked. We'll learn more in the deployment

PR #5007, opened 2024-11-20
f[WIP] eat(sdf): Protect Admin Routes behind a new Administer role

stack72:
We have an admin page controlled by a feature flag. We locked the API endpoints down to be those with systeminit.com emails, but we want to take advantage of spicedb to make this work correctly

Even if a user gets to the WorkspaceAdmin page, we have locked down all of the API endpoints for them based on a new role in spicedb

This role will be manually granted and can be done so using the following Zed cli command:

```
zed relationship create system:system admin administer:<user_pk>
```

This will be granted by TechOps on the SystemInitiative users that need access to it so that it can be tracked correctly

### DIFF
--- a/lib/permissions/src/lib.rs
+++ b/lib/permissions/src/lib.rs
@@ -26,6 +26,7 @@ type Result<T> = result::Result<T, Error>;
 pub enum ObjectType {
     User,
     Workspace,
+    System,
 }
 
 #[derive(Clone, Copy, strum::Display)]
@@ -33,6 +34,7 @@ pub enum ObjectType {
 pub enum Permission {
     Approve,
     Manage,
+    Administer,
 }
 
 #[derive(Clone, Copy, strum::Display, Debug)]
@@ -40,6 +42,7 @@ pub enum Permission {
 pub enum Relation {
     Approver,
     Owner,
+    Admin,
 }
 
 /// RelationBuilder allows defining a relationship in SpiceDb.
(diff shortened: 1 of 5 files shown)


lib/permissions/src/lib.rs
```diff
@@ -26,20 +26,23 @@ type Result<T> = result::Result<T, Error>;
...
     Manage,
+    Administer,
 }
 
 #[derive(Clone, Copy, strum::Display, Debug)]
 #[strum(serialize_all = "snake_case")]
 pub enum Relation {
     Approver,
     Owner,
+    Admin,
```
britmyerss, line 45: You know - I actually wonder if `owner` can also just work here. Only it'd be `owner` of the system vs the workspace

lib/permissions/src/lib.rs
```diff
@@ -206,6 +209,10 @@ impl PermissionBuilder {
         self.object(ObjectType::Workspace, id)
     }
 
+    pub fn system_object(self) -> Self {
+        self.object(ObjectType::System, "system")
```
sprutton1, line 213: Do we ever want different permission sets for different systems? Maybe this should be `"sdf"`

stack72:
ok after talking this over with @sprutton1 and @britmyerss we are going to do the following:

* For local installations of SI, we are going to use a wildcard that anyone on a local instance against a LOCAL spicedb will be an admin
* For production system, we are going to add the work to the admin page to be able to make someone an admin
    * this is going to require us to manually seed a user in production and then that user (more than likely Scott / John) will then invite the SI team to be admins where appropriate
    * We will get an audit log event when a user is added / removed from the spicedb definition

PR #5013, opened 2024-11-21
fix(forklift): ensure that config file defaults work

fnichol:
This change also adds a debug level log event showing the computed config.

### DIFF
--- a/lib/forklift-server/src/config.rs
+++ b/lib/forklift-server/src/config.rs
@@ -53,7 +53,7 @@ pub struct Config {
     #[builder(default = "default_data_warehouse_stream_name()")]
     data_warehouse_stream_name: Option<String>,
 
-    #[builder(default)]
+    #[builder(default = "default_enable_audit_logs_app()")]
     enable_audit_logs_app: bool,
 
     #[builder(default)]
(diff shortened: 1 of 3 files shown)


lib/forklift-server/src/config.rs
```diff
@@ -192,6 +196,7 @@ fn buck2_development(config: &mut ConfigFile) -> Result<()> {
 
     config.audit.pg.certificate_path = Some(postgres_cert.clone().try_into()?);
     config.audit.pg.dbname = audit_logs::database::DBNAME.to_string();
+    config.enable_audit_logs_app = true;
```
nickgerace, line 199: Worth calling out: this will enable audit logs being written to the db for local dev.

lib/forklift-server/src/config.rs
```diff
@@ -209,6 +214,7 @@ fn cargo_development(dir: String, config: &mut ConfigFile) -> Result<()> {
...
```
nickgerace, line 217: Here too

nickgerace (dismissed):

nickgerace (approved):

lib/forklift-server/src/config.rs
nickgerace: No more!

lib/forklift-server/src/config.rs
nickgerace: No more!

PR #5017, opened 2024-11-22
fix: incorrectly handle io error with fail to load credential profile

johnrwatson:
Handle IO errors when loading from profile in module index.


fnichol (approved):
Yas!

nickgerace (approved):

PR #5018, opened 2024-11-22
feat: add resourcePayloadToValue by default to all variants

sprutton1:
This ensure that `si:resourcePayloadToValue` is always added to new variants. This is a stub approach in favor of the more complicated effort of making this an Intrinsic function. That work as been started [here](https://github.com/systeminit/si/tree/resource_to_payload). 

Intrinsics are not installed in existing workspaces, so there are edges around managing the existing javascript function and the new intrinsic. That work is better pushed off until after the upcoming holiday.


zacharyhamm (dismissed):
🚀

zacharyhamm (approved):

PR #5019, opened 2024-11-22
feat: management functions understand views

zacharyhamm:
Management funcs are now run in the context of a view. They can only create components in that view. But they can manage components in other views. The position of created components is relative to the manager component, but the position specified in updates is absolute.

If a manager component is in more than one view, a the "play" button will open a dropdown allowing you to select the view to run the function in.

This also makes position (x,y) optional, allowing you to update just the size of a component.

### DIFF
--- a/lib/cyclone-client/src/client.rs
+++ b/lib/cyclone-client/src/client.rs
@@ -1325,6 +1325,7 @@ mod tests {
         let req = ManagementRequest {
             execution_id: "1234".to_string(),
             handler: "manage".to_string(),
+            current_view: "DEFAULT".to_string(),
             this_component: ComponentViewWithGeometry {
                 kind: None,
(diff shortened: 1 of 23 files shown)


lib/cyclone-client/src/client.rs
```diff
@@ -1325,6 +1325,7 @@ mod tests {
         let req = ManagementRequest {
             execution_id: "1234".to_string(),
             handler: "manage".to_string(),
+            current_view: "DEFAULT".to_string(),
```
jobelenus, line 1328: The "default view" is named DEFAULT, but that is editable. It might be better to find the default view and send its name??

jobelenus (approved):
There might be one problem I commented on... but I also might not be understanding it perfectly.

lib/cyclone-client/src/client.rs
zacharyhamm: This is just in the test harness. We need something there (anything will do)

PR #5027, opened 2024-11-26
Add SDF endpoints to update AI gen prompts on the server

jkeiser:
This allows prompts to updated (stored in SQL) on the server so that they can be iterated upon. UI forthcoming.

### DIFF
--- a/lib/asset-sprayer/src/lib.rs
+++ b/lib/asset-sprayer/src/lib.rs
@@ -25,16 +25,16 @@
     while_true
 )]
 
-use std::path::PathBuf;
+use std::{borrow::Cow, path::PathBuf};
 
 use async_openai::{config::OpenAIConfig, types::CreateChatCompletionRequest};
 use config::AssetSprayerConfig;
-use prompt::Prompt;
 use telemetry::prelude::*;
 use thiserror::Error;
 
 pub mod config;
 pub mod prompt;
+pub use prompt::Prompt;
 
 #[remain::sorted]
 #[derive(Debug, Error)]
@@ -77,7 +77,7 @@ impl AssetSprayer {
     }
 
     pub async fn prompt(&self, prompt: &Prompt) -> Result<CreateChatCompletionRequest> {
-        prompt.prompt(&self.prompts_dir).await
+        prompt.prompt(self).await
     }
 
     pub async fn run(&self, prompt: &Prompt) -> Result<String> {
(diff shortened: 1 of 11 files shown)


lib/asset-sprayer/src/lib.rs
```diff
@@ -25,16 +25,16 @@
     while_true
 )]
 
-use std::path::PathBuf;
+use std::{borrow::Cow, path::PathBuf};
```
nickgerace, line 28: Wow! Haven't seen `Cow` in a long term. Might've even been perl code 😄

lib/asset-sprayer/src/lib.rs
```diff
@@ -95,6 +95,33 @@ impl AssetSprayer {
...
+        &self,
+        prompt: &(impl RawPromptYamlSource + std::fmt::Display),
+    ) -> Result<CreateChatCompletionRequest> {
+        Ok(serde_yaml::from_str(&self.raw_prompt_yaml(prompt).await?)?)
+    }
+
+    pub async fn raw_prompt_yaml(
+        &self,
+        prompt: &(impl RawPromptYamlSource + std::fmt::Display),
+    ) -> Result<Cow<'static, str>> {
```
nickgerace, line 109: I suspect that could be a `String`, but if you want clone-on-write semantics, this might be the right choice!

lib/dal/src/migrations/U3402__prompt_overrides.sql
nickgerace, line 1: Looks good!

lib/dal/src/prompt_override.rs
```diff
@@ -0,0 +1,127 @@
...
+    ) -> WsEventResult<Self> {
+        WsEvent::new(
+            ctx,
+            WsPayload::PromptUpdated(PromptUpdatedPayload { kind, overridden }),
+        )
+        .await
+    }
+}
+
+pub struct PromptOverride;
```
nickgerace, line 41: Scoping things by a unit struct is a great idea in this instance because it appears that we do not want to return the entire row, but rather the prompt override itself, which is "just" a string. Is that right?

lib/dal/tests/integration_test/prompt_overrides.rs
```diff
@@ -0,0 +1,72 @@
+use std::collections::HashSet;
+
+use dal::prompt_override::PromptOverride;
+use dal::DalContext;
+use dal_test::{color_eyre::Result, test};
+use pretty_assertions_sorted::assert_eq;
+
+#[test]
+async fn update_prompt(ctx: &mut DalContext) -> Result<()> {
```
nickgerace, line 9: Very nice!

nickgerace (approved):

lib/dal/src/prompt_override.rs
jkeiser: Yep. It felt better having a unit struct to scope in this instance for `PromptOverride::get` rather than `prompt_override::get` or `prompt_override::get_prompt_override`.

lib/asset-sprayer/src/lib.rs
jkeiser: In production mode, these strings are always embedded in the executable, so there's no point copying them--the reason for Cow is mainly because of dev mode.

PR #5033, opened 2024-11-27
fix(web): wait for all components to be loaded into the components.store before hydrating each view's geometry mapping lazily

britmyerss:
<div><br/>via <a href="https://giphy.com/pusheen/">Pusheen</a> on <a href="https://giphy.com/gifs/lazy-BBkF0eR41FsoU">GIPHY</a></div>

### DIFF
--- a/app/web/src/store/views.store.ts
+++ b/app/web/src/store/views.store.ts
@@ -804,8 +804,6 @@ export const useViewsStore = (forceChangeSetId?: ChangeSetId) => {
                 views,
               });
               this.selectView(response.view.id);
-
-              componentsStore.FETCH_ALL_COMPONENTS();
             },
           });
         },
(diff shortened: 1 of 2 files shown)


app/web/src/store/views.store.ts
```diff
@@ -1846,6 +1847,7 @@ export const useViewsStore = (forceChangeSetId?: ChangeSetId) => {
                 // then refetch (i.e. there might be updates!)
                 if (data.toRebaseChangeSetId === changeSetId) {
                   this.FETCH_VIEW();
+                  componentsStore.FETCH_ALL_COMPONENTS();
                   this.LIST_VIEWS();
```
vbustamante, line 1851: It feels like this could cause the same race that was this PR fixes, so it makes sense to make this func async and await these calls too

britmyerss: Yeah I had a similar thought too - good shout.

PR #5035, opened 2024-11-27
feat: add a NATS "dead letter queue" stream for failing messages

fnichol:
This change introduces a new NATS Jetstream stream called `DEAD_LETTER_QUEUES` which will contain metadata of messages that reached their consumer's `max_deliver` limit. Currently, only the `forklift-server` consumer for the `AUDIT_LOGS` stream is configured, but more streams and consumers could be configured to use this in the future.

The forklift/audit logs NATS consumer is configured to attempt `4` deliveries of each message with a linear backoff (i.e. when a message is `nack`d) of 5 seconds, then 10, then 15 for a total of 30 seconds before the message metadata is added to the "dead letter queue" stream. Note that a "failed" message is *not* deleted from its source stream, but rather skipped over by that consumer. This allows us to inspect each failed message in place in the stream and decide how to triage any remediations.


lib/forklift-server/src/server/app/audit_logs.rs
```diff
@@ -74,6 +80,12 @@ pub(crate) async fn build_and_run(
             .create_consumer(async_nats::jetstream::consumer::pull::Config {
                 durable_name: Some(durable_consumer_name),
                 filter_subject: consumer_subject.into_string(),
+                max_deliver: 4,
+                backoff: vec![
+                    Duration::from_secs(5),
+                    Duration::from_secs(10),
+                    Duration::from_secs(15),
+                ],
```
nickgerace, line 88: Excellent!

nickgerace (approved):

PR #5036, opened 2024-11-27
Add an API test for basic authoring scenario

britmyerss:
The `doubler` is now enshrined in API tests! 

This test does the following: 
- create the doubler
- Add an attribute function (that doubles a prop)
- Add a qualification
- Create a component
- Set a prop
- See that it doubles as expected
- Add another prop to the schema variant
- Add another attribute function
- See that the component auto-upgrades, and the new attribute func runs!

A bit of pain here in feature flags and adding a new test, but we'll get it figured out! 

<div><br/>via <a href="https://giphy.com/latenightseth/">Late Night with Seth Meyers</a> on <a href="https://giphy.com/gifs/latenightseth-seth-meyers-lnsm-q5KSpsbhXp3lw7HxXl">GIPHY</a></div>

### DIFF
--- a/bin/si-api-test/tests/7-emulate_authoring.ts
+++ b/bin/si-api-test/tests/7-emulate_authoring.ts
@@ -1,0 +1,462 @@
+// deno-lint-ignore-file no-explicit-any
+import assert from "node:assert";
+import { SdfApiClient } from "../sdf_api_client.ts";
+import {
+    runWithTemporaryChangeset,
+    sleep,
+    sleepBetween,
+    createComponent,
+    createAsset,
+    updateAssetCode,
+    nilId,
+    getQualificationSummary,
+    getPropertyEditor,
+    attributeValueIdForPropPath,
+    setAttributeValue,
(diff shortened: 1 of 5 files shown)


bin/si-api-test/tests/7-emulate_authoring.ts
```diff
@@ -0,0 +1,462 @@
...
+    nilId,
+    getQualificationSummary,
+    getPropertyEditor,
+    attributeValueIdForPropPath,
+    setAttributeValue,
+} from "../test_helpers.ts";
+import { ulid } from "https://deno.land/x/ulid@v0.3.0/mod.ts";
+
+export default async function emulate_authoring(sdfApiClient: SdfApiClient) {
+    await sleepBetween(0, 750);
```
johnrwatson, line 20: nitpick: personally I'd remove this line, it just slows down the tests.

If we are looking to stagger the tests it should be in the harness above, not the test itself.

johnrwatson (approved):

PR #5042, opened 2024-11-29
chore(sdf): always migrate `si_audit` database

fnichol:
This change removes the "attempt to migrate and log errors" behavior when migrating the `si_audit` database. With this change a migration is performed, identical to migrations for the `si` and `si_layer_cache` databases. When an error is encountered, it early-returns with the error.


johnrwatson (approved):
I really liked this way of shipping safely

fnichol:
We're good to revert this try-it-and-log-it behvior as we've confirmed that we're migrating successfully in production

PR #5054, opened 2024-12-03
Fix: Change component type on views other than the default

vbustamante:
Issue 1:
The set type endpoint sent the `ComponentUpdated` event with the geometry for the default views, instead of no geometry, so set_type outside of default always failed. Leading to 
Issue 2:
The nodes would disappear on change type, since we store components and groups on separate arrays in both the components and views stores. On type changes we need to move the data to the correct array.

### DIFF
--- a/app/web/src/store/components.store.ts
+++ b/app/web/src/store/components.store.ts
@@ -43,6 +43,7 @@ import { useAssetStore } from "./asset.store";
 import { useRealtimeStore } from "./realtime/realtime.store";
 import { useWorkspacesStore } from "./workspaces.store";
 import { useFeatureFlagsStore } from "./feature_flags.store";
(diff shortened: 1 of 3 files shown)


johnrwatson (approved):

jobelenus (changes requested):
I do not want to have component store reliant on view store. Only view store uses component store. Since its the same data, I believe we can do it all in the view store. I'll take a whack at it.

vbustamante:
@jobelenus The real fix here I believe would be to not separate groups and components in separate arrays, but this was a quick fix. And since useViewStore is only called inside the event handler, it won't cause any breaking dependency loops. It could also be dealt with in the views store event handler, just moving most of the code there, I believe

Anyhow, feel free to take a stab at it

PR #5056, opened 2024-12-03
Unify view logic in one place

jobelenus:

### DIFF
--- a/app/web/src/store/views.store.ts
+++ b/app/web/src/store/views.store.ts
@@ -1622,28 +1622,38 @@ export const useViewsStore = (forceChangeSetId?: ChangeSetId) => {
                 // don't update
                 if (metadata.change_set_id !== changeSetId) return;
                 const { viewId, geometry } = { ...data.component.viewData };
-                if (!viewId || !geometry)
-                  // this is expected in many situations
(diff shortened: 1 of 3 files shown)


app/web/src/store/views.store.ts
```diff
@@ -1622,28 +1622,38 @@ export const useViewsStore = (forceChangeSetId?: ChangeSetId) => {
...
-                  view.groups[data.component.id] = geometry as IRect;
-                }
+                // changed component type means book-keeping for all views
+                // PSA: currently SDF does not change any geometry when you change component types
+                // if it starts to change geometry, it must return new geometry for every view
+                // the structure doesn't currently support that
+                Object.values(this.viewsById).forEach((view) => {
+                  const groupGeo = view.groups[data.component.id];
+                  const componentGeo = view.components[data.component.id];
+                  const _geo = groupGeo ?? componentGeo;
```
vbustamante, line 1633: _geo should only be used for unused variables, I don't know if the linter will complain but it should 😁

jobelenus: That isn't the JS convention.

vbustamante (approved):

PR #5057, opened 2024-12-03
chore(deps): Updating Rust Deps

stack72:
Required so we can upgrade our version of rustls for a vulnerability

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -172,7 +172,7 @@ remain = "0.2.13"
 reqwest = { version = "0.12.2", default-features = false, features = ["rustls-tls", "json", "multipart"] }
 ring = "=0.17.5" # Upgrading this is possible, but a pain, so we don't want to pick up every new minor version (see: https://github.com/facebook/buck2/commit/91af40b66960d003067c3d241595fb53d1e636c8)
 rust-s3 = { version = "0.34.0-rc4", default-features = false, features = ["tokio-rustls-tls"] }
-rustls = { version = "0.23.18", default-features = false}
+rustls = { version = "0.23.18", default-features = false } # NOTE(nick,fletcher): rustls switched to "aws-lc-rs" as its default crypto provider, but we want ring (i.e. we disable the default feature for "aws-lc-rs")
 rustls-native-certs = "0.7.0"
 rustls-pemfile = { version = "2.1.1" }
 sea-orm = { version = "1.1.0", features = ["sqlx-postgres", "runtime-tokio-rustls", "macros", "with-chrono", "debug-print"] }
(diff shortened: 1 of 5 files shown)


nickgerace (dismissed):

third-party/rust/BUCK
```diff
@@ -15495,57 +15525,23 @@ cargo.rust_library(
...
     crate = "rustls",
-    crate_root = "rustls-0.22.4.crate/src/lib.rs",
-    edition = "2021",
-    features = [
-        "default",
-        "log",
-        "logging",
-        "ring",
-        "tls12",
-    ],
```
fnichol, line 15522: I'm reasonbly sure we don't want `default-features = false` as `rustls` won't have feature support anymore for `tls12`/`ring`/etc

Cargo.toml
```diff
@@ -172,7 +172,7 @@ remain = "0.2.13"
 reqwest = { version = "0.12.2", default-features = false, features = ["rustls-tls", "json", "multipart"] }
 ring = "=0.17.5" # Upgrading this is possible, but a pain, so we don't want to pick up every new minor version (see: https://github.com/facebook/buck2/commit/91af40b66960d003067c3d241595fb53d1e636c8)
 rust-s3 = { version = "0.34.0-rc4", default-features = false, features = ["tokio-rustls-tls"] }
-rustls = { version = "0.22.3" }
+rustls = { version = "0.23.18", default-features = false}
```
fnichol, line 175: Was there a reason to exclude the default features? I think that's not going to work for the rest of the codebase

fnichol: For example, Buck2's [Cargo.toml](https://github.com/facebook/buck2/blob/a5dc441ea44e2eb3198b8d604fb7b2e65f6e3a1b/Cargo.toml#L251C1-L251C18) doesn't exclude the default features

fnichol (commented):
This may work, but I'm not 100% sure that everything works as before (it's the `default-features = false` change that's giving me this sense). I guess if a `/try` passes then we know we're probably fine though

johnrwatson:
we should add the vulnerabillity-fix label if this is what that's intended for

Cargo.toml
nickgerace: We discussed offline, but the short version is that `aws-lc-sys` (required by `aws-lc-rs`) has C/C++ code to compile and we want to use `ring`.

nickgerace:
We added a second commit to provide context to the change.

And @johnrwatson just did it!

Cargo.toml
fnichol: Perfect, I understand now--we wanted to avoid the `aws-lc-rs` dep!

fnichol (dismissed):
Nice, I understand now!

fnichol (approved):

PR #5064, opened 2024-12-04
feat: migrate local module cache on sdf boot, in dev mode

zacharyhamm:
Use the same technique for "development run" detection used to set up the signing keys to set a dev_mode flag on the SDF config. If this dev_mode flag is true, and we are running migrations, also update the local module cache. This ensures we have an asset list locally on SDF boot.

### DIFF
--- a/lib/sdf-server/src/config.rs
+++ b/lib/sdf-server/src/config.rs
@@ -146,6 +146,9 @@ pub struct Config {
 
     #[builder(default)]
     audit: AuditDatabaseConfig,
+
+    #[builder(default)]
+    dev_mode: bool,
 }
 
 impl StandardConfig for Config {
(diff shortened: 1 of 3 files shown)


stack72 (approved):
this is awesome!

lib/sdf-server/src/config.rs
```diff
@@ -513,6 +524,7 @@ fn buck2_development(config: &mut ConfigFile) -> Result<()> {
     config.spicedb.enabled = true;
     config.audit.pg.certificate_path = Some(postgres_cert.clone().try_into()?);
     config.audit.pg.dbname = audit_logs::database::DBNAME.to_string();
+    config.dev_mode = true;
```
nickgerace, line 527: Yep, this is the right approach: adding it to the cargo and buck2 sections!

nickgerace (approved):

PR #5072, opened 2024-12-05
Accurately indicate which prompts are overridden with Reset button

jkeiser:
This makes it so the Reset button is a solid indicator of whether a prompt is overridden even if you keep the application open for a while. Before this, you would have to reload to be sure.

This:
* Subscribes to the `PromptUpdated` event and updates the `overridden` state of prompts when it is received.
* Renames `PromptEditor.vue` -> `PromptAdmin.vue` for consistency
* Adds an `"all"` option to `realtimeStore.subscribe()` that lets you subscribe to workspace-independent events.


app/web/src/store/prompt.store.ts
```diff
@@ -65,26 +60,48 @@ export const usePromptStore = () => {
...
 
-        // Fetch selectedPromptText when selectedPromptKind changes
-        const { selectedPromptKind } = storeToRefs(this);
-        watch(selectedPromptKind, (kind) => {
-          this.selectedPromptText = "";
-          if (kind) this.FETCH_PROMPT(kind);
+        // Default selectedPromptKind to the first one (if available)
+        // TODO this is almost a computed value, except it needs to be settable as well to be
+        // used in a v-model. There must be a better way to do that.
+        watchEffect(() => {
```
jobelenus, line 75: In general, this is the correct approach. However, there is one way to make this more imperative. Instead of using `watch`, have action methods on the store that change `this.selectedPromptKind`... and thats the *only* watch to change that value... same for the second watcher you've got.

jobelenus (approved):

app/web/src/store/prompt.store.ts
jkeiser: Totally on board there; I didn't do it for two reasons:
1. `v-model` wants a ref (something you can both get and set), so I'd need to make some kind of computed thingy, wouldn't I? I see there is a way to say `computed({ get: () => ..., set(value) => ... })` so maybe I could use that.
2. For "clear out text when kind changes," I didn't want text to get cleared out if you set it to the same value it already had (and watch protects against that). I could totally add that in the setter, though!

jobelenus (approved):

PR #5076, opened 2024-12-06
docs: Documenting management functions as a segment of asset funcs

stack72:


app/docs/src/reference/asset/function.md
```diff
@@ -779,77 +780,587 @@ async function main(component: Input): Promise<Output> {
...
+
+Management functions are typically used to achieve the following things:
+
+1. Importing existing resources into System Initiative
+2. Creation of a template
+3. Configuring pre-existing components
+
+These functions will also be able to be used for a host of other types of use-cases, like:
+
+* Being able to run analysis across a segment of components
```
zacharyhamm, line 849: Passive voice. "Running analyses" instead of "Being able to run"

app/docs/src/reference/asset/function.md
```diff
@@ -779,77 +780,587 @@ async function main(component: Input): Promise<Output> {
...
+  coordinates for the component.
+
+- `components`
+
+  This is an object that contains all of the components that a management
+  function is connected to, keyed by the component id. Each of these components
+  expose the component type, this is largely the schema name, the `properties`,
+  `geometry`, `parent` and an array `connections`.
+
+The entire structure of the input is:
```
zacharyhamm, line 877: Since the type structure automatically adapts if the schemas change, it might be worth pointing out that this type is what it looks like if it's a VPC management component, and that it will change if it manages other kinds of components

zacharyhamm (commented):
A few notes, but this should give people a good start on management funcs!

app/docs/src/reference/asset/function.md
stack72: ah you are correct - I will make it a bit more generic!

stack72: ok @zacharyhamm made it a bit more generic for now

zacharyhamm (approved):

PR #5078, opened 2024-12-06
feat: add foyer metrics to our metrics pipeline

sprutton1:
With the latest version of foyer we can export usage metrics to opentelemetry. This requires an opentelemetry update to `0.26.0`. I originally attempted to update to `0.27.0`, but we have some other dependencies on the older version that made it trickier than it was worth. The API also changes pretty drastically, so I think that undertaking would be better left to a dedicated effort. 

[I borrowed a graph ](https://github.com/foyer-rs/foyer/blob/main/etc/grafana/dashboards/foyer.json)from our friends over at the foyer repo. It'll need some minor changes to adapt it to our environment, but it works decently enough locally.

cc @MrCroxx Thanks for the related PR. These metrics should help us get to the bottom of our memory issues. I'll let you know what we turn up next week!

image

### DIFF
--- a/lib/telemetry-application-rs/src/lib.rs
+++ b/lib/telemetry-application-rs/src/lib.rs
@@ -7,6 +7,10 @@
 )]
 // TODO(fnichol): document all, then drop `missing_errors_doc`
 #![allow(clippy::missing_errors_doc)]
+use telemetry::opentelemetry::{
+    metrics::MetricsError,
+    trace::{TraceError, TracerProvider},
+};
 use tracing_subscriber::{filter::FilterExt, layer::SubscriberExt, util::SubscriberInitExt};
 
 use std::{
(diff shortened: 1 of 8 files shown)


lib/telemetry-application-rs/src/lib.rs
```diff
@@ -434,11 +444,7 @@ fn telemetry_resource(config: &TelemetryConfig) -> Resource {
     // TODO(fnichol): create opentelemetry-resource-detector-aws for ec2 & eks detection
     Resource::from_detectors(
         Duration::from_secs(3),
-        vec![
```
sprutton1, line 437: @fnichol I wasn't sure what these were for exactly. The `ProcessResource` and `OsResource` detectors no longer seem to be available.

fnichol: If I recall, the `ProcessResouce` gave us attributes like `process.command_args` and `process.pid`, and the `OsResource` gave attributes such as `os.type` (maybe that was about it?). There also used to be one for AWS which I never enabled (at the time I wasn't sure of our deployment strategy). Maybe these were split out or retired--maybe I can check as well.

fnichol (approved):
Looking good, it's worth seeing what we get without those missing resource detectors

MrCroxx:
Thanks for letting me know. 🥰 And sorry for the late response, I just got back from a business trip. 

It's my pleasure to see foyer working on your project. Feel free to reach me if there is other problems. 🙏

BTW, may I ask if restricting the admission rate limit works for the OOM issue?

PR #5087, opened 2024-12-08
feat: don't page unless you're absolutely sure it's worth waking someone up for 

johnrwatson:
I'm not particularly happy with this indirect assertion system, but logically it ties together and should hold us in good stead moving forward to prevent unnecessary pages.

It works off the premise that the only reason we should page someone from CI is:
- If we have a valid failure test result
- It's the correct environment

Previously we had a bunch of other failure scenarios which would have also paged, such as:
- NPM outage
- Github orchestration issue/infrastructure issue on their side
- Cypress results not uploading correctly in CI
- Other networking issue/blip
- Other 3rd party dependency issue

Now for every time the system wants to assert whether to page someone, the CI runner MUST have a valid failure artifact or marker pulled from the CI workflow.

`ALL` failures, regardless of cause will still post into #_alerts_internal to let us know there was an orchestration issue of some manner. Should save a lot of noise in paging people with no actionable result.

For the API tests, to assert whether the test actually failed while running our code, I made it throw exit code 53, which is a random code I devised that should not match any other orchestration failure from deno, dependencies or otherwise.

<hr />

Also fixed an issue with the Slack notification for failed tests not working because it didn't have the environment to pull the Slack token.

Developing this was a proper grind
<div>Screenshot 2024-12-09 at 23 04 05</div>

### DIFF
--- a/.github/workflows/e2e-validation.yml
+++ b/.github/workflows/e2e-validation.yml
@@ -95,7 +95,7 @@ jobs:
 
             # Run the npx task and store exit code in a variable
             npx cypress run --spec "cypress/e2e/${{ matrix.tests }}/**" || exit_code=$?
-            
+
             # Check the exit code
             if [ -z "$exit_code" ]; then
               echo "Cypress Test task succeeded!"
@@ -113,7 +113,7 @@ jobs:
 
       - name: 'Upload Cypress Recordings to Github'
         uses: actions/upload-artifact@v4
-        if: always()
+        if: failure()
         with:
           name: cypress-recordings-run-${{ matrix.tests }}
           path: app/web/cypress/videos/**/*.mp4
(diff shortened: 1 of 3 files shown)


.github/workflows/run-api-test.yml
```diff
@@ -109,7 +112,7 @@ jobs:
             
             # Check the exit code
             if [ -z "$exit_code" ]; then
-              echo "Deno task succeeded!"
+              echo "Deno task succeeded [ or the orchestration failed for a totally non-valid reason ]!"
```
britmyerss, line 115: `totally non-valid` 🤣

.github/workflows/run-api-test.yml
```diff
@@ -120,32 +121,67 @@ jobs:
...
           if [ $n -ge $max_retries ]; then
             echo "All $max_retries attempts failed."
-            exit 1
+            exit_code=1
           fi
 
+          echo "last_exit_code=$exit_code" >> "$GITHUB_ENV"
+          exit "$last_exit_code"
+
+      - name: Upload artifact if exit code 53
```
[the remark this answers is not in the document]
johnrwatson: See description
```
For the API tests, to assert whether the test actually failed while running our code, I made it throw exit code 53, which is a random code I devised that should not match any other orchestration failure from deno, dependencies or otherwise.
```

.github/workflows/e2e-validation.yml
```diff
@@ -95,25 +95,25 @@ jobs:
...
             npx cypress run --spec "cypress/e2e/${{ matrix.tests }}/**" || exit_code=$?
-            
+
             # Check the exit code
             if [ -z "$exit_code" ]; then
               echo "Cypress Test task succeeded!"
               break
             fi
 
+            if [ $n -ge $max_retries ]; then
```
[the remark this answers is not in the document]
johnrwatson: Great catch, I'll move it back

.github/workflows/e2e-validation.yml
```diff
@@ -126,24 +126,42 @@ jobs:
...
+          # Check for marker files
+          for marker in artifacts/*/*.mp4; do
+            if [ -f "$marker" ]; then
+              echo "Artifact detected for failed test: $marker"
+              echo "Setting failure to true and breaking"
+              has_artifacts=true
+              break
+            fi
+          done
+          # If at least one valid failure marker is present, then page
```
britmyerss, line 147: one thought here - we could send different metadata to firehydrant and have firehydrant do the routing if it's page-able... I'm like 50/50 on this though, so your call

johnrwatson: The tricky bit here was actually isolating when a signal should be sent at all into firehydrant. We could possibly split it so that for each CI failure there are various failure modes, i.e.:
- If there is a github failure we pass a "github failure" metadata key
- If there is a cypress orchestration failure we send that as a metadata key
- If unmatched, we page(?)

It feels like we would spend a lot of time trying to increase our filter accuracy to more actuely catch + report on different failure modes.

It certainly would help with metrics, i.e. we'd see exactly how often certain failure modes happen, but I think a ping into #_alerts_internal should be sufficient for now. If this doesn't work out I think what you suggested is probably the way to go

PR #5094, opened 2024-12-10
bug: update asset schema and create action prompts

adamhjk:
Fixes the schema prompt for AWS assets, to always include a region property and a corresponding socket.


lib/asset-sprayer/prompts/aws/asset_schema.yaml
```diff
@@ -19,6 +19,25 @@ messages:
 
       If you encounter a JSON Schema, you will use it to construct properties using the System Initiative API, not only as a Joi validation.
 
+      You will always include a region property in the schema, and a corresponding input socket connected to it, using the following property and socket definition, delimited by :::'s
+
```
stack72, line 23: thank you!!

stack72 (dismissed):

stack72 (approved):

PR #5095, opened 2024-12-10
feat(sdf, dal): Adding the ability to set a webhook url for a workspace

stack72:
We currently only post to that webhook from the change set approval flow but we want to make this more of an opt-in to what you can / can't to

### DIFF
--- a/app/web/src/store/workspaces.store.ts
+++ b/app/web/src/store/workspaces.store.ts
@@ -12,6 +12,7 @@ import handleStoreError from "./errors";
 import { AuthApiRequest } from ".";
 
 export type WorkspacePk = string;
+export type WorkspaceIntegrationId = string;
 
 type WorkspaceExportId = string;
 type WorkspaceExportSummary = {
@@ -39,6 +40,12 @@ export type WorkspaceImportSummary = {
   importedWorkspaceName: string;
 };
 
+export type WorkspaceIntegration = {
+  pk: WorkspaceIntegrationId;
+  workspaceId: WorkspacePk;
+  slackWebhookUrl?: string;
+};
+
 const LOCAL_STORAGE_LAST_WORKSPACE_PK = "si-last-workspace-pk";
 
 // Note(victor): The workspace import exists outside a change set context
@@ -60,6 +67,7 @@ export const useWorkspacesStore = () => {
         importId: null as string | null,
         importLoading: false as boolean,
         importError: undefined as string | undefined,
+        integrations: null as WorkspaceIntegration | null,
       }),
       getters: {
         allWorkspaces: (state) => _.values(state.workspacesByPk),
@@ -78,6 +86,9 @@ export const useWorkspacesStore = () => {
             this.urlSelectedWorkspaceId || "",
             null,
           );
+        },
+        getIntegrations(): WorkspaceIntegration | null {
+          return this.integrations || null;
         },
       },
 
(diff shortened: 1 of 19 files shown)


app/web/src/store/workspaces.store.ts
```diff
@@ -181,6 +234,7 @@ export const useWorkspacesStore = () => {
         watch(
           () => this.selectedWorkspacePk,
           () => {
+            this.GET_INTEGRATIONS();
```
nickgerace, line 237: If I'm reading this right, when this route is run, then the below code will execute in this watch?

lib/dal/src/workspace_integrations.rs
```diff
@@ -0,0 +1,126 @@
...
+pub enum WorkspaceIntegrationsError {
+    #[error(transparent)]
+    Pg(#[from] PgError),
+    #[error("transactions error: {0}")]
+    Transactions(#[from] TransactionsError),
+}
+
+pub type WorkspaceIntegrationsResult<T> = Result<T, WorkspaceIntegrationsError>;
+
+pub use si_id::WorkspaceIntegrationId;
```
nickgerace, line 17: Ironically I kept this pattern purely because I didn't want the original `si-id` PR to update **_every_** call site. Though I can't say I hate if we continue this either rather than just `use si_id::<my-id>`... hmm... at any rate, looks good!

lib/dal/src/workspace_integrations.rs
```diff
@@ -0,0 +1,126 @@
...
+        ctx: &DalContext,
+    ) -> WorkspaceIntegrationsResult<Option<Self>> {
+        let workspace_pk = ctx.workspace_pk()?;
+
+        let maybe_row = ctx
+            .txns()
+            .await?
+            .pg()
+            .query_opt(
+                "SELECT * FROM workspace_integrations AS w WHERE workspace_pk = $1",
```
nickgerace, line 96: Are we using the `AS w` here? Just want to make sure I'm not missing anything

lib/sdf-server/src/service/v2/change_set.rs
```diff
@@ -76,6 +89,41 @@ pub type ChangeSetAPIError = Error;
...
+
+            let response = client.post(webhook_url).json(&slack_message).send().await?;
+
+            let status = response.status();
+            let body = response.text().await?;
+
+            if status.is_success() {
+                return Ok(());
+            } else {
+                eprintln!(
```
nickgerace, line 115: This should probably be `error!`

lib/si-id/src/lib.rs
```diff
@@ -240,3 +240,4 @@ id!(ViewId);
 id!(WorkspaceId);
 id!(WorkspacePk);
 id!(WorkspaceSnapshotNodeId);
+id!(WorkspaceIntegrationId);
```
nickgerace, line 243: Nice!

lib/dal/src/migrations/U3410__workspace_integration.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE workspace_integrations
+(
+    pk                          ident primary key default ident_create_v1(),
+    slack_webhook_url           text NULL,
```
britmyerss, line 4: I would add a name and then make this webhook_url

stack72: gotcha!

nickgerace (approved):

PR #5098, opened 2024-12-10
Feat: adding some view operations to command bar

jobelenus:
Adding:
- Go to View (by name)
- New View (type the name)

### DIFF
--- a/app/web/src/components/Workspace/CommandModal.vue
+++ b/app/web/src/components/Workspace/CommandModal.vue
@@ -22,6 +22,8 @@
           &lt;socket&gt;
         </li>
         <li>[Q]ueue a &lt;component&gt; &lt;action&gt;</li>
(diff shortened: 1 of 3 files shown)


app/web/src/components/Workspace/CommandModal.vue
```diff
@@ -78,6 +80,50 @@ class Pan implements Command {
...
+    return new GotoView();
+  }
+}
+
+// eslint-disable-next-line @typescript-eslint/no-empty-interface
+interface NewView extends Command {}
+class NewView implements Command {
+  name = "New View";
+  shortcut = "N";
+  expects: CommandArg[] = ["stringInput"];
```
jobelenus, line 109: @jkeiser you might find this part interesting—I added support to type (and hit enter) on an arbitrary string input (here: a name for a view) that is not a part of a dropdown menu

PR #5099, opened 2024-12-10
Split core si-id macro into multiple macros

nickgerace:
## Description

This PR splits the core si-id macro into four macros and is a follow-up to #5093.

The macros are based on the existence of PostgreSQL conversion implementations and the "nil" ID. The macros using "nil" should not be used by any new IDs and exist for backwards compatibility.

## Additional Changes

- All macros have moved into their own module
- All conversions between IDs have been moved into their own module
- Our standard set of lints for new crates have been added to "si-id" (along with resolutions to failed lints)

## Potential Future Work

The below list has been carried over and modified from from #5093:

- Migrate all IDs in "si-layer-cache" (currently not reliant on a macro)
- Migrate all IDs in "module-index-server" (currently not reliant on a macro)
- Clean-up ID call sites (e.g. reduce "pub use" statements, but we may not actually want this)
- Remove all usages of the third party ulid crate and go through "si-id" (we may not actually want this)

### DIFF
--- a/lib/si-id/src/lib.rs
+++ b/lib/si-id/src/lib.rs
@@ -1,228 +1,47 @@
+//! Provides a centralized location for constructing identifiers for SI.
+
+#![warn(
+    bad_style,
+    clippy::missing_panics_doc,
+    clippy::panic,
+    clippy::panic_in_result_fn,
+    clippy::unwrap_in_result,
+    clippy::unwrap_used,
+    dead_code,
+    improper_ctypes,
+    missing_debug_implementations,
+    missing_docs,
+    no_mangle_generic_items,
+    non_shorthand_field_patterns,
+    overflowing_literals,
+    path_statements,
+    patterns_in_fns_without_body,
+    unconditional_recursion,
+    unreachable_pub,
+    unused,
+    unused_allocation,
+    unused_comparisons,
+    unused_parens,
+    while_true
+)]
+
+#[macro_use]
+pub(crate) mod macros;
+pub(crate) mod conversions;
 pub mod ulid;
 
-macro_rules! id {
-    (
-        $(#[$($attrss:tt)*])*
-        $name:ident
-    ) => {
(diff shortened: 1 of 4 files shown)


lib/si-id/src/lib.rs
```diff
@@ -1,242 +1,75 @@
...
+id_with_pg_types!(ComponentId);
+id_with_pg_types!(FuncId);
+id_with_pg_types!(FuncRunId);
+id_with_pg_types!(SchemaId);
+id_with_pg_types!(UserPk);
+
+// Please keep these alphabetically sorted!
+id_with_none!(SchemaVariantId);
+
+// Please keep these alphabetically sorted!
```
nickgerace, line 71: This comment appears multiple times and I think there's a world where we could read from an enum with the `#[remain::sorted]` macro to handle sorting here within another macro, but for now, this should suffice.

lib/si-id/src/macros.rs
```diff
@@ -0,0 +1,299 @@
...
+            ) -> Result<postgres_types::IsNull, Box<dyn std::error::Error + Sync + Send>> {
+                postgres_types::ToSql::to_sql(&self.0.to_string(), ty, out)
+            }
+        }
+    };
+}
+
+// TODO(nick): figure out how to scope this for only macros in this module.
+#[allow(missing_docs)]
+macro_rules! do_not_use_directly_id_inner {
```
nickgerace, line 173: Open to ideas on how to ensure this macro isn't exposed outside of this _file_. Low priority for this PR though.

jhelwig (approved):

lib/si-id/src/macros.rs
nickgerace: This isn't possible in that way that we want it with stable Rust right now. Amended the comment accordingly

jhelwig (approved):

PR #5103, opened 2024-12-11
Remove unused Rust dependencies

nickgerace:
This PR removes as many unused Rust dependencies as I could find within the repository.


jkeiser (approved):
Thank you for doing this :) I looked through and there are no version changes I can see anywhere; this should be a "as long as it compiles it can't change anything" pr. One step closer to being able to regenerate Cargo.lock without having to surgically edit everything because it changes a bunch of stuff I didn't expect!

johnrwatson (approved):

PR #5104, opened 2024-12-11
Migrate all remaining IDs to si-id

nickgerace:
## Description

This PR migrates all remaining IDs that I could find to si-id. It adds sea-orm as a dependency to si-id, which will only be used if the IDs use it, though it is now included in the dependency chain.

## Methodology

The methodology to finding these was searching for `struct.*Id` and `struct.*Ulid` throughout the codebase. Eventually, it may be beneficial to remove the ulid dependency from all crates except for si-id to see if any hidden ones remain, but this should have covered all of them.

## I think we're done!

This PR is a follow-up to https://github.com/systeminit/si/pull/5099 and https://github.com/systeminit/si/pull/5093. 

There is potential future work to clean-up ID call sites, which would involve reducing "pub use" statements, but we may not actually want that. In addition, we may also want to restrict the usage of the third party ulid crate throughout the repository and make crates go through si-id, but we may not actually want that either.

Because neither of these action items are desirable in the near term, we can call the si-id work done... for now!


lib/module-index-server/src/routes/upsert_module_route.rs
```diff
@@ -121,11 +121,13 @@ pub async fn upsert_module_route(
...
     };
 
-    let new_schema_id = Some(SchemaId(Ulid::new()));
+    let new_schema_id = Some(SchemaId::new());
     let schema_id = match module_kind {
         ModuleKind::WorkspaceBackup => None,
         ModuleKind::Module => match module_schema_id {
-            Some(schema_id_string) => Some(SchemaId(Ulid::from_string(&schema_id_string)?)),
+            Some(schema_id_string) => {
+                Some(SchemaId::from_raw_id(Ulid::from_string(&schema_id_string)?))
```
jhelwig, line 129: Why a `from_raw_id` `fn` instead of a `impl From<ulid::Ulid> for SchemaId`? (Pretty much the same question for the other `from_raw_id`, too.)

nickgerace: Just an oversight! Fixing...

jhelwig (approved):

PR #5108, opened 2024-12-11
Fix: Restore geometries too

vbustamante:

### DIFF
--- a/lib/dal/src/component.rs
+++ b/lib/dal/src/component.rs
@@ -3332,6 +3332,10 @@ impl Component {
         ctx.add_dependent_values_and_enqueue(component.input_socket_attribute_values(ctx).await?)
             .await?;
 
+        Geometry::restore_all_for_component_id(ctx, component_id)
+            .await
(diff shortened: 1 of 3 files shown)


britmyerss:
Looks good to me - but we should either add a new test or expand an existing one to cover this going forward.

jobelenus (approved):

lib/dal/src/component.rs
```diff
@@ -3332,6 +3332,10 @@ impl Component {
         ctx.add_dependent_values_and_enqueue(component.input_socket_attribute_values(ctx).await?)
             .await?;
 
+        Geometry::restore_all_for_component_id(ctx, component_id)
```
britmyerss, line 3335: The reason you went here vs. expanding the `import_component_subgraph` is because the geometries aren't in the subgraph of the component?

britmyerss (approved):

lib/dal/src/component.rs
vbustamante: @britmyerss yeah, the subgraph code is very generic. Since we've structured our geometries the other way around I had to implement a special case here

PR #5110, opened 2024-12-11
Add audit trail reference and entity vocab

nickgerace:
## Description

This PR adds an audit trail reference document and entity-based vocabulary words.

### DIFF
--- a/app/docs/src/reference/audit-trail.md
+++ b/app/docs/src/reference/audit-trail.md
@@ -1,0 +1,124 @@
+# Audit Trail Reference
+
+The Audit Trail feature gives you the ability to know what is going on inside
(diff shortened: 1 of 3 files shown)


app/docs/src/reference/audit-trail.md
```diff
@@ -0,0 +1,6 @@
+# Audit Trail Reference
+
+The Audit Trail feature gives you the ability to know what is going on inside
```
stack72, line 3: We should add a sample of the structured log format IMO and show that metadata changes per type

stack72 (dismissed):
This is so good!!

stack72 (approved):

PR #5111, opened 2024-12-12
Add ability to run iterative prompts

jkeiser:
This adds two distinct capabilities to the system:
* The ability to include the current function text in a prompt with `{FUNCTION_TEXT}`. This allows ChatGPT to "fill in" a function you've already written rather than just generating a new one.
* The ability to run a prompt iteratively: prompts are now an array of messages rather just one. After each prompt, the response from ChatGPT will be taken as the new function text, and fed back into the next prompt. This allows you to break up the problem for ChatGPT and/or cut files in pieces:
   1. Figure out the property names and hierarchy in this schema.
   2. Add validations to the schema.
   3. Add sockets.
   4. etc.

More sophisticated outputs from ChatGPT (other things than function text) may be useful here, and this should let us experiment with different docs at different times.

An example second prompt:

```yaml
- ...
  <normal first prompt>
  ...
- model: gpt-4o-mini
  temperature: 0.0
  messages:
    - role: system
      content: >
        The following is a function in typescript, between three ^ characters. 

        ^^^
        {FUNCTION_TEXT}
        ^^^
        
        Please reproduce the entire function and add the comment "hi mom".

        Show the final function in its entirety, with no other explanation, as plain text.
        
        Do not wrap the code in markdown delimiters.
```

And there's "hi mom" at the bottom of the generated function!

### DIFF
--- a/bin/asset-sprayer-prompts/src/main.rs
+++ b/bin/asset-sprayer-prompts/src/main.rs
@@ -73,11 +73,16 @@ async fn main() -> Result<()> {
     let raw_prompt = asset_sprayer.raw_prompt(args.prompt_kind).await?;
     match args.action {
         Action::Show => {
-            let prompt = prompt.generate(&raw_prompt).await?;
-            println!("{}", serde_yaml::to_string(&prompt)?);
+            for raw_request in prompt.read(&raw_prompt)? {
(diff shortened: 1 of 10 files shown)


jkeiser:
We can resurrect if we go further with AI, but we've gone a new direction with asset generation.

PR #5117, opened 2024-12-12
fix: delete components on head if they are not in default view

zacharyhamm:
Sends a WsEvent without geometry so views are not taken into account when deleting a component.

Fixes BUG-699

To test this: create two components in a view that is not the default view. One that will have a resource (successfully get created) and one that does not (sec group and docker image were my test). Apply to head, be sure that a resource is created for the resource one. Then in a new change set, attempt to delete them. Delete should succeed.


jobelenus (changes requested):
as we talked about, we just need the `into_frontend_type` once, not once per view 👍

jobelenus (approved):

PR #5119, opened 2024-12-12
Hotfix: watcher is too aggressive

jobelenus:

### DIFF
--- a/app/web/src/store/views.store.ts
+++ b/app/web/src/store/views.store.ts
@@ -1559,7 +1559,7 @@ export const useViewsStore = (forceChangeSetId?: ChangeSetId) => {
             )
               this.syncUrlIntoSelection();
             const viewId = router.currentRoute.value.params.viewId;
-            if (viewId) {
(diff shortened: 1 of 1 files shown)


stack72 (approved):

app/web/src/store/views.store.ts
```diff
@@ -1559,7 +1559,7 @@ export const useViewsStore = (forceChangeSetId?: ChangeSetId) => {
             )
               this.syncUrlIntoSelection();
             const viewId = router.currentRoute.value.params.viewId;
-            if (viewId) {
+            if (viewId && this.selectedViewId !== viewId) {
```
wendybujalski, line 1562: ah yeah as soon as you mentioned the watcher I realized that we didn't add a check for if you are already in that view! 🤪

wendybujalski (approved):
👍

app/web/src/store/views.store.ts
jobelenus: 🧠

PR #5123, opened 2024-12-12
Try color-eyre in dal integration tests

nickgerace:


nickgerace:
Closing since we already have `dal_test::Result` and we will migrate to using result types and `color-eyre` as we go.

PR #5128, opened 2024-12-13
feat: start all layer_caches in parallel

sprutton1:
This is to help improve the speed at which we start the application when the disk caches are relatively full.


zacharyhamm (approved):
Worth a shot

PR #5130, opened 2024-12-13
Add pipeline to restart SDF/Rebaser/Pinga units on relevant nodes

johnrwatson:
Adds a new pipeline which we can trigger that will restart the service replicas that are impacted by our current run-away memory issue. It will basically quickly drop the replicas and bring them back. For SDF it pushes the FE into maintenance mode first to give a bit of better user experience, all the other services are gracefully stopped so they should finish up their current work before attempting to be started again.

In theory, it's much better to take just one node offline at a time, but it would require a fair bit more re-work in our toolbox to let us do it.

We can schedule this like a maintenance event (using the maintenance mode) and it should be very minimally impactful for the users.

This is untestable until I merge it, because the workflow doesn't exist in the UI/API until it's present on main at least once

### DIFF
--- a/.github/workflows/deploy-service-restart.yml
+++ b/.github/workflows/deploy-service-restart.yml
@@ -1,0 +1,70 @@
+name: Restart Services to refresh memory
+
+run-name: Refresh Service Unit in ${{ inputs.environment }} by @${{ github.actor }}
+
+on:
+  workflow_call:
+    inputs:
+      environment:
+        type: string
+        default: "tools"
+
+  workflow_dispatch:
+    inputs:
+      environment:
+        type: choice
+        required: true
+        description: "Where to deploy"
+        default: "tools"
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy-service-restart.yml
```diff
@@ -0,0 +1,70 @@
...
+# next one.
+# Note that this will also wait for E2E tests to finish, as per the crons workflows
+concurrency:
+  group: deploy-${{ inputs.environment }}
+  cancel-in-progress: false
+
+jobs:
+
+  restart-rebaser:
+    uses: ./.github/workflows/instance-refresh.yml
```
sprutton1, line 34: do you mean to use the instance-refresh here or the new service-restart you wrote?

sprutton1 (approved):

PR #5137, opened 2024-12-16
chore(deps): Update nanoid dependency

stack72:


bin/auth-api/package.json
```diff
@@ -44,7 +44,7 @@
     "koa-bodyparser": "^4.3.0",
     "lodash": "^4.17.21",
     "luxon": "^3.3.0",
-    "nanoid": "^4.0.1",
```
stack72, line 47: Auth API works perfectly with this lower version - we are not using anything specific for the 4.0.1 - upgrading to 5.x.x (which has the fix) is way more intrusive to our deps than downgrading to 3.3.8

zacharyhamm (approved):

PR #5138, opened 2024-12-16
Add search bar and timestamp sort to audit logs UI

nickgerace:


lib/vue-lib/src/design-system/menus/DropdownMenuItem.vue
```diff
@@ -69,6 +70,7 @@
       </div>
     </div>
     <div
+      v-if="!centerHeader && !header"
```
wendybujalski, line 73: ah I just realized this conditional isn't quite right - it should be `!(centerHeader && header)`

wendybujalski: my mistake - fix this and everything else looks great!

wendybujalski (changes requested):
just gotta fix that conditional and you'll be good to go!

lib/vue-lib/src/design-system/menus/DropdownMenuItem.vue
nickgerace: Done!

wendybujalski (approved):

PR #5139, opened 2024-12-16
BUG-697: z-index frames based on size, not parentage. resize parents to contain a larger child

jobelenus:


wendybujalski (approved):
Clever!

wendybujalski (approved):
👍

PR #5143, opened 2024-12-17
feat: Add button to create an "automation token" and use it in SDF API

jkeiser:
This adds the ability to create "automation tokens" intended for use with API clients in hands-free contexts like github.

* "Create API Token" button on Manage Workspace to generate a token and copy it to your buffer. This is *not* intended as a final UI; just a way to get this usable.
  image
* Automation tokens:
  - Not revocable
  - Have an expiration of 1 day (which SDF does not appear to honor)
  - Are rejected by all SDF and module-index endpoints
* Token format for existing users has not yet changed, to ensure we don't interrupt service. All services will be rolled out with the ability to handle the both the old token format and the new one, but only automation tokens will be minted with the new format until the next deployment.
* Automation tokens are initially heavily restricted to a very small set of permissions. We may open it up to things we deem safe for automation.

### Testing

I have tested:

* Old and new format for user tokens both work when:
  - Creating a new workspace from the auth portal.
  - Installing components in a fresh workspace (installing new modules from a local module-index server).
  - Multiplayer and creating / moving components around.
  - Hitting the SDF and local module-index server.
* Tokens generated by the old Auth API allow access to workspaces as well.
* Automation tokens are rejected by the SDF API as well as the module-index API.

Automated tests: added tests for the new token formats. There is no new coverage in the SDF, DAL or auth api automated tests.

It is critical we ensure SDF and module-index are working after deploy in preprod. I have not tested the new tokens against preprod or prod module-index / SDF, because that requires private keys from those environments to create the tokens.

### Deployment

auth api, SDF, and module-index will all have to be deployed. It doesn't matter what order they are deployed in, but all need to be deployed to ensure the new code is generating and accepting tokens correctly to ensure service doesn't get interrupted.

### Caveats

* Tokens have a 1 day expiration date, but SDF does *not* honor it. jwt.io indicates the token should be expired. Something to be aware of: I don't expect many people to create these tokens yet until they, you know, do something, so I'll address this soon after.

### Quickstart

You too can have your API requests denied due to lack of permission:
1. Run the auth api and portal locally (see instructions in `bin/auth-api/README.md`).
2. Create a new workspace.
3. Click the "Generate Automation Token" button (see image above). Key is now in your buffer.
4. Hit `curl -H "Authorization: Bearer $MY_AUTH_API_TOKEN" http://localhost:5156/api/v2/workspaces/01JF8X1BCPAFTFR18J24JHQGBE/integrations`

See your very own 401 saying you don't have a `web` role.

token

### DIFF
--- a/bin/auth-api/src/routes/auth.routes.ts
+++ b/bin/auth-api/src/routes/auth.routes.ts
@@ -74,7 +74,11 @@ router.post("/auth/login", async (ctx) => {
     throw new ApiError("Forbidden", "You do not have access to that workspace");
   }
 
-  const token = createSdfAuthToken(user.id, workspaceId);
+  const token = createSdfAuthToken({
+    userId: user.id,
+    workspaceId,
+    role: "web",
+  });
 
   ctx.body = { token };
 });
--- a/bin/auth-api/src/routes/workspace.routes.ts
+++ b/bin/auth-api/src/routes/workspace.routes.ts
@@ -69,19 +69,21 @@ async function extractOwnWorkspaceIdParam(ctx: CustomRouteContext) {
   return workspace;
 }
 
-async function isWorkspaceOwner(ctx: CustomRouteContext) {
+async function authorizeWorkspaceRoute(ctx: CustomRouteContext, role?: RoleType) {
   const workspace = await extractWorkspaceIdParam(ctx);
-
   const authUser = extractAuthUser(ctx);
-  const memberRole = await userRoleForWorkspace(authUser.id, workspace.id);
-  if (memberRole !== RoleType.OWNER) {
-    throw new ApiError(
-      "Forbidden",
-      "You do not have the correct permisison to edit this workspace",
-    );
+
+  if (role) {
+    const memberRole = await userRoleForWorkspace(authUser.id, workspace.id);
+    if (memberRole !== role) {
+      throw new ApiError(
+        "Forbidden",
+        "You do not have the correct permission to edit this workspace",
+      );
+    }
   }
 
-  return true;
+  return { authUser, workspace };
 }
 
 router.get("/workspaces/:workspaceId", async (ctx) => {
@@ -141,10 +143,7 @@ router.post("/workspaces/new", async (ctx) => {
 });
 
 router.patch("/workspaces/:workspaceId", async (ctx) => {
-  const authUser = extractAuthUser(ctx);
-  await isWorkspaceOwner(ctx);
-
-  const workspace = await extractOwnWorkspaceIdParam(ctx);
+  const { authUser, workspace } = await authorizeWorkspaceRoute(ctx, RoleType.OWNER);
 
   const reqBody = validate(
     ctx.request.body,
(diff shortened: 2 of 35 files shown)


bin/auth-api/src/routes/auth.routes.ts
```diff
@@ -74,7 +74,11 @@ router.post("/auth/login", async (ctx) => {
     throw new ApiError("Forbidden", "You do not have access to that workspace");
   }
 
-  const token = createSdfAuthToken(user.id, workspaceId);
+  const token = createSdfAuthToken({
+    userId: user.id,
+    workspaceId,
+    role: "web",
```
stack72, line 80: do we really want to hard code a role of "web" - can't this at least be an enum we extend?

bin/auth-api/src/routes/workspace.routes.ts
```diff
@@ -424,6 +404,20 @@ router.get("/workspaces/:workspaceId/go", async (ctx) => {
   ctx.redirect(redirectUrl);
 });
 
+router.post("/workspaces/:workspaceId/createAutomationToken", async (ctx) => {
+  const { authUser, workspace } = await authorizeWorkspaceRoute(ctx, RoleType.OWNER);
+
+  const token = createSdfAuthToken({
+    userId: authUser.id,
+    workspaceId: workspace.id,
+    role: "automation",
```
stack72, line 413: this should also be part of the enum

bin/auth-api/src/routes/auth.routes.ts
jkeiser: The enum is here: https://github.com/systeminit/si/pull/5143/files#diff-45a5a885b573adda0e5349e1af441abf982f4f977141755f48dd5780d52a17bdR52

Basically in TS (at least in my experience) unless an enum has extra data they are generally a type union of the possible values, and you specify those values directly. TS will yell at you if you don't use one of them.

britmyerss, line 80: I'm a little confused by this - we're scoping this token to only be used by the web app?

jkeiser: Sort of. Roles are a general indication of what you're using the token for, and that tells you about both what you can access as well as the scope of the token. The current tokens we give people are designed for use during active sessions in a web browser, so we called it `"web"`. Alternate names are absolutely accepted :) What we didn't want was to leave it blank for this purpose; giving you essentially maximal access by default seemed undesirable.

jkeiser: TS is being weird about the enum, but the line is in auth.service.ts:

```typescript
export type SdfAuthTokenRole = "web" | "automation";
```

And it's part of the type on the parameter you pass to `createSdfAuthToken`, which is how TS knows what values are allowed.

bin/auth-api/src/routes/workspace.routes.ts
```diff
@@ -424,6 +404,20 @@ router.get("/workspaces/:workspaceId/go", async (ctx) => {
...
+  }, {
+    expiresIn: "1 day",
```
britmyerss, line 415: it's been a while since I was in JWT spec land - does this add the `exp` claim to the token? I would have expected that to get automatically enforced by SDF's middleware but I could be wayyy off here.

jkeiser: I would have too; looking at the token at `https://jwt.io` it appears correct, but I'll definitely be revisiting this.

PR #5144, opened 2024-12-17
Split out approvals API

nickgerace:
## Description

This PR splits out the approvals API for the first pass of the fine grained access control work. For now, this will only work with SchemaVariants and no other entities in SI will require approvals. In addition, the number of required approvers will be "1" and the approvers will be the workspace approvers rather than a specific group.

A new feature flag named "fine-grained-access-control" toggle this feature, which is currently driven by two new sdf routes: "approve_v2" (POST) and "approval_status" (GET).

## Frontend changes

- Load users and approval requirements and reload when they change
- Display a list of approvers and submitted approvals
- Rename "approvers" to "defaultApprovers"

## Test harness and helpers changes

- Add "prelude" and refactor module naming and setup as a result
- Add SpiceDbClient for "sdf_test" macro expansion

## Changes to the dal

- Add ChangeSetApproval struct and PG table
- Add change set approval round trip integration test

## Changes to WorkspaceSnapshot and WorkspaceSnapshotGraph

- Add ability to detect changes (not updates)
- Add ability to collect approval requirements for a given list of changes
- Add ability to calculate the checksum for a given list of IDs corresponding to nodes on the graph
- Add a new EntityKind trait to snapshot and snapshot graph
- Organize workspace snapshot module and workspace snapshot graph module imports
- Consolidate "detect_updates" and "detect_changes" logic into one Detector struct and corresponding module
- Add minimal unit test for detecting changes

## Changes to si-events and si-id

- Add EntityKind and Checksum to a new module for global workspace snapshot concepts
- Add ChangeSetApprovalStatus enum with only the "Approved" variant (used by both the frontend and the PG table)
- Add ChangeSetApprovalId corresponding to the new PG table

## Changes to sdf

- Consolidate latest approval and requirements logic for the frontend in a new crate-only module called "dal_wrapper", which contains logic that would otherwise be in the dal or a similar crate, but cannot be because it must explicitly be in sdf-server
- Rename "util" module to "key_generation" since that is the domain its contents serves
- Remove unused variants in the v2 change set error enum
- Add an integration test (not a service test, despite the module name) for testing the core feature of this commit using "dal_wrapper" functions and the SpiceDB client
- Add "approve_v2" POST route
- Add "approval_status" GET route


lib/dal/src/migrations/U3500__change_set_approvals.sql
```diff
@@ -0,0 +1,22 @@
+CREATE TABLE change_set_approvals
+(
+    id ident primary key NOT NULL DEFAULT ident_create_v1(),
+    created_at timestamp with time zone NOT NULL DEFAULT CLOCK_TIMESTAMP(),
+    change_set_id ident NOT NULL,
+    status text NOT NULL,
+    user_id ident NOT NULL,
```
britmyerss, line 7: is it worth adding an `applyWhenReady` flag now?  Or follow up migration?

nickgerace: Discussed offline, but we should not add this yet given that the "apply when ready" feature's shape is unclear yet.

nickgerace:
Rebased with main after https://github.com/systeminit/si/pull/5205

app/web/src/store/auth.store.ts
```diff
@@ -61,6 +68,21 @@ export const useAuthStore = () => {
       },
     },
     actions: {
+      // stolen from admin store for a second
+      async LIST_WORKSPACE_USERS(workspaceId: string) {
+        return new ApiRequest<{ users: WorkspaceUser[] }>({
+          method: "get",
+          // TODO(nick): move workspace users list out of the admin routes in sdf.
```
britmyerss, line 75: just to be clear - until this happens, only SI users can use this feature, yeah? That's ok - just wanted to make sure I understood (as the route is protected by admin permissions in sdf)

app/web/src/store/change_sets.store.ts
```diff
@@ -696,6 +747,12 @@ export function useChangeSetsStore() {
               this.changeSetsWrittenAtById[changeSetId] = new Date();
             },
           },
+          {
+            eventType: "ChangeSetApprovalStatusChanged",
+            callback: (changeSetId) => {
+              this.FETCH_APPROVAL_STATUS(changeSetId);
```
britmyerss, line 753: I always forget - do we need to filter by `changeSetId === currentChangeSetId` here?

lib/dal/src/workspace_snapshot/graph/v4/entity_kind.rs
britmyerss, line 1: why do we need entity kind here vs. using the NodeWeight Enum?

app/web/src/store/auth.store.ts
nickgerace: @britmyerss yes this is true and we will need to move the logic into another endpoint. We believe the inner logic is okay for users to access and does not need to come from the admin endpoints. However, we'll address that in a follow-up PR as the `TODO` is pretty literal... need to do it 😄

app/web/src/store/change_sets.store.ts
nickgerace: @jobelenus may be able to provide more context here, but I believe if we can avoid updating things unnecessarily for change sets you don't have selected, that is a good idea. _However,_ in this case, you might be approving multiple change sets or working in another change set while waiting for an approval for your original one... and in both cases, getting the latest approval status is a good idea, especially if we want to wire those updates through to notifications in the nav bar.

lib/dal/src/workspace_snapshot/graph/v4/entity_kind.rs
nickgerace: Yep :) `Content` nodes are the primary culprit here, but also `EntityKind` may allow for easier exposure of graph concepts than say... `NodeWeightDiscriminants`. But yeah, mainly `Content` nodes!

app/web/src/store/change_sets.store.ts
jobelenus: Probably need that filter. Getting change set status data w/o it being selected doesn't matter—since as soon as you select it, we re-load it anyway

lib/sdf-server/tests/service_tests/change_set_approval.rs
britmyerss, line 1: 👏 👏 👏

britmyerss (dismissed):

app/web/src/store/change_sets.store.ts
nickgerace: @jobelenus Great, I'll add a reminder to do so in future work (also this too https://github.com/systeminit/si/pull/5144#discussion_r1915457766)

britmyerss (dismissed):

nickgerace:
DEAR GOD WORK. PLEASE.

britmyerss (approved):

nickgerace:
The test error is literally unreproducible locally. I suspect something isn't being committed or the rebaser or spicedb is having a bad time in CI with the shared load, but that test failure should be impossible. Let's investigate tomorrow @jhelwig.

nickgerace:
Disabling auto-merge since we likely need to account for CI-specific behavior in the test.

britmyerss (approved):

PR #5145, opened 2024-12-17
toasts and other UI adjustments for template generation

wendybujalski:

### DIFF
--- a/app/web/src/components/ManagementRunPrototype.vue
+++ b/app/web/src/components/ManagementRunPrototype.vue
@@ -129,6 +129,9 @@ const runPrototype = async (viewId: ViewId) => {
   );
 
   if (result.result.success) {
+    // TODO(WENDY) - for template functions, select the returned componentIds
+    // modelingEventBus.emit("setSelection", componentIds);
+
     lastExecution.value = result.result.data;
     if (result.result.data.message) {
       const toastOptions = {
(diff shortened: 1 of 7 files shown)


app/web/src/components/ManagementRunPrototype.vue
```diff
@@ -129,6 +129,9 @@ const runPrototype = async (viewId: ViewId) => {
   );
 
   if (result.result.success) {
+    // TODO(WENDY) - for template functions, select the returned componentIds
+    // modelingEventBus.emit("setSelection", componentIds);
```
jobelenus, line 133: Does this need to be done? Also, I dont think we need the event bus... its just an action in the view store now...

app/web/src/store/components.store.ts
```diff
@@ -1169,21 +1171,57 @@ export const useComponentsStore = (forceChangeSetId?: ChangeSetId) => {
...
+              },
+              {
+                timeout: false,
+                closeOnClick: false,
+                position: POSITION.TOP_CENTER,
+                toastClassName: "si-toast-no-defaults",
+              },
+            );
 
-            return new ApiRequest({
```
jobelenus, line 1176: I prefer the `return` style, b/c things start to break if `ApiRequest` isn't returned.

jobelenus (approved):

PR #5147, opened 2024-12-17
fix: schemas should be created with lineage_id == id 

zacharyhamm:
On demand asset installation can produce identical Schema nodes by id, but they will have different lineage ids. Since the lineage ids are different, we will treat them as new nodes on the graph. The idempotency of our updates usually means this is fine. But the interaction with the code to ensure there is only one default edge caused an issue.

This fixes both, by ensuring new schemas are always created with the lineage id equal to the schema id. This will ensure we compare the nodes instead of treating the schema node as a new node.

It also ensures that if a new edge update comes in that sets a schema variant to the default, but that variant is already default, we preserve that default.

### DIFF
--- a/lib/dal/src/schema.rs
+++ b/lib/dal/src/schema.rs
@@ -149,9 +149,10 @@ impl Schema {
 
         let workspace_snapshot = ctx.workspace_snapshot()?;
 
-        let lineage_id = workspace_snapshot.generate_ulid().await?;
+        // Lineage id has to match id here, otherwise every new schema will be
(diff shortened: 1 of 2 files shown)


zacharyhamm:
To test this:

Add a new kind of component to a change set (one that has never been installed).
Create a new change set, and repeat with the exact same schema variant kind.

Apply one of the two change sets.
Then apply the second one (it should have two of those components on it).

HEAD should be in a working state.

stack72 (approved):

PR #5149, opened 2024-12-17
fix: allow moving components inside of their parents

zacharyhamm:
This ensures we don't attempt to correct the size of a parent if there are not in fact new parents.


fnichol (approved):
Looks good!

PR #5151, opened 2024-12-18
Instrument more audit log events

britmyerss:
This PR adds coverage to the following areas:
- Function Authoring (Test, Execute, Attach for each function kind, Detach, Delete, Create, Updating Metadata )
- Schema Variant Authoring (Create, Regenerate, Delete, Unlock,)
- Contributing a module
- Components (Creating/Deleting Connections, Attaching/Orphaning) 
- Views (Create, Update, Delete)

To do still include: 
- Handling code changes
- Handling View Diagram Objects
- Updating Schema Variant Metadata 
- Installing a Module
- More granular Func Binding updates (depending on how we want to approach this, for ex: updating the arguments for an attribute function)

<div><br/>via <a href="https://giphy.com/spongebob/">SpongeBob SquarePants</a> on <a href="https://giphy.com/gifs/spongebob-season-4-spongebob-squarepants-26nfp8HGGHLPGY2KQ">GIPHY</a></div>

### DIFF
--- a/lib/dal/src/func/binding.rs
+++ b/lib/dal/src/func/binding.rs
@@ -31,8 +31,8 @@ use crate::{
     FuncError, FuncId, PropId, SchemaVariantError, SchemaVariantId,
 };
 use crate::{
-    ComponentId, InputSocketId, OutputSocketId, SchemaError, SchemaId, SchemaVariant,
-    WorkspaceSnapshotError, WsEventError,
+    ComponentId, InputSocket, InputSocketId, OutputSocket, OutputSocketId, Prop, SchemaError,
+    SchemaId, SchemaVariant, WorkspaceSnapshotError, WsEventError,
 };
 
 pub use attribute_argument::AttributeArgumentBinding;
@@ -127,7 +127,7 @@ pub enum AttributeFuncDestination {
 }
 
 impl AttributeFuncDestination {
-    pub(crate) async fn find_schema_variant(
+    pub async fn find_schema_variant(
         &self,
         ctx: &DalContext,
     ) -> FuncBindingResult<SchemaVariantId> {
@@ -143,6 +143,25 @@ impl AttributeFuncDestination {
             }
         };
         Ok(schema_variant_id)
+    }
+
+    pub async fn get_name_of_destination(&self, ctx: &DalContext) -> FuncBindingResult<String> {
+        let name = match self {
+            AttributeFuncDestination::Prop(prop_id) => Prop::get_by_id(ctx, *prop_id).await?.name,
+            AttributeFuncDestination::OutputSocket(output_socket_id) => {
+                OutputSocket::get_by_id(ctx, *output_socket_id)
+                    .await?
+                    .name()
+                    .to_string()
+            }
+            AttributeFuncDestination::InputSocket(input_socket_id) => {
+                InputSocket::get_by_id(ctx, *input_socket_id)
+                    .await?
+                    .name()
+                    .to_string()
+            }
+        };
+        Ok(name)
     }
 }
 
--- a/lib/sdf-server/src/service/diagram/create_connection.rs
+++ b/lib/sdf-server/src/service/diagram/create_connection.rs
@@ -8,12 +8,13 @@ use axum::{
     extract::{Host, OriginalUri},
     Json,
 };
-use dal::diagram::SummaryDiagramInferredEdge;
 use dal::{
     change_status::ChangeStatus, diagram::SummaryDiagramEdge, ChangeSet, Component, ComponentId,
-    InputSocketId, OutputSocketId, Visibility, WsEvent,
+    InputSocket, InputSocketId, OutputSocketId, Visibility, WsEvent,
 };
+use dal::{diagram::SummaryDiagramInferredEdge, OutputSocket};
 use serde::{Deserialize, Serialize};
+use si_events::audit_log::AuditLogKind;
 
 #[derive(Deserialize, Serialize, Debug)]
 #[serde(rename_all = "camelCase")]
@@ -112,6 +113,28 @@ pub async fn create_connection(
                 .await?;
         }
     }
+    let to_component_name = to_component.name(&ctx).await?;
+    let to_socket_name = InputSocket::get_by_id(&ctx, request.to_socket_id)
+        .await?
+        .name()
+        .to_string();
+    ctx.write_audit_log(
+        AuditLogKind::CreateConnection {
+            from_component_id: request.from_component_id,
+            from_component_name: from_component.name(&ctx).await?,
+            from_socket_id: request.from_socket_id,
+            from_socket_name: OutputSocket::get_by_id(&ctx, request.from_socket_id)
+                .await?
+                .name()
+                .to_string(),
+            to_component_id: request.to_component_id,
+            to_component_name: to_component_name.clone(),
+            to_socket_id: request.to_socket_id,
+            to_socket_name: to_socket_name.clone(),
+        },
+        format!("{to_component_name} --- {to_socket_name}"),
+    )
+    .await?;
 
     ctx.commit().await?;
 
--- a/lib/sdf-server/src/service/v2/func/binding/attribute/reset_attribute_binding.rs
+++ b/lib/sdf-server/src/service/v2/func/binding/attribute/reset_attribute_binding.rs
@@ -4,8 +4,9 @@ use axum::{
 };
 use dal::{
     func::binding::{attribute::AttributeBinding, EventualParent},
-    ChangeSet, ChangeSetId, Func, FuncId, SchemaVariant, WorkspacePk, WsEvent,
+    ChangeSet, ChangeSetId, Component, Func, FuncId, SchemaVariant, WorkspacePk, WsEvent,
 };
+use si_events::audit_log::AuditLogKind;
 use si_frontend_types as frontend_types;
 
 use crate::{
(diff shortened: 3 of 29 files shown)


lib/dal/src/func/binding.rs
```diff
@@ -144,6 +144,25 @@ impl AttributeFuncDestination {
...
+    pub async fn get_name_of_destination(&self, ctx: &DalContext) -> FuncBindingResult<String> {
+        let name = match self {
+            AttributeFuncDestination::Prop(prop_id) => Prop::get_by_id(ctx, *prop_id).await?.name,
+            AttributeFuncDestination::OutputSocket(output_socket_id) => {
+                OutputSocket::get_by_id(ctx, *output_socket_id)
+                    .await?
+                    .name()
+                    .to_string()
+            }
+            AttributeFuncDestination::InputSocket(input_socket_id) => {
```
nickgerace, line 157: nit for the future: I like to use `Self` when possible for readability and suppleness for refactoring. So the first one would be `Self::Prop` instead of `AttributeFuncDestination::Prop`.

lib/sdf-server/src/service/diagram/create_connection.rs
```diff
@@ -112,6 +113,28 @@ pub async fn create_connection(
...
+            from_socket_name: OutputSocket::get_by_id(&ctx, request.from_socket_id)
+                .await?
+                .name()
+                .to_string(),
+            to_component_id: request.to_component_id,
+            to_component_name: to_component_name.clone(),
+            to_socket_id: request.to_socket_id,
+            to_socket_name: to_socket_name.clone(),
+        },
+        format!("{0}-{1}", to_component_name, to_socket_name),
```
nickgerace, line 135: nit for the future: I'm a bit surprised the linter didn't catch this, but you can do the following:
```rust
format!("{to_component_name}-{to_socket_name}"),
```

lib/sdf-server/src/service/v2/func/binding/attribute/reset_attribute_binding.rs
```diff
@@ -51,15 +52,42 @@ pub async fn reset_attribute_binding(
...
+                        func_id,
+                        func_display_name: func.display_name.clone(),
+                        schema_variant_id: None,
+                        component_id: Some(component_id),
+                        subject_name: component.name(&ctx).await?.to_owned(),
+                    },
+                    func.name.clone(),
+                )
                 .await?;
+            }
```
nickgerace, line 90: Great use of breaking out an `if let` into a `match`... sometimes you'll see uses of `match` when unnecessary to expose all the variants left unused to avoid bugs if more variants are added to the enum. Example:
```rust
enum MyEnum {
    One
    Two
    Three
}

impl MyEnum {
    pub fn name(&self) -> &'static str {
        match self {
            One => "one"
            Two => "one"
            Three => "one"
        }
    }
}
```

By using a `match`, we cover all the variants by default unless we use a wildcard, variable or match guard.

If that function instead used `if let One = self { // code } else { // other code }`, then adding new variants could cause undesired behavior in the `else` block.

I bring this all up since this is a great example of a sneaky `if let` that might not have been obvious from static code analysis, but you found it for this PR. Nice!

lib/sdf-server/src/service/v2/func/binding/create_binding.rs
```diff
@@ -288,7 +364,18 @@ pub async fn create_binding(
                             .await?;
                             let schema_variant =
                                 SchemaVariant::get_by_id_or_error(&ctx, schema_variant_id).await?;
-
+                            let func = Func::get_by_id_or_error(&ctx, func_id).await?;
+                            ctx.write_audit_log(
+                                AuditLogKind::AttachManagementFunc {
```
nickgerace, line 369: Unrelated to your PR but this route is frightening with all the nested SHEESH! We should refactor this later...

lib/sdf-server/src/service/v2/func/create_func.rs
```diff
@@ -151,14 +249,39 @@ pub async fn create_func(
...
+                ctx.write_audit_log(
+                    AuditLogKind::AttachCodeGenFunc {
+                        func_id: func.id,
+                        func_display_name: func.display_name.clone(),
+                        schema_variant_id: Some(schema_variant_id),
+                        component_id: None,
+                        subject_name: schema_variant_name,
+                    },
+                    func.name.clone(),
+                )
```
nickgerace, line 282: Right now we create audit logs outside of the `dal`, but I'm wondering if `CreateFunc` should just move to `Func::new`. Then again, some of the `new` methods are run by other `copy` or `clone` or `duplicate` methods... so the audit log doesn't always reveal the user intent...

Just rambling on your PR, but I'm thinking about making this less manual in some way. Feels like there's room for a crate in between the `dal` low level and the route authoring that could exist.

lib/sdf-server/src/service/v2/func/create_unlocked_copy.rs
```diff
@@ -54,6 +55,11 @@ pub async fn create_unlocked_copy(
     let code = get_code_response(&ctx, new_func.id).await?;
     let summary = new_func.into_frontend_type(&ctx).await?;
 
+    let variant = if let Some(schema_variant_id) = request.schema_variant_id {
+        SchemaVariant::get_by_id(&ctx, schema_variant_id).await?
+    } else {
+        None
+    };
```
nickgerace, line 62: Aren't all funcs attached to assets these days? How is this possible? (unrelated to your PR, just curious if you know)

lib/si-events-rs/src/audit_log/v1.rs
```diff
@@ -37,34 +37,151 @@ pub enum AuditLogKindV1 {
...
+    CreateFunc {
+        func_display_name: Option<String>,
+        func_kind: FuncKind,
+    },
+    CreateFuncArgument {
+        func_id: FuncId,
+        func_display_name: Option<String>,
+        func_name: String,
+        // strings for now I suppose
+        kind: String,
```
nickgerace, line 122: Why? 😄

lib/si-events-rs/src/audit_log/v1.rs
```diff
@@ -368,8 +717,26 @@ pub enum AuditLogMetadataV1 {
...
+        new_description: String,
+        old_category: String,
+        new_category: String,
+        old_link: String,
+        new_link: String,
+        old_color: String,
+        new_color: String,
+        old_component_type: String,
+        new_component_type: String,
+        //todo: what to do about the code?
```
nickgerace, line 733: What is this referring to?

lib/sdf-server/src/service/diagram/create_connection.rs
fnichol: You know, I didn't realize you could use positional numbers like `0` and `1`. Before `format!` was able to use `{variable}` style syntax, the best you could do was `format!("{}-{}", to_component_name, to_socket_name)`. Anyway, @nickgerace's idea is the most current Rustic way, agreed!

nickgerace: @fnichol year the positional numbers are interesting! I've used them in the past, moreso when I want to display something multiple times. They're cool!

fnichol (dismissed):
Not a small amount of change! Good news is that most of it is net-new.

lib/si-events-rs/src/audit_log/v1.rs
britmyerss: oh - this was from we discussed adding types to si-events that are only for audit events. I had this in there before that convo and never came back. I will update!

lib/si-events-rs/src/audit_log/v1.rs
britmyerss: Just calling out that for updating the code for a schema variant definition, I wasn't sure how to handle that and punted :)

lib/sdf-server/src/service/diagram/create_connection.rs
britmyerss: I will update this too!

lib/sdf-server/src/service/v2/func/create_unlocked_copy.rs
britmyerss: Funcs can technically be attached to only components, it's in a bit of a half state though as we don't expose it in the UI today.

lib/sdf-server/src/service/v2/func/create_func.rs
britmyerss: yeah I had similar thoughts - it's also notable we usually end up with 3 things per route, ws event(s), posthog events, and now audit events and usually the posthog and audit events are quite similar. Not sure what to do with it exactly, but it crossed my mind as I was in here.

lib/sdf-server/src/service/v2/func/binding/create_binding.rs
britmyerss: Yeah, definitely worth a pass.

lib/sdf-server/src/service/v2/func/create_func.rs
nickgerace: Yep... not this PR but it's starting to bubble over.

lib/si-events-rs/src/audit_log/v1.rs
nickgerace: Ah gotcha. I remember this. Sounds good!

lib/si-events-rs/src/audit_log/v1.rs
nickgerace: Gotcha!

nickgerace (approved):

lib/sdf-server/src/service/v2/module/contribute.rs
```diff
@@ -83,6 +93,7 @@ pub async fn contribute(
             "pkg_hash": response.latest_hash,
         }),
     );
+    ctx.commit().await?;
```
fnichol, line 96: I'm guessing we want this now because `ctx.write_audit_log().await?` will flush its events on `ctx.commit().await?`?

fnichol (approved):
All right!

lib/sdf-server/src/service/v2/module/contribute.rs
nickgerace: There was already a commit in the route! The bad GitHub diff got me too :/ took me a bit to find it

nickgerace: There's also an equivalent to "publish immediately" for audit logs to skip the rebaser _I think_. I had it at some point... can't remember if it's still there.

PR #5153, opened 2024-12-18
Fix: navbar remembers what view you're in

jobelenus:


zacharyhamm (approved):

wendybujalski (approved):
Oh, this is a clever and simple way to remember which view a user is in! Nice.

PR #5154, opened 2024-12-18
chore: disable the foyer disk cache

sprutton1:
This disables the disk cache portion of Foyer as we investigate issues with memory growing unbounded, leading to OOMs in production environments. Local testing indicates that the disk portion of the cache and the rate at which we flush to it may be causing us to greatly over allocate memory. More investigation needs to happen in this area, so this is mainly to stabilize prod.

### DIFF
--- a/lib/si-layer-cache/src/hybrid_cache.rs
+++ b/lib/si-layer-cache/src/hybrid_cache.rs
@@ -1,22 +1,16 @@
 use foyer::opentelemetry_0_26::OpenTelemetryMetricsRegistry;
-use foyer::{
-    DirectFsDeviceOptions, Engine, FifoPicker, HybridCache, HybridCacheBuilder, LargeEngineOptions,
-    RateLimitPicker, RecoverMode,
-};
-use std::cmp::max;
+use foyer::{Cache as MemCache, CacheBuilder};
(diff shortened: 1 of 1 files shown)


lib/si-layer-cache/src/hybrid_cache.rs
```diff
@@ -104,44 +78,24 @@ where
...
+                )
+                .build();
 
         Ok(Self { cache })
     }
 
     pub async fn get(&self, key: &str) -> Option<V> {
-        match self.cache.obtain(key.into()).await {
-            Ok(Some(entry)) => match entry.value() {
+        match self.cache.get(key) {
```
fnichol, line 97: If memory serves this doesn't really change the sematics, but rather a change in API because we're dealing with a `MemCache` rather than a `HybridCache`?

fnichol (approved):
Yeah I like it, pretty straight forward and a good experiment

lib/si-layer-cache/src/hybrid_cache.rs
sprutton1: Correct, obtain de-dups calls to the disk, so no disk, no obtain.

PR #5157, opened 2024-12-18
Ensure all Geometry for a View are removed when removing a View

jhelwig:
Because of the edge directionality, removing the `DiagramObject` for a `View` is insufficient to also remove all `Geometry` that represent that `DiagramObject`/`View`. To account for this, we now use the source of all `Represents` edges pointing at the `DiagramObject` to explicitly remove the `Geometry` that represent the `DiagramObject`/`View`.

Additionally, when we check to see if we would be orphaning anything by removing a `View`, if we encounter a `Geometry` that has no `Represents` edge, we ignore it and move on, as it's not representing anything to orphan.


zacharyhamm (approved):
Fixes the bug!

PR #5159, opened 2024-12-18
UI improvements to the generate panel

wendybujalski:
Screenshot 2024-12-18 at 2 47 47 PM

### DIFF
--- a/app/web/src/store/func/funcs.store.ts
+++ b/app/web/src/store/func/funcs.store.ts
@@ -955,22 +955,22 @@ export const GenerateAwsFunctionKinds = {
   AssetSchema: {
     description: "AWS Asset Schema",
     exampleCommand: {
-      command: "sqs",
-      subcommand: "create-queue",
+      service: "sqs",
+      command: "create-queue",
     },
   },
   Action: {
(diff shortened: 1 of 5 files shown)


app/web/src/store/func/funcs.store.ts
```diff
@@ -955,22 +955,22 @@ export const GenerateAwsFunctionKinds = {
   AssetSchema: {
     description: "AWS Asset Schema",
     exampleCommand: {
```
jkeiser, line 957: I don't mind this nomenclature (it sounds more correct!), but there are a lot of places we use the command/subcommand structure; we'll need to go through and fix the others. There's a type called AwsCommand, I believe, that could help fix the stuff on the AWS side.

Not necessary for this PR though, perfect being the enemy of the good and all that :)

jkeiser (approved):
Yay flex-grow!

app/web/src/store/func/funcs.store.ts
stack72: Service and command *are* the correct terminolgy here - it's what AWS specifies. The internals of command and subcommand should not leak out

PR #5161, opened 2024-12-18
fix: retrieve from memory when waiting for snapshots

sprutton1:
The original implementation of these calls went directly to the memory cache, knowing that we were waiting for something new and it couldn't be on the disk yet. When foyer was added, this changed to go directly to the cache, meaning that we are knowingly spawning a thread that reads from disk with every attempt here, every millisecond, for up to 2000 milliseconds. This creates a TON of disk pressure and additional allocations. I _think_ this might be part of the source of our memory pressure issues, but this really leads me to believe that there is a leak internal to foyer around how they load things from the disk (maybe specifically around misses?).

Another curious thing is that I don't seem to have any traces that point to the disk misses, of which there should be many thousand.

Not that this undoes what was done earlier to move us to a just-memory cache. I would like to see how this performs in tools prod overnight at least to see if it has a significant impact on the memory allocations.

### DIFF
--- a/lib/si-layer-cache/src/db/rebase_batch.rs
+++ b/lib/si-layer-cache/src/db/rebase_batch.rs
@@ -102,7 +102,7 @@ where
         let mut tried = 0;
         let read_wait = Instant::now();
         while tried < MAX_TRIES {
-            if let Some(v) = self.cache.cache().get(&key).await {
+            if let Some(v) = self.cache.cache().get_from_memory(key.clone()).await {
                 span.record("si.layer_cache.memory_cache.hit", true);
                 span.record(
                     "si.layer_cache.memory_cache.read_wait_ms",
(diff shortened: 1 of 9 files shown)


zacharyhamm (approved):
The changes to the read wait are great imo. I like using the `&str` better, but it's not a blocker.

sprutton1:
> The changes to the read wait are great imo. I like using the `&str` better, but it's not a blocker.

@zacharyhamm 
Just a style preference or something else?

zacharyhamm:
> > The changes to the read wait are great imo. I like using the `&str` better, but it's not a blocker.
> 
> 
> 
> @zacharyhamm 
> 
> Just a style preference or something else?
> 
> 
> 
> 

&str means you can pass a String as easily as a &str or an Arc<str>, and means you don't have to do a ref count operation in the read loop

PR #5162, opened 2024-12-18
button on template generated toast to view func code

wendybujalski:
Screenshot 2024-12-18 at 5 04 45 PM

### DIFF
--- a/app/web/src/components/ModelingView/TemplateSelectionModal.vue
+++ b/app/web/src/components/ModelingView/TemplateSelectionModal.vue
@@ -86,6 +86,7 @@ import {
 } from "@si/vue-lib/design-system";
 import { computed, onBeforeUnmount, onMounted, ref } from "vue";
 import { storeToRefs } from "pinia";
+import { useRouter } from "vue-router";
 import { useComponentsStore } from "@/store/components.store";
 import { useViewsStore } from "@/store/views.store";
 import { useFeatureFlagsStore } from "@/store/feature_flags.store";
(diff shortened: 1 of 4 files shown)


zacharyhamm:
@wendybujalski I'll just make a commit to this PR, then someone else can review it and we'll merge it

zacharyhamm (dismissed):

app/web/src/components/ModelingView/TemplateSelectionModal.vue
```diff
@@ -169,7 +171,7 @@ const onCreateTemplate = () => {
     category: category.value,
   };
 
-  componentsStore.CREATE_TEMPLATE_FUNC_FROM_COMPONENTS(templateData);
+  componentsStore.CREATE_TEMPLATE_FUNC_FROM_COMPONENTS(templateData, router);
```
jobelenus, line 174: is there a reason to pass the router, rather than import the router in component store?

wendybujalski: For some reason, we were not able to get the router imported successfully within the component store. It would only result in "undefined" no matter where we imported it within the store. Same issue within the toast itself. This workaround avoids that problem.

zacharyhamm: useRouter does not work inside of the toast component, something about they way they are mounted

zacharyhamm: also does not work inside stores!

jobelenus: Ohhhhh, how interesting... because we definitely do router.push in view store for example?? Or am I losing my mind? Highly likely

jobelenus (approved):

PR #5165, opened 2024-12-18
Fix: load all components at the right time

jobelenus:
After first paint, but before the data is required by other things! And don't do it twice!

Tested this against all the regressions we've been bashing away on (e.g. no blank modeling screen, no missing component on apply)


zacharyhamm:
Taking a look! But note the failed lints

zacharyhamm (approved):

PR #5168, opened 2024-12-19
BUG-652: send inferred edges along with created component

jobelenus:
Thank you Zack!

### DIFF
--- a/lib/dal/src/component.rs
+++ b/lib/dal/src/component.rs
@@ -3646,6 +3646,7 @@ impl Component {
 #[serde(rename_all = "camelCase")]
 pub struct ComponentCreatedPayload {
     pub component: DiagramComponentView,
+    pub inferred_edges: Option<Vec<SummaryDiagramInferredEdge>>,
     change_set_id: ChangeSetId,
 }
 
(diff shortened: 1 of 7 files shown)


zacharyhamm (dismissed):

lib/dal/src/component/frame.rs
```diff
@@ -159,31 +159,28 @@ impl Frame {
     }
 
     /// Provides the ability to attach or replace a child [`Component`]'s parent
-    #[instrument(level = "info", skip(ctx))]
+    //#[instrument(level = "info", skip(ctx))]
```
zacharyhamm, line 162: Let's restore this instrumentation

lib/dal/src/component.rs
```diff
@@ -3852,6 +3853,23 @@ impl WsEvent {
...
+    pub async fn component_created_with_inferred_edges(
+        ctx: &DalContext,
+        component: DiagramComponentView,
+        inferred_edges: Option<Vec<SummaryDiagramInferredEdge>>,
+    ) -> WsEventResult<Self> {
+        WsEvent::new(
+            ctx,
+            WsPayload::ComponentCreated(ComponentCreatedPayload {
+                change_set_id: ctx.change_set_id(),
+                inferred_edges: inferred_edges,
```
zacharyhamm, line 3872: clippy might ding you here, since you can just do `inferred_edges` like in JS

zacharyhamm (approved):
🚀

PR #5169, opened 2024-12-19
feat: auto select components created by mgmt fns

zacharyhamm:
If you create a bunch of components via a management function, (say, via a template func), we will now replace the selection with them when the components are created.


app/web/src/store/func/funcs.store.ts
```diff
@@ -318,6 +322,9 @@ export const useFuncStore = () => {
               { componentId },
               { viewId },
             ]),
+            optimistic: (requestUlid) => {
+              this.pendingManagementRunRequestUlid = requestUlid;
```
jobelenus, line 326: what happens if there are multiple in flight requests?

app/web/src/store/func/funcs.store.ts
```diff
@@ -940,6 +947,39 @@ export const useFuncStore = () => {
...
+              { requestUlid, status, message, createdComponentIds, funcName },
+              _metadata,
+            ) => {
+              if (
+                requestUlid &&
+                requestUlid === this.pendingManagementRunRequestUlid &&
+                createdComponentIds?.length
+              ) {
+                const componentsStore = useComponentsStore();
+                componentsStore.eventBus.emit(
```
jobelenus, line 962: i dont think we need all this event bus nonsense... the `setSelectedComponentIds` fn is now in the view store...

app/web/src/store/func/funcs.store.ts
zacharyhamm: Last one wins. If the order is bonked, where the latest execution's event arrives before the previous execution's event (because of thread scheduling or because one operator is a lot more complex than another), then the selection just won't happen. I think that's ok, since I'm not sure what the ideal solution is there: we don't want it to flap between selections if a bunch arrive at once.

jobelenus (dismissed):

jobelenus:
🚀

jobelenus (dismissed):

wendybujalski (dismissed):

wendybujalski (approved):

PR #5170, opened 2024-12-19
Allow removal of Views contining Components also in other Views when applying a set of Updates

jhelwig:
The previous version of the `CorrectTransorms` for a `ViewNodeWeight` was incorrectly determining that a `View` should not actually be deleted if it still had any `Geometry` for `Component`s, regardless of whether those `Component`s also had any `Geometry` in other `View`s. This meant that while the `SDF` API endpoint would succeed, the `View` was never removed when the `Update`s were applied by the rebaser.

When operating within a `ChangeSet`, we do allow removing a `View` while it still has `Geometry` for `Component`s, as long as the `View` being removed is not the only one that has `Geometry` for any of those `Component`s. This change mirrors that logic in `CorrectTransforms` to allow the generated `Updates` sent to the `rebaser` to apply with the `View` removal intact.

### DIFF
--- a/lib/dal/src/workspace_snapshot/graph/v4/component.rs
+++ b/lib/dal/src/workspace_snapshot/graph/v4/component.rs
@@ -1,9 +1,11 @@
 use petgraph::prelude::*;
+use si_id::ViewId;
 
 use crate::{
     component::ComponentResult,
     workspace_snapshot::{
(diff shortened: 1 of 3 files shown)


jhelwig:
Confirmed that the new test fails in the way expected without the change to the `CorrectTransforms`, and passes with those changes in place.

zacharyhamm (approved):
works great

PR #5173, opened 2024-12-19
BUG-725: remove selections from head prior to applying.

jobelenus:
1. A race is happening with the router where the change set applied pushes to head
2. That push brings the head stores back
3. The head stores attempt to sync selected data back into the URL
4. But the route URL has not been set to "head"
5. So the head stores sync overwrite the push to head
6. And it pushes you to the changeset you were on previously
7. Which is merged, and causes tons of breakage


wendybujalski (approved):

fnichol (approved):
👍🏽

PR #5175, opened 2024-12-20
Fix: dont make parents smaller, also fixed the bug why we thought new children didnt fit inside the parent

jobelenus:


fnichol (approved):
Very nice, I don't believe I'm seeing the bug behavior anymore with this change.

PR #5176, opened 2024-12-20
fix(dal): The order in which we process dependencies between components is inconsistent, so we must always check if the component is in the dependency graph before adding it as Petgraph doesn't prevent nodes with duplicate node weights from being inserted

britmyerss:
In order to calculate the dependencies between actions, we must account for dependencies between components based on inferred connections and connections.  The issue was, if a component (A) with an enqueued action is a dependency for another component  (B) with an enqueued action, and we processed component A first, we'd end up inserting Component B into the graph, then inserting it again as it has an action enqueued. This would make the edges between the Actions all screwed up depending on what other components are in the mix.  Because Petgraph doesn't guarantee the order in which nodes are retrieved from the graph, this would sometimes be just fine, and other times, be incorrect. Fun!

<div><br/>via <a href="https://giphy.com/gifs/cat-memory-NusOH30J7QiJy">GIPHY</a></div>

### DIFF
--- a/lib/dal/src/action/dependency_graph.rs
+++ b/lib/dal/src/action/dependency_graph.rs
@@ -86,8 +86,10 @@ impl ActionDependencyGraph {
         // things they are feeding data into.
         for component_id in actions_by_component_id.keys().copied() {
             let component = Component::get_by_id(ctx, component_id).await?;
-            let component_index = component_dependencies.add_node(component_id);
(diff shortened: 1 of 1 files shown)


fnichol (approved):
We reviewed and ran the system locally (creation and deletion) to confirm it's behaving as we expect.

PR #5184, opened 2024-12-20
Fix: apply move to head from a view that doesnt exist there

jobelenus:


app/web/src/store/change_sets.store.ts
```diff
@@ -603,6 +604,21 @@ export function useChangeSetsStore() {
                 const route = useRouterStore().currentRoute;
 
                 if (route?.name) {
+                  // am i on a view, and about to move to head
+                  // where that view does not yet exist
+                  if ("viewId" in route.params) {
```
zacharyhamm, line 609: Should this be generic, for switching between any change set? Or does it only matter when going to head

jobelenus: good point!

jobelenus: this spot is only about moving to head after apply... i will add the changeset dropdown

PR #5185, opened 2024-12-20
Fix: creating new View fixes, and two Scenario 1 fixes

jobelenus:
1. make a changeset
2. create a new view
3. be looking at your new view
4. hit apply
5. you get this error
image

We also had a subtler 404 bug in step 2 when creating a view that is fixed.

We also fixed:
- sockets did not draw connections correctly after a component was changed into a frame
- when selecting a frame that had "to be deleted" child frames in it the children were "behind" the selected frame, not on top


fnichol (approved):
We ran this branch through "scenario 1" to confirm the fixes

PR #5186, opened 2024-12-20
feat: replace vm2 with deno and web workers

sprutton1:
R.I.P. vm2

Okay, there's a lot here, but the premise is:
* remove vm2 (deprecated)
* use deno to allow for dynamic imports
* construct a web worker to allow dynamic function execution with the appropriate scope

The gory details:
* [Add buck2 targets for deno](https://github.com/systeminit/si/pull/5186/commits/ea5fb252b8505de682207adaf062e126df1d5370)
  * Targets for compiling, formatting/linting, and compiling
* [Create deno version of ts-lib](https://github.com/systeminit/si/pull/5186/commits/0e3f4023a543f7c1569e4faa09d3a1768daa98d9)
  * deno really wants imports to end with `.ts` and the existing bits of the stack seem to hate it. This create another library with a different `index.ts` and symlinks for the actual source code. I hate it, but I couldn't come up with anything better
* [migrate from vm2 to deno and web workers](https://github.com/systeminit/si/pull/5186/commits/fe663b7f327dd9f68c7f438f16197a6b52aace83)
  * Replace the pnpm buck2 targets with deno targets
  * Replace tsvfs transpilation with deno_emit (deno compile would not play nice otherwise)
  * Add a web worker that takes user code and constructs it into a runnable function... and then runs it. This includes constructing the sandbox, managing the timeout, and bubbling up errors
  * Pass user code to the worker and do stuff with the response. Note that we can manage deno permissions at this level, so if we want to lockdown certain URLs or filepaths, we can do it here
  * Create a new custom function in the sandbox that can be used to import packages dynamically

The remaining commits are tests/lints/chores.

In order to actually do an import in an asset:

```typescript
    const imported = await pkg.importPackage("node:crypto");
    if (imported.error) {
      return {
        result: 'failure',
        message: `Failed to import: ${imported.error}`
      }
    }
    const crypto = imported.value;
    console.log(crypto.randomBytes(8))
```

The biggest thing missing from here in my opinion is that imported types are not included in the code editor, so you don't get any hinting on import. I think this may be a little tricky, but maybe possible if editor can parse and import in the background.


bin/lang-js/src/debug.ts
```diff
@@ -1,13 +1,46 @@
...
   const debugActive = process.env.SI_LANG_JS_LOG || process.env.SI_LOG;
-
-  return (msg: any) => {
+  return (msg: unknown) => {
     if (debugActive) {
-      const pretty_json = JSON.stringify(msg, null, 2);
-      for (const line of pretty_json.split("\n")) {
-        process.stderr.write(`${namespace} ${line}\n`);
+      try {
+        const safeStringify = (obj: unknown): string => {
```
jobelenus, line 10: :heart_hands::skin-tone-3: that WeakSet, it looks like the safeStringify method could be a private fn all by itself there, rather than re-created with every Debug() call?

bin/lang-js/src/function_kinds/management.ts
```diff
@@ -17,19 +18,19 @@ export interface ManagementFunc extends Func {
...
-  from: string,
+  from: string;
   to: {
     component: string;
     socket: string;
-  }
+  };
 }
 
 export interface ManagementCreate {
```
jobelenus, line 36: looks like this is no longer used and can be deleted?

bin/lang-js/src/function_kinds/resolver_function.ts
```diff
@@ -62,29 +63,39 @@ export interface TypeCheckSuccess {
...
 
-const isBoolean = (value: unknown): TypeCheckResult => (_.isBoolean(value)
+const isBoolean = (
+  value: unknown,
+): TypeCheckResult => (_.isBoolean(value)
   ? { valid: true }
   : { valid: false, message: "Return type must be a boolean." });
 
-const isInteger = (value: unknown): TypeCheckResult => (_.isInteger(value)
+const isInteger = (
```
jobelenus, line 78: it doesn't look like these 3 helper functions are used? below it just uses the lodash fns

bin/lang-js/src/function_kinds/resolver_function.ts
```diff
@@ -127,7 +138,7 @@ const isQualification = (value: unknown): TypeCheckResult => {
     };
   }
 
-  if (!qualificationStatuses.includes(value.result)) {
+  if (!qualificationStatuses.includes(value.result as string)) {
```
jobelenus, line 141: I do wonder if we can use an else statement, so TS "knows"/proves that `value.result` is a string, without casting it.. but NBD.. just trying to get the most of out TS checks

bin/lang-js/src/function_kinds/management.ts
sprutton1: ah, a bad rebase on my part I think

bin/lang-js/src/function_kinds/resolver_function.ts
sprutton1: If I understand correctly, they are used by`typeChecks`  [here](https://github.com/systeminit/si/blob/main/bin/lang-js/src/function_kinds/resolver_function.ts#L223). I didn't actually change this bit, just formatting.

bin/lang-js/BUCK
```diff
@@ -1,113 +1,84 @@
...
-    dev_deps_srcs = dev_deps_srcs,
+export_file(
+    name = "deno.json",
 )
 
-typescript_runnable_dist(
+alias(
     name = "lang-js",
+    actual = ":bin"
 )
```
fnichol, line 25: This works with the `alias` rule, but I believe you could also drop this `alias` rule and rename the `deno_compile` below to have `name = "lang-js",` and then the `bin/lang-js:lang-js target would be the `deno_compile` rule directly. If memory serves I had an `alias` rule in because of all the dist/runnable_dist/whatever stack of rules to make the prior version work.

bin/lang-js/BUCK
```diff
@@ -1,113 +1,84 @@
...
+    ]) +
+    ["//lib/ts-lib-deno:ts-lib-deno"],
+    permissions = [
+        "allow-all",
+    ],
+    unstable_flags = [
+        "worker-options",
+    ],
+    visibility = ["PUBLIC"],
 )
```
fnichol, line 43: That's a nice looking rule!

bin/lang-js/BUCK
```diff
@@ -1,113 +1,84 @@
...
-typescript_check(
-    name = "check-type",
-    srcs = [":src"],
-    prod_deps_srcs = prod_deps_srcs,
-    dev_deps_srcs = dev_deps_srcs,
+deno_format(
+    name = "check-format",
+    srcs = glob(["**/*.ts", "**/*.js"]),
+    check = True,
 )
```
fnichol, line 55: Eventually I think we could push the default declarations of these rules into the `deno_compile` *macro* definition, similar to our `rust_binary`/`rust_library` rules (they help to define format checks/fixes, unit tests, etc). That can be a later refactoring though (and would be more useful for our second/third Deno sub-project).

bin/lang-js/BUCK
```diff
@@ -1,113 +1,84 @@
...
-    name = "docs",
-    command = "docs",
-    srcs = glob(["src/**/*"]),
-    path = "bin/lang-js",
-    deps = [
-        "//:node_modules",
-    ],
-    visibility = ["PUBLIC"],
+    srcs = {
+        "//:deno.json": ".",
```
fnichol, line 79: I was curious so see that `//:deno.lock` wasn't in the root `BUCK` file, so I presume that the lock file isn't needed to build?

bin/lang-js/src/index.ts
```diff
@@ -1,22 +1,21 @@
 #!/usr/bin/env node
 
-import fs from "fs";
-import { Command } from "commander";
-import { makeConsole } from "./sandbox/console";
-import { Request } from "./request";
+import process from "node:process";
+import { Command } from "jsr:@cliffy/command@^1.0.0-rc.7";
```
fnichol, line 4: Nice! Wow, there's so much API similarities between these 2

bin/lang-js/src/index.ts
```diff
@@ -69,7 +66,11 @@ async function main() {
   let errorFn = makeConsole(executionId).error;
 
   try {
-    const requestJson = fs.readFileSync(STDIN_FD, "utf8");
+    const decoder = new TextDecoder();
+    let requestJson = "";
+    for await (const chunk of Deno.stdin.readable) {
+      requestJson += decoder.decode(chunk);
+    }
```
fnichol, line 73: Cool, streaming input too 😉

bin/lang-js/src/base64.ts
```diff
@@ -1,36 +1,5 @@
...
-    system,
-    Array.from(fsMap.keys()),
-    ts,
-    compilerOptions,
-  );
-  return (
-    tsEnv.languageService.getEmitOutput("index.ts", false, true).outputFiles[0]
-      ?.text ?? ""
-  );
-}
```
fnichol, line 36: Hee hee, I kept forgetting the TypeScript compilation was added in here, always a surprise!

bin/lang-js/tests/requestStorage.spec.ts
```diff
@@ -1,34 +1,35 @@
...
+
+const KEY = "a";
+const VALUE = "this is a test value";
+const EXEC_ID = "execId";
+
+function setupStorage() {
+  makeBeforeRequestStorage(EXEC_ID).setItem(KEY, VALUE);
+}
+
+Deno.test("requestStorage", async (t) => {
```
fnichol, line 15: Cool!

bin/lang-js/src/sandbox/exec.ts
```diff
@@ -34,10 +34,14 @@ const defaultOptions: Options = {
...
  */
 function mergedOptions(userOptions?: Options): Options {
+  // we do some LD_LIBRARY_PATH tomfoolery to ensure a deno compile'd bin works
+  // in alpine, but this breaks sub-processes, so we need to unset it here.
+  // See flake.nix and rootfs_build.sh for more details.
   return {
     ...defaultOptions,
     ...userOptions,
     stdin: userOptions?.input ? "pipe" : "ignore",
+    env: { LD_LIBRARY_PATH: "" },
```
fnichol, line 44: Oh....this must have been *painful* to discover. And it does make sense that we wouldn't want to propagate the fix for `lang-js` itself beyond it's own process since it's a runtime envvar.

bin/lang-js/src/worker.js
```diff
@@ -1,5 +1,8 @@
 import { createSandbox } from "./sandbox.ts";
 import { rawStorage } from "./sandbox/requestStorage.ts";
+import { Debug } from "./debug.ts";
+
+const debug = Debug("langJs:worker");
```
fnichol, line 5: That is sweet!

flake.nix
```diff
@@ -220,7 +220,27 @@
...
+            extraInstallPhase = ''
+              wrapProgram $out/bin/lang-js \
+                --set LD_LIBRARY_PATH "/usr/local/lib" \
+                --set SI_LANG_JS_LOG "debug" \
+                --prefix PATH : ${pkgs.lib.makeBinPath langJsExtraPkgs} \
+                --run 'cd "$(dirname "$0")"'
+                # ^ deno falls over trying to resolve libraries if you don't set
+                # the working path
+            '';
+          };
```
fnichol, line 250: I'm going to guess this fixed the path to, say glibc (like libc.so but also the static `ld` linker)? The `wrapProgram` is a very nice way to solve this and something I kind of forgot about in the last iteration (I may have been able to create a program wrapper as the entry point for the lang-js "program").

prelude-si/rootfs/rootfs_build.sh
```diff
@@ -116,6 +117,11 @@ for dir in / run etc usr/local/etc home/app/.config; do
     mkdir -pv "/\$dir/$BIN"
 done
 
+# for glibc, added later
+mkdir -p /usr/local/lib
+mkdir -p /lib64
+ln -sf /usr/local/lib/ld-linux-x86-64.so.2 /lib64/
```
fnichol: Ok, this does partially answer my question above!

prelude-si/rootfs/rootfs_build.sh
```diff
@@ -178,6 +184,13 @@ EOZ
 EOL
 # LEAVE CHROOT
 
+# See https://github.com/denoland/deno_docker/blob/main/alpine.dockerfile for
+# the inspiration here. We get the glibc bits directly from distroless/cc
```
fnichol: Oh, I see, this build/linking issue is related to Deno specifically (at least in part)! It was these kinds of challenges that had me drop the use of the `pkg` npm package--I just could not get the library paths to work correctly.

flake.nix
```diff
@@ -221,19 +221,26 @@
...
+          # known place. Note that LD_LIBRAY_PATH is unset when using siExec to
+          # ensure the binaries our binaries run don't inherit it.
           lang-js = binDerivation {
             pkgName = "lang-js";
             dontAutoPatchELF = true;
             dontStrip = true;
             extraInstallPhase = ''
+              # since we can't patchelf, we need to ensure the dynamic linker
+              # is where we expect it. This gets droppped into the rootfs as
+              # /lib64/ld-linux-x86-64.so.2 -> /nix/store/*/ld-linux-x86-64.20.2
```
fnichol, line 235: Everything's built "hermetically" until it turns out that it doesn't portable run, eh? Pragmatism to the rescue, I love it

fnichol (approved):
Well done, I can tell this took serious effort in a few dimensions. I only found one script that maybe could be dropped and nothing else that would block this merging.

Awesome work!

prelude-si/deno/deno_install.py
```diff
@@ -0,0 +1,95 @@
+#!/usr/bin/env python3
```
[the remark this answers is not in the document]
sprutton1: Yep, just a straggler

flake.nix
sprutton1: Yah, we need to know:

1. Where is my glibc stuff since I'm in alpine?
2. Where is the linker (and since we can't patchelf, it's in `/lib64/` by default, hence the linking)

PR #5188, opened 2024-12-20
fix(sdf): Ensure we match intrinsics schema variants as well as components

stack72:


fnichol (approved):
We confirmed the corrected behavior locally, and as this code is exercised in a function authoring context (i.e. not modelling) running the scenario 1flow shouldn't call this change

fnichol:
I ran through our "scenario 1" anyway to confirm it behaves as expected

PR #5203, opened 2025-01-03
feat: Add sdf whoami endpoint accessible to user or automation tokens

jkeiser:
This adds a "whoami" endpoint that you can hit to check your token and email (who SI thinks you are). It mainly exists to validate that we can actually authorize automation tokens for an endpoint (since all other endpoints reject automation tokens).

## Extractor Funnel

This also factors the token and authorization extractors so they always use the same code. web/automation authorization still work via extractors, they just all run through a single funnel now.

`EndpointAuthorization` is the main authorization extractor you will use (or `AccessBuilder`, which uses it), which:
  1. Extracts the access token from the `Authorization: ` header using the RawAccessToken extractor.
  2. Validates the access token and checks expiration date using the ValidatedToken extractor.
  3. Checks that the user is a member of the workspace using the WorkspaceMember extractor.
  4. Checks that the token gives the user "web" permissions using the AuthorizedRole extractor.
  5. Returns the workspace_id, user record, and the role that was authorized.

### Customization For Different Endpoints

The big reason to break this up was to allow different authorization for different endpoints, as well as to support WS endpoints using the same extractors. The `AuthorizedRole` and `RawAccessToken` extractors check if they already have a value before doing their default behavior, so you can invoke other extractors that set the value first. To wit:
- The whoami endpoint doesn't need maximal permissions (the web role), so it adds the `_: AuthorizeForAutomationRole` extractor to its endpoint function, causing endpoint validation to check for the automation role instead.
- WS endpoints add the `_: TokenFromQueryParam` extractor to the endpoint function, which retrieves the token from the "token" query parameter instead of the Authorization header.

Where appropriate, these endpoints cache their results so that if you use multiple extractors (which we do frequently) you don't pay a double or triple database access or public key decryption penalty.

This also adds `Deref` and `Into` to a lot of the extractors, which means instead of saying things like `Nats(nats): Nats` you can just say `nats: Nats` and still use the nats object like normal.

## Future

Going forward, we'll want to take some of this (RawAccessToken and ValidatedToken in particular) and move it to a common library (quite possibly repurpose `si-jwt-public-key` for it) so that the module index can use them. I started to do that, but there's a minor snarl in that it needs to be able to access the public key we grab as part of `AppState` (which isn't shared between module index and sdf). Will get back to that at a later time.

### DIFF
--- a/lib/sdf-server/src/middleware/workspace_permission.rs
+++ b/lib/sdf-server/src/middleware/workspace_permission.rs
@@ -8,11 +8,10 @@ use axum::{
 };
 use futures::future::BoxFuture;
 use permissions::{Permission, PermissionBuilder};
-use si_jwt_public_key::SiJwtClaimRole;
 use tower::{Layer, Service};
 
 use crate::{
-    extract::{self, Authorization},
+    extract::{self, EndpointAuthorization},
     AppState,
 };
 
@@ -66,24 +65,24 @@ where
         Box::pin(async move {
             let (mut parts, body) = req.into_parts();
 
-            let Authorization(claim) =
-                match Authorization::from_request_parts(&mut parts, &me.state).await {
-                    Ok(claim) => claim,
-                    Err(err) => return Ok(err.into_response()),
-                };
+            let auth = match EndpointAuthorization::from_request_parts(&mut parts, &me.state).await
+            {
+                Ok(auth) => auth,
+                Err(err) => return Ok(err.into_response()),
+            };
 
             if let Some(client) = me.state.spicedb_client() {
                 let is_allowed = match PermissionBuilder::new()
-                    .workspace_object(claim.workspace_id())
+                    .workspace_object(auth.workspace_id)
                     .permission(me.permission)
-                    .user_subject(claim.user_id())
+                    .user_subject(auth.user.pk())
                     .has_permission(client)
                     .await
                 {
                     Ok(is_allowed) => is_allowed,
                     Err(e) => return Ok(extract::unauthorized_error(e).into_response()),
                 };
-                if !is_allowed || !claim.authorized_for(SiJwtClaimRole::Web) {
+                if !is_allowed {
                     return Ok(
                         extract::unauthorized_error("not authorized for web role").into_response()
                     );
--- a/lib/si-jwt-public-key/src/lib.rs
+++ b/lib/si-jwt-public-key/src/lib.rs
@@ -86,6 +86,16 @@ pub enum SiJwtClaimRole {
     Automation,
 }
 
+impl SiJwtClaimRole {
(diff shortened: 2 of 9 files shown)


lib/sdf-server/src/middleware/workspace_permission.rs
```diff
@@ -66,24 +65,24 @@ where
...
-                    .user_subject(claim.user_id())
+                    .user_subject(auth.user.pk())
                     .has_permission(client)
                     .await
                 {
                     Ok(is_allowed) => is_allowed,
                     Err(e) => return Ok(extract::unauthorized_error(e).into_response()),
                 };
-                if !is_allowed || !claim.authorized_for(SiJwtClaimRole::Web) {
+                if !is_allowed {
```
zacharyhamm, line 85: I think I'm missing something. Why was the authorized_for call dropped here?

zacharyhamm (commented):
A few ?s

lib/si-jwt-public-key/src/lib.rs
```diff
@@ -130,7 +133,7 @@ impl SiJwtClaims {
             Self::V2(SiJwtClaimsV2 { role, .. }) => *role,
             Self::V1(SiJwtClaimsV1 { .. }) => SiJwtClaimRole::Web,
         };
-        role == required_role
+        role <= required_role
```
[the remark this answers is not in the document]
jkeiser: An is_subset_of() or subsumes() or something is a good idea, I'll do that. That's what we're getting at here.

lib/sdf-server/src/middleware/workspace_permission.rs
jkeiser: Because it's already authorized, which you can see above in `EndpointAuthorization` (if you get that object, you know the user is a member of the workspace, and you know the role has been checked).

lib/si-jwt-public-key/src/lib.rs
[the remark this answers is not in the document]
jkeiser: Fixed!

lib/sdf-server/src/middleware/workspace_permission.rs
jkeiser: (This was already the case, it turns out, btw--`Authorization` was also checking that you were authorized for the web role.)

zacharyhamm (approved):
Tested this by replacing my local workspace token with one for another user (my gmail account) but for the same "workspace id" (without modifying the token, just the key on the data bag in the local storage). This seems good to me, although for something this sensitive another set of eyes might be worthwhile

jkeiser:
I think I will wait to merge our auth stack until it's NOT 5PM on Friday. :)

PR #5204, opened 2025-01-03
feat: mgmt fns can create views, create components in views

zacharyhamm:
The return type for management functions has been extended to support creating views, and creating new components in views other than the current view.

The "geometry" property on a create can now *either* be a plain geometry object, which will create the component in the current view, or it can be a map of view names to geometries, which will create the component in the specified view with that geometry.

Additionally, you can add to your ops return an object like so:

```
ops: {
  views: {
    create: ["view a", "view b", "view c"]
  }
}
```

to create new views in a mangement function. The views will be available for create and update operations in the management func itself.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1939,6 +1939,7 @@ dependencies = [
  "si-data-nats",
  "si-data-pg",
  "si-events",
+ "si-id",
  "si-jwt-public-key",
  "si-layer-cache",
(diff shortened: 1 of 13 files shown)


stack72 (approved):
this is great! :)

nickgerace (approved):

jobelenus (approved):
🚀

PR #5208, opened 2025-01-06
Feat: Make automation token UI conform to figma

vbustamante:
image

image

### DIFF
--- a/app/auth-portal/src/pages/WorkspaceAuthTokensPage.vue
+++ b/app/auth-portal/src/pages/WorkspaceAuthTokensPage.vue
@@ -13,96 +13,65 @@
             :to="{ name: 'workspace-settings', params: { workspaceId } }"
           >
             <Icon
+              class="flex-none"
+              iconBgActiveTone="action"
+              iconIdleTone="shade"
+              iconTone="warning"
               name="settings"
+              size="lg"
               tooltip="Settings"
               tooltipPlacement="top"
-              size="lg"
-              class="flex-none"
-              iconTone="warning"
-              iconIdleTone="shade"
-              iconBgActiveTone="action"
             />
           </RouterLink>
         </WorkspacePageHeader>
 
         <Stack>
           <ErrorMessage :asyncState="createAuthToken" />
-          <div class="flex flex-row flex-wrap items-end gap-xs">
+          <div
+            v-if="workspace.role === 'OWNER'"
+            class="flex flex-row flex-wrap items-end gap-md"
+          >
             <VormInput
               v-model="createAuthTokenName"
-              :disabled="workspace.role !== 'OWNER'"
-              required
-              label="Token Name"
-              defaultValue="My API Token"
+              inlineLabel
+              label="Token Name*"
+              @keydown.enter.prevent="createAuthToken.execute()"
             />
             <VButton
-              :disabled="workspace.role !== 'OWNER'"
+              :disabled="!(createAuthTokenName?.length > 0)"
               :loading="createAuthToken.isLoading.value"
               loadingText="Creating ..."
-              iconRight="chevron--down"
               tone="action"
               variant="solid"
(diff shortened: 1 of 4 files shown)


app/auth-portal/src/pages/WorkspaceAuthTokensPage.vue
```diff
@@ -13,103 +13,111 @@
...
+                    @revoked="delete authTokens.state.value[authToken.id]"
                   />
                 </tbody>
               </table>
             </Stack>
           </div>
         </Stack>
       </div>
+      <Modal ref="tokenDisplayModalRef" size="lg" title="Token Generated">
+        <ErrorMessage
```
jkeiser, line 83: Nice touch making it red until you copy!

... though is there a `SuccessMessage` or something one could use for the happy case?

app/auth-portal/src/pages/WorkspaceAuthTokensPage.vue
```diff
@@ -124,14 +132,19 @@ import {
...
 } from "@si/vue-lib/design-system";
 import { useAsyncState } from "@vueuse/core";
 import { apiData } from "@si/vue-lib/pinia";
+import { useHead } from "@vueuse/head";
 import { useWorkspacesStore, WorkspaceId } from "@/store/workspaces.store";
 import WorkspacePageHeader from "@/components/WorkspacePageHeader.vue";
 import { useAuthTokensApi } from "@/store/authTokens.store";
 import AuthTokenListItem from "@/components/AuthTokenListItem.vue";
 
+useHead({ title: "Automation Tokens" });
```
jkeiser, line 146: Should I be doing this on all pages?

stack72 (approved):

app/auth-portal/src/pages/WorkspaceAuthTokensPage.vue
```diff
@@ -175,8 +199,14 @@ const createAuthToken = useAsyncState(
...
+const tokenDisplayModalRef = ref<InstanceType<typeof Modal> | null>(null);
+
+const tokenCopied = ref(false);
 async function copyToken() {
-  if (createAuthToken.state.value)
-    await navigator.clipboard.writeText(createAuthToken.state.value);
+  if (createAuthToken.state.value) {
+    await navigator.clipboard?.writeText(createAuthToken.state.value);
+  }
+  tokenCopied.value = true;
```
jkeiser, line 210: It seems unlikely to trigger as a bug, but this should probably be inside that `if`

jkeiser (approved):

app/auth-portal/src/pages/WorkspaceAuthTokensPage.vue
```diff
@@ -175,8 +199,14 @@ const createAuthToken = useAsyncState(
 /** Name of token to create */
 const createAuthTokenName = ref("");
 
+/** Token modal */
+const tokenDisplayModalRef = ref<InstanceType<typeof Modal> | null>(null);
```
jkeiser, line 203:
```suggestion
const tokenDisplayModalRef = ref<InstanceType<typeof Modal>>();
```

Other places in the codebase do this too. It's DEFINITELY just a nit.

app/auth-portal/src/pages/WorkspaceAuthTokensPage.vue
vbustamante: I think the move here would be to have make this into a statusMessage or messageBox component eventually, since we don't have anything else

app/auth-portal/src/pages/WorkspaceAuthTokensPage.vue
vbustamante: @jkeiser This can be done on the router too, but on the auth portal we've been using this pattern. I'd say useHead on the auth portal, the router title on the webapp

PR #5213, opened 2025-01-07
chore: Add auth-api and auth-portal to Tilt

jkeiser:
This adds auth-api and auth-portal to Tilt in manual mode, so that you can host them as part of your local stack. They must be started manually, similar to module-index.

### Later work

You still have to initialize the database and set the proper env variables; and it requires you to enable the services soon after starting tilt (since the database creation gets angry if there are other services connected to postgres). We'll take a look at these papercuts in the future.

image


fnichol (approved):
Tiltfile changes look good after chatting!

PR #5215, opened 2025-01-07
feat: delete components from management functions

zacharyhamm:
Components managed by a management component can now be deleted in a management func. Two new properties have been added to the ops return value: `delete?: string[]` and `erase?: string[]`. Delete will mark components with resources as to_delete, while erase will delete components regardless of their resource status. Components to delete or erase can be specified either by their name, or their id (available on the components input array). id will be more reliable if there is more than one managed component with the same name.


zacharyhamm:
This PR refactors deletion so that the management function and the delete components endpoint use mostly the same code. To test it one should delete and erase a combination of components on head, not on head, with and without resources to see if they delete properly

jobelenus (approved):

PR #5223, opened 2025-01-08
feat: remove views and components from views in mgmt fns

zacharyhamm:
Adds the ability to remove a view with a management func, and to remove a component from a view with a management func.


stack72 (approved):
This is great :)

PR #5231, opened 2025-01-10
fix(dal): When creating props, include documentation if it exists

britmyerss:
We were always defaulting the documentation to `None` when creating Props.  This fixes that, and adds an integration test to ensure they show up as expected. 

I also hacked in a tool tip for the front end so when you hover over the name of a prop, the documentation shows (see screenshot, I'm hovering over the `region` prop and this is what it looks like). I'm not sure if I like it, but this at least let me test that it works end to end and I can remove it before merging. We already have a documentation icon when there is a `docLink` for the prop, which when you click on it, opens the link.  

image


app/web/src/components/AttributesPanel/TreeFormItem.vue
```diff
@@ -402,6 +402,7 @@
           size="none"
         />
         <div
+          v-tooltip="propDocumentation"
```
wendybujalski, line 405: Definitely put a TODO near here to mention that we need to truncate this string for the tooltip and add a way to pop out a modal to read longer documentation strings.

wendybujalski: This also will only work for props that are not headers - when we implement the final version, make sure to do it for headers as well.

stack72 (approved):

PR #5237, opened 2025-01-10
Add features to `si-id`

nickgerace:
This may or may not work given that this will produce `N` buck2 buildable units based on the number of features, but I wanted to post the diff nonetheless.


nickgerace:
I'm going to close this given the fact that the number of combinations with features could be explosive given how buck2 works, but the idea is here in case folks want to revive it someday.

PR #5240, opened 2025-01-11
chore(docs): Updating the roadmap based on current opportunities

stack72:
Fuse FS isn't in this list until I blog post it next week


mahirl (approved):
looks good to me.

PR #5245, opened 2025-01-13
fix(dal,sdf): when finding change sets, ensure it belongs to the relevant workspace

britmyerss:
This change introduces `ChangeSet::find_across_workspaces` to differentiate from `ChangeSet::find` to ensure we're not acting on ChangeSets that don't belong to the Workspace in question. 

<div><br/>via <a href="https://giphy.com/fallontonight/">The Tonight Show Starring Jimmy Fallon</a> on <a href="https://giphy.com/gifs/fallontonight-weird-al-yankovic-i-dont-belong-here-iQ5xPFRDM2jnsjpWj2">GIPHY</a></div>

### DIFF
--- a/lib/dal/src/change_set.rs
+++ b/lib/dal/src/change_set.rs
@@ -556,8 +556,10 @@ impl ChangeSet {
         Ok(())
     }
 
+    /// Finds a [`ChangeSet`] across all workspaces, ignoring the provided [`WorkspacePk`] on the
+    /// current [`DalContext`]
     #[instrument(
-        name = "change_set.find",
+        name = "change_set.find_across_workspaces",
         level = "debug",
         skip_all,
         fields(
@@ -565,7 +567,7 @@ impl ChangeSet {
             si.workspace.id = Empty,
         ),
     )]
-    pub async fn find(
+    pub async fn find_across_workspaces(
         ctx: &DalContext,
         change_set_id: ChangeSetId,
     ) -> ChangeSetResult<Option<Self>> {
@@ -591,6 +593,52 @@ impl ChangeSet {
                 Ok(Some(change_set))
             }
             None => Ok(None),
+        }
+    }
+
+    /// Find a change set within the [`WorkspacePk`] set for the current [`DalContext`]
+    #[instrument(
+        name = "change_set.find",
+        level = "debug",
+        skip_all,
+        fields(
+            si.change_set.id = %change_set_id,
+            si.workspace.id = Empty,
+        ),
+    )]
+    pub async fn find(
+        ctx: &DalContext,
+        change_set_id: ChangeSetId,
+    ) -> ChangeSetResult<Option<Self>> {
+        let span = current_span_for_instrument_at!("debug");
+        let workspace_id = ctx.workspace_pk()?;
+        let row = ctx
+            .txns()
+            .await?
+            .pg()
+            .query_opt(
+                "SELECT * FROM change_set_pointers WHERE id = $1 AND workspace_id = $2",
+                &[&change_set_id, &workspace_id],
+            )
+            .await?;
+
+        match row {
+            Some(row) => {
+                let change_set = Self::try_from(row)?;
+
+                if let Some(workspace_id) = change_set.workspace_id {
+                    span.record("si.workspace.id", workspace_id.to_string());
+                }
+                Ok(Some(change_set))
+            }
+            None => {
+                // warn here so we can see if something is requesting change sets cross workspace
+                warn!(
+                    si.workspace.id = %workspace_id,
+                    "Change Set Id: {change_set_id} not found for Workspace: {workspace_id}",
+                );
+                Ok(None)
+            }
         }
     }
 
--- a/lib/dal/src/context.rs
+++ b/lib/dal/src/context.rs
@@ -403,9 +403,11 @@ impl DalContext {
         Ok(workspace)
     }
 
-    /// Update the context to use the most recent snapshot pointed to by the current `ChangeSetId`.
+    /// Update the context to use the most recent snapshot pointed to by the current [`ChangeSetId`].
+    /// Note: This does not guarantee that the [`ChangeSetId`] is contained within the [`WorkspacePk`]
+    /// for the current [`DalContext`]
     pub async fn update_snapshot_to_visibility(&mut self) -> TransactionsResult<()> {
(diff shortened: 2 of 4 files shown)


lib/dal/src/func/runner.rs
```diff
@@ -1043,7 +1043,7 @@ impl FuncRunner {
                     maybe_action_originating_change_set_id
                 {
                     if let Some(original_change_set) =
-                        ChangeSet::find(ctx, action_originating_change_set_id).await?
+                        ChangeSet::find_across_workspaces(ctx, action_originating_change_set_id).await?
```
britmyerss, line 1046: I didn't dive deep in how we're constructing the `DalContext` to execute functions, this might not need to be here

lib/dal/src/workspace_snapshot/migrator.rs
```diff
@@ -87,7 +87,7 @@ impl SnapshotGraphMigrator {
         info!("Migrating {} snapshot(s)", open_change_sets.len(),);
 
         for change_set in open_change_sets {
-            let mut change_set = ChangeSet::find(ctx, change_set.id)
+            let mut change_set = ChangeSet::find_across_workspaces(ctx, change_set.id)
```
britmyerss, line 90: I believe this is correct - @zacharyhamm could use a check

lib/sdf-server/src/service/v2/admin/get_snapshot.rs
```diff
@@ -23,7 +23,7 @@ pub async fn get_snapshot(
 ) -> AdminAPIResult<Response<Body>> {
     let ctx = builder.build_head(access_builder).await?;
 
-    let change_set = ChangeSet::find(&ctx, change_set_id)
+    let change_set = ChangeSet::find_across_workspaces(&ctx, change_set_id)
```
britmyerss, line 26: this might change after the other PR lands - will see.

lib/dal/src/context.rs
```diff
@@ -403,9 +403,11 @@ impl DalContext {
         Ok(workspace)
     }
 
-    /// Update the context to use the most recent snapshot pointed to by the current `ChangeSetId`.
+    /// Update the context to use the most recent snapshot pointed to by the current [`ChangeSetId`].
+    /// Note: This does not guarantee that the [`ChangeSetId`] is contained within the [`WorkspacePk`]
+    /// for the current [`DalContext`]
     pub async fn update_snapshot_to_visibility(&mut self) -> TransactionsResult<()> {
-        let change_set = ChangeSet::find(self, self.change_set_id())
+        let change_set = ChangeSet::find_across_workspaces(self, self.change_set_id())
```
britmyerss, line 410: Could do a similar move here to have a tenancy-scoped version of this method, but not in this PR. Note that scoping this to the workspace breaks all of our integration tests at the moment.

lib/dal/tests/integration_test/change_set.rs
```diff
@@ -255,6 +255,78 @@ async fn build_from_request_context_limits_to_change_sets_of_current_workspace(
...
+    let user_1_tenancy = dal::Tenancy::new(*user_1_workspace.pk());
+    let access_builder = AccessBuilder::new(user_1_tenancy, HistoryActor::User(user_1.pk()), None);
+
+    let user_1_dal_context = ctx_builder
+        .build_head(access_builder)
+        .await
+        .expect("could not build dal context");
+
+    //first, let's ensure we can't find it when using the user 1 dal ctx
+    let user_2_change_set_unfound = ChangeSet::find(&user_1_dal_context, user_2_change_set.id)
```
britmyerss, line 316: before this change, we would User-1-Workspace-1 would successfully find (and be able to abandon) the change set in User-2's Workspace.

lib/dal/src/change_set.rs
```diff
@@ -698,7 +739,7 @@ impl ChangeSet {
     #[instrument(level = "info", skip_all)]
     pub async fn apply_to_base_change_set(ctx: &mut DalContext) -> ChangeSetApplyResult<ChangeSet> {
         // Apply to the base change with the current change set (non-editing) and commit.
-        let mut change_set_to_be_applied = Self::find(ctx, ctx.change_set_id())
+        let mut change_set_to_be_applied = Self::find_across_workspaces(ctx, ctx.change_set_id())
```
jkeiser, line 742: This one *might* be able to be `ChangeSet::find()` (it seems like if we really need find_across_workspaces than we have been given a busted `DalContext`). But if you're worried about it or noticed it breaks things, I'm 100% on board with "fix most stuff and then look at these in another PR").

lib/dal/src/workspace_snapshot/migrator.rs
jkeiser: It's hard to imagine this isn't correct, given the fact that these change set IDs came from `list_open_for_all_workspaces()`!. (Again, it at least doesn't make things worse.)

jkeiser (approved):
Super glad to see it narrowed down to just a few callers!

lib/dal/src/change_set.rs
britmyerss: yeah I think you're right! Will fix

lib/dal/src/workspace_snapshot/migrator.rs
zacharyhamm: this is what we want, yes. The migrator will set up the dal context for the right visibility and tenancy but only *after* it finds the workspace, so this is definitely necessary

lib/dal/src/change_set.rs
```diff
@@ -594,6 +596,45 @@ impl ChangeSet {
...
+        match row {
+            Some(row) => {
+                let change_set = Self::try_from(row)?;
+
+                if let Some(workspace_id) = change_set.workspace_id {
+                    span.record("si.workspace.id", workspace_id.to_string());
+                }
+                Ok(Some(change_set))
+            }
+            None => Ok(None),
```
zacharyhamm, line 634: If we emitted an info or error here, we could track any time this returns None. Which should be almost never, except maybe stale URLs in cache for applied/abandoned change sets

zacharyhamm: (or set an attribute, whichever makes the most sense for a honeycomb query)

jkeiser (dismissed):

jkeiser (approved):
Lock it down!

PR #5249, opened 2025-01-14
feat(sdf): Add automation API for "create changeset"

jkeiser:
This is a "create change set" endpoint for automation tokens (`POST /api/public/v0/workspaces/:workspace_id/change-set`). It takes `changeSetName` and is identical to the v2 create changeset api.

**Versioning**: This is designed to support a GitHub Action (which is the real public API we're shipping and supporting). The API endpoints themselves are not something we're exposing and documenting. (This is part of why `v0` is in the URL, to give a sense of semantic versioning independence.)

## The Code

This largely copies the v2 create changeset API.

It also has a couple of extractor ergonomics improvements (used in the new endpoint) that we can fold back into v2 endpoints. (I did not do the folding right now, to limit blast radius and to make this more reviewable.) Specifically:
* `WorkspaceDalContext`: Most workspace endpoints can use the WorkspaceDalContext extractor to get the DalContext, instead of extracting most of that information manually and constructing the DalContext in the endpoint code. This has the benefit of ensuring the DalContext is only given when authorization is complete, as well.
* `PosthogEventTracker`: Most endpoints have to extract and track the posthog client, original URI and host in order to send tracking events. This folds those three into one extractor. It then lets you call `tracker.track()` which will automatically include those in the posthog event.

### Example

As an example, create_change_set (the existing one) would now have this preamble:

```rust
pub async fn create_change_set(
    WorkspaceDalContext(ctx): WorkspaceDalContext,
    Json(request): Json<CreateChangeSetRequest>,
    tracker: PosthogEventTracker,
) -> ChangeSetResult<Json<CreateChangeSetResponse>> {
    let change_set = ChangeSet::fork_head(&ctx, &request.change_set_name).await?;
    ...
```

Where it currently looks like this:

```rust
pub async fn create_change_set(
    HandlerContext(builder): HandlerContext,
    AccessBuilder(access_builder): AccessBuilder,
    PosthogClient(posthog_client): PosthogClient,
    OriginalUri(original_uri): OriginalUri,
    Host(host_name): Host,
    Json(request): Json<CreateChangeSetRequest>,
) -> ChangeSetResult<Json<CreateChangeSetResponse>> {
    let ctx = builder.build_head(access_builder).await?;

    let change_set = ChangeSet::fork_head(&ctx, &request.change_set_name).await?;
    ...
```

### DIFF
--- a/lib/sdf-server/src/extract/request.rs
+++ b/lib/sdf-server/src/extract/request.rs
@@ -16,13 +16,10 @@ use super::{internal_error, unauthorized_error, ErrorResponse};
 pub struct RequestUlidFromHeader(pub Option<Ulid>);
 
 #[async_trait]
-impl FromRequestParts<AppState> for RequestUlidFromHeader {
+impl<S> FromRequestParts<S> for RequestUlidFromHeader {
     type Rejection = ErrorResponse;
 
-    async fn from_request_parts(
-        parts: &mut Parts,
-        _state: &AppState,
-    ) -> Result<Self, Self::Rejection> {
+    async fn from_request_parts(parts: &mut Parts, _state: &S) -> Result<Self, Self::Rejection> {
         let request_ulid = parts
             .headers
             .get("X-SI-REQUEST-ULID")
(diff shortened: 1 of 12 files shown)


vbustamante (approved):
Nice additions! The ctx extractor should be pretty useful

PR #5251, opened 2025-01-14
feat: a fuse filesystem for si

zacharyhamm:
Adds an initial spike of a fuse filesystem for si. Currently able to
list change sets, make change sets, and list (uninstalled) schemas.

Grab a bearer token from your dev tools for the workspace you want to
mount as an si-fs, and do:

export SI_BEARER_TOKEN="the token base64 stuff"

Then,

```
buck2 run //bin/si-fs:si-fs -- --workspace-id <THE WORKSPACE ID> --endpoint https://app.systeminit.com/ <THE MOUNT POINT>
```

The si-fs program will become a daemon handling fuse filesystem
operations. Do ls <THE MOUNT POINT> to see your open change sets.


lib/sdf-server/src/service/change_set/create_change_set.rs
```diff
@@ -58,6 +47,7 @@ pub async fn create_change_set(
         .publish_on_commit(&ctx)
         .await?;
 
+    let change_set = change_set.into_frontend_type(&ctx).await?;
```
zacharyhamm, line 50: @jobelenus  or @vbustamante  or @wendybujalski 

We moved the request and response into the frontend types so we could use them in our SI fs client. I tested create change set with this, and it seems to be happy with the "frontend" type on the frontend. But let me know if you think this will cause problems

jobelenus: Yea this shouldn't be a problem. Excellent to keep following the pattern

wendybujalski: Looks good to me, I do not see any potential issues.

fnichol (approved):
We're looking good!

fnichol (dismissed):

fnichol (approved):

PR #5253, opened 2025-01-14
feat(web) - UI for revoking auth tokens, active and inactive token lists

wendybujalski:
Screenshot 2025-01-14 at 6 01 07 PM

### DIFF
--- a/app/auth-portal/src/components/AuthTokenListItem.vue
+++ b/app/auth-portal/src/components/AuthTokenListItem.vue
@@ -14,7 +14,7 @@
     "
   >
     <td>
-      <TruncateWithTooltip>{{ authToken.name }}</TruncateWithTooltip>
+      <TruncateWithTooltip>{{ token.name }}</TruncateWithTooltip>
     </td>
     <!-- TODO show user of token if it's not current user--right now only owner can create -->
     <td><Timestamp size="long" :date="createdAt" enableDetailTooltip /></td>
@@ -69,16 +69,17 @@ import {
   TruncateWithTooltip,
   Timestamp,
 } from "@si/vue-lib/design-system";
-import { computed, onMounted, onUnmounted, ref } from "vue";
+import { computed } from "vue";
 import { useAsyncState } from "@vueuse/core";
 import { apiData } from "@si/vue-lib/pinia";
 import { Workspace } from "@/store/workspaces.store";
-import { AuthToken, useAuthTokensApi } from "@/store/authTokens.store";
+import { useAuthTokensApi } from "@/store/authTokens.store";
+import { AuthTokenWithRealtimeData } from "./AuthTokenList.vue";
 
 const api = useAuthTokensApi();
 
 const props = defineProps<{
-  authToken: AuthToken;
+  authToken: AuthTokenWithRealtimeData;
   workspace: Readonly<Workspace>;
   active: boolean;
 }>();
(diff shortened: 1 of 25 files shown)


app/auth-portal/src/components/MemberListItem.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <tr
-    class="children:px-md children:py-sm children:truncate text-sm font-medium text-gray-800 dark:text-gray-200"
```
wendybujalski, line 3: Frontend Best Practices:

Only use the approved abstract colors - shade, neutral, action, success, warning, and destructive. Do not use named Tailwind colors like red or gray.

app/auth-portal/src/components/AuthTokenListItem.vue
```diff
@@ -56,26 +90,43 @@ const emit = defineEmits<{
...
 const expiresAt = computed(() =>
-  props.authToken.expiresAt
-    ? new Date(props.authToken.expiresAt).toLocaleString()
-    : undefined,
+  props.authToken.expiresAt ? new Date(props.authToken.expiresAt) : undefined,
+);
+
+const expired = computed(() => {
+  // eslint-disable-next-line @typescript-eslint/no-unused-expressions
+  checkExpirationRecomputer.value;
```
jkeiser, line 111: What does having this line here do? ((It seems like checkExpirationRecomputer will run no matter what?)

jkeiser: Oh. Never mind. This is to make it retrigger!

wendybujalski: This is there to force `expired` to recompute when `checkExpirationRecomputer` has its value changed, which happens once every five seconds (see the `onMounted` code near the bottom of this file). This ensures that the `expired` boolean will be accurate over time without needing to refresh the page, since otherwise the computed value will not update as time passes.

app/auth-portal/src/components/AuthTokenListItem.vue
```diff
@@ -92,4 +142,22 @@ const expiresAt = computed(() =>
 // }
 // /** Name of token to create */
 // const name = ref(authToken.name);
+
+// This pokes the computed values to check if any tokens have expired every 5 seconds
+const EXPIRATION_CHECK_INTERVAL = 5000;
+const checkExpiration = ref();
```
jkeiser, line 148: checkExpiration isn't used, does it need to be a ref?

Ooh, if you set a `now` variable instead of a tick counter, it gets rid of the need for a bunch of code as well as for the warning itself:

```ts
const now = ref(Date.now())
const expired = computed(() => expiresAt.value?.getTime() <= now.value);

let checkExpiration;
onMounted(() => {
  checkExpiration = setInterval(() => {
    now.value = Date.now(), EXPIRATION_CHECK_INTERVAL));
  }
}
onUnmounted(() => clearInterval(checkExpiration));
```

The current code works fine though, but probably needs a comment in the spot where you ignore the warning.

app/auth-portal/src/components/AuthTokenListItem.vue
```diff
@@ -6,48 +6,81 @@
...
+      <template v-else>Never</template>
+    </td>
+    <td v-if="workspace.role === 'OWNER'" class="text-center">
+      <template v-if="active">
+        <IconButton
+          v-if="revoke.error.value"
+          icon="alert-triangle"
+          iconTone="destructive"
+          class="w-min mx-auto"
+          tooltip="Error. Token not revoked!"
```
jkeiser, line 37: I like this way of reporting the error! Is it on purpose not to put the error message?

wendybujalski: Yes, the error message that comes through is just as unhelpful but much longer, so I made it a lil bit prettier 😅

stack72 (approved):

app/auth-portal/src/pages/WorkspaceAuthTokensPage.vue
```diff
@@ -180,9 +167,50 @@ const authTokens = useAsyncState(
   { shallow: false },
 );
 
-const listedTokens = computed(() =>
-  _.reverse(_.sortBy(_.values(authTokens.state.value), "createdAt")),
-);
+// This pokes the computed values to check if any tokens have expired every 5 seconds
```
jkeiser, line 170: I wonder if the thing to do here is have this be the only recomputer. Instead of storing simple `AuthToken`, we could store `AuthToken & { isExpired: boolean }`. Then we update isExpired to true whenever we notice a token becomes expired.

That way, the listedTokens / activeTokens split can just use `token.isExpired || token.revokedAt` and the AuthTokenListItem would automatically get isExpired passed in as well (and could drive itself off of that instead of running its own loop).

jkeiser (approved):
This looks great!

I think we might be able to streamline the expiration timer stuff a bit (down to one loop, and make it set `now` so the computation is straightforward) but we can mull that over tomorrow.

app/auth-portal/src/pages/WorkspaceAuthTokensPage.vue
wendybujalski: Yeah that sounds like a much simpler system - I can make those changes in my next PR 😅

bin/auth-api/src/routes/auth_token.routes.ts
```diff
@@ -92,8 +92,16 @@ router.put("/workspaces/:workspaceId/authTokens/:authTokenId", async (ctx) => {
   ctx.body = { authToken };
 });
 
+// revoke the given authToken for the given workspace
+router.post("/workspaces/:workspaceId/authTokens/:authTokenId/revoke", async (ctx) => {
+  const { authToken } = await authorizeAuthTokenRoute(ctx, RoleType.OWNER);
```
stack72, line 97: this needs to be [RoleType.OWNER]

app/auth-portal/src/pages/WorkspaceAuthTokensPage.vue
wendybujalski: (or fold them into this one now!)

stack72 (approved):

PR #5254, opened 2025-01-15
feat(auth-api): Deploy new posthog event for emails

stack72:
This is the format of the new posthog event so that we can send a transactional email


britmyerss (approved):

nickgerace (dismissed):

wendybujalski (approved):
👍

PR #5255, opened 2025-01-15
Fix: to determine what mgmt edges are removed, we need all mgmt edges, not just components & edges in the view

jobelenus:
Same fix as the one before, but very different
image

### DIFF
--- a/lib/dal/src/diagram.rs
+++ b/lib/dal/src/diagram.rs
@@ -13,7 +13,7 @@ use std::{
 use telemetry::prelude::*;
 use thiserror::Error;
 
-use crate::FuncError;
+use crate::workspace_snapshot::node_weight::NodeWeight;
 use crate::{
     attribute::{
         prototype::argument::{AttributePrototypeArgumentError, AttributePrototypeArgumentId},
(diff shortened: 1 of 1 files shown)


lib/dal/src/diagram.rs
```diff
@@ -729,13 +742,9 @@ impl Diagram {
...
             let removed_edges = Self::assemble_removed_edges(ctx).await?;
             diagram_component_views.diagram_edges.extend(removed_edges);
-            let removed_management_edges = Self::assemble_removed_management_edges(
-                ctx,
-                base_snapshot,
-                &diagram_component_views.management_edge_set,
-                &component_info_cache,
-            )
-            .await?;
+            let removed_management_edges =
```
zacharyhamm, line 745: This always returns an empty vec in my testing.

jobelenus:
@zacharyhamm this code needs a lot of love. I just grabbed helper code from places that used ctx, but we can't use ctx because we're using base_snapshot. Would love to get a sense of the better rust-y way to accomplish this!

zacharyhamm (approved):
ship it!

zacharyhamm:
@jobelenus if this isn't on up to date a rebase might make the tests more reliable

PR #5256, opened 2025-01-15
feat(sdf): Management APIs to update components+trigger management functionsAdd "update component properties" and "trigger management functions"

jkeiser:
This adds two management API endpoints:

* Component update: PUT /api/public/v0/workspaces/:workspace_id/change-sets/:change_set_id/components/:component_id/properties
  Lets you pass a set of properties with either `{ "propId": <value> }` or `{ "<prop name/path>": <value> }`, and sets those properties to the given values.

  This is patchy at the top level in that it will only set properties you specify at the top level, but it will replace the entire value of anything you set. So if you have a property like `{ a: { b: 1, c: 2 } }` and you POST with `{ "a": { b: 10 } }`, then `c` will be reset to its default value. But if you set `{ "a/b": 10 }`, then `c` will retain its value.

  The final interface is something we intend to experiment with; this gets us to where we have some flexibility writing the github action.
* Trigger management function: this is an exact analog of the normal endpoint to trigger a management function. You pass the view, component and prototype id (in the same format). I have copied the endpoint itself. This interface will likely evolve to support the github action over time as well, but our initial plan is to have you pass these IDs yourself.

This puts us in spitting distance of a useful management-function-triggering github action (apply changeset and the action itself are the remaining items before we can really put it into practice).

### Code

I also found an extractor pattern that lets us use some of our extractors more easily (`parts.extract()` instead of `extract_with_state()`), and applied it where it could be. Doesn't affect more than a few callers.

management!

### DIFF
--- a/lib/sdf-server/src/extract/change_set.rs
+++ b/lib/sdf-server/src/extract/change_set.rs
@@ -1,0 +1,103 @@
+use axum::{
+    async_trait,
+    extract::{FromRequestParts, Path},
+    http::request::Parts,
+    RequestPartsExt as _,
+};
+use dal::{ChangeSet, ChangeSetId, DalContext};
+use derive_more::{Deref, Into};
+use serde::Deserialize;
+
+use crate::app_state::AppState;
+
+use super::{bad_request, internal_error, workspace::WorkspaceAuthorization, ErrorResponse};
+
+///
+/// Gets a DalContext pointed at the TargetChangeSet.
+///
+/// This ensures the user is authorized to access the workspace, has the correct role, and
+/// that the change set is in fact a part of the workspace.
+///
+#[derive(Clone, derive_more::Deref, derive_more::Into)]
+pub struct ChangeSetDalContext(pub DalContext);
+
+#[async_trait]
+impl FromRequestParts<AppState> for ChangeSetDalContext {
+    type Rejection = ErrorResponse;
+
+    async fn from_request_parts(
+        parts: &mut Parts,
+        state: &AppState,
+    ) -> Result<Self, Self::Rejection> {
+        // Get the workspace we are accessing (and authorized for)
+        let WorkspaceAuthorization {
+            mut ctx,
+            workspace_id,
+            ..
+        } = parts.extract_with_state(state).await?;
+
+        // Validate the change set id is within that workspace
+        let TargetChangeSetId(change_set_id) = parts.extract().await?;
+        let change_set = ChangeSet::find(&ctx, change_set_id)
+            .await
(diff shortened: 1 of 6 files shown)


lib/sdf-server/src/service/public/components.rs
```diff
@@ -0,0 +1,186 @@
...
+
+#[derive(Deserialize, Serialize, Debug)]
+#[serde(rename_all = "camelCase")]
+struct UpdateComponentPropertiesRequest {
+    domain: HashMap<ComponentPropKey, serde_json::Value>,
+}
+
+#[derive(Deserialize, Serialize, Debug)]
+#[serde(rename_all = "camelCase")]
+struct UpdateComponentPropertiesResponse {}
```
vbustamante, line 154: Since these structs are not public, I have been naming them just Response/Request (and by this logic, Path), since this reduces boilerplate on usage and avoids weird scenarios were we copy controllers to create new ones and end up with names that make zero sense accidentaly. 

Whatever you choose is fine though, just a suggestion

vbustamante (dismissed):
Left some nits as comments but looks great

lib/sdf-server/src/service/public/management.rs
```diff
@@ -0,0 +1,186 @@
...
+    ChangeSetDalContext(ctx): ChangeSetDalContext,
+    tracker: PosthogEventTracker,
+    Path(RunPrototypePath {
+        management_prototype_id: prototype_id,
+        component_id,
+        view_id,
+    }): Path<RunPrototypePath>,
+    Json(request): Json<RunPrototypeRequest>,
+) -> Result<Json<RunPrototypeResponse>> {
+    // let force_change_set_id = ChangeSet::force_new(&mut ctx).await?;
```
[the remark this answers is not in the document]
jkeiser: I think it's a TBD ... I commented out functionality that exists in the current endpoint but not in the new one. But I'll just remove this comment for now, I think it's pretty clear what has to happen if we bring it back.

lib/sdf-server/src/service/public/components.rs
jkeiser: I love that convention! The reason I have them named is that I'm experimenting with the readability of having all change_set routes in one file (since they are often super related). That just requires either inline modules or names. I'll experiment with inline modules (I think there is a bunch of value in having the CRUD that manipulates a single object together in a single module) in the next PR!

vbustamante (approved):

lib/sdf-server/src/extract/change_set.rs
britmyerss, line 1: 👏

PR #5258, opened 2025-01-15
feat(auth-api): Adding tracking for API Token Create and Revoke

stack72:


wendybujalski (approved):
👍

bin/auth-api/src/routes/auth_token.routes.ts
```diff
@@ -75,38 +86,41 @@ router.get("/workspaces/:workspaceId/authTokens/:authTokenId", async (ctx) => {
...
+  const { authToken, authUser, workspace } = await authorizeAuthTokenRoute(ctx, [RoleType.OWNER]);
 
   await updateAuthToken(authToken.id, { revokedAt: new Date() });
 
+  tracker.trackEvent(authUser, "workspace_api_token_revoked", {
+    workspaceId: workspace.id,
+    workspaceName: workspace.displayName,
+    tokenName: authToken.name,
+    tokenCreated: authToken.createdAt,
+    tokenRevoked: new Date(),
```
jkeiser, line 100: We probably also want to report `tokenExpires` for consistency (and because I can imagine wanting to know how close to the expiration date people revoke tokens).

This also makes me realize the endpoint has a bug--if we're going to return an authToken it should be the *updated* one with `authToken.revokedAt`. If it had the new date, you would have been able to use a common method to construct the event properties across these endpoints. (With the exception of tokenDeleted). (It is obviously not necessary to fix this bug in your PR, I'm just writing to remind myself.)

jkeiser (approved):
One minor comment. Thanks for adding this, I'll make sure I'm adding posthog tracking to future api requests!

PR #5261, opened 2025-01-15
chore(dal): Use ChangeSet::get_by_id() anywhere change set is already required

jkeiser:
This adds a ChangeSet::get_by_id() that calls ChangeSet::find() and yields ChangeSetNotFound if it fails. It replaces all callers that were doing exactly the same thing (and removes the ChangeSetNotFound error from many of their error enums).

I also moved DefaultChangeSetNotFound to workspace's `default_change_set()` method while I was at it; that is only called in one place, and was also a rote replacement.

Wanted this while writing an endpoint, but it seems worth breaking out. This is entirely a textual replacement of callers (no change in behavior).

### DIFF
--- a/lib/sdf-server/src/service/change_set.rs
+++ b/lib/sdf-server/src/service/change_set.rs
@@ -42,8 +42,6 @@ pub enum ChangeSetError {
     ActionPrototype(#[from] ActionPrototypeError),
     #[error("cannot abandon head change set")]
     CannotAbandonHead,
-    #[error("change set not found")]
-    ChangeSetNotFound,
     #[error("component error: {0}")]
     Component(#[from] ComponentError),
     #[error("dal change set error: {0}")]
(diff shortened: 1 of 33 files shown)


lib/sdf-server/src/service/change_set.rs
```diff
@@ -80,7 +78,9 @@ impl IntoResponse for ChangeSetError {
             ChangeSetError::ActionAlreadyEnqueued(_) => {
                 (StatusCode::NOT_MODIFIED, self.to_string())
             }
-            ChangeSetError::ChangeSetNotFound => (StatusCode::NOT_FOUND, self.to_string()),
+            ChangeSetError::DalChangeSet(DalChangeSetError::ChangeSetNotFound(..)) => {
```
jkeiser, line 81: I considered doing an `inner.into_response()` here, but changeset not existing could be a 500 in some callsites (it's only a 404 if the changeset is one you specified in the request) so I left it specific here.

britmyerss (approved):

britmyerss (approved):

PR #5263, opened 2025-01-15
fix(veritech): When errors occur that don't cause us to retry the function, inform subscribers

britmyerss:
This change ensures that for all functions types, when errors occur (such as killing a function, or timing out), we inform subscribers of the failure so they can handle them appropriately. 

The main issue this fixes is during DVU jobs, if a function times out or is killed, the job will continue to run until it hits its own time out, exacerbating hung change sets

<div><br/>via <a href="https://giphy.com/gifs/fail-failed-HNEmXQz7A0lDq">GIPHY</a></div>

### DIFF
--- a/lib/dal/tests/integration_test/func/kill_execution.rs
+++ b/lib/dal/tests/integration_test/func/kill_execution.rs
@@ -11,7 +11,7 @@ use dal_test::helpers::{
 };
 use dal_test::test;
 use pretty_assertions_sorted::assert_eq;
-use si_events::FuncRunState;
+use si_events::{ActionResultState, FuncRunState};
 
 #[test]
 async fn kill_execution_works(ctx: &mut DalContext) {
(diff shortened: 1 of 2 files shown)


lib/veritech-server/src/handlers.rs
```diff
@@ -242,47 +241,76 @@ where
...
+            kill_sender_remove_blocking(&state.kill_senders, execution_id.to_owned()).await?;
             func_result
         },
     };
 
     match result {
+        // Got an Ok - let anyone subscribing to a reply know
         Ok(function_result) => {
             if let Err(err) = publisher.publish_result(&function_result).await {
-                metric!(counter.function_run.action = -1);
```
sprutton1, line 268: This is scooped up by the dec call below, yeah?

zacharyhamm: yeah were we double decrementing here?

zacharyhamm (dismissed):
This keeps the DVU from getting stuck! 🚀

lib/veritech-server/src/handlers.rs
britmyerss: yeah I think just leftover from pre-refactor

zacharyhamm (approved):

lib/dal/tests/integration_test/func/kill_execution.rs
```diff
@@ -104,8 +116,21 @@ async fn kill_execution_works(ctx: &mut DalContext) {
         .expect("could not get func run")
         .expect("no func run found")
         .state();
+
+    // This should be [`FuncRunState::Killed`]
+    // but because this is getting bubbled back to the Pinga Job running the
+    // action, and the error is now embedded in a successful response,
+    // (because we did get a response from Veritech)
```
nickgerace, line 123: UGH! So sorry about this... that explains a lot.

PR #5279, opened 2025-01-16
Add ability to set approvers on views in sdf

nickgerace:
## Introduction

This PR adds the ability to set approvers on the views with a new set of `sdf` routes, backed by a new integration test among existing ones.

## Description

To support this change, we now divide `ApprovalRequirements` into two separate categories: explicit and virtual.

- Explicit requirements are derived from `ApprovalRequirementDefinition` nodes (new to `WorkspaceSnapshotGraphV4`!) on the graph from any node with an outgoing `HasApprovalRequirement` edge
- Virtual requirements are derived from examining a list of changes on the fly and are useful for determining if meta changes to permissions require approvals themselves (e.g. if you add a `ApprovalRequirementDefinition`, we need to determine who can approve that change)

Both types of requirements share an inner "rule" type with covers the universal elements of all approval requirements:

- The ID of the entity requiring an approval
- The kind of entity requiring an approval
- The minimum number of approvals for this rule
- The approvers of this rule, which can either be individual users or a "permission lookup" bag used to query SpiceDB for who can approve the requirement

Back to virtual requirements: not only are they useful for determining meta changes to permissions, but also for creating requirements when nodes are deleted and creating the checksum for new approvals involving deleted nodes. In fact, that is why checksum calculation now requires the hashes alongside the IDs... you can't get the merkle tree hash for a deleted node!

As a result of these changes, detecting changes between two graphs is now fallible (we need to get the node weight in order to get the merkle tree hash).

## Other Changes

Other changes in this PR include, but are not limited to, the following:

- Use `UserPk` instead of `String` for return type involving frontend approvals
- Re-order enum variants and their discriminants where possible when dealing with `ApprovalRequirementDefinition` and `HasApprovalRequirement`
- Rename "geometry" dummy data to "view" in deserialize graph integration test because it could be confusing (although it is entirely cosmetic for the purposes of the test)

## Relevant PRs

This is a follow-up to #5144


lib/dal/src/workspace_snapshot/graph/traits/approval_requirement.rs
britmyerss, line 1: 👏

jkeiser:
Reading through after check-in: one thing I will be looking for is whether any work needs to be done to support the automation API's force_apply and request_approval endpoints.

nickgerace:
@jkeiser funny you mention that! @stack72 gave me a heads up that we'd need to be aware of it.

Fortunately, this work is 100% feature flagged other than the new node weight addition (which adding them is feature flagged anyway), so we can solve it in the next turn.

lib/dal/src/workspace_snapshot/graph/detector.rs
```diff
@@ -88,16 +93,34 @@ impl<'a, 'b> Detector<'a, 'b> {
...
         let mut changes = Vec::new();
 
         petgraph::visit::depth_first_search(
             self.updated_graph.graph(),
             Some(self.updated_graph.root()),
             |event| self.calculate_changes_dfs_event(event, &mut changes),
-        );
+        )?;
+
+        // Add changes for IDs that only exist in the base graph because our DFS walk of the
```
britmyerss, line 105: can you elaborate on this a bit for me?

lib/dal/src/workspace_snapshot/graph/v4/approval_requirement.rs
britmyerss, line 1: I'm not following the virtual vs. explicit here - let's chat tomorrow so I can understand

lib/dal/src/workspace_snapshot/graph/detector.rs
jhelwig: Because we're walking the change set's graph, we won't encounter things that have been deleted, such as an approval requirement definition. Because we want all approval requirement definition changes (add/remove/modify) to require approval from the workspace approvers we need to know which things only exist in the base graph (and what kind of thing they are) to be able to later generate a "virtual" approval requirement where appropriate for them.

britmyerss: ahhh makes total sense!

lib/dal/src/workspace_snapshot/graph/v4/approval_requirement.rs
britmyerss: this is exactly what I was looking for - super super helpful, thank you!!

britmyerss (approved):
I pulled this branch and tested locally the existing approval flow.  Approvers can still force apply, and contributors still get bumped to the existing approval flow.  Tested both with creating new components/updating existing and creating schema variants.

nickgerace:
Thank you very much for the review @britmyerss!!

PR #5283, opened 2025-01-17
fix: arm rootfs needs an arm linker

sprutton1:
I created this link manually in an arm64 rootfs and things started to work

### DIFF
--- a/bin/lang-js/src/function.ts
+++ b/bin/lang-js/src/function.ts
@@ -19,6 +19,7 @@ import { Debugger } from "./debug.ts";
 import { transpile } from "jsr:@deno/emit";
 import { Debug } from "./debug.ts";
 import * as _worker from "./worker.js";
+import { bundleCode } from "./transpile.ts";
 
 const debug = Debug("langJs:function");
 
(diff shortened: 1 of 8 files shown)


johnrwatson (approved):

zacharyhamm (approved):

bin/lang-js/src/function_kinds/management.ts
```diff
@@ -44,7 +44,7 @@ export interface ManagementCreate {
 }
 
 export interface ManagementOperations {
-  create?: ManagementCreate,
```
zacharyhamm, line 47: lol

bin/lang-js/src/function.ts
```diff
@@ -19,6 +19,7 @@ import { Debugger } from "./debug.ts";
 import { transpile } from "jsr:@deno/emit";
 import { Debug } from "./debug.ts";
 import * as _worker from "./worker.js";
+import { bundleCode } from "./transpile.ts";
```
zacharyhamm, line 22: What's going on here exactly?

bin/lang-js/src/transpile.ts
```diff
@@ -0,0 +1,77 @@
+import * as _ from "npm:lodash-es";
+import { transpile } from "jsr:@deno/emit";
+import { Debug } from "./debug.ts";
+import * as _worker from "./worker.js";
+
+const debug = Debug("langJs:transpile");
+
+const LOCK_FILE = "/tmp/lang-js-transpile.lock";
```
zacharyhamm, line 8: So lang-js will sync now on every transpilation?

sprutton1: Yeah, that's the idea I'm going for. This is really only a thing in tests and local dev. In prod we run firecracker, so we won't ever have contention on the lang-js binary

bin/lang-js/src/function.ts
sprutton1: We need to go from ts -> js so we can build it out as a function to run in the web worker

zacharyhamm (approved):

PR #5286, opened 2025-01-17
Fix: if no view data is received, but component exists on this view, we need to allow it to process

jobelenus:
We missed this case.


zacharyhamm (approved):

zacharyhamm:
@jobelenus doesn't this mean if we receive no view data we delete the component from all views?

zacharyhamm (approved):

PR #5293, opened 2025-01-20
build(tilt): refactor build cmds to speed up initial dev:up

fnichol:
This change changes the build step for a Tilt local resource (called the
`cmd` attribute) to a common `buck2 build` command with all core backend
targets enumerated. In other words, all the following services:

- `//app/web:dev`
- `//bin/forklift:forklift`
- `//bin/pinga:pinga`
- `//bin/rebaser:rebaser`
- `//bin/sdf:sdf`
- `//bin/veritech:veritech`

will use the *same* `buck2 build ...` command.

Why and How?

Prior to this change, each service only tried to build itself (that is,
the Rebaser only attempted to build `//bin/rebaser:rebaser`). The
problem with this approach is that each `buck2 build` subcommand would
interfere with other in-flight `buck2 build` commands as they share a
single `buckd` daemon (this is enforced by design in the Buck2 daemon).
With this change, the first service to start building will build *all*
the other core services in parallel via a single `buck2 build` command.
That means that once the first build is done, all subsequent `buck2
build` commands will immediately return as the daemon knows no inputs
have been modified since the last build invocation for those targets.

### DIFF
--- a/dev/BUCK
+++ b/dev/BUCK
@@ -29,16 +29,16 @@ tilt_up(
 tilt_up(
     name = "up-standard",
     args = [
-        "--standard-rustc-build-mode"
-    ]
+        "--standard-rustc-build-mode",
+    ],
 )
 
 # Bring up the full set of services for development with debug build optimizations
 tilt_up(
     name = "up-debug-all",
     args = [
-        "--debug-rustc-build-mode"
-    ]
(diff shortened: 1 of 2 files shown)


sprutton1 (approved):
Nice!

jkeiser (approved):

PR #5294, opened 2025-01-21
Make generator into pipeline, add asset func step

vbustamante:
- Make generator create spec files on a folder
- Add command to make specs into tars on another folder
- Change architecture a little bit so the generator can have steps added easily
- Add asset func generator step, that only generates top level props for now
- Fix issue that made generated input sockets not send values to props
- Make socket generation into its own step

### DIFF
--- a/bin/si-gen-cloud-control/src/commands/generateSiSpecs.ts
+++ b/bin/si-gen-cloud-control/src/commands/generateSiSpecs.ts
@@ -1,0 +1,55 @@
+import { getServiceByName, loadCfDatabase } from "../cfDb.ts";
+import { pkgSpecFromCf } from "../specPipeline.ts";
+import { PkgSpec } from "../bindings/PkgSpec.ts";
+import { generateAssetFuncs } from "../pipeline-steps/generateAssetFuncs.ts";
+import {
+  generateSocketsFromDomainProps,
+} from "../pipeline-steps/generateSocketsFromDomainProps.ts";
+
+export function generateSiSpecForService(serviceName: string) {
+  const cf = getServiceByName(serviceName);
+  return pkgSpecFromCf(cf);
+}
+
+export async function generateSiSpecs() {
+  const db = await loadCfDatabase();
+
+  let imported = 0;
+  const cfSchemas = Object.values(db);
+
+  let specs = [] as PkgSpec[];
+
+  for (const cfSchema of cfSchemas) {
+    try {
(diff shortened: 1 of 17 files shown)


deno.json
```diff
@@ -1,5 +1,9 @@
 {
-  "workspace": ["./bin/lang-js", "./lib/ts-lib-deno"],
+  "workspace": [
+    "./bin/lang-js",
+    "./lib/ts-lib-deno",
+    "./bin/si-gen-cloud-control"
```
sprutton1, line 5: I think this is going to cause issues with the firecracker rootfs build. I'll dig in a bit to verify it.

bin/si-gen-cloud-control/src/commands/generateSiSpecs.ts
```diff
@@ -0,0 +1,55 @@
...
+
+      specs.push(pkg);
+    } catch (e) {
+      console.log(`Error Building: ${cfSchema.typeName}: ${e}`);
+    }
+  }
+
+  // EXECUTE PIPELINE STEPS
+  specs = generateSocketsFromDomainProps(specs);
+  specs = generateAssetFuncs(specs);
```
sprutton1, line 34: This is a great change. My only nit is that I think it's weird we do the bulk of the work (generating the domain, resource, and related inputs) in the loop and then call the rest pipeline steps. I feel like we should break creating the domain and the resource into their own pipeline steps. I also think step should work over a single spec and we should iterate over each one here instead of internally to the step itself. We can talk about it.

deno.json
vbustamante: I can't run deno locally without it :/

bin/si-gen-cloud-control/src/commands/generateSiSpecs.ts
vbustamante: Some steps will need to take a look at all specs for every spec, like generating the subschemas, so to give each step maximum flexibility, looping internally is necessary

vbustamante: But yeah we should break the initial step as much as we can, this is just a first pass. Although It maybe is enough for first release? We'll see

deno.json
sprutton1: https://github.com/systeminit/si/pull/5295

We'll want to merge this first and then rebase

sprutton1 (approved):

PR #5295, opened 2025-01-21
feat: add buck2 to the generator

sprutton1:
buck2! tests! formatting!

### DIFF
--- a/bin/lang-js/BUCK
+++ b/bin/lang-js/BUCK
@@ -24,6 +24,8 @@ alias(
     actual = ":bin"
 )
 
+# we include si-gen-cloud-control here because the deno workspace
+# requires resolution of all members even if we don't use it
 deno_compile(
     name = "bin",
     main = "src/index.ts",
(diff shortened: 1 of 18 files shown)


bin/lang-js/BUCK
```diff
@@ -24,6 +24,8 @@ alias(
     actual = ":bin"
 )
 
+# we include si-gen-cloud-control here because the deno workspace
+# requires resolution of all members even if we don't use it
```
fnichol, line 28: Yeah that's a bummer, but cool.

I bet we might be able to figure something out, similar to using `turbo`'s trick of simulating our mono repo with the appropriate directories and files to make dependency installs work.

fnichol (approved):
Looks great, nice to have the consistency of these checks!

bin/lang-js/BUCK
sprutton1: That's a great idea. I want to take another pass over the deno jobs in general, so I'll take a look at this then

PR #5298, opened 2025-01-21
feat: create default action funcs

sprutton1:
image


bin/si-gen-cloud-control/src/pipeline-steps/generateActionFuncs.ts
```diff
@@ -0,0 +1,39 @@
...
+export function generateDefaultActionFuncs(specs: PkgSpec[]): PkgSpec[] {
+  const newSpecs = [] as PkgSpec[];
+  const defaultActionFuncs = createDefaultActionFuncs();
+
+  for (const spec of specs) {
+    const schemaVariant = spec.schemas[0]?.variants[0];
+    const funcs = spec.funcs;
+    const actionFuncs = schemaVariant.actionFuncs;
+
+    if (!schemaVariant || !actionFuncs || !funcs) {
```
vbustamante, line 18: Are actions funcs and funcs optional here? I though they were non optional arrays

bin/si-gen-cloud-control/src/pipeline-steps/generateAssetFuncs.ts
```diff
@@ -19,7 +19,7 @@ export function generateAssetFuncs(specs: PkgSpec[]): PkgSpec[] {
       continue;
     }
 
-    const assetFuncUniqueKey = schemaVariant.data.funcUniqueId;
+    const assetFuncUniqueKey = schemaVariant.data?.funcUniqueId;
```
vbustamante, line 22: Data shouldn't be optional on this, if it really is, we should change the type definition. (but I believe you can just drop the `?`)

bin/si-gen-cloud-control/src/pipeline-steps/generateAssetFuncs.ts
```diff
@@ -40,7 +40,7 @@ export function generateAssetFuncs(specs: PkgSpec[]): PkgSpec[] {
 
     const assetFunc: FuncSpec = {
       name: assetFuncName,
-      uniqueId: assetFuncUniqueKey,
+      uniqueId: assetFuncUniqueKey!,
```
vbustamante, line 43: We should avoid this assertions generally. If it's optional, fail if not found. See comment above though

bin/si-gen-cloud-control/src/spec/sockets.ts
```diff
@@ -20,7 +20,7 @@ export function createSocketFromProp(
     return socket;
   } else if (prop.metadata.writeOnly || prop.metadata.createOnly) {
     const socket = createSocket(prop.name, "input");
-    if (socket.data) {
+    if (prop.data) {
```
vbustamante, line 23: Same nit as above, I don't think .data is optional

bin/si-gen-cloud-control/src/pipeline-steps/generateAssetFuncs.ts
sprutton1: The tests will complain without this check

```
error: TS18047 [ERROR]: 'schemaVariant.data' is possibly 'null'.
    const assetFuncUniqueKey = schemaVariant.data.funcUniqueId;
                               ~~~~~~~~~~~~~~~~~~
    at file:///home/scott/projects/si/bin/si-gen-cloud-control/src/pipeline-steps/generateAssetFuncs.ts:24:32
```

sprutton1: Yeah, looking at the spec, data is nullable, so we need to check that it exists when we check the schemaVariant

bin/si-gen-cloud-control/src/spec/sockets.ts
sprutton1: Data is also nullable here as data is an Option in the PkgSpec

bin/si-gen-cloud-control/src/pipeline-steps/generateAssetFuncs.ts
vbustamante: Ok, my IDE is messing something up then, sorry

vbustamante (approved):
Ship it!

vbustamante (dismissed):

vbustamante (approved):

PR #5300, opened 2025-01-22
feat(actions): Add "get component" management function API

jkeiser:
This adds `GET /api/public/v0/components/:componentId`, which is similar (but not identical) to the mainline /api/v2/components/:componentId endpoint. This endpoints gets the data we most need to handle graceful input validation and defaults in the GitHub Action:

```typescript
{
  // The usual component data.
  component: { displayName, ... },
  // List of views this component is in.
  viewData: { viewId, name, geometry }[],
  // List of management functions associated with this component.
  managementFunctions: { managementPrototypeId, name }[],
  // Domain property values. This isn't presently used, but being able to see this is really useful debug information in logs.
  domain: <JS value of domain props>
}
```

This gives us enough information that the github action can:

* Allow the user to omit the `view` parameter when the component is in only one view
* Allow the user to omit the `managementFunction` parameter when the component only has one management function
* Specify the `view` and/or `managementFunction` parameters as *names* instead of IDs, in cases where there are multiple and disambiguation is necessary.

With this endpoint deployed locallyI was able to run the VPC template with this github action spec:

```yaml
      - uses: systeminit/actions/trigger@main
        with:
          changeSetName: Testing 123
          componentId: 01JJ4TXDYTKY91R04ET8GCM675
          domain: |
            Region: us-east-1
            BaseCidrBlock: 10.0.0.0/16
          apiToken: ${{ secrets.SI_API_TOKEN }}

          # These don't have to be specified when there's only one, but this is how you would do it by name
          # view: DEFAULT
          # managementFunction: vpcTemplateCreate
```

### Design

Philosophically, instead of creating this endpoint, we could have handled validation/search/defaulting *inside* the endpoints. But given the cycle time for updating / adding endpoints, we're focusing on building composable endpoints that can be used to build more functionality without having to go back and redeploy so frequently. This is part of why it's a private API: down the road, when we've seen a lot more use cases, we'll have a much stronger idea of what we need to compose operations and reference views/components/functions/props/etc.

### Testing

The github action has been tested using this endpoint to:

* Look up default view and management function (IDs are correct) and successfully trigger management function
* Find associated view and function by name (names are correct)
* Emits the correct names of components/views/functions in logs and errors (component name and information is correct).
* 404 when component is not found
* 401/403 when token is missing/not authorized

### DIFF
--- a/lib/sdf-server/src/service/public/components.rs
+++ b/lib/sdf-server/src/service/public/components.rs
@@ -4,16 +4,21 @@ use axum::{
     extract::Path,
     http::StatusCode,
     response::{IntoResponse, Response},
-    routing::put,
+    routing::{get, put},
     Json, Router,
 };
 use dal::{
+    diagram::{geometry::Geometry, view::View},
+    management::prototype::ManagementPrototype,
     prop::{PropPath, PropResult, PROP_PATH_SEPARATOR},
-    AttributeValue, Component, ComponentId, Prop, PropId, SchemaVariantId, WsEvent,
+    AttributeValue, Component, ComponentError, ComponentId, DalContext, Prop, PropId,
+    SchemaVariantId, WsEvent,
 };
 use serde::{Deserialize, Serialize};
 use serde_json::json;
 use si_events::audit_log::AuditLogKind;
+use si_frontend_types::{DiagramComponentView, GeometryAndView};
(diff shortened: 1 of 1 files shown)


britmyerss (approved):
LGTM

PR #5301, opened 2025-01-22
feat(dal): Add the ability to set an annotation on a prop in the asset builder that it's create only

stack72:
When this value is set, if the component has a resource associated with it, then we won't allow a user to set the value in the attribute editor

We know a VPC CidrBlock can't be modified after the resource is created. So we can change the prop definition to be:

```
    const cidrBlockProp = new PropBuilder()
        .setKind("string")
        .setName("CidrBlock")
        .setValidationFormat(Joi.string().required().regex(/^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])(\/)(1[6-9]|2[0-8])$/).messages({
            'string.pattern.base': 'Must be a valid IPv4 CIDR with CIDR Blocks between /16 and /28'
        }))
        .setWidget(new PropWidgetDefinitionBuilder().setKind("text")
            .setCreateOnly()
            .build())
        .build();
```

Notice the `.setCreateOnly()` as part of the PropWidget

That causes the UI to react as follows (when a resource exists for the component)

Screenshot 2025-01-21 at 23 53 19

This value can't be overridden like a value that has been created via a socket or an attribute func so there's special casing to this

### DIFF
--- a/lib/dal/src/property_editor/schema.rs
+++ b/lib/dal/src/property_editor/schema.rs
@@ -22,6 +22,7 @@ impl PropertyEditorSchema {
     pub async fn assemble(
         ctx: &DalContext,
         schema_variant_id: SchemaVariantId,
+        has_resource: bool,
     ) -> PropertyEditorResult<Self> {
         let mut props = HashMap::new();
         let mut child_props = HashMap::new();
@@ -32,7 +33,7 @@ impl PropertyEditorSchema {
         let root_prop_id =
             Prop::find_prop_id_by_path(ctx, schema_variant_id, &PropPath::new(["root"])).await?;
         let root_prop = Prop::get_by_id(ctx, root_prop_id).await?;
-        let root_property_editor_prop = builder.build(ctx, root_prop).await?;
+        let root_property_editor_prop = builder.build(ctx, root_prop, has_resource).await?;
         let root_property_editor_prop_id = root_property_editor_prop.id;
         props.insert(root_property_editor_prop_id, root_property_editor_prop);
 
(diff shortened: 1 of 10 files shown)


sprutton1:
So, to use this in the spec we would need something like this?

```
...
{
  widgetOptions: {
    label: "si_create_only_prop"
  }
}
```

lib/dal/src/property_editor/schema.rs
```diff
@@ -145,6 +163,7 @@ impl PropertyEditorPropBuilder {
                 Some(prop_id) => prop_id == prop.id,
                 None => false,
             },
+            create_only: is_create_only && has_resource,
```
zacharyhamm, line 166: I don't think I understand the connection to the resource

zacharyhamm: nevermind i get it

stack72:
> So, to use this in the spec we would need something like this?
> 
> ```
> ...
> {
>   widgetOptions: {
>     label: "si_create_only_prop"
>   }
> }
> ```

No, we provide a helper func in the widget:

```
.setWidget(new PropWidgetDefinitionBuilder()
            .setKind("text")
            .setCreateOnly() <----------- ME!!!
            .build())
```

Means we can control the label and this is not a label that any other cloud provider will have as a dropdown variable

zacharyhamm (approved):
Tested with a new asset. Works as expected! No frontend regressions but may not fully grok the impact.

sprutton1:
> > So, to use this in the spec we would need something like this?
> > ```
> > ...
> > {
> >   widgetOptions: {
> >     label: "si_create_only_prop"
> >   }
> > }
> > ```
> 
> No, we provide a helper func in the widget:
> 
> ```
> .setWidget(new PropWidgetDefinitionBuilder()
>             .setKind("text")
>             .setCreateOnly() <----------- ME!!!
>             .build())
> ```
> 
> Means we can control the label and this is not a label that any other cloud provider will have as a dropdown variable

I mean within the pkgspec we generate. We aren't using the asset builder to create the specs. I'll mess around with it now that it's merged.

PR #5302, opened 2025-01-22
feat: write to funcs, reorg a bit

zacharyhamm:
Writes to funcs (tested in vim, more editors should be tested).

Reorgs so asset defs and funcs have their own locked/unlocked subfolders (per func).


fnichol (approved):
New code paths so we aren't affecting existing code. Test drove this and looks good!

PR #5307, opened 2025-01-22
fix(sdf, web): Protect against deleting last view

stack72:
There was no protection to stop a user from deleting ALL views. We now check to see if there's at least 1 view available before allowing a deletion to proceed

### DIFF
--- a/app/web/src/components/ViewCard.vue
+++ b/app/web/src/components/ViewCard.vue
@@ -88,6 +88,7 @@
       />
       <DropdownMenuItem
         label="Delete View"
+        :disabled="viewsStore.viewListCount === 1"
         icon="trash"
         :onSelect="() => deleteView(view)"
       />
--- a/lib/sdf-server/src/service/v2/view/remove_view.rs
+++ b/lib/sdf-server/src/service/v2/view/remove_view.rs
@@ -12,7 +12,7 @@ use crate::{
     track,
 };
 
-use super::ViewResult;
+use super::{ViewError, ViewResult};
 
 pub async fn remove_view(
     HandlerContext(builder): HandlerContext,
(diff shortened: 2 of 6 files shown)


lib/sdf-server/src/service/v2/view/remove_view.rs
```diff
@@ -28,6 +28,11 @@ pub async fn remove_view(
 
     let force_change_set_id = ChangeSet::force_new(&mut ctx).await?;
 
+    let current_view_list = View::list(&ctx).await?;
+    if current_view_list.len() == 1 {
```
jkeiser, line 32: This is a reasonable thing by itself, but don't we want to actually prevent users from deleting the default view? Like, ```
if (view_id == View::get_id_for_default(ctx)) {
  return Err(ViewError::CantDeleteDefaultView);
}
```

It seems like even if there are multiple views, we don't want the default view deleted.

Comment (stack72) [reply lib/sdf-server/src/service/v2/view/remove_view.rs:32]: We do not want to restrict that - we are going to default to another view if there is one

**Review by wendybujalski — APPROVED** [review]
👍

File: app/web/src/components/ViewCard.vue [hunk 2]
```diff
@@ -88,6 +88,7 @@
       />
       <DropdownMenuItem
         label="Delete View"
+        :disabled="viewsStore.viewListCount === 1"
```
britmyerss, line 91: 👏 this is such a simple but brilliant quality of life thing

britmyerss:
how did you test this? This is an area where we should add integration tests for - though it can be a follow on. 

I haven't pulled locally for testing just yet but the code looks correct to me.

stack72:
> how did you test this? This is an area where we should add integration tests for - though it can be a follow on.
> 
> I haven't pulled locally for testing just yet but the code looks correct to me.

Opened a change set
Tried to delete the DEFAULT view - button is locked
Added a new view called test
Delete the DEFAULT view and it successfully removes
Button is blocked for deleting the "test" view
Ensure that when I merge, the change set merges correctly and the "test" view is now set as the default view

nickgerace (approved):

PR #5313, opened 2025-01-23
fix(dal): Ensure we're only updating what is actually impacted by component deletion

britmyerss:
During rebase, when we discover a component has been removed that had connections to other components, only enqueue those impacted attribute values (and not all attribute values with the same socket id, as it was before)

This ensures that when change sets are applied to head with deleted components, we do detect that values need to update in other change sets and enqueue them appropriately.

<div><br/>via <a href="https://giphy.com/gifs/moodman-WQCv56djMTCmQkySAg">GIPHY</a></div>


zacharyhamm (approved):
Tested by removing components on head, ensuring only the connected component had its values recomputed in the change set and in another change set with the components that were on head. Great work!

nickgerace (commented):
Great work!!

PR #5318, opened 2025-01-23
BUG-710: don't throw, reset copying

jobelenus:
The current scheme for copy & paste won't work across views... just nullifying the bad stuff here


stack72 (approved):
So nice find!

PR #5320, opened 2025-01-23
feat(si-fs): create funcs except attr and action, edit bindings

zacharyhamm:
Also adds a cache to the si-fs client to prevent hammering the backend
when filesystem clients make lots of syscalls.


fnichol (approved):
Aside from changing some SDF function visibilities, these changes are still confined to si-fs

PR #5321, opened 2025-01-23
Feat: connections panel now has management sockets

jobelenus:
Testing Steps:

1. I had the "Basic Template Pattern" mgmt component on the diagram in a separate view
2. I ensured that another component could connect its `Manager` input socket to that mgmt component.
3. I refreshed the page to make sure it would be there. I deleted the connection. And refreshed the page again
4. I made the connection and deleted it (w/o refresh) to ensure that would work.
5. I added two Tailscale Users to the diagram because they can manage one another
6. I made sure I could set and delete both output and input sockets accordingly
7. Try to create a loop in the graph, and see the error message that you can't, and you see the optimistic edge removed
8. I made sure that the dropdown did not have duplicate entries (e.g. output socket "Manage" dropdown had *both* input and output sockets in it)

### DIFF
--- a/app/web/src/components/ComponentConnectionsPanel.vue
+++ b/app/web/src/components/ComponentConnectionsPanel.vue
@@ -230,21 +230,19 @@ const sockets = computed(() => {
 
   const peersFunction = componentsStore.possibleAndExistingPeerSocketsFn;
   const sockets =
-    selectedComponent.def.sockets
-      .filter((s) => !s.isManagement)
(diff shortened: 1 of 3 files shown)


jobelenus:
@zacharyhamm I'd love if we could reconsider a few decisions we made in the past... just don't know their ramifications...

stack72 (approved):
DWIGHT SCHRUTE IS MANAGER!

PR #5323, opened 2025-01-23
feat(clover): Make create action work

vbustamante:
- Rename generator to Clover
- Add credential and propmap to domain.extra
- fixup codegeneration for update and create
- Make sure the create actions works, update doesn't yet


bin/lang-js/BUCK
```diff
@@ -24,6 +24,8 @@ alias(
     actual = ":bin"
 )
 
+# we include clover here because the deno workspace
+# requires resolution of all members even if we don't use it
```
sprutton1, line 28: this is out of date, you might want to rebase main

sprutton1: On that note, after a rebase, you'll have another spot to update in the root BUCK file

bin/clover/src/pipeline-steps/addDefaultPropsAndSockets.ts
```diff
@@ -0,0 +1,149 @@
...
+        } else if (!prop.metadata.readOnly) {
+          propUsageMap.updatable.push(prop.name);
+        }
+
+        if (prop.kind === "object") {
+          prop.entries.forEach((p) => queue.unshift(p));
+        }
+      }
+
+      propUsageMapProp.data.defaultValue = JSON.stringify(propUsageMap);
```
sprutton1, line 78: I wonder how this handles on updates. Like, if we generate a new asset with a new createOnly prop, will it update the value?

sprutton1 (approved):

bin/clover/src/pipeline-steps/addDefaultPropsAndSockets.ts
vbustamante: It will skip the value when creating the payload

sprutton1 (dismissed):

britmyerss (approved):

britmyerss (approved):

PR #5327, opened 2025-01-24
Support local auth api on module-index

jkeiser:
We talk to the auth api from the module-index to figure out if you are a systeminit.com user. This honors the SI_AUTH_API_URL environment variable in the module-index (similarly as sdf). It defaults to the production API server.

### DIFF
--- a/dev/Tiltfile
+++ b/dev/Tiltfile
@@ -329,13 +329,13 @@ si_buck2_resource(
     resource_deps = [
         "postgres",
     ],
-    readiness_probe = probe(
-        period_secs = 5,
-        http_get = http_get_action(
-            port = 9001,
-            path = "/",
-        ),
-    ),
+    # readiness_probe = probe(
+    #     period_secs = 5,
+    #     http_get = http_get_action(
+    #         port = 9001,
+    #         path = "/",
+    #     ),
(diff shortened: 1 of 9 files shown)


dev/Tiltfile
```diff
@@ -329,13 +329,13 @@ si_buck2_resource(
     resource_deps = [
         "postgres",
     ],
-    readiness_probe = probe(
```
britmyerss, line 332: why is this commented out?

britmyerss:
@jkeiser how did you test this?

jkeiser:
> @jkeiser how did you test this?

@britmyerss I made sure the Modules tab stopped reporting an error locally, and I ran it with the new variable *not* set and verified that the whoami request went to the same auth-api URL as before. (Said request failed because the token was one of my local tokens and not a public token.)

dev/Tiltfile
jkeiser: Because it sucks watching it fill the logs, makes it hard to see what's going on. I added this check when I added auth-api to this file. Actually detecting whether the auth-api is up doesn't affect anything, so while it'd be nice to have, it's not as important as being able to debug :)

PR #5328, opened 2025-01-24
Deny SDF access to revoked tokens

jkeiser:
This denies automation API requests for revoked API tokens with 401.

It does this by making a status (`/`) request to the auth API with the given token. If the auth API denies the request, we return a 401 Unauthorized.

### Risk mitigation

This is code that runs on every single endpoint. However, the new check is wrapped in if (role != "web"), which means web UI never even checks if tokens are revoked.

The PR also modifies the auth-api-client, which is used by module-index to hit `whoami` and find out your email. The code flow is slightly different (we now strip `Bearer` when we create the client instead of when we make the request), and mitigation here was to test the module-index locally. The `/modules?su=true` endpoint successfully gets the email.

### Testing

* Tested that web requests (with web tokens) generally work: logging in, logging out, creating changesets and components, running management functions, and merging a changeset.
* Tested that github actions with API tokens *can* reach the automation API
* Tested that revoked API tokens can't reach the automation API
* Tested module index locally (the whoami requests succeed in getting your email to check `su`)

This is based on top of the Auth API revocation PR.


dev/Tiltfile
```diff
@@ -329,13 +329,13 @@ si_buck2_resource(
...
         "postgres",
     ],
-    readiness_probe = probe(
-        period_secs = 5,
-        http_get = http_get_action(
-            port = 9001,
-            path = "/",
-        ),
-    ),
+    # readiness_probe = probe(
```
jkeiser, line 332: This change is just because it's annoying seeing these pings in the logs. It only affects local dev and only for folks who are running local auth apis.

lib/si-jwt-public-key/src/lib.rs
```diff
@@ -124,6 +126,13 @@ pub struct SiJwtClaimsV1 {
 }
 
 impl SiJwtClaims {
+    pub fn token_id(token: &SiJwt) -> JwtKeyResult<Option<AuthTokenId>> {
+        match token.jwt_id {
```
zacharyhamm, line 130: you can wrap the whole match in Ok() and then you don't need an Ok in each match arm

lib/sdf-server/src/extract/request.rs
```diff
@@ -59,12 +59,21 @@ impl FromRequestParts<AppState> for ValidatedToken {
...
 
-        let RawAccessToken(raw_token) = parts.extract_with_state(state).await?;
+        let RawAccessToken(raw_token) = parts.extract().await?;
 
         let jwt_public_signing_key = state.jwt_public_signing_key_chain().clone();
         let token = validate_raw_token(jwt_public_signing_key, raw_token)
             .await
             .map_err(unauthorized_error)?;
+
+        // If it has role: Automation, check with auth API if the token has been revoked
```
jkeiser, line 69: This is the new code that runs everywhere. It's scoped specifically to avoid impact to web ui token usage (which can't be revoked anyway).

lib/sdf-server/src/service/public/change_sets.rs
```diff
@@ -39,7 +39,7 @@ pub fn routes(state: AppState) -> Router<AppState> {
             .route("/request_approval", post(request_approval))
             .route(
                 "/force_apply",
-                post(force_apply).layer(WorkspacePermissionLayer::new(
+                post(force_apply).route_layer(WorkspacePermissionLayer::new(
```
zacharyhamm, line 42: To be clear, the problem before was the layer applied to all routes and you just want it on this route ("force_apply")

zacharyhamm (approved):
No breaking changes for SDF when tested locally. I did not boot up the module index but the change doesn't look like it could cause problems (famous last words I guess).

lib/sdf-server/src/service/public/change_sets.rs
jkeiser: It only applies to this route; I was getting a situation where this layer was running before the parent's `route_layer()` (which set up the workspace id for it). My hypothesis was that all layer()s run before all route_layer()s. Either way, this fixed the ordering issue.

lib/si-jwt-public-key/src/lib.rs
jkeiser: I go back and forth over which of these looks better! I can never decide :(

britmyerss (approved):

PR #5330, opened 2025-01-24
Move list workspace users of out the admin routes

nickgerace:
## Introduction

This PR moves the "list workspace users" functionality out of the admin routes and into the workspace routes.

## Description

We do not want to use the admin DalContext nor the admin store unless we're in the AdminDashboard and this change allows anyone in the workspace to list workspace users, which is required for fine grained access control (and is generally useful).

However, the auth store location in the frontend and the workspace route addition in the backend are not necessarily the best places for this functionality to live either. This commit serves as a quick stepping stone that can be refactored when the domain begins to boil over.

## Additional Changes

Additional work in this commit includes, but is not limited to cleaning up workspace API error types and general re-organization of errors for readability and clarity.


app/web/src/store/change_sets.store.ts
```diff
@@ -674,7 +674,9 @@ export function useChangeSetsStore() {
           {
             eventType: "ChangeSetApprovalStatusChanged",
             callback: (changeSetId) => {
-              this.FETCH_APPROVAL_STATUS(changeSetId);
+              if (this.selectedChangeSet?.id === changeSetId) {
+                this.FETCH_APPROVAL_STATUS(changeSetId);
+              }
```
nickgerace, line 679: Long-standing "TODO" for this feature, which is flagged.

app/web/src/store/admin.store.ts
```diff
@@ -56,10 +52,11 @@ export const useAdminStore = () => {
             url: `${API_PREFIX}/workspaces/${workspaceId}/change_sets`,
           });
         },
+        // TODO(nick): remove in favor of the auth store call.
         async LIST_WORKSPACE_USERS(workspaceId: string) {
-          return new ApiRequest<{ users: AdminUser[] }>({
+          return new ApiRequest<{ users: WorkspaceUser[] }>({
             method: "get",
-            url: `${API_PREFIX}/workspaces/${workspaceId}/users`,
+            url: WORKSPACE_API_PREFIX.concat([workspaceId, "users"]),
```
nickgerace, line 59: Removing this would introduce the auth store to the admin dashboard, which we should avoid until we know for sure where this route will land for the longer-term future.

lib/sdf-server/src/service/v2/workspace/list_workspace_users.rs
```diff
@@ -0,0 +1,47 @@
...
+pub struct Response {
+    users: Vec<User>,
+}
+
+pub async fn list_workspace_users(
+    HandlerContext(builder): HandlerContext,
+    AccessBuilder(access_builder): AccessBuilder,
+    Path(workspace_id): Path<WorkspacePk>,
+) -> Result<Json<Response>, WorkspaceAPIError> {
+    let ctx = builder.build_head(access_builder).await?;
```
nickgerace, line 38: Like before, the context is built using the HEAD change set, which we may or may not want in the long term. However, this reflects existing behavior minus the requirement for an "admin" DalContext.

app/web/src/store/admin.store.ts
jobelenus: Endpoints can live in multiple stores without a problem...

jobelenus (approved):

wendybujalski (approved):
👍

britmyerss:
@nickgerace how did you test this?

nickgerace:
@britmyerss the following:

1. Access the admin dashboard
2. Enter my current workspace
3. Observe that I am in the users list and the route returns 200 OK
4. Also observe that no errors appear in the console

Then...

1. Open a second browser with a different user
2. Open the console and network tabs in each browser for error observation
3. In your original browser, add this user to your local workspace as a collaborator
4. In your second brwoser, create an asset, drag it on the diagram
5. In your second brwoser, Request approval
6. In your original browser, approve the change set
7. In your second browser, merge it
8. Observe that the new route was never called during this flow

britmyerss (approved):

lib/sdf-server/src/service/v2/workspace/list_workspace_users.rs
jkeiser: FYI, for future endpoints under v2, you can remove the Path() from this entirely and use `ctx.workspace_id()`, which already gets it from path. I didn't make this change for most of the endpoints simply because I wanted that churn to happen separately in a refactor PR (and my authentication PR was already big enough). What you are doing right now is perfectly in line with all the rest of v2.

PR #5337, opened 2025-01-24
Fix: we have the workspaceId in the URL, dont need to wait & race for the fetch all workspaces call here

jobelenus:


stack72 (approved):

nickgerace (approved):
OOOOOOOOOH that makes sense!!

PR #5339, opened 2025-01-25
docs: Adding reference docs for the GitHub Action

stack72:


app/docs/src/reference/github-action.md
```diff
@@ -0,0 +1,96 @@
...
+### Optional Parameters
+
+- **domain:** An object containing the domain properties to set for the
+  management function component. You may specify property keys as names,
+  property IDs, or paths (e.g. `{ "name/first": "John", "name/last": "Doe" }`).
+  Each property's value will be replaced with the new value. Properties not
+  specified will not be changed in the component. If not specified then no
+  properties will be set.
+
+- **workspaceId**: The ID of the workspace containing the component. If not
```
britmyerss, line 35: hmm what happens if you pass in a different workspace id? I thought this wouldn't work...

app/docs/src/reference/github-action.md
```diff
@@ -0,0 +1,96 @@
...
+  specified, the workspace that the token was generated for will be used.
+
+- **managementFunction:** The name of the management function to execute.
+  Optional if the component has a single management function.
+
+- **view:** The name of view in which the component should be updated. Optional
```
britmyerss, line 41: might be worth calling out this view is also where the mgmt function will execute in

app/docs/src/reference/github-action.md
stack72: I was working off the action.yml but I removed it now :)

app/docs/src/reference/github-action.md
stack72: great shout!

britmyerss (approved):

PR #5340, opened 2025-01-25
Protect against unforced apply without reqs met

nickgerace:
## Description

This PR adds a new route for applying a change set with fine grained access control. This route contains new dal wrapper logic to ensure that applying a change set (unforced) only happens if all requirements have been met. There's a new integration test for this as well.

## Other Changes

Other changes include, but are not limited to, the following:

- Split apply prepare step into two versions because sdf now handles the protection layer
- Rename the "change_set_approval" module to "change_set" in dal wrapper
- Add sdf-test helpers for mimicking protected apply behavior in the route, which is used by the new test

### DIFF
--- a/lib/dal/src/change_set.rs
+++ b/lib/dal/src/change_set.rs
@@ -420,8 +420,22 @@ impl ChangeSet {
     /// [`ChangeSetStatus::Approved`]. Finally,
     /// lock every [`SchemaVariant`] and [`Func`] that is currently unlocked
     pub async fn prepare_for_apply(ctx: &DalContext) -> ChangeSetResult<()> {
-        let change_set = ChangeSet::get_by_id(ctx, ctx.change_set_id()).await?;
+        Self::prepare_for_apply_inner(ctx, false).await
+    }
 
+    /// This is a copy of [Self::prepare_for_apply], but skips the status check. This is because
+    /// sdf now handles the approvals flow as part of the fine grained access control work (i.e.
+    /// SpiceDB is intentionally not accessible in the DAL).
+    pub async fn prepare_for_apply_without_status_check(ctx: &DalContext) -> ChangeSetResult<()> {
+        Self::prepare_for_apply_inner(ctx, true).await
+    }
+
+    // TODO(nick): when the fine grained access control feature flag dies, get rid of the separate
+    // functions (and this inner one) and collapse them into one again.
+    async fn prepare_for_apply_inner(
+        ctx: &DalContext,
+        dangerous_skip_status_check: bool,
+    ) -> ChangeSetResult<()> {
         // Ensure that DVU roots are empty before continuing.
         if !ctx
             .workspace_snapshot()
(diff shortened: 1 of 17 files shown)


lib/dal/src/change_set.rs
```diff
@@ -439,12 +453,16 @@ impl ChangeSet {
...
+        // WARNING(nick): we should only skip this status check if using sdf's protected apply logic.
+        if !dangerous_skip_status_check {
+            // if the change set status isn't approved, we shouldn't go
+            // locking stuff
+            let change_set = ChangeSet::get_by_id(ctx, ctx.change_set_id()).await?;
+            if change_set.status != ChangeSetStatus::Approved {
+                return Err(ChangeSetError::ChangeSetNotApprovedForApply(
+                    change_set.status,
+                ));
+            }
```
nickgerace, line 465: We are using approval requirements instead of a database row's column's based value. We now know if approvals have gone stale and we need to know if the requirements are met at an _**sdf**_ level because it has access to SpiceDB. This will appear later in the PR.

lib/sdf-server/src/dal_wrapper/change_set.rs
```diff
@@ -15,6 +15,49 @@ use super::DalWrapperError;
...
+    ChangeSet::prepare_for_apply_without_status_check(ctx).await?;
+
+    // We need to commit our preparations before performing the apply.
+    ctx.commit().await?;
+
+    // With the requirement check and preparations committed, we can finally perform the apply.
+    ChangeSet::apply_to_base_change_set(ctx).await?;
+
+    Ok(())
+}
```
nickgerace, line 59: This is the primary change of the PR. Now, we will block on unsatisfied requirements rather than the status field.

lib/sdf-server/src/service/v2/change_set/apply_v2.rs
```diff
@@ -0,0 +1,69 @@
...
+) -> Result<()> {
+    let mut ctx = builder
+        .build(request_ctx.build(change_set_id.into()))
+        .await?;
+    let spicedb_client = state
+        .spicedb_client()
+        .ok_or(ChangeSetAPIError::SpiceDBClientNotFound)?;
+
+    // Perform the protected apply flow.
+    dal_wrapper::change_set::protected_apply_to_base_change_set(&mut ctx, spicedb_client).await?;
```
nickgerace, line 30: One-liner check

lib/sdf-server/tests/service_tests/change_set_apply.rs
```diff
@@ -0,0 +1,184 @@
...
+                    unsatisfied_requirements,
+                )) => assert_eq!(
+                    vec![(todd_view_id.into_inner().into(), EntityKind::View)], // expected
+                    unsatisfied_requirements.to_owned()                         // actual
+                ),
+                Some(err) => return Err(eyre!("unexpected error: {err}")),
+                None => return Err(eyre!("unexpected report: {report:?}")),
+            },
+            other => return Err(eyre!("unexpected result: {other:?}")),
+        }
```
nickgerace, line 120: This is what happens when you try to apply with unsatisfied requirements

lib/sdf-server/tests/service_tests/change_set_apply.rs
```diff
@@ -0,0 +1,184 @@
...
+            }], // expected
+            frontend_requirements // actual
+        );
+    }
+
+    // Scenario 5: apply the changes used the protected flow and observe that it works.
+    {
+        SdfTestHelpers::protected_apply_change_set_to_base(ctx, &mut spicedb_client).await?;
+        ChangeSetTestHelpers::fork_from_head_change_set(ctx).await?;
+
```
nickgerace, line 166: Smooth sailing if everything has been satisfied

lib/sdf-test/src/helpers.rs
```diff
@@ -0,0 +1,33 @@
...
+        // We need to prepare for apply, but use the new version that skips the status check. Why?
+        // Our new approval requirements check above replaces the old status check.
+        ChangeSet::prepare_for_apply_without_status_check(ctx).await?;
+
+        // We can mostly follow the DAL test flow for applying a change set while factoring in
+        // approvals. The "mostly" part comes from the fact that we need to perform our own
+        // "prepare" step above rather than using the one from the DAL test helpers.
+        ChangeSetTestHelpers::apply_change_set_to_base_approvals_without_prepare_step(ctx).await?;
+        Ok(())
+    }
```
nickgerace, line 32: This is a very similar version to what's in the new sdf route, but it looks similar to `ChangeSetTestHelpers` apply patterns in `dal-test`. Why isn't this in `dal-test` then? It needs SpiceDB.

britmyerss (approved):
Great work!

PR #5342, opened 2025-01-25
feat: generate sub-assets from specs

sprutton1:

### DIFF
--- a/bin/clover/src/commands/generateSiSpecs.ts
+++ b/bin/clover/src/commands/generateSiSpecs.ts
@@ -10,6 +10,8 @@ import { generateDefaultLeafFuncs } from "../pipeline-steps/generateLeafFuncs.ts
 import { generateDefaultManagementFuncs } from "../pipeline-steps/generateManagementFuncs.ts";
 import { addDefaultPropsAndSockets } from "../pipeline-steps/addDefaultPropsAndSockets.ts";
 import { generateSocketsFromResourceProps } from "../pipeline-steps/generateSocketsFromResourceProps.ts";
+import { generateSubAssets } from "../pipeline-steps/generateSubAssets.ts";
+import { generateIntrinsicFuncs } from "../pipeline-steps/generateIntrinsicFuncs.ts";
 
 export function generateSiSpecForService(serviceName: string) {
   const cf = getServiceByName(serviceName);
@@ -38,10 +40,14 @@ export async function generateSiSpecs() {
   specs = generateSocketsFromDomainProps(specs);
   specs = generateSocketsFromResourceProps(specs);
   specs = addDefaultPropsAndSockets(specs);
-  specs = generateAssetFuncs(specs);
   specs = generateDefaultActionFuncs(specs);
   specs = generateDefaultLeafFuncs(specs);
   specs = generateDefaultManagementFuncs(specs);
+  // subAssets should not have any of the above, but need an asset func and
+  // intrinsics
+  specs = generateSubAssets(specs);
+  specs = generateIntrinsicFuncs(specs);
+  specs = generateAssetFuncs(specs);
 
   // WRITE OUTS SPECS
   for (const spec of specs) {
--- a/bin/clover/src/pipeline-steps/generateIntrinsicFuncs.ts
+++ b/bin/clover/src/pipeline-steps/generateIntrinsicFuncs.ts
@@ -1,0 +1,25 @@
+import { PkgSpec } from "../bindings/PkgSpec.ts";
+import _ from "npm:lodash";
+import {
+  createNormalizeToArray,
+  createResourcePayloadToValue,
+  createSiFuncs,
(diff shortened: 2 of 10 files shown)


bin/clover/src/commands/generateSiSpecs.ts
```diff
@@ -38,10 +40,14 @@ export async function generateSiSpecs() {
...
   specs = addDefaultPropsAndSockets(specs);
-  specs = generateAssetFuncs(specs);
   specs = generateDefaultActionFuncs(specs);
   specs = generateDefaultLeafFuncs(specs);
   specs = generateDefaultManagementFuncs(specs);
+  // subAssets should not have any of the above, but need an asset func and
+  // intrinsics
+  specs = generateSubAssets(specs);
+  specs = generateIntrinsicFuncs(specs);
+  specs = generateAssetFuncs(specs);
```
nickgerace, line 50: I see

bin/clover/src/pipeline-steps/generateIntrinsicFuncs.ts
```diff
@@ -0,0 +1,25 @@
...
+  const newSpecs = [] as PkgSpec[];
+
+  for (const spec of specs) {
+    const funcs = spec.funcs;
+
+    spec.funcs = [
+      ...funcs,
+      ...createSiFuncs(),
+      createResourcePayloadToValue(),
+      createNormalizeToArray(),
```
nickgerace, line 19: 👍

bin/clover/src/pipeline-steps/generateSubAssets.ts
```diff
@@ -0,0 +1,134 @@
...
+import { SchemaVariantSpec } from "../bindings/SchemaVariantSpec.ts";
+import { attrFuncInputSpecFromSocket } from "../spec/sockets.ts";
+import { createSocket } from "../spec/sockets.ts";
+import { attrFuncInputSpecFromProp } from "../spec/sockets.ts";
+import { getSiFuncId } from "../spec/siFuncs.ts";
+
+export function generateSubAssets(specs: PkgSpec[]): PkgSpec[] {
+  const newSpecs = [] as PkgSpec[];
+  for (const spec of specs) {
+    const schema = spec.schemas[0];
```
nickgerace, line 18: Why pop off the first schema rather than iterate?

bin/clover/src/pipeline-steps/generateSubAssets.ts
```diff
@@ -0,0 +1,134 @@
...
+    if (!schema) {
+      console.log(
+        `Could not generate default props and sockets for ${spec.name}: missing schema`,
+      );
+    }
+    const schemaVariant = schema.variants[0];
```
nickgerace, line 24: I assume the zero indexed variant is the default. Is that always true and if not, is that even relevant?

bin/clover/src/pipeline-steps/generateSubAssets.ts
```diff
@@ -0,0 +1,134 @@
...
+          newSpecOutputSocket.inputs.push(attrFuncInputSpecFromProp(newDomain));
+        }
+
+        const variantData = _.cloneDeep(schemaVariant.data);
+        const variant: SchemaVariantSpec = {
+          ...schemaVariant,
+          data: {
+            ...variantData,
+            displayName: name,
+            funcUniqueId: ulid(),
```
nickgerace, line 84: Okay I can buy this. It mimics what import in the dal does.

bin/clover/src/pipeline-steps/generateSubAssets.ts
```diff
@@ -0,0 +1,134 @@
...
+          prop.data.funcUniqueId = getSiFuncId("si:normalizeToArray");
+          schemaVariant.sockets.push(propInputSocket);
+        }
+
+        // output the domain of this new spec
+        const newSpecOutputSocket = createSocket(objName, "output", "one");
+        if (newSpecOutputSocket.data) {
+          newSpecOutputSocket.data.funcUniqueId = getSiFuncId("si:identity");
+          newSpecOutputSocket.inputs.push(attrFuncInputSpecFromProp(newDomain));
+        }
```
nickgerace, line 76: These funcs should already exist based on the order of your code so 👍

nickgerace (approved):
I understand `clover` at a high level, and approve this based on logic flow and the architecture. Nice!

bin/clover/src/pipeline-steps/generateSubAssets.ts
vbustamante: we know the generator should always generate one schema per spec always

bin/clover/src/pipeline-steps/generateSubAssets.ts
vbustamante: similar to above, it's a requirement of the generator to create one schema per spec and a single variant per schema for every variant, and that one will be default.

PR #5346, opened 2025-01-25
feat(sdf): Record whether automation tokens did an action in audit logs

jkeiser:
This adds an "authenticationMethod" to the audit logs, which tells you whether the user authenticated via web or API token, and also contains the token id so we can filter and do forensics to figure out what token did bad stuff and needs revoking as well as look over what revoked tokens *did*.

image

## Testing

* Created change set and components in UI, applied change set, checked that audit log ui showed them as web usage
* Upgraded existing database, checked that System and User logs had the right authentication method in audit log ui
* Ran automated actions, checked that audit log ui showed them as automation tokens with token id


app/web/src/components/AuditLogCell.vue
```diff
@@ -36,29 +36,13 @@ import { IconButton, themeClasses } from "@si/vue-lib/design-system";
...
-          entityType: string;
-          entityName: string;
-          changeSetId?: string;
-          changeSetName: string;
-          metadata: Record<string, unknown>;
-        },
-        unknown
-      >
-    >,
+    type: Object as PropType<Cell<AuditLogDisplay, unknown>>,
```
nickgerace, line 45: Thanks!

app/web/src/store/logs.store.ts
```diff
@@ -121,16 +130,8 @@ export const useLogsStore = (forceChangeSetId?: ChangeSetId) => {
                 this.logs = response.logs.map(
                   (log) =>
                     ({
-                      title: log.title,
+                      ...log,
```
nickgerace, line 133: Thanks again!

lib/audit-database/src/migrations/U0003__authentication_method.sql
```diff
@@ -0,0 +1,10 @@
+ALTER TABLE audit_logs
+ADD COLUMN authentication_method JSONB;
+-- Set authentication_method = 'Jwt(Web)' for rows where user_id is not null
+UPDATE audit_logs
+SET authentication_method = '{"method":"Jwt","role":"Web"}'
+WHERE user_id IS NOT NULL;
```
nickgerace, line 6: Will this always be true for everything prior to this PR? Or will there be example where different methods and roles were used?

lib/dal/src/context.rs
```diff
@@ -1063,6 +1075,14 @@ impl DalContext {
     }
 }
 
+// We have a few patterns of request context
+// enum dRequestContext {
+//     ChangeSet(ChangeSetScope),
+//     Workspace(WorkspaceScope),
+//     Admin(AdminScope),
+//     Account(AccountScope),
+// }
```
nickgerace, line 1084: We should delete this

lib/si-events-rs/src/authentication_method.rs
```diff
@@ -0,0 +1,28 @@
...
+
+#[derive(AsRefStr, Clone, Debug, Copy, Deserialize, Eq, PartialEq, Serialize, Display, Hash)]
+#[serde(tag = "method")]
+pub enum AuthenticationMethodV1 {
+    /// For revocable SDF tokens (particularly Automation tokens)
+    /// Appears as { "method": "Jwt", "role": "Web" | "Automation", "token_id": "<...>" }
+    Jwt {
+        role: AuthenticationMethodRoleV1,
+        token_id: Option<AuthTokenId>,
+    },
```
nickgerace, line 16: Does the token come based on the role? If so, we may want to explicitly group the token. Examples:

```rust
enum Method {
    JwtWeb,
    JwtAutomation(AuthTokenId),
    System
}
```

Or...

```rust
enum Method {
    Jwt(JwtMethodRole),
    Automation,
}

enum JwtMethodRole {
    Web,
    Automation(AuthTokenId)
}
```

lib/audit-database/src/migrations/U0003__authentication_method.sql
jkeiser: Everything prior to this PR was either a web token or a system thing, since automation roles are still under feature flag.

(Well, except for a few of our prod workspaces under feature flag, but we're OK with those old audit logs lying to us that it was a webui-initiated change.)

lib/si-events-rs/src/authentication_method.rs
jkeiser: I went through this thought process too, and landed where I did mainly because the permutations were odd.

For example:
* we anticipate the ability to add more roles in the future. I'd rather not have to upgrade again just for that.
* web tokens will likely need the ability to have token IDs in the future. (In my opinion we can't live without this: you have to be able to revoke web tokens without disabling a user entirely, because they can be stolen just like an automation token can.)

Both of those suggest a generic method like `Jwt(Role, Option<TokenId>)`.

There is a reasonable compromise, however:

```
enum Method {
  System,
  JwtV1, // V1 web tokens
  Jwt(Role, AuthTokenId), // V2 web tokens
}
```

This would mean we can't move to V2 tokens unless we also enable revocability for user tokens, but it would work pretty well. You can even tweak the serialization so they serialize like NULL = System, "JwtV1" = JwtV1, and ["<role>", "<token id>"] = Jwt(Role, AuthTokenId)

jkeiser: We went with the simpler form for now because we don't really know what the future holds, and we versioned these things so we could make different decisions later.

nickgerace: We talked a bit offline and while there are potential ways to improve this further, we need to try this in the real world first.

nickgerace (approved):

PR #5347, opened 2025-01-25
fix(web): Drop propertyWidget from prop snippet

stack72:
This is not required for the majority of our props


adamhjk (approved):
we hardly needed ye

PR #5350, opened 2025-01-26
Let web rebuild without rebuilding all the rust

jkeiser:
`web` doesn't get much startup build time improvement from being included in the pack of rust apps, and including it in the frontend apps means restarting just web will wait for sdf and friends to build. This breaks that dependency. web still waits for sdf, so we're not losing anything in terms of dependency. Initial build likely isn't affected at all, but the web startup is fast enough that it's not worth trying to parallelize it :)

### DIFF
--- a/dev/Tiltfile
+++ b/dev/Tiltfile
@@ -44,7 +44,6 @@ groups = {
 }
 
 core_backend_targets = [
-    "//app/web:dev",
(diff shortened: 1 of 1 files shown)


fnichol (approved):
I have to say, that's a beautiful change!

PR #5356, opened 2025-01-27
feat(si-fs): create schemas with mkdir

zacharyhamm:
`mdkir schema-name` in the schemas folder will create a new schema


fnichol (approved):
Only net-new code, doesn't affect existing functionality

PR #5357, opened 2025-01-27
chore(tilt): Only rebuild individual service on Restart in tilt

jkeiser:
This makes it so when you click the little Restart button next to rebaser, forklift, pinga, veritech, or sdf, they will only rebuild what is necessary for *themselves* to restart. This speeds up restart.

At the same time, we preserve the new "build everything in a single command for performance" on tilt startup by creating a new "rust-backend-build" resource, which those 5 services all depend on. On startup, they will wait for this; but on restart, this won't be retriggered (this is how tilt works in general, inter-service dependencies only matter on startup).

The net effect is to speed up service rebuilds without slowing down initial startup.

### DIFF
--- a/dev/Tiltfile
+++ b/dev/Tiltfile
@@ -23,6 +23,7 @@ groups = {
         "rebaser",
         "sdf",
         "veritech",
+        "rust-initial-build",
     ],
     "frontend": [
         "auth-portal",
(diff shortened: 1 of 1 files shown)


britmyerss (approved):
I used this locally the last few days - great idea!

PR #5358, opened 2025-01-27
feat(si-fs): create action and attribute functions with mkdir

zacharyhamm:
Call mkdir in the actions or attributes folder to create a "pending"
action or attribute function. A file "PENDING_BINDINGS_EDIT_ME.json"
will appear in the new folder. Edit that file with a correct action or
correct output location for a function, and the save will create the
pending function.

You can also configure function arguments in the bindings for attribute
functions now.


fnichol (approved):
We're evolving the `v2/fs` SDF endpoints and don't have any shared call paths with existing behavior, so this again boils down to net-new code.

PR #5360, opened 2025-01-28
fix(dal): Allow attribute values set by dynamic funcs to be set by management funcs if the value is not controlled by an ancestor

britmyerss:
This change means that if you try to set a value from a management function, and the value would otherwise be set by another function (for example the region prop takes it's input from an input socket), we will override the value and override the default prototype.  

Tested by writing management functions that do this, and added a integration test.

<div><br/>via <a href="https://giphy.com/power/">Power</a> on <a href="https://giphy.com/gifs/power-starz-season6-episode605-SqTa7KFVfBJ9b1QFPo">GIPHY</a></div>

### DIFF
--- a/lib/dal/tests/integration_test/management.rs
+++ b/lib/dal/tests/integration_test/management.rs
@@ -709,6 +709,48 @@ async fn deeply_nested_children(ctx: &DalContext) {
 }
 
 #[test]
+async fn override_values_set_by_sockets(ctx: &DalContext) {
+    let small_odd_lego = create_component_for_default_schema_name_in_default_view(
+        ctx,
+        "small odd lego",
+        "small odd lego",
+    )
+    .await
+    .expect("could not create component");
+
+    let (execution_result, result) =
+        exec_mgmt_func(ctx, small_odd_lego.id(), "Override Props", None).await;
+    assert_eq!(result.status, ManagementFuncStatus::Ok);
+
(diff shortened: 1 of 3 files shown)


lib/dal/tests/integration_test/management.rs
```diff
@@ -708,6 +708,48 @@ async fn deeply_nested_children(ctx: &DalContext) {
...
+    let component = Component::get_by_id(ctx, child_id).await.expect("get comp");
+
+    let props = component
+        .domain_prop_attribute_value(ctx)
+        .await
+        .expect("could not get domain");
+    let domain = AttributeValue::get_by_id(ctx, props)
+        .await
+        .expect("could not get attribute value");
+    let view = domain.view(ctx).await.expect("could not get view");
```
zacharyhamm, line 749: This test doesn't seem to assert that the value was overriden?

zacharyhamm (commented):
Building locally to test it out and confirm, but the test doesn't seem to assert the override

zacharyhamm (approved):
I added an assertion to the test. This approval can't mark this as mergeable because I'm on the branch now.

zacharyhamm:
I've tested this PR and it looks good. I wanted to make sure the test asserted the override condition, but forgot that that means my approval is no bueno.

stack72 (approved):

PR #5361, opened 2025-01-28
fix: reorganize sockets logic

sprutton1:
* move socket create logic out to separate steps
* create output sockets for props in the resource
* create input sockets by walking through all created output sockets and creating an input socket for props with a matching name
* consolidate most socket creation to using the same function calls


bin/clover/src/spec/props.ts
```diff
@@ -289,7 +289,7 @@ export function createScalarProp(
     validationFormat: null,
     defaultValue: null,
     funcUniqueId: null,
-    inputs: null,
+    inputs: [],
```
vbustamante, line 292: I wonder if this can break something on import, making all props think they have values set by functions. Did you test for that?

bin/clover/src/pipeline-steps/generateSubAssets.ts
```diff
@@ -88,6 +88,15 @@ export function generateSubAssets(incomingSpecs: PkgSpec[]): PkgSpec[] {
...
         }
 
+        // reuse existing assets so we don't recreate the same asset over and
+        // over again
+        const maybeExistingSubAsset = newSpecsByHash[hash];
+        if (maybeExistingSubAsset) {
+          logger.debug(`Existing subasset found: ${name}`);
+          maybeExistingSubAsset.names.push(name);
+          continue;
+        }
```
vbustamante, line 98: any reason to move this lower? I feel like stopping it as early as possible is best

bin/clover/src/spec/props.ts
sprutton1: Not having this breaks setting region and aws credential as we don't set the input on the props at all if the list is null. Imported assets seem good

bin/clover/src/pipeline-steps/generateSubAssets.ts
sprutton1: because we weren't creating the input sockets on the parent prop by skipping where we were

vbustamante (approved):
LGTM

PR #5365, opened 2025-01-28
Feat: Show approvers the breakdown of changes

jobelenus:
image

Here is a breakdown of the changes:

- there is no more change set status === `rejected` or `approved`
- we look at "are all requirements `isSatisfied===true` to determine if someone can apply
- any one can withdraw
- we've updated how we determine "who is an approver" and "how many changesets they can approve"
- any one can "apply", so long as every requirement `isSatisfied`
- you can offer a vote if you're in any approver group of any requirement.

This needs solid testing both with `WORKSPACE_FINE_GRAINED_ACCESS_CONTROL = true` as well as `false`.


stack72 (approved):
ok, I've tested this with the **feature flag turned off** - it is a new modal but we know that will change this week - the fact it works is the most important part!

I verified with @britmyerss that we are good to launch this for a day or 2 before we turn on the feature flag for all

nickgerace (approved):
Tested locally with the flag on and off. I'm approving @jobelenus's changes and not mine.

Super nice job @jobelenus!! 🔥 this worked great. Swapping from reject to approve and back and forth was great too.

PR #5368, opened 2025-01-29
fix(web): Small turn to expose the prop path in the intrinsics panel

stack72:
This is a stop gap solution until we redesign that panel. This is going to be useful in the interim

Screenshot 2025-01-29 at 00 19 40


wendybujalski (approved):
👍

PR #5371, opened 2025-01-29
feat(hoist): reject and promote new modules

sprutton1:
This ensures that when we push a new module, if a module already exists with the same schema, we reject the existing module and then promote ours as a built-in.

Also adds a progress bar for hoist and some basic throttling to ensure we don't lock up when uploading many packages to the module index. The module index should really handle S3 throttling .


stack72 (approved):
SHIP IT!!

PR #5372, opened 2025-01-29
Revert "fix(sdf, web): Protect against deleting last view"

stack72:
Reverts systeminit/si#5307


stack72:
Not the source of an issue we are investigating!

PR #5375, opened 2025-01-29
fix(dal, rebaser, pinga): If a change set is abandoned, do not process rebase requests or execute jobs. 

britmyerss:
If it's a DVU job, we need to remove the roots so it's not enqueued again.

I will follow up with an integration test for this if it works but want to get this into Tools ASAP.

<div><br/>via <a href="https://giphy.com/AppleTV/">Apple TV</a> on <a href="https://giphy.com/gifs/AppleTV-apple-tv-app-iPRnsx9RzW5e0oQOsJ">GIPHY</a></div>

### DIFF
--- a/lib/dal/src/job/consumer.rs
+++ b/lib/dal/src/job/consumer.rs
@@ -15,7 +15,6 @@ use crate::billing_publish::BillingPublishError;
 use crate::diagram::DiagramError;
 use crate::prop::PropError;
 use crate::validation::ValidationError;
-use crate::FuncError;
 use crate::{
     action::prototype::ActionPrototypeError, action::ActionError,
     attribute::value::AttributeValueError,
@@ -24,6 +23,7 @@ use crate::{
     ActionPrototypeId, ComponentError, ComponentId, DalContext, DalContextBuilder,
     StandardModelError, TransactionsError, Visibility, WorkspaceSnapshotError, WsEventError,
 };
+use crate::{ChangeSetError, FuncError};
 
 #[remain::sorted]
 #[derive(Error, Debug)]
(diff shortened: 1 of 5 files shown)


jhelwig:
This wouldn't help for any jobs already enqueued, but we should add a check to the rebaser to not dispatch any DVU or actions for abandoned change sets in the first place, so it doesn't clog up the queue & pinga capacity for pinga to do filtering there.

So, we'll probably want both pinga-side filtering, and rebaser-side not-dispatching-in-the-first-place for now.

johnrwatson:
@britmyerss @jhelwig should we maybe change to logic to be != Open, instead of specifically just Abandoned exclusion. I'm thinking it's probably still possible for an Applied changeset to fall into the same trap/similar

zacharyhamm (approved):

jhelwig:
The plan is to use `is_active()` on the status for the filtering as we also want waiting for approval to be processed in addition to open.

zacharyhamm:
image

hitting the abandon button right after making a change that would enqueue a dvu and it bails

johnrwatson:
> image
> 
> hitting the abandon button right after making a change that would enqueue a dvu and it bails

Minor nit, but I'm not sure it should be ERROR if it's handled in the code as a proper exception

britmyerss:
> > image
> 
> > 
> 
> > hitting the abandon button right after making a change that would enqueue a dvu and it bails
> 
> 
> 
> Minor nit, but I'm not sure it should be ERROR if it's handled in the code as a proper exception

It's an error in pinga right now because the job started before the change set was abandoned, so when it submits the rebase request it fails as the rebaser knows not to handle that request. We can live with it for now as we know we need deeper work here.

PR #5378, opened 2025-01-29
Check to see if Components have been moved to other Views when removing a View

jhelwig:
When removing a View, we make sure that it isn't the only View that has Geometry for any of the Components in it. We weren't taking into account that a Component might be moved from only being in the View we're removing to being in another View (or multiple other views).

We now track which Components got new Geometry for Views other than the one being removed, and if we see that the View being removed had the only Geometry for the Component, we see if it is in the list of Components that are getting new Geomerty. If it is in that list, then we allow the removal to proceed as we won't be orphaning the Component from all Views.

### DIFF
--- a/lib/dal/src/workspace_snapshot/node_weight/view_node_weight/v1.rs
+++ b/lib/dal/src/workspace_snapshot/node_weight/view_node_weight/v1.rs
@@ -1,4 +1,4 @@
-use std::collections::HashSet;
+use std::collections::{HashMap, HashSet};
 
 use crate::{
     workspace_snapshot::{
@@ -58,6 +58,8 @@ impl CorrectTransforms for ViewNodeWeightV1 {
         let mut maybe_view_removal_update_idx = None;
         let mut removed_geometries = HashSet::new();
         let mut removed_components = HashSet::new();
+        let mut components_with_new_geometry: HashMap<Ulid, HashSet<Ulid>> = HashMap::new();
+        let mut new_geometry_for_other_views: HashSet<Ulid> = HashSet::new();
 
         for (update_idx, update) in updates.iter().enumerate() {
             match update {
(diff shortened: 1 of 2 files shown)


lib/dal/src/workspace_snapshot/node_weight/view_node_weight/v1.rs
```diff
@@ -146,9 +174,25 @@ impl CorrectTransforms for ViewNodeWeightV1 {
...
+                            // If the Component is either in other views, or is being added to (at least)
+                            // another View in the same set of Transforms, then removing this view is fine
+                            // since it won't orphan any Components.
                             if !appears_in_views.is_empty() {
                                 continue;
                             }
+                            // The Component needs to both be getting new Geometry from the set of Update,
+                            // _and_ at least one of those Geometry need to be for a View other than the
+                            // one being removed.
+                            dbg!(&components_with_new_geometry, &new_geometry_for_other_views);
```
nickgerace, line 186: We should probably remove this, but it's non-blocking

lib/dal/tests/integration_test/node_weight/view.rs
```diff
@@ -309,3 +310,69 @@ async fn correct_transforms_remove_view_with_component_in_another_view(ctx: &mut
...
+
+    assert_eq!(
+        1,
+        View::list(ctx).await.expect("Unable to list Views").len(),
+    );
+    assert_eq!(
+        1,
+        Geometry::list_ids_by_component(ctx, component.id())
+            .await?
+            .len(),
```
nickgerace, line 359: I don't _think_ we need to commit here since nothing is relying on DVU/rebases... should be good

lib/dal/tests/integration_test/node_weight/view.rs
```diff
@@ -309,3 +310,69 @@ async fn correct_transforms_remove_view_with_component_in_another_view(ctx: &mut
...
+    );
```
nickgerace, line 375: 👍

nickgerace (approved):
Approving since my comments do not warrant blocking given the severity of the bug this change fixes.

PR #5382, opened 2025-01-30
feat(mgmt): Allow management functions to connect their input to their progeny

jkeiser:
This allows your management function to pass on its input connections to any components it creates or manages. It adds three basic functionalities to achieve this:

1. **Reading your input sockets:** `thisComponent` now has a new `incomingConnections` property, containing all incoming connections for each input socket, with `component` and `socket` for each one.
2. **Connecting input sockets:** `ops.create.<component>.connect` now supports input sockets in addition to output sockets (just specify `{ from: { component, socket }, to: string }`). (`ops.update.<component>.connect.add/remove` support this as well.)
3. **Connecting to non-managed components:** You can now connect sockets to arbitrary components by component ID. If `component` is set to a ULID, it will connect to that component (managed and created components are still supported).

image

## Example

This management function creates a VPC and connects its "AWS Credentials" and Region inputs to the same ones hooked up to the management component itself:

```ts
async function main({ thisComponent: Input): Promise<Output> {
    return {
        status: "ok",
        ops: {
            create: {
                MyVpc: {
                    kind: "VPC",
                    connect: [
                        {
                            from: thisComponent.incomingConnections["AWS Credential"],
                            to: "AWS Credential",
                        },
                        {
                            from: thisComponent.incomingConnections.Region,
                            to: "Region",
                        },
                    ]
                }
            }
        }
    }
}
```

## For Later

There are some things we could do to make this easier to use in another turn:

* **Support undefined/array in from:** The recommended pattern for single-arity sockets like Region is to just shove Region into `ops.create.connect[].from` like in the example. However, if the user does not hook up the socket, the error message is confusing (can't deserialize undefined). We should make connect treat `from: undefined` as "don't connect anything here" so the user doesn't have to. We should also let it support `from: <array>`, so you can directly shove arity-many sockets in here and have that work too.
* **Pass output socket connections:** Output sockets, while less useful, seem perfectly doable here and probably should happen for parity.
* **Mirror incomingConnections on create:** If we allowed you to specify `incomingConnections: { <socket>: { component; socket } }` on ops.create[] instead of just `connect`, it would improve the experience:
  - It would let you write `incomingConnections: { Region }` instead of `connect: [ { from: Region, to: "Region" } ]`.
  - Components with all the same input sockets (very common) could write `{ incomingConnections }` instead of `{ connect: [ { from: Region, to: "Region" }, { from: AWSCredential, to: "AWS Credential" } ] }`.
  - If we then allowed incomingConnections to ignore extra sockets, you could use incomingConnections even more places.
* **inferred:** We should pass inferred connections, but add `inferred: true` so that you can decide whether to connect based on that or not.
* **Error case:** When you have two connections to a socket with arity="one", we pass you the first one instead of telling you there is an error.
* **Name bikeshed:** `connections.input` might work better for this (incoming is a mouthful to say and type; less capitals is nice; it gives an obvious place to hang output connections). `sockets.input` maybe too.

## Testing

This still needs regression tests, but we'd like to get this up quickly for John (I'll fast follow today). What I've tested:

1. Check that we get the right incomingConnections:
   - For 0, 1, N connections
   - For arity: one and arity: many
   - Inferred connections
   - Error message if there are multiple connections and arity: one
   - Right value for connections with secrets, strings, object and array typed data
2. Check that we can create and remove connections from ops.reate.connect and ops.update.connect.add/remove:
   - Referencing existing managed components
   - Referencing other components being added in ops.create
   - Referencing external component ids
   - Referencing sockets whose names are different than the other side of the connection

### DIFF
--- a/lib/dal/src/management/mod.rs
+++ b/lib/dal/src/management/mod.rs
@@ -74,14 +74,14 @@ pub enum ManagementError {
     Func(#[from] FuncError),
     #[error("input socket error: {0}")]
     InputSocket(#[from] InputSocketError),
-    #[error("Cannot connect component {0} to component {1} because component {1} does not have an input socket with name {2}")]
-    InputSocketDoesNotExist(ComponentId, ComponentId, String),
+    #[error("Component {0} does not have an input socket with name {1}")]
+    InputSocketDoesNotExist(ComponentId, String),
     #[error("No existing or created view could be found named: {0}")]
     NoSuchView(String),
     #[error("output socket error: {0}")]
     OutputSocket(#[from] OutputSocketError),
-    #[error("Cannot connect component {0} to component {1} because component {0} does not have an output socket with name {2}")]
-    OutputSocketDoesNotExist(ComponentId, ComponentId, String),
+    #[error("Component {0} does not have an output socket with name {1}")]
+    OutputSocketDoesNotExist(ComponentId, String),
     #[error("prop error: {0}")]
     Prop(#[from] PropError),
     #[error("schema error: {0}")]
@@ -92,6 +92,8 @@ pub enum ManagementError {
     StandardModel(#[from] StandardModelError),
     #[error("transactions error: {0}")]
     Transactions(#[from] TransactionsError),
+    #[error("ulid decode error: {0}")]
+    UlidDecodeError(#[from] ulid::DecodeError),
     #[error("workspace snapshot error: {0}")]
     WorkspaceSnapshot(#[from] WorkspaceSnapshotError),
     #[error("ws event error: {0}")]
@@ -179,16 +181,17 @@ impl From<RawGeometry> for ManagementGeometry {
 
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
-pub struct ConnectionIdentifier {
+pub struct SocketRef {
     pub component: String,
     pub socket: String,
 }
 
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[serde(untagged)]
 #[serde(rename_all = "camelCase")]
-pub struct ManagementConnection {
-    from: String,
-    to: ConnectionIdentifier,
+pub enum ManagementConnection {
+    Input { from: SocketRef, to: String },
+    Output { from: String, to: SocketRef },
 }
 
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
@@ -464,7 +467,7 @@ pub struct ManagementOperator<'a> {
 
 #[derive(Clone, Debug)]
 struct PendingConnect {
-    from_component_id: ComponentId,
+    component_id: ComponentId,
     connection: ManagementConnection,
 }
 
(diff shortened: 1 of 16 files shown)


lib/dal/src/management/mod.rs
```diff
@@ -721,61 +724,78 @@ impl<'a> ManagementOperator<'a> {
...
-
-        let destination_input_socket_id = self
-            .socket_map
-            .input_socket_id(to_variant_id, &connection.to.socket)
-            .ok_or(ManagementError::OutputSocketDoesNotExist(
-                source_component_id,
-                destination_component_id,
-                connection.to.socket.to_owned(),
-            ))?;
+    async fn prepare_for_connection(
```
jkeiser, line 765: prepare_for_connection now supports both input and output directions, so this change lets it figure out which side the `component_id` goes on ("from" for output, "to" for input).

lib/dal/src/management/mod.rs
```diff
@@ -971,19 +991,30 @@ impl<'a> ManagementOperator<'a> {
...
-            ))
+    }
+
+    fn get_managed_component_id(&self, placeholder: &str) -> ManagementResult<ComponentId> {
+        self.get_managed_component_id_opt(placeholder).ok_or(
+            ManagementError::ComponentWithPlaceholderNotFound(placeholder.to_owned()),
+        )
+    }
+
+    fn get_managed_or_external_component_id(
```
jkeiser, line 1007: This is how we support component ID in socket refs. We could expand this support to all places we look up components (i.e. just modify the existing get_real_component_id), but I wanted to be intentional about that--for example, I'm not sure whether we would want `ops.update` to be able to update arbitrary components in the graph.

lib/dal/src/management/prototype.rs
```diff
@@ -108,6 +119,31 @@ pub struct ManagedComponent {
...
 
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct ManagerComponent {
+    #[serde(flatten)]
+    component: ManagedComponent,
+    incoming_connections: HashMap<String, SocketRefsAndValues>,
+}
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
```
jkeiser, line 130: incoming connections are passed as a singleton if the socket arity is "one", and an array if the arity is "many". Even if there is only one connection, sockets with arity="many" get an array with one value, for consistency.

lib/dal/src/management/prototype.rs
```diff
@@ -108,6 +119,31 @@ pub struct ManagedComponent {
     geometry: HashMap<String, ManagementGeometry>,
 }
 
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct ManagerComponent {
+    #[serde(flatten)]
```
jkeiser, line 125: #[serde(flatten)] is an easy way to handle "extending" a serialized type without having to rewrite all the fields.

dev/Tiltfile
zacharyhamm, line 1: Is this a necessary part of this PR?

zacharyhamm (commented):
We have pretty good test coverage of management functions. This feels like a case that needs an added test (see the functions in the small odd lego schema in the tests)

dev/Tiltfile
jkeiser: Nope, it is something that just merged a few seconds ago, I was based on another PR for dev speed. I'll rebase on main!

britmyerss: No, I think we ship it and immediately start the tests in a separate PR so we can start using this

lib/sdf-server/src/config.rs
```diff
@@ -410,7 +410,7 @@ pub enum IncomingStream {
 
 impl Default for IncomingStream {
     fn default() -> Self {
-        Self::TcpSocket(SocketAddr::from(([0, 0, 0, 0], 5156)))
+        Self::TcpSocket(SocketAddr::from(([127, 0, 0, 1], 5157)))
```
britmyerss, line 413: probably don't want this to stay?

lib/veritech-core/src/lib.rs
```diff
@@ -153,7 +153,7 @@ impl GetNatsSubjectFor for ValidationRequest {
 pub enum VeritechRequest {
     ActionRun(ActionRunRequest),
     KillExecution(KillExecutionRequest),
-    Management(ManagementRequest),
+    Management(Box<ManagementRequest>),
```
britmyerss, line 156: why is this boxed?

lib/sdf-server/src/config.rs
jkeiser: oh, thanks! Yeah, this is coordination between me and Wendy. Removing.

lib/veritech-core/src/lib.rs
jkeiser: Because it's gotten above like 400 bytes, and Rust emits a lint noting that that variant is making the whole enum too big (like, if you create an ActionRun variant, it'll still take up all that same space because an enum is as big as its biggest variant).

britmyerss (approved):
Tested locally by creating/removing connections from inputs!

PR #5383, opened 2025-01-30
feat: make template converge more idempotent

adamhjk:
* It does not correctly deal with geometry
* Everything else should work, including deletes


adamhjk:
still needs a little work - adding the udpate func names to enqueue updates

adamhjk:
this also sucked in some formatting fixes!

johnrwatson (approved):

PR #5386, opened 2025-01-30
BUG-768: only show child frames in the diagram outliner when an ancestor is in the view

jobelenus:
Currently the diagram outliner shows all ancestors that are frames even if they are not in the current view. Thats not the intention.

Testing steps, easier to use pictures, but the basic setup is:
image
image

- have a child of region (VPC outside frame) in TEST that has **no children** in DEFAULT... it should not show in DEFAULT outliner (primary bug)
- have children of region (generic frame, VPC inside frame) in TEST that **have** children in DEFAULT (key pair, ec2 instance)... it should show in DEFAULT outliner (this _was_ working, but needed to make sure the fix to above didn't break this)
- make sure that everything else in both views looks right/didn't break (e.g. root components, frames with no children)

### DIFF
--- a/app/web/src/components/DiagramOutline/DiagramOutlineNode.vue
+++ b/app/web/src/components/DiagramOutline/DiagramOutlineNode.vue
@@ -306,14 +306,32 @@ const inView = computed(() =>
   viewComponentIds.value.includes(props.component.def.id),
 );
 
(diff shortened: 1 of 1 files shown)


wendybujalski (approved):
👍

PR #5389, opened 2025-01-30
Fix: a better toast fix for never-ending spinner

jobelenus:

### DIFF
--- a/app/web/src/store/components.store.ts
+++ b/app/web/src/store/components.store.ts
@@ -1408,10 +1408,7 @@ export const useComponentsStore = (forceChangeSetId?: ChangeSetId) => {
                 callback: (data, metadata) => {
                   if (metadata.change_set_id !== changeSetId) return;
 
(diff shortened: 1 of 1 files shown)


wendybujalski (approved):

app/web/src/store/components.store.ts
```diff
@@ -1428,7 +1425,7 @@ export const useComponentsStore = (forceChangeSetId?: ChangeSetId) => {
                       },
                       component: CreatingTemplate,
                     },
-                    options: { timeout: 3 * 1000, closeOnClick: true },
+                    options: { timeout: false, closeOnClick: true },
```
johnrwatson, line 1428: **Apologies for the silly question:** Is the logic here that basically if we try and update a toast that doesn't exist, it now doesn't create it? So for the case where the toast never went away, it was because we were trying to update it but it didn't exist yet(?)

johnrwatson (approved):

PR #5391, opened 2025-01-30
feat(si-fs): attach existing functions with `mv`

zacharyhamm:
`mv` a function folder into the correct kind folder for a schema
(actions, attributes, managements, etc.), either the `locked` or
`unlocked` folder in the schema functions folder, or the root folder
(the one with the function) name in the change set functions list.

Attributes and actions will be marked pending, the same as when creating
new ones. Edit the PENDING_BINDINGS_EDIT_ME.json with a correct output
location or correct action kind to finish attaching the function.


adamhjk (approved):
lgtm!

PR #5393, opened 2025-01-30
Fix: edge not selectable

jobelenus:
Can't believe that assignment to rawEdgesById didn't get noticed and cleaned up. Adding the delete pattern we used above as well, just for posterity.

Testing steps:
- we found this during John's workshop steps.
- a views edges were clobbering edges already stored


wendybujalski (approved):
👍

PR #5394, opened 2025-01-30
Do not filter requirements by kind and other fixes

nickgerace:
## Description

This PR ensures that the InsertApprovalModel no longer filters requirements based on entity kind. Now, it shows a requirement per entity and the labelling is expanded for this feature. This is NOT the final state we want to be in due to showing potentially too many and overwhelming results for the same approver groups and individuals, but this is a step in the right direction.

This PR also ensures that the approval status data is up to date on the requester's side when requesting approvals. This is due to a potential race or missing edge case when the change set moves from an "open" to a "needsapproval" state. A corresponding "TODO" has been added to the problem area, but the current fix is sufficient in the interim.

In addition to these changes, both the requirements and actions are now scrollable. We now also show the email for a user (though truncation may not work at this time) as well as right padding for the status text (e.g. "Waiting..." which has now been capitalized to match the other states).

One final change, we now always allow users to reject the changes, even if all approvals have been met.

### DIFF
--- a/app/web/src/components/Actions/ActionCard.vue
+++ b/app/web/src/components/Actions/ActionCard.vue
@@ -4,7 +4,7 @@
       clsx(
         'flex flex-row items-center text-sm relative p-2xs min-w-0 w-full border border-transparent',
         !props.noInteraction
-          ? 'cursor-pointer hover:border-action-500 dark:hover:border-action-300'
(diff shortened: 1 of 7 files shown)


nickgerace:
Closed in favor of https://github.com/systeminit/si/pull/5403

PR #5401, opened 2025-01-31
test(sdf): Add regression tests for incomingConnections

jkeiser:
This adds regression tests for incomingConnections. In particular, it tests that incomingConnections values can be used in ops.create.connect and ops.update.connect.add/remove. It also tests that incomingConnections contains the correct values for the given sockets.


zacharyhamm (dismissed):
These look great

stack72 (approved):

PR #5405, opened 2025-02-01
feat(clover): input socket for id arrays

vbustamante:


bin/clover/src/pipeline-steps/createInputSocketsAcrossAssets.ts
```diff
@@ -76,6 +106,43 @@ export function createInputSocketsBasedOnOutputSockets(
...
+                    ),
+                  );
+                }
+              }
+            }
+          }, { skipTypeProps: true });
+        }
+      }
+    }, { skipTypeProps: true });
+
```
vbustamante, line 145: This is the core of the feature

sprutton1 (approved):

PR #5406, opened 2025-02-01
fix(dal): respect socket arity, connection annotations, and other socket metadata when it changes from variant to variant

britmyerss:
Before this change, if you created a variant with a socket, then modified the socket arity or connection annotations, the updates would not be reflected on regeneration. 

<div><br/>via <a href="https://giphy.com/gifs/metal-while-protractor-gWyhZM3XKcyaI">GIPHY</a></div>

Tested this manually by modifying connection annotations and arity with components on the diagram and seeing them update, and wrote an integration test for it.

### DIFF
--- a/lib/si-pkg/src/spec/variant.rs
+++ b/lib/si-pkg/src/spec/variant.rs
@@ -12,7 +12,7 @@ use crate::{
 
 use super::{
     ActionFuncSpec, LeafFunctionSpec, ManagementFuncSpec, PropSpec, PropSpecData,
-    PropSpecWidgetKind, RootPropFuncSpec, SiPropFuncSpec, SocketSpec, SpecError,
+    PropSpecWidgetKind, RootPropFuncSpec, SiPropFuncSpec, SocketSpec, SocketSpecData, SpecError,
 };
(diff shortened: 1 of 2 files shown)


lib/si-pkg/src/spec/variant.rs
```diff
@@ -377,9 +377,24 @@ impl SchemaVariantSpec {
...
                 );
-
+                // If there is nothing to skip, we still need to include the data from the newer this_socket
+                // and copy it to the other_socket so we don't lose any changes here
                 if new_merge_skips.is_empty() {
-                    merged_sockets.push(other_socket.to_owned());
+                    if let (Some(this_socket_data), Some(other_socket_data)) =
+                        (this_socket.data.as_ref(), other_socket.data.as_ref())
+                    {
+                        // keep other_socket's func_unique_id, but everything else from the newer this_socket
```
nickgerace, line 386: Ah good find!

nickgerace (commented):
Not approving since I'm on thumbs, but this looks great. Logic and corresponding test are sound. Nice job!

PR #5408, opened 2025-02-01
feat: arn sockets, code gen, normalizeArrays

sprutton1:


bin/clover/src/cloud-control-funcs/code-gen/awsCloudControlCodeGenCreate.ts
```diff
@@ -55,13 +55,50 @@ async function main(component: Input): Promise<Output> {
     }
   }
 
+  const cleaned = removeEmpty(payload);
+
   const cloudControlPayload: CloudControlPayload = {
     TypeName: cloudControlType,
-    DesiredState: payload,
+    DesiredState: cleaned,
```
sprutton1, line 62: We need to disappear empty arrays and such or values that will conflict will cause failures even if one is empty

stack72 (approved):

PR #5411, opened 2025-02-02
fix(clover): .setDefaultValue in asset definition func if there is a default value in the prop spec

britmyerss:
If there is a default value on the prop spec, we need to add `.setDefaultValue({defaultValue})` to the asset definition function, otherwise on regenerate, this is missing, and we interpret that the user wants to remove the default value for that prop. 

I have not yet uploaded this to a local module index and installed for testing, but hoping this gets us a head start.

### DIFF
--- a/bin/clover/src/pipeline-steps/generateAssetFuncs.ts
+++ b/bin/clover/src/pipeline-steps/generateAssetFuncs.ts
@@ -71,9 +71,8 @@ function generateAssetCodeFromVariantSpec(variant: SchemaVariantSpec): string {
 
     for (const prop of variant.domain.entries) {
       const varName = `${prop.name}Prop`.replace(" ", "");
-      propDeclarations += `${indent(1)}const ${varName} = ${
-        generatePropBuilderString(prop, 2)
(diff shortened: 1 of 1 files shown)


bin/clover/src/pipeline-steps/generateAssetFuncs.ts
```diff
@@ -235,14 +231,14 @@ function generatePropBuilderString(
...
-            prop.data?.widgetOptions,
-          )
+        `${generateWidgetString(
+          prop.data?.widgetKind,
+          is_create_only,
+          indent_level,
+          prop.data?.widgetOptions,
+        )
         }` +
+        `${prop.data?.defaultValue ? `${indent(indent_level)}.setDefaultValue(${JSON.stringify(prop.data.defaultValue)})\n` : ''}` +
```
vbustamante, line 241: This is the only functional change right?

vbustamante (approved):
Looks good! The VS code formatter and the intellij one seem to disagree about some things, we probably should fix that for Clover at some point

PR #5413, opened 2025-02-03
fix(clover): add secret props in asset definition func

britmyerss:
When we create the schema spec, for schemas that consume secrets, we must add a prop and socket with the correct configuration.  However, we provide abstraction over this in the asset definition function, so we must detect when secret props are being used, build a `SecretProp`, ensuring we don't add the corresponding `InputSocket` as the `SecretProp` handles adding both the prop and the socket.

### DIFF
--- a/bin/clover/src/pipeline-steps/generateAssetFuncs.ts
+++ b/bin/clover/src/pipeline-steps/generateAssetFuncs.ts
@@ -81,6 +81,29 @@ function generateAssetCodeFromVariantSpec(variant: SchemaVariantSpec): string {
     adds += propAdds;
   }
 
+  // Code for Secret Props
+
+
+  {
+    if (variant.secrets.kind !== "object") {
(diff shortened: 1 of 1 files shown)


stack72 (approved):

bin/clover/src/pipeline-steps/generateAssetFuncs.ts
```diff
@@ -81,6 +81,28 @@ function generateAssetCodeFromVariantSpec(variant: SchemaVariantSpec): string {
     adds += propAdds;
   }
 
+  // Code for Secret Props
+
+
+  {
+    if (variant.secrets.kind !== "object") {
+      console.log(
```
sprutton1, line 89: I think we should error here as we'll generate a bad asset func without this, right?

sprutton1 (approved):

PR #5414, opened 2025-02-03
fix(web): Only add favourite or default workspaces to the workspaces list

stack72:
We created 50 workspaces for the SI Camp and it has shown that the list grows to include everything.

We are going to show the most important user workspaces in that list and a user can add what is important to them from the auth portal


britmyerss (approved):

fnichol:
Are we shooting to merge this one?

PR #5419, opened 2025-02-03
feat(clover): Generate sockets for string props

vbustamante:
- Generate output sockets for all primaryIdentifiers
- Generate input sockets for string props that match other asset names, which already happened for string arrays


sprutton1 (approved):
nice!

jkeiser (approved):

PR #5421, opened 2025-02-03
feat(clover): Allow generation of just specific specs with argument

jkeiser:
Allows you to add arguments to `generate-specs` to restrict which services will be run.

```
# Generates all specs
deno run task generate-specs
# Generates all ECS and S3 specs
deno run task generate-specs ECS S3
```

And gets rid of the hardcoded list of specs.

It also adds `SI_MODULE_INDEX_URL` environment variable support so you can point at an arbitrary module index server. Defaults to what it did before. Probably should default to something else, however :)


vbustamante (approved):
Great!

PR #5423, opened 2025-02-04
Add max deliver config to pinga (default is 1)

nickgerace:
## Introduction

This PR adds a configurable max delivery field to pinga, which defaults to "1".

## Description

First, why does the new configuration option default to "1"? The single naxum handler in pinga server relies on the default AckLayer and its inner logic is infallible. How is that the case? The job execution code is wrapped with an infallible wrapper.

There are two feasible ways that the handler can fail, which causes a NACK to be sent via the code in the default AckLayer... the JSON extractor can fail (error 422 on deserialization) or the subject is not a valid UTF8 string. The other extractors are fallible, but their default implementations immediately return "Ok". Therefore, either the request payload is a poison pill or the subject is invalid. In either case, we should not retry the job and we should immediately send the message to the Dead Letter Queue stream, which is already configured to handle this.

As a result of this paradigm, there is no custom backoff or increased number of max deliveries. So... why does this commit make it configurable? Due errant test behavior when adding a custom backoff and deserialization errors recently seen when pulling messages off the "PINGA_JOBS" stream, we have not confirmed whether or not the system relies on re-deliveries at the pinga-level or not. Therefore, if we discover that re-delivery is required, the max delivery should be increased accordingly with no custom backoff.

An example max delivery increase would be from 1 (default) to 60,000. WTF? WHY? Well, we have observed in live production scenarios that we can see around ~500 re-deliveries per second. Setting max deliveries to 60,000 allows for approximately 2 minutes worth of pinga trying to execute a job. This math is inadequate at best, but the point is this: if we find that we do require retries, then bump that number up.

In addition to these changes, the "initial DVU" step of the rebaser has been removed. It was added in the quiescence work, but it may actually be the cause of the poison pill messages. Working with @britmyerss revealed that it is likely unnecessary and both testing and static analysis have indicated as such.

This was tested with 201 components with the root component being a Region as a configurationDownFrame. There were VPCs, Subnets, EC2 Instances, and all kinds of nesting. Flipping the region in the root set off pinga into a frenzy and none of the errors were observed.

Fixes BUG-776

### DIFF
--- a/bin/pinga/src/args.rs
+++ b/bin/pinga/src/args.rs
@@ -119,6 +119,10 @@ pub(crate) struct Args {
     #[arg(long)]
     pub(crate) concurrency: Option<u32>,
 
+    /// Configures the max delivery number NATS JetStream consumer(s)
+    #[arg(long)]
+    pub(crate) max_deliver: Option<i64>,
+
     /// The path at which the layer db cache is created/used on disk [e.g. /banana/]
     #[arg(long)]
     pub(crate) layer_db_disk_path: Option<String>,
@@ -217,6 +221,9 @@ impl TryFrom<Args> for Config {
             }
             if let Some(concurrency) = args.concurrency {
                 config_map.set("concurrency_limit", i64::from(concurrency));
+            }
+            if let Some(max_deliver) = args.max_deliver {
+                config_map.set("max_deliver", max_deliver);
             }
             if let Some(layer_cache_disk_path) = args.layer_db_disk_path {
                 config_map.set("layer_db_config.disk_path", layer_cache_disk_path);
--- a/lib/pinga-server/src/config.rs
+++ b/lib/pinga-server/src/config.rs
@@ -55,6 +55,9 @@ pub struct Config {
     #[builder(default = "default_concurrency_limit()")]
     concurrency_limit: usize,
 
+    #[builder(default = "default_max_deliver()")]
+    max_deliver: i64,
+
     #[builder(default = "random_instance_id()")]
     instance_id: String,
 
@@ -102,6 +105,11 @@ impl Config {
         self.concurrency_limit
     }
 
+    /// Gets the config's max delivery setting for NATS JetStream consumer(s).
+    pub fn max_deliver(&self) -> i64 {
+        self.max_deliver
+    }
+
     /// Gets the config's instance ID.
     pub fn instance_id(&self) -> &str {
         self.instance_id.as_ref()
@@ -123,6 +131,8 @@ pub struct ConfigFile {
     crypto: VeritechCryptoConfig,
     #[serde(default = "default_concurrency_limit")]
     concurrency_limit: usize,
+    #[serde(default = "default_max_deliver")]
+    max_deliver: i64,
     #[serde(default = "random_instance_id")]
     instance_id: String,
     #[serde(default = "default_layer_db_config")]
@@ -137,6 +147,7 @@ impl Default for ConfigFile {
             pg: Default::default(),
             nats: Default::default(),
             concurrency_limit: default_concurrency_limit(),
+            max_deliver: default_max_deliver(),
             crypto: Default::default(),
             instance_id: random_instance_id(),
             layer_db_config: default_layer_db_config(),
@@ -160,6 +171,7 @@ impl TryFrom<ConfigFile> for Config {
         config.nats(value.nats);
         config.crypto(value.crypto);
         config.concurrency_limit(value.concurrency_limit);
+        config.max_deliver(value.max_deliver);
         config.instance_id(value.instance_id);
         config.symmetric_crypto_service(value.symmetric_crypto_service.try_into()?);
         config.layer_db_config(value.layer_db_config);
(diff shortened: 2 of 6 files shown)


bin/pinga/src/args.rs
```diff
@@ -119,6 +119,10 @@ pub(crate) struct Args {
     #[arg(long)]
     pub(crate) concurrency: Option<u32>,
 
+    /// Enables NATS JetStream consumer(s) to have poison pill protection
+    #[arg(long)]
+    pub(crate) jetstream_consumer_poison_pill_protection: Option<bool>,
```
nickgerace, line 124: This is an "opt-out" option. If it gives us trouble, we can disable it without a git-based deployment.

lib/pinga-server/src/config.rs
```diff
@@ -183,6 +198,10 @@ fn default_concurrency_limit() -> usize {
     DEFAULT_CONCURRENCY_LIMIT
 }
 
+fn default_jetstream_consumer_poison_pill_protection() -> bool {
+    true
+}
```
nickgerace, line 203: Again, it is on by default

lib/pinga-server/src/server.rs
```diff
@@ -264,12 +270,31 @@ impl Server {
...
+        // Kaboom. Maybe you add upper bounds? Failure on misconfiguration? Who knows? Not me at
+        // the time of writing.
+        if jetstream_consumer_poison_pill_protection {
+            config.max_deliver = 4;
+            config.backoff = vec![
+                Duration::from_secs(5),
+                Duration::from_secs(10),
+                Duration::from_secs(15),
+            ];
         }
```
nickgerace, line 295: This is the main guts of the PR. It is copied from the forklift settings, which were used in conjunction with the introduction of the dead letter queue.

lib/rebaser-server/src/serial_dvu_task.rs
```diff
@@ -65,10 +65,6 @@ impl SerialDvuTask {
     pub(crate) async fn try_run(self) -> Result<Shutdown> {
         metric!(counter.serial_dvu_task.change_set_in_progress = 1);
 
-        // Attempt to run an initial dvu in case there are processed requests that haven't yet been
-        // finished with a dvu run
-        self.maybe_run_initial_dvu().await?;
```
nickgerace, line 70: @fnichol hey! Two things here

1. @britmyerss and I looked at this and we believe it might not have been doing anything.
2. We believe it may also be a potential culprit for the `PINGA_JOBS` poison pill issues, though there is not hard evidence yet other than static analysis, talking it through and timeline.

We believe this is safe to remove and it is not essential for use in the rebaser, but I wanted to check with you for review!

lib/pinga-server/src/server.rs
britmyerss: Does this mean the first nack will be retried after 5 seconds? Or the second will?

nickgerace: The first!

CONSUME --> NACK (1) --> 5 SEC --> CONSUME --> NACK (2) --> 10 SEC --> CONSUME --> NACK (3) --> 15 SEC --> CONSUME --> NACK (4)

britmyerss (dismissed):

lib/rebaser-server/src/serial_dvu_task.rs
fnichol: I rolled this change over and over in my head and landed on: "yeah, that's going to work for us" 😄

lib/rebaser-server/src/serial_dvu_task.rs
```diff
@@ -156,38 +152,4 @@ impl SerialDvuTask {
...
-                si.change_set.id = %self.change_set_id,
-                "enqueuing *initial* dependent_values_update",
-            );
-            ctx.enqueue_dependent_values_update().await?;
-            ctx.blocking_commit_no_rebase().await?;
-            metric!(counter.serial_dvu_task.initial_dvu_running = -1);
-        }
-
-        Ok(())
-    }
```
fnichol, line 192: I was a bit surprised of the differences between this now-removed `maybe_run_initial_dvu` and `run_dvu` functions. At least there's no difference now, now that there's only one

fnichol (approved):
Ok, making sense to me after talking it over and taking a read here!

nickgerace:
@britmyerss those test failures look suspicious... we might want to take a look.

nickgerace:
Bringing this back to drafts because we've found the bug this PR tries to prevent _and_ we have learned new things about pinga behavior that require some further thought.

britmyerss (approved):

fnichol (approved):
The tests are passing in a ci `try` and using the product locally was good. As we haven't yet found/resolved the underlying reason for these errors (which we suspect is in deserializing), it's hard to game out how this may affect the behavior of the full system. However, this is closer to the behavior before these requests were on a durable Jetstream and any "failed" requests will now be marked and preserved (and then skipped over) meaning we will have more information to find and ultimately resolve the underlying issue.

PR #5427, opened 2025-02-04
fix(hoist): Show real total numbers for uploads

vbustamante:
We were showing the counts including skipped modules before so the progress bar didn't make sense


jkeiser (approved):
Happy UI improvement!

PR #5433, opened 2025-02-05
feat: run as subcommand to support imports

sprutton1:
https://github.com/denoland/deno/issues/20945

Apparently, imports in deno compiled binaries are only dynamic if they are statically analyzable, i.e. available at compile time. I, for one, hardly call this dynamic. 

Anyways, in order to get around this, we use deno to call deno to deno run our function code. In order to make this work as it does today, we generate a fully bundled version of the packages in `sandbox.ts` and import it at runtime. Also, since we are now running in a subprocess, we have to use stdout/err to communicate with the parent, so we now pass around the `console.log`s, `requestStorage` bits, and returns from the functions.

The big advantage here is that we no longer need to transpile ourselves as deno can just run the typescript code we write. And, regular imports work! We stick to the deno prefix paradigm for marking imports.

Here's us doing various kinds of imports and the results.

```typescript
import { say } from "jsr:@morinokami/deno-says/say";
import moment from "npm:moment";
import { basename } from "https://deno.land/std@0.218.0/path/mod.ts";

async function main(): Promise < Output > {
    const timestamp = moment('2025-02-07 20:02:41').format('YYYY-MM-DD HH:mm:ss');
    const username = basename('/home/sprutton1');
    say(`Current Date and Time (UTC - YYYY-MM-DD HH:MM:SS formatted): ${timestamp}\nCurrent User's Login: ${username}\n`);


    return {
        result: 'success',
        message: "Logged user and time"
    };
}
```

```
 _________________________________________
< Current Date and Time (UTC - YYYY-MM-DD >
< HH:MM:SS formatted): 2025-02-07         >
< 20:02:41
Current User's Login: sprutton1 >
 -----------------------------------------
        \     _
         \  ( ･ヽ
               \ \
               |  ｀ヽ
               ⎩      ﾄ､
                u¯u︶u
```

**EDIT**: it's actually a little bit faster than main!  Approximately 37.1% faster when running very simple functions.
```bash
BEFORE
Benchmark 1: cat test.json | ./lang-js resolverfunction
  Time (mean ± σ):     528.7 ms ±   7.1 ms    [User: 295.1 ms, System: 305.6 ms]
  Range (min … max):   514.7 ms … 562.9 ms    100 runs

AFTER
Benchmark 1: cat test.json | ./lang-js resolverfunction
  Time (mean ± σ):     332.6 ms ±   9.6 ms    [User: 192.7 ms, System: 156.0 ms]
  Range (min … max):   314.8 ms … 363.7 ms    100 runs
```


BUCK
```diff
@@ -84,7 +84,6 @@ pnpm_workspace(
         "//app/auth-portal:package.json",
         "//app/web:package.json",
         "//bin/auth-api:package.json",
-        "//bin/lang-js:package.json",
```
sprutton1, line 87: cleaning up cruft

bin/lang-js/deno.json
```diff
@@ -2,6 +2,18 @@
   "imports": {
     "@cliffy/command": "jsr:@cliffy/command@^1.0.0-rc.7",
     "@deno/emit": "jsr:@deno/emit@^0.46.0",
-    "@std/assert": "jsr:@std/assert@1"
-  }
+    "@std/assert": "jsr:@std/assert@1",
```
sprutton1, line 5: necessary for the build.ts bits to work

bin/lang-js/src/function_kinds/action_run.ts
```diff
@@ -118,14 +119,15 @@ async function execute(
...
-async function run(arg) {
+async function main(arg) {
+  let payload = null;
+  let resourceId = null;
   try {
     ${code}
-    arg = Array.isArray(arg) ? arg : [arg];
-    const resourceId = arg[0]?.properties?.si?.resourceId;
-    const payload = arg[0]?.properties?.resource?.payload ?? null;
+    resourceId = arg?.properties?.si?.resourceId;
```
britmyerss, line 127: we don't need to normalize the arg anymore?

flake.nix
```diff
@@ -233,7 +236,24 @@
...
+
+              # Cache the dependencies
+              ${pkgs.deno}/bin/deno cache \
+                --node-modules-dir \
+                --reload \
+                bin/lang-js/src/sandbox.ts \
+
+            '';
             extraInstallPhase = ''
+              # build a cache of our deps so we don't need to donwload them in
```
britmyerss, line 251: 👏

britmyerss (approved):

bin/lang-js/src/function_kinds/action_run.ts
sprutton1: I'm actually not sure why we ever were. The Input for action funcs is an object, not a list of objects. I am expanding the args in a slightly different way, so it could be that I'm stomping out a bug I didn't realize was there.

sprutton1: I did verify action runs with components of various shapes and everything seems fine

PR #5435, opened 2025-02-05
fix(pinga): This fixes up metrics so it doesn't look like we've got runaway jobs/unexecuted funcs by inc/dec counters in the right spots

britmyerss:
<div><br/>via <a href="https://giphy.com/channel/BenJammins/">BenJammins</a> on <a href="https://giphy.com/gifs/calculator-calculating-calculated-IokAQCByJS254Dmw6f">GIPHY</a></div>

### DIFF
--- a/lib/dal/src/job/definition/dependent_values_update.rs
+++ b/lib/dal/src/job/definition/dependent_values_update.rs
@@ -243,8 +243,6 @@ impl DependentValuesUpdate {
     ) -> DependentValueUpdateResult<JobCompletionState> {
         let start = tokio::time::Instant::now();
         let span = Span::current();
-        metric!(counter.dvu_concurrency_count = 1);
-
         let roots = ctx.workspace_snapshot()?.take_dependent_values().await?;
 
         let mut unfinished_values: HashSet<Ulid> = HashSet::new();
@@ -285,7 +283,6 @@ impl DependentValuesUpdate {
             }
         }
         let all_value_ids = dependency_graph.all_value_ids();
-        metric!(counter.dvu.values_to_run = all_value_ids.len());
 
         let mut tracker = StatusUpdateTracker::new_for_values(ctx, all_value_ids).await?;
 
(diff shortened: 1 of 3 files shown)


lib/dal/src/job/definition/dependent_values_update.rs
```diff
@@ -285,7 +283,6 @@ impl DependentValuesUpdate {
             }
         }
         let all_value_ids = dependency_graph.all_value_ids();
-        metric!(counter.dvu.values_to_run = all_value_ids.len());
```
britmyerss, line 288: this doesn't map 1:1 with the funcs that actually have to run, so removing it.

lib/dal/src/job/definition/dependent_values_update.rs
```diff
@@ -334,6 +331,7 @@ impl DependentValuesUpdate {
                             .await?
                             .value(ctx)
                             .await?;
+                        metric!(counter.dvu.function_execution = 1);
 
                         update_join_set.spawn(values_from_prototype_function_execution(
```
nickgerace, line 336: Yeah this makes sense to me. That's where we actually execute, I believe.

lib/pinga-server/src/handlers.rs
```diff
@@ -199,12 +200,14 @@ async fn execute_job_inner(mut ctx_builder: DalContextBuilder, job_info: JobInfo
...
     };
-
+    let kind_str = job_info.kind.as_str();
+    metric!(counter.pinga_job_in_progress = 1, label = kind_str);
     info!("Processing job");
 
     job.run_job(ctx_builder.clone()).await?;
 
     info!("Finished processing job");
+    metric!(counter.pinga_job_in_progress = -1, label = kind_str);
```
nickgerace, line 210: Clean!

nickgerace (approved):

lib/pinga-server/src/handlers.rs
```diff
@@ -180,6 +182,7 @@ async fn execute_job(
             };
         }
     }
+    metric!(counter.pinga_job_in_progress = -1, label = job_kind);
```
britmyerss, line 185: moving this out here means even if the job "fails", we decrement the counter.

For example, right now if the change set is abandoned during DVU, the rebase will fail, which fails the job, and we weren't decrementing the count!

nickgerace (approved):

PR #5437, opened 2025-02-06
feat(si-fs): types for all functions!

zacharyhamm:
Types for all functions!

Uses a tsconfig.json to ensure that types are compiled per-function,
instead of across the entire fuse filesytem.

With dashmap, type compilation is much faster for the whole workspace


zacharyhamm:
Testing notes: boot up an editor that supports the typescript language server in the fuse root. Try two at once! I ran it with Zed + VsCode + Helix all at once, to see how it handled it. Wait for the types to compile, and edit some functions to see the typing at work

fnichol (approved):
Nice to see the Posthog tracking in as well, nice!

PR #5439, opened 2025-02-06
fix(Clover): output sockets were not getting extra annotations

vbustamante:
Fixes the bug cloudlife saw today when trying to connect bucket name

### DIFF
--- a/bin/clover/src/pipeline-steps/createInputSocketsAcrossAssets.ts
+++ b/bin/clover/src/pipeline-steps/createInputSocketsAcrossAssets.ts
@@ -4,11 +4,9 @@ import {
   ConnectionAnnotation,
   propHasSocket,
   propPathToString,
+  setAnnotationOnSocket,
 } from "../spec/sockets.ts";
-import {
-  bfsExpandedPropTree,
-  ExpandedPropSpec,
-} from "../spec/props.ts";
+import { bfsExpandedPropTree, ExpandedPropSpec } from "../spec/props.ts";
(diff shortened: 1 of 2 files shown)


bin/clover/src/pipeline-steps/createInputSocketsAcrossAssets.ts
```diff
@@ -119,15 +120,22 @@ export function createInputSocketsBasedOnOutputSockets(
     // wanting to connecting something like "TaskArn" or "Arn" -> "TaskRoleArn"
     for (const prop of domain.entries as ExpandedPropSpec[]) {
       if (!prop.name.toLowerCase().endsWith("arn")) continue;
-      getOrCreateInputSocketFromProp(schemaVariant, prop, "many", ["Arn"]);
+      const socket = getOrCreateInputSocketFromProp(
```
jkeiser, line 123: Does this change (not passing "Arn") change behavior in some way?

jkeiser (commented):
What was the actual bug? I can't find the behavior change in here. Nothing looks particularly objectionable, but if this fixes the problem, now I don't know what else might be broken :)

vbustamante:
@jkeiser the getOrAddInput function that got created was adding the annotation to the input socket instead of the output. Although very similar, adding the annotation was somewhat different for every case where that func was called so I extracted the logic

jkeiser (approved):
Oh WOW, that's subtle! Nice catch!

PR #5440, opened 2025-02-07
chore(clover): Strongly type Clover working types (specs)

jkeiser:
Clover is passing a bunch of PkgSpecs that it has dotted with extra cloudformation metadata, which actually contain `ExpandedPropSpec`s deep underneath. But because the PkgSpec *type* only has normal PropSpecs, we have to cast or check all props each time we look at them. The casting is dangerous (can be wrong), and the checking is unnecessary.

This PR adds an ExpandedPkgSpec that is just like PkgSpec except all props underneath are ExpandedPropSpecs. Typescript infers the types everywhere, therefore, and we can remove casts.

* Adds explicit types for everything in the pipeline. This gives us a lot more type safety, completion and editor type features, as many types showed up as "any" due to how ts-rs generates the bindings.
* Removes "type assumption" casts `as *Spec` everywhere (so we don't assume things are a certain type when they actually aren't
* Removes isExpandedPropSpec() and all calls to it, since we now know for sure that's what it is
* Gets us compiling correctly when the bindings actually load (there were a few issues that occur once the bindings get fixed to have `.ts` extensions on the imports)

While this makes us compile correctly when the si-pkg bindings are fixed, this PR does not fix them. That will come when we fix ts-rs.

### Testing

I regenerated all pkg files, and diffed them, and there are no differences.

### DIFF
--- a/bin/clover/README.md
+++ b/bin/clover/README.md
@@ -1,21 +1,84 @@
-# Useful Link
+clover generates full AWS asset coverage for all resource types in the
+[Cloudformation schema](https://docs.aws.amazon.com/cloudformation-cli/latest/userguide/resource-type-schema.html).
 
-- [Cloudformation resource type schema](https://docs.aws.amazon.com/cloudformation-cli/latest/userguide/resource-type-schema.html)
+# Initial Setup
+
+Before you can run the schema generator, you must download the latest [Cloudformation resource type schema](https://docs.aws.amazon.com/cloudformation-cli/latest/userguide/resource-type-schema.html)
+to cloudformation-schema/*.json:
+
+```sh
+$ deno task run fetch-schema
+```
+
+# Generating Assets
+
+This will fetch the current ID for each schema, and then regenerate all schemas into the
+si-specs directory.
+
+```sh
+$ deno task run generate-specs [...SCHEMA]
+```
+
+* `generate-specs` with no arguments to regenerate all specs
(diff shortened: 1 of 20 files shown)


bin/clover/README.md
```diff
@@ -1,21 +1,84 @@
...
 
-- [Cloudformation resource type schema](https://docs.aws.amazon.com/cloudformation-cli/latest/userguide/resource-type-schema.html)
+# Initial Setup
 
-# Running tests
+Before you can run the schema generator, you must download the latest [Cloudformation resource type schema](https://docs.aws.amazon.com/cloudformation-cli/latest/userguide/resource-type-schema.html)
+to cloudformation-schema/*.json:
+
+```sh
+$ deno task run main.ts fetch-schema
```
vbustamante, line 10:
```suggestion
$ deno task run fetch-schema
```

bin/clover/README.md
```diff
@@ -1,21 +1,84 @@
...
+```
+
+# Generating Assets
+
+This will fetch the current ID for each schema, and then regenerate all schemas into the
+si-specs directory.
 
 ```sh
-$ deno test -A
+$ deno task run main.ts generate-specs [...SCHEMA]
```
vbustamante, line 19:
```suggestion
$ deno task run generate-specs [...SCHEMA]
```

vbustamante (dismissed):

vbustamante (approved):

PR #5447, opened 2025-02-08
fix(Clover): Multiple socket fixes

vbustamante:
- Overrides for CloudTrail::Trail sockets
- Stop creating input sockets for the same component that has the output sockets, it that's the only matching socket
- Always create output socket for `${VariantName}Name` and `${VariantName}Id`

### DIFF
--- a/bin/clover/src/pipeline-steps/assetSpecificOverrides.ts
+++ b/bin/clover/src/pipeline-steps/assetSpecificOverrides.ts
@@ -1,7 +1,11 @@
 import _ from "npm:lodash";
 import _logger from "../logger.ts";
-import { createInputSocketFromProp } from "../spec/sockets.ts";
+import {
+  createInputSocketFromProp,
+  setAnnotationOnSocket,
+} from "../spec/sockets.ts";
 import { ExpandedPkgSpec } from "../spec/pkgs.ts";
+import { ExpandedPropSpec } from "../spec/props.ts";
 
 const logger = _logger.ns("assetOverrides").seal();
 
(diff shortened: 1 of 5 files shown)


bin/clover/src/pipeline-steps/assetSpecificOverrides.ts
```diff
@@ -25,31 +28,49 @@ type OverrideFn = (spec: ExpandedPkgSpec) => void;
...
+    const variant = spec.schemas[0].variants[0];
 
-function addGatewayIdSocketToEC2Route(spec: ExpandedPkgSpec) {
-  const schema = spec.schemas[0];
-  const variant = spec.schemas[0].variants[0];
-  const domain = variant.domain;
+    for (const prop of variant.domain.entries) {
+      switch (prop.name) {
+        case "S3BucketName": {
+          const socket = createInputSocketFromProp(prop, "one", [{
```
jkeiser, line 50: If you call getOrCreateInputSocketFromPop, it will actually take care of the push as well (and prevent dups from appearing).

bin/clover/src/pipeline-steps/createInputSocketsAcrossAssets.ts
```diff
@@ -46,11 +46,18 @@ export function createInputSocketsBasedOnOutputSockets(
           socket.data?.connectionAnnotations,
         ) as ConnectionAnnotation[];
 
-        for (const annotations of existingAnnotations) {
-          for (const annotation of annotations.tokens) {
-            foundOutputSockets[annotation] ??= [];
-            foundOutputSockets[annotation].push(schemaVariant);
+        for (const { tokens } of existingAnnotations) {
+          if (tokens.length !== 1) continue;
```
jkeiser, line 50: If there is more than one token we don't want to add the schema variant for them?

bin/clover/src/spec/sockets.ts
```diff
@@ -47,6 +47,14 @@ export function createInputSocketFromProp(
   return socket;
 }
 
+export function getSocketOnVariant(
```
jkeiser, line 50: If we're adding something here anyway, might be better to add getOrCreateOutputSocketFromProp() instead, to match getOrCreateInputSocketFromProp(). But just a nit.

jkeiser (approved):

PR #5449, opened 2025-02-10
feat(clover): Add doc links to all props; don't lose documentation when regenerating

jkeiser:
This PR:
* Adds doc links to all props that go to the cloudformation documentation (they go directly to the prop when the prop exists, otherwise to the page for the asset). e.g. https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ec2-subnetroutetableassociation.html#cfn-ec2-subnetroutetableassociation-routetableid
* Preserves documentation when regenerating assets (docs weren't in the asset function)
* Uses the modern Node API for base 64 encoding, so that we can support Unicode documentation in asset functions

## Testing

* Uploaded assets to local module index
* Regenerated the Subnet asset to ensure it regenerated, and clicked the documentation
* Clicked some doc links to make sure they went to the right assets

### DIFF
--- a/bin/clover/src/pipeline-steps/generateAssetFuncs.ts
+++ b/bin/clover/src/pipeline-steps/generateAssetFuncs.ts
@@ -276,9 +276,7 @@ function generateWidgetString(
     return "";
   }
 
-  const kind = widgetKind === "ComboBox"
-    ? "comboBox"
-    : widgetKind.toLowerCase();
+  const kind = `${widgetKind[0].toLowerCase()}${widgetKind.slice(1)}`;
 
   let widgetStr =
     `${indent(indentLevel)}.setWidget(new PropWidgetDefinitionBuilder()\n` +
--- a/bin/clover/src/pipeline-steps/generateAssetFuncs.ts
+++ b/bin/clover/src/pipeline-steps/generateAssetFuncs.ts
@@ -52,11 +52,6 @@ export function generateAssetFuncs(
 function generateAssetCodeFromVariantSpec(
   variant: ExpandedSchemaVariantSpec,
 ): string {
-  if (variant.domain.kind !== "object") throw "Domain prop is not object";
-  if (variant.resourceValue.kind !== "object") {
-    throw "ResourceValue prop is not object";
-  }
-
   let declarations = "";
   let adds = "";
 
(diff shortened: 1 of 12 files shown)


bin/clover/src/pipeline-steps/generateAssetFuncs.ts
```diff
@@ -52,11 +52,6 @@ export function generateAssetFuncs(
 function generateAssetCodeFromVariantSpec(
   variant: ExpandedSchemaVariantSpec,
 ): string {
-  if (variant.domain.kind !== "object") throw "Domain prop is not object";
```
jkeiser, line 55: We don't need these checks anymore because domain and resourceValue are specified as objects in the type system.

bin/clover/src/pipeline-steps/generateAssetFuncs.ts
```diff
@@ -205,27 +198,7 @@ function generatePropBuilderString(
...
-        `${
-          prop.data?.defaultValue
-            ? `${indent(indent_level)}.setDefaultValue(${
-              JSON.stringify(prop.data.defaultValue)
-            })\n`
-            : ""
-        }` +
-        `${entryBlock}` +
-        `${indent(indent_level)}.build()`;
+      return generatePropBuilderStringInner(prop.kind, entryBlock);
```
jkeiser, line 201: Gathered all of the variants of generatePropBuilder() into a single generatePropBuilderStringInner(). I verified that it generates identical output (as long as you don't include the new setDocumentation / setDocLink calls).

bin/clover/src/pipeline-steps/generateAssetFuncs.ts
```diff
@@ -240,70 +213,55 @@ function generatePropBuilderString(
...
+        prop.data?.widgetOptions,
+      ) +
+      (
+        prop.data?.defaultValue
+          ? `${indent(indent_level)}.setDefaultValue(${
+            JSON.stringify(prop.data.defaultValue)
+          })\n`
+          : ""
+      ) +
+      (
```
jkeiser, line 249: docLinks and coumentation here

dev/Tiltfile
```diff
@@ -258,6 +258,9 @@ si_buck2_resource(
             path = "/api/",
         ),
     ),
+    links = [
```
jkeiser, line 261: Unrelated, just nice to see the link in tilt and I was in there anyway.

sprutton1 (dismissed):

vbustamante (approved):

PR #5451, opened 2025-02-10
Move special case funcs to intrinsics

nickgerace:
## Introduction

This PR moves the "si:resourcePayloadToValue" and "si:normalizeToArray" funcs to the intrinsics API and package. This was primarily motivated by the performance upside: we no longer run these funcs in veritech, but in whoever invokes the func runner, which is pinga. Now, we move from hundreds of milliseconds of execution time to nearly instant execution of these two funcs.

## Description

The Javascript-based functions are re-written in Rust with as close of a one-to-one logic mapping as possible with "null" and "empty" values and live type evaluation in play.

As a result of these changes, two major things have happened:

1. We have expanded the asset intrinsics panel in the customization interface to allow you to use "si:normalizeToArray" (though you will have to regenerate your asset to install the new func for the first time, which the UI will instruct you to do).
1. We are allowing the old special case funcs and the new intrinsics to co-exist.

Since a chunk of the system relies on func names being unique, achieving those two major "things" was... a total pain in the ass. As you comb this PR, you'll notice all the edge cases. However, we've found them all... famous last words, etc. etc.

What else has changed? This PR fixes the ordering of intrinsic output options in the corresponding panel. They were ordered by name, but now that we are showing the full path (recent change unrelated to this PR), the ordering now relies on the path. The panel also has a fancy new dropdown to explain what each intrinsic func does.

It's worth noting: we now hide the old special case funcs from the user for _any_ asset. They'll be auto-migrated upon regeneration, but you'll notice that only "si:normalizeToArray" is configurable. This is intentional. We are treating the new "si:resourcePayloadToValue" as an internal-only func that we setup for you as part of creating an asset with a real world resource.

Another side effect of these changes is that that test exclusive schemas no longer need to create their own special case funcs. Now, that will get set up automatically. This indirectly fixed some partially finished schemas, like the "Starfield" schema, which is why you now see that the approvals SDF integration test now accounts for auto-enqueued actions. Pretty sick.


app/web/src/components/AssetDetailIntrinsicInput.vue
```diff
@@ -30,17 +28,38 @@
...
+          :disabled="!normalizeToArrayFuncId"
+          :checked="selectedFilter === 'inputSocketForNormalizeToArray'"
+          @select="selectFilter('inputSocketForNormalizeToArray')"
+        />
+        <DropdownMenuItem
+          label="Bind to Prop"
+          checkable
+          :disabled="!normalizeToArrayFuncId"
+          :checked="selectedFilter === 'propForNormalizeToArray'"
+          @select="selectFilter('propForNormalizeToArray')"
```
nickgerace, line 62: Inform the user to regenerate and gray-out/disable the selectable options if the new intrinsic func is not yet available. This is true for old workspaces without any new imports.

app/web/src/components/AssetDetailIntrinsicInput.vue
```diff
@@ -117,7 +141,7 @@ watch(
   { immediate: true },
 );
 
-const emit = defineEmits(["change", "changeToUnset", "changeToIdentity"]);
+const emit = defineEmits(["change", "changeIntrinsicFunc"]);
```
nickgerace, line 144: Simplify the emits. This probably could have been an additional one, but this PR originally had `resourcePayloadToValue` customization. Even though it doesn't now, I think this change is still for the better for the future.

app/web/src/components/AssetDetailsPanel.vue
```diff
@@ -144,21 +144,49 @@
...
+              <li>
+                <strong>Identity:</strong> get the value from another prop or
+                input socket without modifying it
+              </li>
+              <li>
+                <strong>Normalize to Array:</strong> get the value from another
+                prop or input socket, but modify it if the input is empty
+                (becomes "[]") or not an array (wrapped with "[]")
+              </li>
+            </ul>
```
nickgerace, line 188: New toggle-able expansion box explaining your intrinsic options

app/web/src/components/AssetDetailsPanel.vue
```diff
@@ -328,12 +371,7 @@ const _configurableProps = computed(() => {
   variant?.funcIds.forEach((funcId) => {
     const summary = funcStore.funcsById[funcId];
     if (summary?.kind === FuncKind.Intrinsic)
-      if (
-        [FuncBackendKind.Identity, FuncBackendKind.Unset].includes(
-          summary.backendKind,
-        )
-      )
-        return; // don't ignore identity or unset, ignore set string, etc
+      if (INTRINSICS_DISPLAYED.includes(summary.backendKind)) return; // ignore set string, etc
```
nickgerace, line 374: Use the existing const!

app/web/src/components/AssetDetailsPanel.vue
```diff
@@ -398,7 +438,7 @@ const _configurableProps = computed(() => {
     };
     config.push(d);
   });
-  config.sort((a, b) => a.name.localeCompare(b.name));
+  config.sort((a, b) => a.path.localeCompare(b.path));
```
nickgerace, line 441: Subtle, but this fixes the ordering regression unrelated to this work when the props were changed to use full paths recently

lib/si-pkg/src/spec/func.rs
```diff
@@ -77,7 +77,9 @@ pub enum FuncSpecBackendKind {
     JsValidation,
     Management,
     Map,
+    NormalizeToArray,
```
sprutton1, line 80: @vbustamante do we need to regenerate the clover bindings because of this change?

app/web/src/components/AssetFuncListPanel.vue
```diff
@@ -87,10 +87,20 @@ const variantSummaries = computed(() => {
...
     : {};
+
+  // NOTE(nick): filter out deprecated attribute funcs that have become intrinsic funcs.
+  if (r[FuncKind.Attribute]) {
+    r[FuncKind.Attribute] = r[FuncKind.Attribute].filter((f) => {
+      if (f.name === "si:resourcePayloadToValue") return false;
+      else if (f.name === "si:normalizeToArray") return false;
+      return true;
+    });
+  }
```
nickgerace, line 102: We only want to filter out the old special case funcs at the last possible moment. This just removes them from the user's view, but they still come up in the route, which is acceptable. We only want to remove them from the authoring experience.

lib/si-pkg/src/spec/func.rs
nickgerace: You do and don't. I wrote the import to override this so that clover doesn't have to worry about it, but we should change clover anyway

lib/dal-test/src/lib.rs
```diff
@@ -964,7 +964,7 @@ async fn migrate_local_builtins(
     Workspace::setup_builtin(&mut ctx).await?;
 
     info!("migrating intrinsic functions");
-    func::migrate_intrinsics(&ctx).await?;
+    func::migrate_intrinsics_for_tests(&ctx).await?;
```
nickgerace, line 967: Turns out, no one else uses it! Let's make that clear.

lib/dal-test/src/test_exclusive_schemas/fallout.rs
```diff
@@ -50,8 +48,6 @@ pub(crate) async fn migrate_test_exclusive_schema_fallout(
         )
         .build()?;
 
-    let resource_payload_to_value_func = build_resource_payload_to_value_func()?;
```
nickgerace, line 53: You'll see this a lot. Test exclusive schemas can now rely on the intrinsics.

lib/dal-test/src/test_exclusive_schemas/mod.rs
```diff
@@ -243,75 +236,11 @@ pub(crate) async fn migrate(ctx: &DalContext) -> BuiltinsResult<()> {
     .await?;
     Ok(())
 }
-// TODO(nick): remove this if "si:resourcePayloadToValue" becomes an instrinsic func.
```
nickgerace, line 246: 🤜 🤛

lib/sdf-server/src/dal_wrapper/change_set.rs
```diff
@@ -73,6 +73,7 @@ pub async fn status(
         .await?;
     let (requirements, ids_with_hashes_for_deleted_nodes) =
         ApprovalRequirement::list(ctx, &changes).await?;
+    dbg!(&requirements, &ids_with_hashes_for_deleted_nodes, &changes);
```
sprutton1, line 76: Does this need to be cleaned up?

lib/dal/src/attribute/value.rs
```diff
@@ -857,15 +857,6 @@ impl AttributeValue {
             .is_dynamic())
     }
 
-    pub async fn is_set_by_intrinsic(
```
nickgerace, line 860: Unused and potentially dangerous

lib/dal/src/func/runner.rs
```diff
@@ -1669,23 +1672,52 @@ impl FuncRunnerExecutionTask {
...
+                            response_type: self.func.backend_response_type.try_into()?,
+                        };
+                        FuncBackendJsAttribute::create_and_execute(
+                            self.func_dispatch_context,
+                            &self.func,
+                            &serde_json::to_value(args)?,
+                            self.before,
+                        )
+                        .await
+                    }
```
nickgerace, line 1719: Handles the execution of the old version of these funcs running dynamically with the new intrinsic execution code. We `info!` log this intentionally because you will not see it on the veritech resolver function side now.

lib/dal/src/func/runner.rs
```diff
@@ -1739,6 +1771,12 @@ impl FuncRunnerExecutionTask {
                 )
                 .await
             }
+            FuncBackendKind::ResourcePayloadToValue => {
+                FuncBackendResourcePayloadToValue::create_and_execute(&self.args).await
+            }
+            FuncBackendKind::NormalizeToArray => {
+                FuncBackendNormalizeToArray::create_and_execute(&self.args).await
+            }
```
nickgerace, line 1779: This is where the new intrinsics run.

lib/dal/src/pkg.rs
```diff
@@ -252,20 +261,26 @@ where
...
+        }
     }
 
     pub fn insert(&mut self, key: Key, thing: Thing) -> Option<Thing> {
         self.inner.insert(key, thing)
     }
+
+    pub fn insert_override(&mut self, new: Key, old: Key) {
+        self.overrides.insert(new, old);
+    }
```
nickgerace, line 283: Very subtle, but very import. This ensures we are importing/exporting the right intrinsics. Without this, new systems will never migrate to the new intrinsics (also due to the unique func name assumptions, but that's a long story).

lib/dal/src/pkg/export.rs
```diff
@@ -810,9 +810,18 @@ impl PkgExporter {
...
+        if apas.is_empty() {
+            match IntrinsicFunc::maybe_from_str(proto_func.name.as_str()) {
+                Some(IntrinsicFunc::ResourcePayloadToValue)
+                | Some(IntrinsicFunc::NormalizeToArray)
+                    if proto_func.backend_kind == FuncBackendKind::JsAttribute => {}
+                Some(_) => {
+                    return Ok(None);
+                }
+                None => {}
+            }
```
nickgerace, line 824: You'll see this `IntrinsicFunc::maybe_from_str` pattern a lot where we check the func kind or backend kind. That is us accounting for the old funcs. That contributed a huge part to why this PR stayed in drafts  for a chunk of time: lot of papercuts.

lib/dal/src/pkg/import.rs
```diff
@@ -94,28 +92,55 @@ async fn import_change_set(
...
+                    return Err(PkgError::IntrinsicFuncSpecsMultipleForName(
+                        intrinsic.name().to_owned(),
+                    ));
+                }
+
+                // We need to override the unique ID so that accessors grab the correct func.
+                thing_map.insert_override(
+                    override_intrinsic_func_spec.unique_id().into(),
+                    func_spec.unique_id().into(),
+                );
```
nickgerace, line 132: Does a lot of lifting here

lib/dal/src/pkg/import.rs
```diff
@@ -94,28 +92,55 @@ async fn import_change_set(
     Vec<bool /*ImportEdgeSkip*/>,
 )> {
     // let default_change_set_id = ctx.get_workspace_default_change_set_id().await?;
+
+    // Cache the intrinsic funcs pkg in case we need it.
+    let unsafe_to_install_intrinsic_funcs_pkg = SiPkg::load_from_spec(IntrinsicFunc::pkg_spec()?)?;
```
nickgerace, line 97: Do not... install this... just use the func specs. Turns out, you need a full package to get `SiPkgFunc`... so I brought a whole package with me.

lib/dal/src/pkg/import.rs
```diff
@@ -1066,11 +1107,65 @@ pub async fn import_only_new_funcs(
...
+                // Create all arguments for the new intrinsic func. It should be impossible to
+                // find existing arguments, but we should check to avoid duplicate arguments.
+                for argument in override_intrinsic_func_spec.arguments()? {
+                    if FuncArgument::find_by_name_for_func(ctx, argument.name(), func.id)
+                        .await?
+                        .is_none()
+                    {
+                        create_func_argument(ctx, func.id, &argument).await?;
+                    }
+                }
```
nickgerace, line 1152: You'll get the new intrinsics on asset regeneration here.

lib/sdf-server/src/dal_wrapper/change_set.rs
nickgerace: Yes. I'll delete

lib/sdf-server/src/service/v2/func/list_all_funcs.rs
nickgerace, line 1: This route was straight up unused.

vbustamante: We created whits route for debugging why some functions were disappearing months ago, it was only being used by SI people directly. I don't have a problem with deleting it for now, but we should have the habit of leaving comments for those cases when creating them, indeed

vbustamante (dismissed):
This all checks out, we'll regenerate the types for clover at a later point

stack72 (approved):

vbustamante (approved):

stack72 (dismissed):

zacharyhamm (dismissed):

stack72 (approved):

PR #5452, opened 2025-02-10
feat(clover): create input sockets for PolicyDocument props

jkeiser:
Any prop ending with PolicyDocument, which is a text or json type, will become a textare (if it isn't already) and have an input socket with type `policydocument<aws>` on it, to connect to the new Policy Document asset.

It also fixes asset functions for Clover assets with text area props (right now we generate `setWidgetKind("textarea")` when it needs to be `textArea`).

This affects:
* AWS::CodeArtifact::Domain/Repository
* AWS::DynamoDB::Table
* AWS::EC2::VerifiedAccessEndpoint/VerifiedAccessGroup/VPCEndpoint
* AWS::IAM::GroupPolicy/ManagedPolicy/Policy/Role/RolePolicy/UserPolicy
* AWS::IoT::Policy
* AWS::Logs::AccountPolicy/ResourcePolicy
* AWS::NetworkManager::CoreNetwork
* AWS::S3::BucketPolicy
* AWS::S3Express::BucketPolicy
* AWS::S3ObjectLambda::AccessPointPolicy
* AWS::S3Outposts::BucketPolicy
* AWS::SNS::TopicInlinePolicy/TopicPolicy
* AWS::SQS::QueueInlinePolicy/QueuePolicy
* AWS::XRay::ResourcePolicy

## Not Covered

This does not cover props that end with `Policy`. Many cases of this are *not* policy documents (and should not be textarea), however, so I didn't include the ones that are (there are a few) in this PR.

## Testing

* Ran regeneration function in prod and verified sockets connect
* Uploaded locally
* Validated that only the above files are affected

### DIFF
--- a/bin/clover/src/pipeline-steps/createPolicyDocumentInputSockets.ts
+++ b/bin/clover/src/pipeline-steps/createPolicyDocumentInputSockets.ts
@@ -1,0 +1,49 @@
+import { setAnnotationOnSocket } from "../spec/sockets.ts";
+import { ExpandedPropSpec } from "../spec/props.ts";
+import { getOrCreateInputSocketFromProp } from "../spec/sockets.ts";
+import { ExpandedPkgSpec, ExpandedSchemaVariantSpec } from "../spec/pkgs.ts";
+
+//
+// Any prop ending with PolicyDocument is considered an IAM policy document, with these properties:
+//
+// 1. It has a socket with annotation policydocument, which connects to the composable
+//    Policy Document component.
+// 2. It has type=text, widget=TextArea, so the user can copy/paste the JSON from howtos,
+//    websites, or other tools by selecting "set manually."
+//
+// TODO arrays of policies (such as in the IAM Group resource).
+// TODO *Policy (too many props that aren't really policy documents here)
+//
+export function createPolicyDocumentInputSockets(
+  specs: ExpandedPkgSpec[],
+): ExpandedPkgSpec[] {
+  for (const { schemas: [schema] } of specs) {
+    const { variants: [variant] } = schema;
+    createPolicyDocumentInputSocketsFromProp(variant, variant.domain);
+    createPolicyDocumentInputSocketsFromProp(variant, variant.resourceValue);
+  }
+  return specs;
+}
+
+function createPolicyDocumentInputSocketsFromProp(
+  variant: ExpandedSchemaVariantSpec,
+  prop: ExpandedPropSpec,
+) {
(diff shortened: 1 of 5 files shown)


bin/clover/src/pipeline-steps/generateAssetFuncs.ts
```diff
@@ -276,9 +276,7 @@ function generateWidgetString(
     return "";
   }
 
-  const kind = widgetKind === "ComboBox"
-    ? "comboBox"
-    : widgetKind.toLowerCase();
+  const kind = `${widgetKind[0].toLowerCase()}${widgetKind.slice(1)}`;
```
jkeiser, line 279: This fixes the asset functions that generate textarea instead of textArea.

bin/clover/src/spec/sockets.ts
```diff
@@ -1,18 +1,21 @@
...
 import { ExpandedSchemaVariantSpec } from "./pkgs.ts";
+import { Extend } from "../extend.ts";
 
 export const SI_SEPARATOR = "\u{b}";
 
-export type ExpandedSocketSpec = SocketSpec & {
-  data: NonNullable<SocketSpec["data"]>;
-};
+export type ExpandedSocketSpec = Extend<SocketSpec, {
+  data: NonNullable<SocketSpecData>;
```
jkeiser, line 16: This makes it so we bring in SocketSpecData to autocomplete even though the binding files don't import each other correctly.

bin/clover/src/pipeline-steps/createPolicyDocumentInputSockets.ts
```diff
@@ -0,0 +1,49 @@
+import { setAnnotationOnSocket } from "../spec/sockets.ts";
```
sprutton1, line 1: I wonder if this logic should happen in `assetSpecificOverrides`. I had some intention to gather those in one place to make it more obvious where we were special casing things

jkeiser: I didn't put them there because it isn't asset-specific--this feels like an awsSpecificHeuristic to me, maybe?

jkeiser: (I think it's something we can probably move after we merge, too--I'm not sure there's a clear answer on where to move it if anywhere)

vbustamante (approved):
LGTM

PR #5453, opened 2025-02-11
feat(clover): generate arity=one for non-array sockets

jkeiser:
Most sockets are arity one (only allow one thing). This changes them to arity=one, so that other UI features (and things such as incomingConnections) will disallow multiple connections and yield the right types.

## Testing

- Validated that Tags stays arity=many
- Validated that "Subnets" and other array-typed props stay arity=many


sprutton1 (dismissed):

vbustamante (dismissed):
yeah looks good

PR #5454, opened 2025-02-11
bug: escape bad variable names in template generation

adamhjk:
Uses a regex to replace any invalid characters with valid ones, and checks that the leading character is correct.


adamhjk:
https://jam.dev/c/34ad412b-60e3-456d-93ba-02354865a1bc

stack72 (approved):

PR #5459, opened 2025-02-11
feat(si-fs): categorize schemas, add env_logger

zacharyhamm:
List schemas in category folders. You can make a new category with `mkdir`, but it won't be a real category until you make a new schema in that category (with `mkdir` as well, in the category directory).

You can also make a new category by setting the category in attrs.json.

We only list the categories for the *default* schema variants of a schema. So if you change the category of an existing schema on the unlocked attributes, it won't appear in that category in the filesystem until you merge that change set with HEAD. For new schemas, where the unlocked is the default, the schema will be relisted in the new category folder on the next uncached read.

You can now get log output by launching si-fs with RUST_LOG=debug (or trace, info, etc.) in the environment.


Cargo.toml
```diff
@@ -158,6 +159,7 @@ jwt-simple = { version = "0.12.11", default-features = false, features = [
 ] }
 krata-loopdev = "0.0.21"
 lazy_static = "1.5.0"
+log = "0.4"
```
fnichol, line 162: This helps us to get more conditional/leveled/structured log output. I think that we can evolve this to use a decent subset of the existing tracing/telemetry crates although I bet it doesn't work 100% out of the box (as this isn't so much our standard long running service). The important part is that we can start using `info!()`/`error!()`/etc in the code.

bin/si-fs/src/main.rs
```diff
@@ -20,8 +20,6 @@ struct Args {
     foreground: bool,
     #[arg(value_name = "MOUNTPOINT")]
     mount_point: String,
-    #[arg(long, short = 'd')]
-    debug: bool,
```
fnichol, line 24: Yeah, I see how `env_logger` helps to make this more expressive

lib/sdf-server/src/service/v2/fs.rs
```diff
@@ -217,11 +218,42 @@ fn check_change_set(ctx: &DalContext) -> FsResult<()> {
...
+    let mut categories = HashSet::new();
+
+    for schema in dal::Schema::list(&ctx).await? {
+        let default_variant = SchemaVariant::get_default_for_schema(&ctx, schema.id()).await?;
+        categories.insert(default_variant.category().to_string());
+    }
+
+    for module in CachedModule::latest_modules(&ctx).await? {
+        categories.insert(module.category.unwrap_or("".into()));
+    }
```
fnichol, line 244: Yeah, that's pretty nice!

lib/si-filesystem/src/client.rs
```diff
@@ -239,26 +246,27 @@ impl SiFsClient {
...
         let mut request_builder = self.client.get(url.clone()).bearer_auth(&self.token);
         request_builder = if let Some(query) = query.clone() {
             request_builder.query(&query)
         } else {
             request_builder
         };
 
-        let start = Instant::now();
-        let response = request_builder.send().await?;
+        let response = tokio::spawn(async move { request_builder.send().await }).await??;
```
fnichol, line 259: I'm not totally sure I understand the reason for spawning this task and then awaiting it. Could this have been from some other work that didn't get reverted--or maybe I don't see it here :)

lib/si-filesystem/src/lib.rs
```diff
@@ -71,6 +70,7 @@ const DIR_STR_FUNCTIONS: &str = "functions";
 const DIR_STR_SCHEMAS: &str = "schemas";
 const DIR_STR_LOCKED: &str = "locked";
 const DIR_STR_UNLOCKED: &str = "unlocked";
+const DIR_STR_BLANK_CATEGORY: &str = "__UNCATEGORIZED__";
```
fnichol, line 73: Yikes, I guess this is a possibility in that I'm guessing it's technically an optional attribute.

lib/si-filesystem/src/lib.rs
```diff
@@ -1917,12 +1936,14 @@ impl SiFileSystem {
...
             .variant_funcs_of_kind(*change_set_id, *schema_id, *kind)
             .await?;
+
         let mut existing_entries: HashMap<Inode, InodeEntry> = self
             .inode_table
             .direct_child_entries(entry.ino)?
-            .iter()
-            .map(|entry| (entry.ino, entry.clone()))
+            .into_iter()
+            .map(|entry| (entry.ino, entry))
```
fnichol, line 1944: 😄

fnichol (dismissed):
Nicely done and I'm kind of loving being able to see `fuser` and `hyer` logging output now that a logging subsystem is running. In a way, I'm amazed we got so much working without the `fuser` logging to help. Nothing blocking in here, just a question or two. I tested this out and was seeing the category directories coming back nice and peppy. I don't think I have my local editor setup yet to load and understand the `tsconfig.json` file but that file is rendered as I'd expect.

lib/si-filesystem/src/client.rs
zacharyhamm: oh! good call. I was experimenting with reqwest which is just *slow* for some reason. Requests that you can make in 50ms with `curl` take 250ms with reqwest and I can't figure out why. I didn't mean to leave this in. I'll clean it up in a separate PR

fnichol (approved):

PR #5464, opened 2025-02-12
feat(hoist, module-index-server): Adding a new upsert_builtin endpoint

stack72:
Hoist was calling 3 API endpoints in an unsafe manner - now we wrap all of those into a new endpoint that has a transaction when we know it's successful


lib/module-index-server/src/routes/reject_module_route.rs
```diff
@@ -107,3 +113,60 @@ pub async fn reject_module(
...
+            ))),
+            rejected_by_display_name: Set(Some(data.clone())),
+            kind: Set(module.kind),
+            is_builtin_at: Set(module.is_builtin_at),
+            is_builtin_at_by_display_name: Set(module.is_builtin_at_by_display_name),
+            schema_id: Set(module.schema_id),
+            ..Default::default() // all other attributes are `NotSet`
+        };
+
+        let _: si_module::Model = dbg!(active_module.update(txn).await)?;
```
vbustamante, line 168: This dbg should probably not stay (but you also don't need to store the returned value on a variable):
```suggestion
        active_module.update(txn).await?;
```

lib/module-index-server/src/routes/upsert_builtin_route.rs
```diff
@@ -0,0 +1,85 @@
...
+    if let Some(schema_id) = multiparts.schema_id.clone() {
+        let existing_schema_id = SchemaId::from_str(&schema_id)?;
+        reject_other_modules_of_a_schema_id("Clover".to_string(), existing_schema_id, &txn).await?;
+    }
+
+    // Upload the new module
+    let new_module = upsert_module(multiparts, &txn, user_claim, s3_bucket).await?;
+
+    // Promote the new module to be a builtin
+    let (_, _) = promote_module(new_module.id, &txn, "Clover".to_string()).await?;
```
vbustamante, line 80: Again, no need to put it in a variable
```suggestion
    promote_module(new_module.id, &txn, "Clover".to_string()).await?;
```

vbustamante (approved):
Nice!

PR #5466, opened 2025-02-12
fix: koa CVE-2025-25200

johnrwatson:
Fixes critical vuln in Koa, used within the auth api https://github.com/systeminit/si/security/dependabot/380


stack72 (approved):

stack72:
Tested the auth=api locally for this and it works

PR #5467, opened 2025-02-12
feat(clover): Prettify socket names

vbustamante:
Make sure socket names have spaces instead of being pascal case

### DIFF
--- a/bin/clover/src/pipeline-steps/prettifySocketNames.ts
+++ b/bin/clover/src/pipeline-steps/prettifySocketNames.ts
@@ -1,5 +1,3 @@
-import _logger from "../logger.ts";
-import _ from "npm:lodash";
 import { ExpandedPkgSpec } from "../spec/pkgs.ts";
 import { setAnnotationOnSocket } from "../spec/sockets.ts";
 
@@ -25,7 +23,6 @@ export function prettifySocketNames(
     }
 
     newSpecs.push(spec);
-    break;
   }
 
   return newSpecs;
--- a/bin/clover/src/pipeline-steps/prettifySocketNames.ts
+++ b/bin/clover/src/pipeline-steps/prettifySocketNames.ts
@@ -1,0 +1,32 @@
+import _logger from "../logger.ts";
+import _ from "npm:lodash";
+import { ExpandedPkgSpec } from "../spec/pkgs.ts";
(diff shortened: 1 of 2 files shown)


bin/clover/src/pipeline-steps/prettifySocketNames.ts
```diff
@@ -0,0 +1,32 @@
...
+        .replace(/([A-Z]+)([A-Z][a-z])/g, "$1 $2");
+
+      socket.name = newName;
+      socket.data.name = newName;
+
+      setAnnotationOnSocket(socket, newName);
+    }
+
+    newSpecs.push(spec);
+    break;
```
vbustamante, line 28: 🙃 
```suggestion
```

bin/clover/src/pipeline-steps/prettifySocketNames.ts
```diff
@@ -0,0 +1,31 @@
+import _logger from "../logger.ts";
+import _ from "npm:lodash";
```
vbustamante, line 2:
```suggestion
```

bin/clover/src/pipeline-steps/prettifySocketNames.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  for (const spec of specs) {
+    const sockets = spec.schemas[0].variants[0].sockets;
+
+    for (const socket of sockets) {
+      const newName = socket.name
+        // separate any sequence of lowercase letters followed by an uppercase letter
+        .replace(/([a-z])([A-Z])/g, "$1 $2")
+        // Separate any sequence of more than 1 of uppercase letters (acronyms) from the next word
+        .replace(/([A-Z]+)([A-Z][a-z])/g, "$1 $2");
```
stack72, line 17: so this will create `Task Role ARN` as a socket name right?

vbustamante: Yes, but aws actually calls arns "Arn" in the sockets that I tested so this wouldn't interact with that. But that's the idea, yeah

stack72: ah yeah of course not ARN but Arn - I gotcha :)

stack72 (approved):

PR #5468, opened 2025-02-13
fix(clover): Hook up inputs to proper space-separated socket names

jkeiser:
Packages aren't uploading right now, so we can't see the beautiful new socket names! The issue was that props that hook up to an input socket were referencing the old names without spaces, while the sockets themselves had spaces added.

Test: uploaded a couple of assets with @vbustamante 's beautiful new module uploader and dragged them onto the canvas. Spaces galore!

image


vbustamante:
Great catch!

vbustamante (approved):

PR #5470, opened 2025-02-13
feat: start lang-js with cyclone

sprutton1:
Instead of starting lang-js at the time we execute a function, we will start it when cyclone starts. This should helps us churn through functions more quickly when pool is full as we won't have to incur the ~80ms lang-js startup time.

### DIFF
--- a/bin/lang-js/src/index.ts
+++ b/bin/lang-js/src/index.ts
@@ -29,8 +29,6 @@ function onError(
 }
 
 async function main() {
-  let kind: FunctionKind | undefined;
-
   const { options } = await new Command()
     .name("lang-js")
     .version("0.0.1")
@@ -39,15 +37,6 @@ async function main() {
       `timeout for a function execution in seconds (default: ${defaultTimeout})`,
       { default: defaultTimeout },
     )
-    .arguments("<kind:string>")
-    .action((_options, kind_arg: string) => {
-      if (functionKinds().includes(kind_arg)) {
-        kind = kind_arg as FunctionKind;
-      } else {
-        console.error(`Unsupported function kind: '${kind_arg}'`);
-        Deno.exit(1);
-      }
-    })
     .parse(Deno.args);
 
   let timeout: number = defaultTimeout;
@@ -73,11 +62,19 @@ async function main() {
     }
     debug({ request: requestJson });
     const request: Request = JSON.parse(requestJson);
+
     if (request.executionId) {
       executionId = request.executionId;
     } else {
       throw Error("Request must have executionId field");
     }
+
+    if (!request.kind) {
+      throw Error("Request must have a kind field");
+    }
+
+    debug({ request });
+
     // Now we have the executionId, so update our console.error() impl
     errorFn = makeConsole(executionId).error;
 
@@ -86,11 +83,7 @@ async function main() {
       onError(errorFn, err, executionId);
     });
 
-    if (kind === undefined) {
-      throw Error(`Kind is undefined`);
-    }
-
-    await executeFunction(kind, request, timeout);
+    await executeFunction(request, timeout);
   } catch (err) {
     onError(errorFn, err as Error, executionId);
   }
--- a/lib/cyclone-core/src/kill_execution.rs
+++ b/lib/cyclone-core/src/kill_execution.rs
@@ -15,6 +15,10 @@ impl CycloneRequestable for KillExecutionRequest {
         &self.execution_id
     }
 
+    fn kind(&self) -> &str {
+        ""
+    }
+
     fn websocket_path(&self) -> &str {
         ""
     }
--- a/lib/cyclone-core/src/request.rs
+++ b/lib/cyclone-core/src/request.rs
@@ -38,6 +38,7 @@ pub trait CycloneRequestable {
     type Response;
 
     fn execution_id(&self) -> &str;
+    fn kind(&self) -> &str;
     fn websocket_path(&self) -> &str;
     fn inc_run_metric(&self);
     fn dec_run_metric(&self);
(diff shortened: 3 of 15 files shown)


bin/lang-js/src/index.ts
```diff
@@ -39,15 +37,6 @@ async function main() {
       `timeout for a function execution in seconds (default: ${defaultTimeout})`,
       { default: defaultTimeout },
     )
-    .arguments("<kind:string>")
```
fnichol, line 42: Yep, that was the tiny change that makes all other changes necessary! Doesn't look as bad as I thought it might

lib/cyclone-core/src/kill_execution.rs
```diff
@@ -15,6 +15,10 @@ impl CycloneRequestable for KillExecutionRequest {
         &self.execution_id
     }
 
+    fn kind(&self) -> &str {
+        ""
```
fnichol, line 19: Funky. But I get it

lib/cyclone-core/src/request.rs
```diff
@@ -38,6 +38,7 @@ pub trait CycloneRequestable {
     type Response;
 
     fn execution_id(&self) -> &str;
+    fn kind(&self) -> &str;
```
fnichol, line 41: I'm a bit surprised we never had this, but I think it's the way we transmitted the requests that allowed us to know the kind of function (that it, the NATS subject that we subscribe to let us know). This does feel clearer and understand that it becomes a lot more required *inside* the running `lang-js` now.

lib/cyclone-core/src/resolver_function.rs
```diff
@@ -62,6 +62,10 @@ impl CycloneRequestable for ResolverFunctionRequest {
         &self.execution_id
     }
 
+    fn kind(&self) -> &str {
+        "resolverfunction"
```
fnichol, line 66: One day, one day, one day we'll rename this

lib/cyclone-server/src/handlers.rs
```diff
@@ -100,155 +96,120 @@ pub async fn ws_execute_ping(
 
 pub async fn ws_execute_resolver(
     wsu: WebSocketUpgrade,
-    State(lang_server_path): State<LangServerPath>,
-    State(telemetry_level): State<TelemetryLevel>,
-    State(lang_server_function_timeout): State<LangServerFunctionTimeout>,
     State(lang_server_process_timeout): State<LangServerProcessTimeout>,
+    State(child): State<LangServerChild>,
```
fnichol, line 100: Oh yeah, wow, that is different--the process is already running and we only need to provide a handle to the child process.

lib/cyclone-server/src/state.rs
```diff
@@ -1,38 +1,59 @@
...
+            cmd.env("SI_LANG_JS_LOG", "*");
+        }
+
+        debug!(cmd = ?cmd, "spawning child process");
+        let child = cmd
+            .spawn()
+            .map_err(|err| ExecutionError::ChildSpawn(err, lang_server_path.into()))?;
+
+        Ok(Self {
+            child: LangServerChild(Arc::new(Mutex::new(child))),
```
fnichol, line 51: I was wonder why you need all this, then remembered that in Axum the `AppState` must impl `Clone` for each request, and so this make sense. It's probably only really used for that single Cyclone request anyway. Neato

lib/cyclone-server/src/state.rs
```diff
@@ -65,6 +86,15 @@ impl LangServerFunctionTimeout {
     }
 }
 
+#[derive(Clone, Debug, FromRef)]
+pub struct LangServerChild(Arc<Mutex<Child>>);
```
fnichol, line 90: 👍🏽

lib/cyclone-server/src/execution.rs
```diff
@@ -202,7 +175,14 @@ where
...
 
     async fn child_send_function_request(stdin: ChildStdin, request: Request) -> Result<()> {
-        let value = serde_json::to_value(&request).map_err(ExecutionError::JSONSerialize)?;
+        let mut value = serde_json::to_value(&request).map_err(ExecutionError::JSONSerialize)?;
+
+        if let serde_json::Value::Object(ref mut map) = value {
+            map.insert(
+                "kind".to_string(),
+                serde_json::Value::String(request.kind().to_string()),
+            );
```
fnichol, line 184: I like this--I believe you managed to avoid changing the Veritech client's request types as a result. A nice spot to add it in, right before it's needed and only until it's needed.

lib/cyclone-server/src/execution.rs
```diff
@@ -302,11 +282,13 @@ where
             Result::<_>::Ok(())
         };
 
+        let inner = self.child.inner();
+        let mut child_lock = inner.lock().await;
```
fnichol, line 286: 👍🏽

fnichol (approved):
This is excellent work, well done! Given how much this code is exercised in *everything* (i.e. integration tests, local dev, etc) I'm reasonably confident in this working without performing a lot of testing myself (assuming that's okay with you).

lib/cyclone-server/src/state.rs
sprutton1: Yes, exactly. I tried with just an Arc, but found ~14 strong references to it when I went to unwrap it. I assumed axum was doing this under the covers.

lib/cyclone-server/src/execution.rs
sprutton1: Yep, I want to limit the blast radius a bit and we can continue to use the request types as-is. Plus, having the the kind on the relevant trait means that we can't forget to implement it

lib/cyclone-core/src/resolver_function.rs
sprutton1: Actually, I'll just do it here and now since I'm already touching both sides of the relationship anyways.

fnichol (approved):

PR #5475, opened 2025-02-13
chore: bump foyer to 14.1

sprutton1:
I don't have definitive evidence, but I suspect this change will help with the memory pressures we have not been able to resolve. Specifically, [this change](https://github.com/foyer-rs/foyer/commit/170a4108b5dcccbfbcfe0912bcd36eb0eef6dd78).

> fix: fix memory leak caused by circle reference count

This has some interesting implications around how the disk cache works and make me wonder if we'll be writing to it as often. The ideal situation here is that we'll be caching old entries without busting our memory limit and flushing all entries when stopping the service.

Note that they broke out the metrics collection to a separate crate, so I brought that in here.


zacharyhamm (approved):
🤞 this improves our memory leak issues

PR #5479, opened 2025-02-14
build(buck2): impl `nix_binary` Buck2 rule to build/publish/promote

fnichol:
This change introduces a `nix_binary` Buck2 rule (and macro) which is used in `bin/si-fs` to build, publish, and promote a "standalone" binary of the `si-fs` FUSE filesystem daemon.


sprutton1 (approved):
LGTM! Looks like a lot of what we already do for the omnibus packages minus the flake deps, which is what I expected. Nice work.

PR #5481, opened 2025-02-14
feat: stream console messages from func executions

sprutton1:
stream `console.log` and `console.error` from lang-js as they occur in the function execution. This should resolve issues with logs not being displayed correctly in the web client as it more closely matches the pre-deno behavior.

I confirmed that this results in us pushing the console.log lines one by one across nats to whomever is listening on the otherside and that long function executions result in the entirety of the log being present on first glance


adamhjk (approved):
lgtm - tested in my dev environment.

PR #5474, opened 2025-02-13
chore: nix flake update

adamhjk:
Upgrade the versions of whats in our nix flake, in particular to catch an updated version of cfn-lint

PR #5485, opened 2025-02-14
Move special case funcs to intrinsics (ROUND TWO)

nickgerace:
## Introduction

This PR moves the "si:resourcePayloadToValue" and "si:normalizeToArray" funcs to the intrinsics API and package. This was primarily motivated by the performance upside: we no longer run these funcs in veritech, but in whoever invokes the func runner, which is pinga. Now, we move from hundreds of milliseconds of execution time to nearly instant execution of these two funcs.

## Description

The Javascript-based functions are re-written in Rust with as close of a one-to-one logic mapping as possible with "null" and "empty" values and live type evaluation in play.

As a result of these changes, two major things have happened:

1) We have expanded the asset intrinsics panel in the customization
   interface to allow you to use "si:normalizeToArray" (though you will
   have to regenerate your asset to install the new func for the first
   time, which the UI will instruct you to do).
2) We are allowing the old special case funcs and the new intrinsics to
   co-exist.

Since a chunk of the system relies on func names being unique, achieving those two major "things" was... a total pain in the ass. As you comb this PR, you'll notice all the edge cases. However, we've found them all... famous last words, etc. etc.

What else has changed? This PR fixes the ordering of intrinsic output options in the corresponding panel. They were ordered by name, but now that we are showing the full path (recent change unrelated to this PR), the ordering now relies on the path. The panel also has a fancy new dropdown to explain what each intrinsic func does.

It's worth noting: we now hide the old special case funcs from the user for _any_ asset. They'll be auto-migrated upon regeneration, but you'll notice that only "si:normalizeToArray" is configurable. This is intentional. We are treating the new "si:resourcePayloadToValue" as an internal-only func that we setup for you as part of creating an asset with a real world resource.

Another side effect of these changes is that that test exclusive schemas no longer need to create their own special case funcs. Now, that will get set up automatically. This indirectly fixed some partially finished schemas, like the "Starfield" schema, which is why you now see that the approvals SDF integration test now accounts for auto-enqueued actions. Pretty sick.

## History

Original PR: https://github.com/systeminit/si/pull/5451
Revert PR: https://github.com/systeminit/si/pull/5478


lib/dal/src/pkg/import.rs
```diff
@@ -1033,17 +1055,62 @@ pub async fn import_only_new_funcs(
...
+                            create_func_argument(ctx, func.id, &argument).await?;
+                        }
+                    }
 
-            Func::get_by_id_or_error(ctx, func_id).await?
+                    // However, use the _provided_ func spec when inserting into the thing map. Why
+                    // not use the override? Because we are initializing the thing map here and the
+                    // users of the thing map don't need the "overrides" re-direction: they just
+                    // have the func they need here.
+                    thing_map.insert(func_spec.unique_id().into(), Thing::Func(func));
```
nickgerace, line 1104: This is the fix to the bug causing the revert. The attached screenshot explains the old behavior and the new behavior.

Screenshot 2025-02-14 at 3 57 57 PM

lib/dal/src/pkg/import.rs
```diff
@@ -94,38 +92,65 @@ async fn import_change_set(
...
+                );
+
                 let func = import_func(
                     ctx,
-                    func_spec,
+                    &override_intrinsic_func_spec,
                     installed_module.clone(),
                     thing_map,
-                    options.create_unlocked,
+                    false,
```
nickgerace, line 139: We learned that you may not need the overrides map at all anymore so long as the `func_spec.unique_id()` is pointing to the right "thing". However, this import code used for `import_change_set` calls `import_func` which handles thing map insertion internally... so you need the override unless you change `import_func` to have a new parameter to say "hey, use this override key to insert into the thing map instead". If you're reading this and your brain is fried, mine is too.

lib/dal/src/pkg.rs
```diff
@@ -252,20 +261,27 @@ where
...
     }
 
     pub fn insert(&mut self, key: Key, thing: Thing) -> Option<Thing> {
         self.inner.insert(key, thing)
     }
+
+    pub fn insert_override(&mut self, to_override_unique_id: Key, override_unique_id: Key) {
+        self.overrides
+            .insert(to_override_unique_id, override_unique_id);
+    }
```
nickgerace, line 284: We also found that these were incorrectly inverted, though we believe that to have not been the core cause of the rever.

nickgerace:
Cancelled my try until @fnichol's work on CI lands and https://github.com/systeminit/si/pull/5474 is merged.

stack72 (approved):

nickgerace:
Waiting on https://buildkite.com/system-initiative/si-merge-main/builds/2805, but that requires no rebasing.

fnichol:
I rebased against main to try this out again, think this'll do it

nickgerace:
Thanks @fnichol!

nickgerace:
Rebased and trying again.

PR #5486, opened 2025-02-14
feat(clover): Give sockets more descriptive names

vbustamante:


bin/clover/src/spec/sockets.ts
```diff
@@ -50,6 +68,36 @@ export function createInputSocketFromProp(
...
+  let socketName = prop.name;
+  // If we get a prop inside an object on domain, let's name its socket a bit better
+  if (
+    propPath.length > 3 && propPath[2] !== "extra"
+  ) {
+    // Remove any unnecessary identifiers so the socket name does not become enormous
+    // Regex says "remove this token if on end of string"
+    const propParentName = propPath.slice(2, -1).map((name) =>
+      name
+        .replace(/Configuration$/, "")
```
jkeiser, line 82: Nice enumeration!

jkeiser (approved):
I think we're good here, nice!

I was briefly worried that we might pull in "fake" names like for children of arrays and maps ... but I don't think we'll ever have sockets for anything *under* an array/map. At least not now.

PR #5487, opened 2025-02-14
fix: console.* calls dropped in sandbox bundle

sprutton1:
image

If you drop the console from the sandbox build, the console will no longer work in the sandbox.


johnrwatson (approved):

fnichol:
I rebased this branch to benefit from the build software updates

PR #5490, opened 2025-02-14
Fix two flake-y tests by waiting for actions

nickgerace:
This PR fixes two flake-y tests by waiting for actions to run related to "Starfield". Interestingly, this should have only started happening after #5451, which was reverted by #5478 and is intended to come back on "main" after #5485 is approved and merged.

That being said, this can happen for a number of reasons, but the key takeaway is that we should be waiting for actions to run whenver possible in our tests.


fnichol (approved):
Looks good and we'll confirm by running the `si/merge-queue` pipeline

PR #5491, opened 2025-02-14
feat(si-fs): support deno

zacharyhamm:
Supports deno by configuring a workspace root deno.json and a .vscode/settings.json with "deno: enable". In order to support deno we prefetch every function so we can fill out the deno.json workspaces list.

If new functions come in, or are created, *you will have to reload your deno workspace to get them parsed by the deno language server*.

`node_modules` for deno are not yet supported. They can be! But a passthrough will have to be implemented so that the node module tree can be written to disk in another folder, and mirrored in the si-filesystem system.

### DIFF
--- a/lib/sdf-server/src/service/v2/fs.rs
+++ b/lib/sdf-server/src/service/v2/fs.rs
@@ -1,4 +1,7 @@
-use std::collections::{HashMap, HashSet};
+use std::{
+    collections::{BTreeMap, HashMap, HashSet},
+    sync::Arc,
+};
 
 use axum::{
     extract::{Path, Query},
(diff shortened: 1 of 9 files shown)


fnichol:
I rebased this work to make sure that it would benefit from updated Nix software

fnichol:
I'll review today, sorry I ran out of time Friday

fnichol (approved):
Looks great! While I don't yet have a decent vscode setup, I see the various `tsconfig.json` and `deno.json` files, I can see their contents look right, and the file permissions are set correctly. Nice!

fnichol:
Note that this PR will produce an `si-fs` that relies on a newly deployed SDF /cc @stack72

PR #5494, opened 2025-02-15
feat(web): Add a link to the schema docs for an asset

stack72:
https://github.com/user-attachments/assets/e8729a76-c3b0-43a1-a5b2-0f7906526c24

### DIFF
--- a/app/web/src/components/ComponentCard.vue
+++ b/app/web/src/components/ComponentCard.vue
@@ -80,6 +80,14 @@
       </div>
 
       <!-- Slot for additional icons/buttons -->
+      <IconButton
+        v-if="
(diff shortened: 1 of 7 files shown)


app/web/src/components/ComponentCard.vue
```diff
@@ -80,6 +80,14 @@
       </div>
 
       <!-- Slot for additional icons/buttons -->
```
wendybujalski, line 82: this comment should move to be next to the slot or be deleted

wendybujalski (approved):
Looks good!

app/web/src/components/ComponentCard.vue
stack72: I will follow up here

PR #5495, opened 2025-02-15
feat(si-fs): Add a default for the endpoint url

stack72:
Also adds env var options for the parameters


fnichol:
I rebased this and slightly tweaked the settings around echoing back the values of environment variables.

fnichol (approved):
This looks great, I tested it out and make a small change and otherwise this is good to go.

britmyerss (approved):

PR #5496, opened 2025-02-15
feat(clover): Infer enum values from descriptions using AI

jkeiser:
This adds `clover infer-ai`, which uses AI to infer enum values from the descriptions of cloud formation properties.

The intent is that we run this outside of the normal generate-specs process (since it requires an OpenAI setup, etc.). When it runs, it produces `inferred-enums.json`, which contains all of the enums generated (keyed by the description in the cloud formation property). This inferred file doubles as a cache; re-running inference will not generate any new values unless you delete the file or delete the enums you want regenerated.

The next step of this process will be to hook the overrides into generation, which can be done by looking up `enum` by description after loading the cloudformation db.

## Draft

This is draft because it's not really complete without hooking up to the generator.


bin/clover/src/cli.ts
```diff
@@ -36,8 +39,11 @@ To generate all specs:
 To generate all specs containing "ECS" or "S3", you can pass some services as arguments:
 
   clover generate-specs ECS S3
-`
+`,
     )
+    .option("--inferred <file:string>", "Inferred database location", {
+      default: "inferred.json",
+    })
```
nickgerace, line 46: nice

bin/clover/src/cli.ts
```diff
@@ -46,11 +52,50 @@ To generate all specs containing "ECS" or "S3", you can pass some services as ar
...
+  clover infer-ai ECS S3
+
+By default, the AI will only run inference for things that have not already been inferred. To
+force the AI to re-infer all specs, pass the --force flag.
+`,
+    )
+    .option("--force", "Force re-inference on cached values")
+    .option("--inferred <file:string>", "Inferred database location", {
+      default: "inferred.json",
+    })
```
nickgerace, line 86: Is this a duplicate field or am I misreading this?

bin/clover/src/commands/inferAi.ts
```diff
@@ -0,0 +1,186 @@
...
+are absolutely sure.
+
+For example, given this CloudFormation documentation for the "Mode" property of AWS::ApiGateway::RestApi:
+
+  This property applies only when you use OpenAPI to define your REST API. The \`\`Mode\`\` determines how API Gateway handles resource updates.
+  Valid values are \`\`overwrite\`\` or \`\`merge\`\`.
+  For \`\`overwrite\`\`, the new API definition replaces the existing one. The existing API identifier remains unchanged.
+  For \`\`merge\`\`, the new API definition is merged with the existing API.
+  If you don't specify this property, a default value is chosen. For REST APIs created before March 29, 2021, the default is \`\`overwrite\`\`. For REST APIs created after March 29, 2021, the new API definition takes precedence, but any container types such as endpoint configurations and binary media types are merged with the existing API.
+  Use the default mode to define top-level \`\`RestApi\`\` properties in addition to using OpenAPI. Generally, it's preferred to use API Gateway's OpenAPI extensions to model these properties.
```
nickgerace, line 26: Fascinating

bin/clover/src/commands/inferAi.ts
```diff
@@ -0,0 +1,186 @@
...
+
+You would respond with:
+
+  [ "overwrite", "merge" ]
+
+You must respond *only* with a JSON array of strings, or a JSON null value, and no other text.
+Do not include formatting or markdown delimiters. Do not include invalid JSON, and triple check
```
nickgerace, line 33: Good shoutout

nickgerace (approved):

bin/clover/src/commands/inferAi.ts
jkeiser: That is actually how we fixed a bug :) "Stop doing that bug."

bin/clover/src/cli.ts
jkeiser: Just an option you can pass to multiple subcommands (but not all of them). Cliffy is kind of unintuitive: each time you see `.command()` it starts another block of options. I looked to see what the right way was, but I think we may be using a different version from the modern docs.

PR #5498, opened 2025-02-17
feat(clover): Adding an override for the secrets manager secret

stack72:
This is the first part of this work. We now need to ensure that we can use this prop as part of the payload we send to AWS
This isn't specific to SecretManager:;Secret, this is a pattern that we will need to replicate with other assets like KmsKey and RDS DBInstance USername/Password

### DIFF
--- a/bin/clover/src/cloud-control-funcs/actions/awsCloudControlCreate.ts
+++ b/bin/clover/src/cloud-control-funcs/actions/awsCloudControlCreate.ts
@@ -16,49 +16,12 @@ async function main(component: Input): Promise<Output> {
     };
   }
 
-  console.log("code", codeString);
   const domain = component.properties?.domain;
   const code = JSON.parse(codeString);
 
   const payload = code["DesiredState"];
   const propUsageMap = JSON.parse(domain.extra.PropUsageMap);
-  if (
-    !Array.isArray(propUsageMap.secrets)
-  ) {
-    throw Error("malformed propUsageMap on asset");
-  }
-
-  for (
-    const {
-      secretKey,
-      propPath,
-    } of propUsageMap.secrets
-  ) {
-    const secret = requestStorage.getItem(secretKey);
-
-    if (!propPath?.length || propPath.length < 1) {
-      throw Error("malformed secret on propUsageMap: bad propPath");
-    }
-    if (secret) {
-      let secretParent = payload;
-      let propKey = propPath[0];
-      for (let i = 1; i < propPath.length; i++) {
-        const thisProp = secretParent[propKey];
-
-        if (!thisProp) {
-          break;
-        }
-
-        secretParent = secretParent[propKey];
-        propKey = propPath[i];
-      }
-
-      // Only add secret to payload if the codegen output has it
-      if (secretParent[propKey] === "[redacted]") {
-        secretParent[propKey] = secret;
-      }
-    }
-  }
+  addSecretsToPayload(payload, propUsageMap);
 
   const inputObject = {
     TypeName: code["TypeName"],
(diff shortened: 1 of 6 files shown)


bin/clover/src/cloud-control-funcs/actions/awsCloudControlCreate.ts
```diff
@@ -19,9 +19,50 @@ async function main(component: Input): Promise<Output> {
...
+
+  const payload = code["DesiredState"];
+  const propUsageMap = JSON.parse(domain.extra.PropUsageMap);
+  if (
+    !Array.isArray(propUsageMap.secrets)
+  ) {
+    throw Error("malformed propUsageMap on asset");
+  }
+
+  for (
```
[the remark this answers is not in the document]
jkeiser: i.e. "this was a copy/paste out of necessity, so if you change how it works, change it everywhere it was copy/pasted"

bin/clover/src/cloud-control-funcs/actions/awsCloudControlCreate.ts
```diff
@@ -19,9 +19,50 @@ async function main(component: Input): Promise<Output> {
...
+        if (!thisProp) {
+          break;
+        }
+
+        secretParent = secretParent[propKey];
+        propKey = propPath[i];
+      }
+
+      // Only add secret to payload if the codegen output has it
+      if (secretParent[propKey] === "[redacted]") {
```
jkeiser, line 57: This is an example why :) (Different frm the CloudControlUpdate one)

bin/clover/src/cloud-control-funcs/code-gen/awsCloudControlCodeGenUpdate.ts
```diff
@@ -13,13 +13,40 @@ async function main(component: Input): Promise<Output> {
...
+      secretKey,
+      propPath,
+    } of propUsageMap.secrets
+  ) {
+    const secret = requestStorage.getItem(secretKey);
+
+    if (!propPath?.length || propPath.length < 1) {
+      throw Error("malformed secret on propUsageMap: bad propPath");
+    }
+    if (secret) {
```
jkeiser, line 36: Was noodling about how to do this with less integer indexes. If this appeals to you, feel free ... code is fine as is!

```suggestion
  // Grab the secret prop name and create its parent
  let secretName = propPath.pop();
  let secretParent = payload;
  for (const name of propPath) {
    secretParent[name] ??= {};
    secretParent = secretParent[name];
  }
  secretParent[secretName] = secret;
```

jkeiser: Not sure how to expand a suggestion to include multiple lines :/

bin/clover/src/pipeline-steps/assetSpecificOverrides.ts
```diff
@@ -104,5 +105,95 @@ const overrides = new Map<string, OverrideFn>([
     setAnnotationOnSocket(socket, { tokens: ["Id"] });
 
     variant.sockets.push(socket);
-  }]
+  }],
+  [
+    "AWS::SecretsManager::Secret",
+    addSecretProp("Secret String", "secretString", ["SecretString"]),
```
jkeiser, line 111: I like this pattern!

bin/clover/src/pipeline-steps/assetSpecificOverrides.ts
```diff
@@ -104,5 +105,95 @@ const overrides = new Map<string, OverrideFn>([
...
+
+      secretProp = thisProp;
+    }
+
+    if (!secretProp) {
+      console.log(`Could not add secret value for ${spec.name}`);
+      return;
+    }
+
+    // Find propUsageMap
```
jkeiser, line 156: Not at all needed right now, but it feels like we're begging for a getPropFromPropPath() method here!

bin/clover/src/pipeline-steps/assetSpecificOverrides.ts
```diff
@@ -104,5 +105,95 @@ const overrides = new Map<string, OverrideFn>([
...
+      propUsageMapProp?.data.defaultValue,
+    ) as PropUsageMap;
+
+    if (!propUsageMapProp || !Array.isArray(propUsageMap?.secrets)) {
+      return;
+    }
+
+    // Remove secret from the domain tree
+    secretParent.entries = secretParent.entries.filter((
+      p: ExpandedPropSpec,
```
jkeiser, line 174: For future ref, you (shouldn't) have to annotate this. (If you do we need to fix the type :) )

bin/clover/src/pipeline-steps/assetSpecificOverrides.ts
```diff
@@ -104,5 +105,95 @@ const overrides = new Map<string, OverrideFn>([
...
+
+function addSecretProp(
+  secretKind: string,
+  secretKey: string,
+  propPath: string[],
+) {
+  return (spec: ExpandedPkgSpec) => {
+    const variant = spec.schemas[0].variants[0];
+
+    const [secretName] = propPath.slice(-1);
```
jkeiser, line 123: Makes you wanna pop() eh :) But I imagine this is to make sure you don't mess up the propPath for the caller

jkeiser (approved):
Looking good! I think you might need the functions inside main(), I'll reapprove if that happens

bin/clover/src/pipeline-steps/assetSpecificOverrides.ts
vbustamante: yup, the JS api for non destructive array operation is annoying but it made no sense to modify it and have something break in the future

jkeiser (approved):

sprutton1 (approved):

PR #5499, opened 2025-02-17
fix(clover): Adding an override to allow a secrets manager secret to connect to a Container Def Secret

stack72:


vbustamante (approved):
Looks good! Is "Value From" a prop that usually gets an ID? Could that be a pattern?

stack72:
> Looks good! Is "Value From" a prop that usually gets an ID? Could that be a pattern?

I'm honestly not sure at this time... I think if I see another then I will know more

PR #5503, opened 2025-02-18
feat(clover): Do not generate specs for types without handlers

jkeiser:
Some AWS resources (such as Route53 RecordSet) cannot be successfully used with cloudformation. This is inferrable by what types of `handlers` they have, showing what sort of actions you can do with them.

This PR:

* Makes it so we don't generate specs for types without create, read, update, delete or list handlers.
* Processes specs in alphabetical order so we can see what is happening
* Keeps the cfSchema around in ExpandedPkgSpec so that it can be used (but removes it at the end of generation so it doesn't bulk up the PkgSpec

## For later:

* Some types have some handlers but not others. This only handles types where we don't want to generate the asset *at all*. We will handle selectively turning off various actions / management functions based on the handlers in a separate PR.
* During this, we discovered that some sub-assets have conflicting names and were overwriting each other. This PR resolves two of those (Channel Destinations and Instance BlockDeviceMappings) but does not fix a third (TargetGroup Targets). That will have to be fixed in a separate PR.

## Testing

* Checked that the list is roughly the right number of resources not being generated (298 = the 231 top-level resources that *should* be unsupported, plus presumably 67 sub-assets)
* Spot checked that the list of unhandled assets included RecordSet and generally seemed reasonable.
* Checked that the specs we are *still* generating didn't change in any way except input sockets (we don't generate input sockets anymore if the assets they connected to no longer exist).

### DIFF
--- a/bin/clover/src/cli.ts
+++ b/bin/clover/src/cli.ts
@@ -44,6 +44,10 @@ To generate all specs containing "ECS" or "S3", you can pass some services as ar
     .option("--inferred <file:string>", "Inferred database location", {
       default: "inferred.json",
     })
+    .option(
+      "--force-update-existing-packages",
+      "Force the existing package list to be updated",
+    )
     .env(
       "SI_BEARER_TOKEN=<value:string>",
       "Auth token for interacting with the module index",
--- a/bin/clover/src/commands/generateSiSpecs.ts
+++ b/bin/clover/src/commands/generateSiSpecs.ts
@@ -17,11 +17,19 @@ import _logger from "../logger.ts";
 import { assetSpecificOverrides } from "../pipeline-steps/assetSpecificOverrides.ts";
 import { addSignatureToCategoryName } from "../pipeline-steps/addSignatureToCategoryName.ts";
 import { generateOutputSocketsFromProps } from "../pipeline-steps/generateOutputSocketsFromProps.ts";
-import { ExpandedPkgSpec } from "../spec/pkgs.ts";
+import {
+  ExpandedPkgSpec,
+  ExpandedSchemaSpec,
+  ExpandedSchemaVariantSpec,
+} from "../spec/pkgs.ts";
 import { createPolicyDocumentInputSockets } from "../pipeline-steps/createPolicyDocumentInputSockets.ts";
 import { prettifySocketNames } from "../pipeline-steps/prettifySocketNames.ts";
 import { loadInferred } from "../spec/inferred.ts";
 import { addInferredEnums } from "../pipeline-steps/addInferredEnums.ts";
+import { PkgSpec } from "../bindings/PkgSpec.ts";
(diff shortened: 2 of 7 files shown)


bin/clover/src/cli.ts
```diff
@@ -63,6 +68,33 @@ To generate all specs containing "ECS" or "S3", you can pass some services as ar
...
     })
+    // load-specs
+    .command(
+      "load-specs [...services:string]",
+      `Load the specs`,
+    )
+    .option("--inferred <file:string>", "Inferred database location", {
+      default: "inferred.json",
+    })
+    .env(
```
sprutton1, line 79: If we are keeping this, I don't think we need the env vars, right?

bin/clover/src/cli.ts
```diff
@@ -63,6 +68,33 @@ To generate all specs containing "ECS" or "S3", you can pass some services as ar
...
+      "SI_MODULE_INDEX_URL=<value:string>",
+      "Set the module index url; defaults to http://0.0.0.0:5157",
+      { prefix: "SI_" },
+    )
+    .action(async (options, ...services: string[]) => {
+      await loadCfDatabase(
+        {
+          ...options,
+          // moduleIndexUrl: options.moduleIndexUrl ?? DEFAULT_MODULE_INDEX_URL,
```
sprutton1, line 93: Or this

bin/clover/src/commands/generateSiSpecs.ts
```diff
@@ -59,6 +68,7 @@ export async function generateSiSpecs(
   }
 
   // EXECUTE PIPELINE STEPS
+  specs = ignoreSpecsWithoutHandlers(specs);
```
sprutton1, line 71: Did we checkin the file with this in it or am I blind?

sprutton1 (changes requested):

bin/clover/src/cli.ts
jkeiser: Oh! I'll remove this.

bin/clover/src/commands/generateSiSpecs.ts
jkeiser: We didn't :) I just added it.

bin/clover/src/pipeline-steps/ignoreSpecsWithoutHandlers.ts
```diff
@@ -0,0 +1,18 @@
+import { ExpandedPkgSpec } from "../spec/pkgs.ts";
+import _logger from "../logger.ts";
+
+const logger = _logger.ns("assetOverrides").seal();
```
sprutton1, line 4:
```suggestion
const logger = _logger.ns("ignoreSpecs").seal();
```

sprutton1 (approved):
One small nit, but we can merge without it and pick it up in the next one

PR #5506, opened 2025-02-19
feat(auth-stack): Prevent XSS based attacks on text inputs 

stack72:
This also prevents by passing the UI and going straight to the API

Screenshot 2025-02-19 at 14 39 35

Screenshot 2025-02-19 at 14 51 40

### DIFF
--- a/app/auth-portal/src/lib/validations.ts
+++ b/app/auth-portal/src/lib/validations.ts
@@ -1,0 +1,1 @@
+export const ALLOWED_INPUT_REGEX = /^[a-zA-Z0-9-,_@/+ ]+$/;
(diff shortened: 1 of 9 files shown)


stack72:
Just to close the loop here - hubspot already sanitize the values we send to them!

https://developers.hubspot.com/changelog/xss-sanitization-for-crm-properties-containing-html

johnrwatson:
> @stack72 did we also update the template for email on the other side, so that even if bad input sneaks through, we don't let you put HTML directly in?

I'm not sure this is technically possible with the transactional email templates we're using, I could be wrong though (conditionals/if's etc aren't permitted)

PR #5507, opened 2025-02-19
fix: make clover tests run in CI

sprutton1:
We needed to include the entirety of the source in order for the tests to be bubbled up when we figure out what changed. Also, we exclude the cc func code so we don't have to bring the types into them.


bin/clover/src/pipeline-steps/prettifySocketNames.ts
```diff
@@ -23,7 +23,9 @@ export function prettifySocketNames(
     bfsPropTree([variant.domain, variant.resourceValue], (prop) => {
       if (prop.data.inputs) {
         for (const input of prop.data.inputs) {
-          input.socket_name = toSpaceCase(input.socket_name);
+          if (input.kind !== "prop") {
```
sprutton1, line 26: tests complain when compiling

sprutton1:
Please don't merge until we see a successful `try`

PR #5508, opened 2025-02-19
fix(clover): fix doc links for nested props

jkeiser:
This fixes the doc links we create for props that are part of nested definitions. In Cloud Formation, nested object definitions are placed in their own pages.

We had code trying to make this work, but it relied on `$ref`, which is stripped from most props during dereferencing. This code populates a "defName" into every prop with the name of its containing definition before the dereferencing happens.

## Testing

* Diffed output, verified that only docLink changed
* Spot checked modified docLink, verified that old link failed and new link succeeded

### DIFF
--- a/bin/clover/src/spec/props.ts
+++ b/bin/clover/src/spec/props.ts
@@ -1,4 +1,4 @@
-import { CfProperty, normalizeProperty } from "../cfDb.ts";
+import { CfObjectProperty, CfProperty, normalizeProperty } from "../cfDb.ts";
 import { ulid } from "https://deno.land/x/ulid@v0.3.0/mod.ts";
 import { PropSpec } from "../bindings/PropSpec.ts";
 import { PropSpecData } from "../bindings/PropSpecData.ts";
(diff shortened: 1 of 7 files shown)


bin/clover/src/spec/props.ts
```diff
@@ -576,6 +576,7 @@ export type DefaultPropType = "domain" | "secrets" | "resource_value";
 export function createObjectProp(
   name: string,
   parentPath: string[],
+  cfProp: CfObjectProperty | undefined,
```
sprutton1, line 579: What's the advantage of declaring cfProp this way instead of `cfProp?: CfObjectProperty`?

jkeiser: Oh, I did it this way so that typescript would yell at me anywhere I tried to call it without a createObjectProp()--it forced me to make a decision in each place we create an object prop. It's not as necessary now that that initial turn is done, but doesn't hurt to show the explicit decision in the code IMO.

jkeiser: i.e. if you don't use `?`, then you can pass `undefined` but you can't omit the parameter.

vbustamante (dismissed):

sprutton1 (approved):

vbustamante (approved):

PR #5509, opened 2025-02-19
fix: lang-js nodule_modules no longer needed

sprutton1:
setting `node_modules_dir=auto` allowed deno to ensure the `node_modules` at the root of the repo had all the stuff it needs in it. However, it also needed to parse every package in the existing `node_modules`. One of the packages contained an intentionally malformed `package.json` for test purposes that deno would throw up over. No amount of `exclude` incantations would allow deno to ignore this file. So, instead, we ignore the local `node_modules` and run `deno cache` before we attempt to compile to ensure the packages are available in the cache, but without needing to walk the node_modules in the root. Oh, the joys of the javascript package ecosystem.


jobelenus (approved):
Tested on my macOS, app came up as expected w/o having to delete the node_modules

nickgerace (approved):
This is awesome. Looked painful to find... big hugs. Excellent work!!

PR #5519, opened 2025-02-20
feat(clover): Change how JSON Numbers are represented in assets

stack72:
We are going to take JSON Numbers as Float. This will allow us to support their precision. We need to ensure we add the setFloat intrinsic for this

We have the new intrinsics for resourcePayloadToValue and NormalizeToArray so we need to use those in the pkgSpecs

We are also changing to use the AWS Colour pallet - #FF9900


vbustamante (approved):
Diddn't test locally, but it all seems right

stack72:
> Diddn't test locally, but it all seems right

I can send you the spec file it generated! :)

PR #5523, opened 2025-02-20
Add "migration-mode" to remove workspace snapshots unused by any active change sets

jhelwig:
While we do evict snapshots when a change set pointer moves, we don't do anything when a change set's status is no longer one of the "active" statuses. We also do not currently allow setting the status of a change set back to one of the "active" statuses once it has been moved out of being "active".

Because a change set is forever "inactive" once it has become inactive, it is not necessary to keep the snapshot around, unless it is also being used by an "active" change set.

This new migration-mode (`garbageCollectSnapshots`) queries all "active" change sets to gather their workspace snapshot addresses, and all of the workspace snapshot addresses in the LayerDb. Any workspace snapshot that is at least an hour old, and is _NOT_ referenced by an "active" change set is removed. We only consider snapshots older than an hour to avoid race conditions where a change set might have been created or modified after we queried the `change_set_pointers` table, and before when we query the `workspace_snapshots` table. Because the tables are in separate databases, we can't rely on normal transactional integrity.

### DIFF
--- a/lib/dal/src/lib.rs
+++ b/lib/dal/src/lib.rs
@@ -256,6 +256,7 @@ pub fn generate_name() -> String {
 )]
 #[strum(serialize_all = "camelCase")]
 pub enum MigrationMode {
+    GarbageCollectSnapshots,
     Run,
     RunAndQuit,
     Skip,
@@ -280,6 +281,10 @@ impl MigrationMode {
     pub fn is_run_and_quit(&self) -> bool {
         matches!(self, Self::RunAndQuit)
     }
+
+    pub fn is_garbage_collect_snapshots(&self) -> bool {
+        matches!(self, Self::GarbageCollectSnapshots)
+    }
 }
 
 #[cfg(test)]
@@ -293,6 +298,10 @@ mod tests {
 
         #[test]
         fn display() {
+            assert_eq!(
+                "garbageCollectSnapshots",
+                MigrationMode::GarbageCollectSnapshots.to_string()
+            );
             assert_eq!("run", MigrationMode::Run.to_string());
             assert_eq!("runAndQuit", MigrationMode::RunAndQuit.to_string());
             assert_eq!("skip", MigrationMode::Skip.to_string());
@@ -300,6 +309,10 @@ mod tests {
 
         #[test]
         fn from_str() {
+            assert_eq!(
+                MigrationMode::GarbageCollectSnapshots,
+                "garbageCollectSnapshots".parse().expect("failed to parse")
+            );
             assert_eq!(MigrationMode::Run, "run".parse().expect("failed to parse"));
             assert_eq!(
                 MigrationMode::RunAndQuit,
--- a/lib/sdf-server/src/garbage_collection.rs
+++ b/lib/sdf-server/src/garbage_collection.rs
@@ -1,0 +1,159 @@
+use std::{collections::HashSet, future::IntoFuture as _};
+
+use dal::{ChangeSetStatus, ServicesContext, TransactionsError, WorkspaceSnapshotAddress};
+use si_data_pg::{PgError, PgPoolError};
+use si_layer_cache::LayerDbError;
+use strum::IntoEnumIterator;
+use telemetry::prelude::*;
+use thiserror::Error;
+use tokio_util::{sync::CancellationToken, task::TaskTracker};
+
+use crate::{init, Config};
+
+#[remain::sorted]
+#[derive(Debug, Error)]
+pub enum GarbageCollectorError {
+    #[error("error while initializing: {0}")]
+    Init(#[from] init::InitError),
+    #[error("Layer DB error: {0}")]
+    LayerDb(#[from] LayerDbError),
+    #[error("Pg error: {0}")]
+    Pg(#[from] PgError),
+    #[error("Pg Pool error: {0}")]
+    PgPool(#[from] PgPoolError),
+    #[error("Transactions error: {0}")]
+    Transactions(#[from] TransactionsError),
+    #[error("Unable to query workspace snapshots")]
+    UnableToQuerySnapshots,
+}
+
+type Result<T> = std::result::Result<T, GarbageCollectorError>;
+
+pub struct SnapshotGarbageCollector {
+    services_context: ServicesContext,
+}
+
+impl SnapshotGarbageCollector {
+    #[instrument(name = "sdf.snapshot_garbage_collector.new", level = "info", skip_all)]
+    pub async fn new(
+        config: Config,
+        task_tracker: &TaskTracker,
+        task_token: CancellationToken,
+    ) -> Result<Self> {
+        let (services_context, layer_db_graceful_shutdown) =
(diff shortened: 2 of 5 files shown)


lib/sdf-server/src/garbage_collection.rs
```diff
@@ -0,0 +1,154 @@
...
+            .map_err(|err| span.record_err(err))?;
+        let ctx = &ctx;
+
+        let mut open_change_set_snapshot_ids = HashSet::new();
+        let mut all_snapshot_ids = HashSet::new();
+
+        // Gather the WorkspaceSnapshotAddress of all open change sets.
+        let open_statuses: Vec<String> = ChangeSetStatus::iter()
+            .filter_map(|status| {
+                if status.is_active() {
```
johnrwatson, line 72: Is it possible that this isn't safe if there is active work going on in the system? i.e. would we have to ensure that there are no changesets being created/applied/deleted while this mode is being executed?

jhelwig: The danger would be:
* `Snapshot X` created more than an hour ago.
* `Snapshot X` is not used by any "active" change sets at the time we query the `change_set_pointers` table.
* We query `change_set_pointers` to gather the snapshot addresses.
* `Change Set B` is created/modified after we queried, is "active", and is now referencing `Snapshot X`.
* The query of `workspace_snapshots` to gather the snapshot addresses could happen either before, or after `Change Set B` is created/modified to reference `Snapshot X`.

I think it's technically possible, but I think the possibility of it happening without specifically trying is pretty small.

lib/sdf-server/src/garbage_collection.rs
```diff
@@ -0,0 +1,156 @@
...
+        // tables are in completely separate databases.
+        let snapshot_id_rows = ctx
+            .layer_db()
+            .workspace_snapshot()
+            .cache
+            .pg()
+            .query(
+                "SELECT key AS snapshot_id FROM workspace_snapshots WHERE created_at < NOW() - '1 hour'::interval GROUP BY key",
+                &[],
+            )
```
fnichol, line 112: Confirmed that this is running a SQL query on a pg client (out of the pool) in a non-transaction query. No other side effects!

lib/sdf-server/src/garbage_collection.rs
```diff
@@ -0,0 +1,156 @@
...
+        );
+
+        let mut counter = 0;
+        for key in snapshot_ids_to_delete {
+            ctx.layer_db()
+                .workspace_snapshot()
+                .cache
+                .pg()
+                .delete(&key.to_string())
+                .await?;
```
fnichol, line 142: The delete here is also a transaction-less delete query to the dataase

lib/sdf-server/src/garbage_collection.rs
```diff
@@ -0,0 +1,156 @@
...
+
+            counter += 1;
+            if counter % 100 == 0 {
+                info!("Deleted {} snapshot addresses.", counter);
+            }
+        }
+        info!("Deleted {} snapshot address(es).", counter);
+
+        ctx.commit().await?;
```
fnichol, line 151: This shouldn't have any interaction with the layer cache's database, only for the DAL db and any pending WS events.

lib/dal/src/lib.rs
```diff
@@ -256,6 +256,7 @@ pub fn generate_name() -> String {
 )]
 #[strum(serialize_all = "camelCase")]
 pub enum MigrationMode {
+    GarbageCollectSnapshots,
```
fnichol, line 259: I like this!

fnichol (dismissed):
Looking good!

lib/sdf-server/src/garbage_collection.rs
```diff
@@ -0,0 +1,156 @@
...
+        let snapshot_ids_to_delete: HashSet<_> = all_snapshot_ids
+            .difference(&open_change_set_snapshot_ids)
+            .collect();
+        info!(
+            "Found {} snapshot address(es) only used by inactive change sets.",
+            snapshot_ids_to_delete.len()
+        );
+
+        let mut counter = 0;
+        for key in snapshot_ids_to_delete {
```
johnrwatson, line 136: Can we add a throttle here? So we can test with ~1000 records or similar before trying all 3TB+ at once
I.e. `sdf mode garbageCollectSnapshots 1000` or similar

johnrwatson: (Or maybe if it's easier set a `LIMIT 10000` or similar and we can just run it multiple times)

johnrwatson:
@jhelwig https://github.com/systeminit/si/pull/5526 <- Adds the relevant CI bits

lib/sdf-server/src/garbage_collection.rs
jhelwig: Setting a limit in the query against `workspace_snapshots` would definitely be easier. The way the migration mode flag works, I'm not sure how reasonable it would be to make it take two arguments if the first is `garbageCollectSnapshots`. (Ex: `--migration-mode garbageCollectSnapshots 1000`) If it were structured more like sub-commands (as in your example), then it would be a lot easier, but the mode is currently a single flag.

lib/sdf-server/src/garbage_collection.rs
```diff
@@ -0,0 +1,161 @@
...
+            .workspace_snapshot()
+            .cache
+            .pg()
+            .query(
+                "SELECT key AS snapshot_id
+                 FROM workspace_snapshots
+                 WHERE created_at < NOW() - '1 hour'::interval
+                 GROUP BY key
+                 ORDER BY created_at
+                 LIMIT 10000",
```
jhelwig, line 115: @johnrwatson & @fnichol: Added `LIMIT 10000` here from @johnrwatson's request in another comment.

lib/sdf-server/src/garbage_collection.rs
```diff
@@ -0,0 +1,159 @@
...
+        let snapshot_ids_to_delete: HashSet<_> = all_snapshot_ids
+            .difference(&open_change_set_snapshot_ids)
+            .collect();
+        info!(
+            "Found {} snapshot address(es) only used by inactive change sets.",
+            snapshot_ids_to_delete.len()
+        );
+
+        let mut counter = 0;
+        for key in snapshot_ids_to_delete.iter().take(10_000) {
```
jhelwig, line 139: Briefly had this as a `LIMIT 10000` in the SQL query, but realized that had the potential to limit the rows returned to only those referenced by active change sets while there might be candidates for deletion outside of the `LIMIT`. By bounding the number of items we process after having fully calculated the list, we should avoid the problem of potentially stalling out while there is work that could be done.

lib/sdf-server/src/garbage_collection.rs
johnrwatson: Yeah nice, this'll definitely do. cheers

johnrwatson (approved):

PR #5524, opened 2025-02-20
Make module sync rely on local cache only

nickgerace:
## Description

This PR makes the module sync route rely on the local cache only without requiring queries to the module index. It also removes the tab to install assets since dynamic installation is now in play. Since the tab is removed and there's only one tab for most users of SI, the tab group will be disabled and the pill counter will move to the title of the assets panel.

## Screenshots

With the flag on:

Screenshot 2025-02-24 at 1 17 27 PM

With the flag off:

Screenshot 2025-02-24 at 1 18 25 PM


lib/dal/src/cached_module.rs
```diff
@@ -187,23 +223,7 @@ impl CachedModule {
...
-            .map(Into::into)
-            .collect();
-
-        let current_modules = CachedModule::latest_modules(ctx).await?;
-        for cached in current_modules {
-            if !builtin_schema_ids.contains(&cached.schema_id) {
-                CachedModule::remove_by_schema_id(ctx, cached.schema_id).await?;
-            }
-        }
+        Self::remove_unused(ctx, &modules).await?;
```
britmyerss, line 226: I wonder if we should tombstone them now instead of removing them - because otherwise when a builtin is rejected, it'll seem like it's contributeable? I'm 50/50

nickgerace: I think this is a good idea, but probably for a follow-up. I think rejected builtins still would be fine to be contributable in the interim.

app/web/src/components/AssetListPanel.vue
```diff
@@ -10,6 +10,10 @@
         <template #label>
           <div class="flex flex-row gap-xs">
             <div>Assets Installed</div>
+            <PillCounter
+              v-if="!featureFlagsStore.MODULES_TAB"
+              :count="assetList.length"
+            />
```
nickgerace, line 16: The counter only appears if the tab group is disabled (see screenshot)

lib/dal/src/cached_module.rs
```diff
@@ -252,7 +272,37 @@ impl CachedModule {
         Ok(new_modules)
     }
 
-    pub async fn latest_by_schema_id(
+    async fn remove_unused(
+        ctx: &DalContext,
+        module_details_by_hash: &HashMap<String, ModuleDetailsResponse>,
+    ) -> CachedModuleResult<()> {
```
nickgerace, line 278: Collapsed into one function for clarity

lib/dal/src/module.rs
```diff
@@ -365,7 +370,7 @@ impl Module {
         Ok(all_schema_variants)
     }
 
-    pub async fn list_installed(ctx: &DalContext) -> ModuleResult<Vec<Self>> {
+    pub async fn list(ctx: &DalContext) -> ModuleResult<Vec<Self>> {
```
nickgerace, line 373: They are installed!

lib/dal/src/module.rs
```diff
@@ -412,148 +417,106 @@ impl Module {
...
-            latest_modules_count = latest_modules.len()
-        )
     )]
-    pub async fn sync(
-        ctx: &DalContext,
-        latest_modules: Vec<frontend_types::LatestModule>,
-        builtin_modules: Vec<frontend_types::ModuleDetails>,
-        all_modules: Vec<frontend_types::ModuleDetails>,
-    ) -> ModuleResult<frontend_types::SyncedModules> {
+    pub async fn sync(ctx: &DalContext) -> ModuleResult<frontend_types::SyncedModules> {
```
nickgerace, line 427: Core of this PR

lib/dal/tests/integration_test/module.rs
```diff
@@ -169,93 +167,6 @@ async fn module_export_simple(ctx: &mut DalContext) {
     assert_eq!(exported_pkg_func_names, expected_func_names);
 }
 
-#[test]
-async fn dummy_sync(ctx: &DalContext) {
```
nickgerace, line 173: This test was already a bit iffy in what it was testing, but now it especially makes less sense without creating a bytestream of "pkg data" and inserting a cached module row. We'll delete it for now since its coverage would be questionable without more test helper investment.

britmyerss (approved):
I tested this locally last week (before the UI changes), it looks great!!

britmyerss (dismissed):

stack72 (approved):

nickgerace:
We need https://github.com/systeminit/si/pull/5538 to be merged before this PR can be merged.

PR #5528, opened 2025-02-21
feat(sdf,web,dal): Allow users to autoconnect a component's input sockets to matching output sockets

britmyerss:
This PR is a spike to enable users to autoconnect a Component's input sockets to valid matching output sockets. 

This is behind a feature flag and a bit of a brute force implementation, but hey, it's a spike! For the selected Component, if it has values that could have been driven via a socket that are overridden (have manual entry, or more likely, set via import/discovery or a management function), we look across all components to see if any of them have a output socket with the same value as the overridden prop.  

There's a bit of sauce in here to try and support multi-arity sockets and connecting to many different output sockets, but I expect there to be edge cases.  If the value is an array, and we find output sockets that could provide each of the entries in the array, we connect them. 

There's reactivity here, but in the case where there aren't any connections made, there's nothing to tell the user this at the moment. Did I mention it's a spike?

I did write a very basic integration test for it but we should invest more here if we find this valuable.

<div><br/>via <a href="https://giphy.com/gifs/Filmeditor--halloween-ashley-olsen-l2YWs1NexTst9YmFG">GIPHY</a></div>


app/web/src/store/components.store.ts
```diff
@@ -176,6 +176,7 @@ type EventBusEvents = {
   };
   setSelection: ComponentId[];
   rename: ComponentId;
+  autoConnect: ComponentId;
```
britmyerss, line 179: will remove this

stack72 (approved):

PR #5530, opened 2025-02-21
feat: amend cors configuration to auth-api

johnrwatson:
Tighten up our CORS configuration for auth-api so that only permitted origins will be respected and only send credentials (if required) within these respected domain boundaries.

<hr/>

**_Author Notes:_** I might need a hand testing this as ideally the auth-api would still be usable locally. I might need to push some variable through for local dev to allow the CORS config to be dynamic. Theoretically this change should be a no-op and only tighten the existing configuration.


johnrwatson:
Do not merge without author present, please. Bit of a sketchy one that's hard to test without deploying

PR #5532, opened 2025-02-21
feat(clover): Colons on category names and deprecation reporting

vbustamante:
Adds "::" to category names and adds a warning log on the clover generator when it notices that the module index has specs that are not generated anymore, meaning they've probably been removed by AWS

### DIFF
--- a/bin/clover/src/specPipeline.ts
+++ b/bin/clover/src/specPipeline.ts
@@ -12,9 +12,9 @@ import {
 } from "./spec/pkgs.ts";
 
 export function pkgSpecFromCf(cfSchema: CfSchema): ExpandedPkgSpec {
-  const [aws, category, name] = cfSchema.typeName.split("::");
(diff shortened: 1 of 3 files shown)


stack72 (approved):
NICE!!

bin/clover/src/specPipeline.ts
```diff
@@ -12,9 +12,9 @@ import {
 } from "./spec/pkgs.ts";
 
 export function pkgSpecFromCf(cfSchema: CfSchema): ExpandedPkgSpec {
-  const [aws, category, name] = cfSchema.typeName.split("::");
+  const [metaCategory, category, name] = cfSchema.typeName.split("::");
```
jkeiser, line 15: I like the carefulness! Just in case we get some stuff that's got more colons, I'd add a `...more` here, and then check that more.length == 0 below.

jkeiser (approved):

PR #5535, opened 2025-02-21
feat: allow for pruning modules based on a list generated by clover

sprutton1:
Clover can now generate a list that of modules that still exist in the module index, but we did not generate a spec for. This allows hoist to take that list and reject those modules with VERY FANCY features like skipping and allowing all.

image


stack72:
@sprutton1 is this still needed?

PR #5536, opened 2025-02-21
chore(dal): simplify tests that get schema/variants by name

jkeiser:
Separating this PR for a few ergonomics changes for tests:

* Make Schema::find_by_name() error if schema is not found since all callers return an error themselves. This eliminates an unnecessary .expect() (Also rename to get_by_name())
* Add SchemaVariant::default_by_schema_name() since that is a primary use case in tests (we almost never need the schema itself). Did not change all tests to use, just regenerate.rs for now since it's not a purely search-and-replace)
* Use color_eyre errors in regenerate.rs to ergonomicize the tests (so we don't expect() all the time). Not doing this globally across all tests as it changes too many lines at once and isn't *quite* seach-and-replaceable

regenerate.rs tests are reduced by 15-20% with no changes; have very few expect()s breaking the flow; and many statements reduce to a single line now that the expect is gone.

## Risk Mitigation

* No behavior change.
* Only one change outside of tests: si-fs no longer has its own bespoke error when the schema is not found by name. The error message still lists the name of the schema not found.
* All tests still pass.

### DIFF
--- a/lib/dal-test/src/expected.rs
+++ b/lib/dal-test/src/expected.rs
@@ -111,10 +111,9 @@ impl SchemaKey for Schema {
 }
 impl SchemaKey for str {
     async fn lookup_schema(&self, ctx: &DalContext) -> SchemaId {
-        Schema::find_by_name(ctx, self)
+        Schema::get_by_name(ctx, self)
             .await
             .expect("find schema by name")
-            .expect("schema exists")
             .id()
     }
 }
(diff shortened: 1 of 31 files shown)


zacharyhamm (dismissed):
I see how this change to find by name simplifies things. Surprised my use was the only one outside of tests... Looks good

zacharyhamm (approved):

PR #5538, opened 2025-02-24
fix: ensure deno_cache rules are run under buck2 build

sprutton1:
The more I learn about buck2, the more I learn that I don't much about buck2.

In order for a target to actually do anything via `buck2 build`, it must have an output. If it doesn't have an output, `buck2 build` just... exits successfully? But does nothing? 

For example, this `deno_cache` rule I wrote works as long as you `buck2 run` it, but if you `buck2 build` it, it looks like it works, but doesn't actually cache anything. This is a problem if you update the `deno.lock` and try to run our dev stack locally as it will appear to work, but the cache rule won't actually do anything. Adding a fake output file to the job makes it so you can `buck2 build` it and it works as you would expect.


bin/lang-js/BUCK
```diff
@@ -47,13 +49,13 @@ deno_compile(
     name = "bin",
     main = "src/index.ts",
     out = "lang-js",
-    srcs = glob([
+    srcs = [
+        ":build",
+    ] +
+    glob([
```
nickgerace, line 55: Clever

nickgerace (dismissed):

nickgerace (approved):

PR #5540, opened 2025-02-24
fix(auth-portal): Ensure a user must be email verified before proceeding

stack72:
Before this commit, if you were an unverified email user you could bypass the check for email verification and create a new workspace. This prevents this.

We redirect an unverified user to the profile page if they try and hit workspaces, workspace details or billing

We also prevent these API calls in the UI by checking that the email is verified on specific routes we want to protect


stack72:
Testing carried out:

* Signup and don't verify email
** Can't access workspace/new, billingdetails or workspaces- redirected to profile path

* Signup with an OAUTH user and can create and patch workspaces as normal

britmyerss (approved):

PR #5546, opened 2025-02-24
fix(clover): ensure that we send an object out of CloudFormation Resource Socket

stack72:
We don't want a string - we want the object

Also removes StackSet to go with Stack - neither work with our templates and we will remove them for now


adamhjk (dismissed):
Yep, we need this :)

sprutton1 (approved):

PR #5551, opened 2025-02-25
feat(sdf): Preserve prop order when uploading packages to the module index

jkeiser:
Right now, packages contributed or uploaded to the module index have their props re-sorted alphabetically. This PR modifies SiPkg and SiPkgNodeit so that going from PkgSpec -> SiPkg -> bytes -> SiPkg -> PackageNode -> dal preserves ordering.

## Testing

* Tested that round-tripping from PkgSpec -> SiPkg -> bytes -> SiPkg -> PksSpec preserves order (it didn't before)
* Added integration test with round tripping that checks the underlying dal functions for:
  - Regenerate (asset func -> dal variant/component)
  - Install Module (bytes -> SiPkg -> dal variant/component)
  - Contribute (dal -> SiPkg -> bytes)
  - Hoist Upload (PkgSpec -> SiPkg)
  - SiPkg -> PkgSpec (prop order not used anywhere except tests at the moment)
* Manual tests:
  * Test that existing uploaded modules still load and are gracefully loaded alphabetically
  * Test that Clover assets uploaded to the module index via hoist preserve prop order after this change.
  * Test that Contribute flow preserves prop order.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6922,6 +6922,7 @@ dependencies = [
  "chrono",
  "derive_builder",
  "indexmap 2.7.1",
+ "itertools 0.13.0",
  "object-tree",
  "petgraph",
  "remain",
--- a/lib/dal-test/src/expected.rs
+++ b/lib/dal-test/src/expected.rs
@@ -334,6 +334,7 @@ impl ExpectSchemaVariant {
         .id
         .into();
         variant.update_asset_func(ctx, asset_func).await;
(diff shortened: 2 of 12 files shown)


adamhjk:
oh my god, you are my fucking hero.

zacharyhamm (approved):
This seems solid! As soon as it hits main, the module index will have to be redeployed so that it is aware of the SiPkg format changes. No new modules will be able to be contributed until it is re-deployed.

PR #5555, opened 2025-02-25
fix(clover): Adding Prop overrides for KMS Key, Logs LogGroup and DBParameterGroup

stack72:


bin/clover/src/pipeline-steps/assetSpecificOverrides.ts
```diff
@@ -192,6 +192,51 @@ const overrides = new Map<string, OverrideFn>([
...
     },
   ],
+  [
+    "AWS::KMS::Key",
+    (spec: ExpandedPkgSpec) => {
+      const variant = spec.schemas[0].variants[0];
+
+      const prop = variant.domain.entries.find((p: ExpandedPropSpec) =>
+        p.name === "KeyPolicy"
+      );
```
sprutton1, line 202: Maybe we should add this as function on ExpandedPropSpec

bin/clover/src/pipeline-steps/assetSpecificOverrides.ts
```diff
@@ -192,6 +192,51 @@ const overrides = new Map<string, OverrideFn>([
...
+      
+      if (!prop) return;
+      prop.kind = "json"
+      prop!.data.widgetKind = "CodeEditor";
+    },
+  ],
+  [
+    "AWS::RDS::DBParameterGroup",
```
sprutton1, line 210: nit: we should alphabetize these

bin/clover/src/pipeline-steps/assetSpecificOverrides.ts
sprutton1: In general

sprutton1 (approved):

PR #5558, opened 2025-02-26
feat(module-index-client): Catch 404 errors on get requests in module index client

stack72:


fnichol (approved):
Nicely done and not even as much work as I was expecting. I forgot about the `error_for_status` method and how that worked which really helps to deal with all remaining errors.

nickgerace (approved):

zacharyhamm (approved):

PR #5566, opened 2025-02-27
Veritech testing ground

nickgerace:
## Description

This PR has become a testing ground for preventing veritech failures to process existing work and take on new work.

### DIFF
--- a/bin/veritech/src/args.rs
+++ b/bin/veritech/src/args.rs
@@ -105,6 +105,27 @@ pub(crate) struct Args {
     /// back to an instance of a Pinga service.
     #[arg(long)]
     pub(crate) instance_id: Option<String>,
+
+    /// Excludes the "pause-resume" stream wrapper around the core stream used for receiving function
+    /// execution requests.
+    #[arg(long, default_value = "false")]
+    pub(crate) exclude_pause_resume_stream_wrapper: bool,
+
+    /// The duration that a naxum stream will be paused on command.
+    ///
+    /// Only used if the "pause-resume" stream is enabled.
+    #[arg(long)]
+    pub(crate) pause_duration_ms: Option<u64>,
+
+    /// The duration for backoff subscribing to a naxum stream with pause-resume semantics.
+    ///
+    /// Only used if the "pause-resume" stream is enabled.
+    #[arg(long)]
+    pub(crate) reconnect_backoff_duration_ms: Option<u64>,
+
+    /// The graceful shutdown timeout in seconds.
+    #[arg(long)]
+    pub(crate) graceful_shutdown_timeout_secs: Option<u64>,
 }
 
 impl TryFrom<Args> for Config {
@@ -151,6 +172,25 @@ impl TryFrom<Args> for Config {
             }
             if let Some(instance_id) = args.instance_id {
                 config_map.set("instance_id", instance_id);
+            }
+            config_map.set(
+                "exclude_pause_resume_stream_wrapper",
+                args.exclude_pause_resume_stream_wrapper,
+            );
+            if let Some(pause_duration_ms) = args.pause_duration_ms {
+                config_map.set("pause_duration_ms", pause_duration_ms);
+            }
+            if let Some(reconnect_backoff_duration_ms) = args.reconnect_backoff_duration_ms {
+                config_map.set(
+                    "reconnect_backoff_duration_ms",
+                    reconnect_backoff_duration_ms,
+                );
+            }
+            if let Some(graceful_shutdown_timeout_secs) = args.graceful_shutdown_timeout_secs {
+                config_map.set(
+                    "graceful_shutdown_timeout_secs",
+                    graceful_shutdown_timeout_secs,
+                );
             }
         })?
         .try_into()
--- a/lib/naxum/src/middleware/ack/maintain_progress.rs
+++ b/lib/naxum/src/middleware/ack/maintain_progress.rs
@@ -3,7 +3,7 @@ use std::{sync::Arc, time::Duration};
 use async_nats::jetstream::{self, message::Acker};
 use tokio::time::{self, Instant, Interval};
 use tokio_util::sync::CancellationToken;
-use tracing::{debug, info, trace, warn};
+use tracing::{trace, warn};
 
 pub struct MaintainProgressTask {
     acker: Arc<Acker>,
@@ -27,8 +27,11 @@ impl MaintainProgressTask {
     }
 
     pub async fn run(mut self) {
-        trace!(si.naxum.task = Self::NAME, "running task");
-        debug!(si.naxum.task = Self::NAME, "first ack message");
+        trace!(
+            si.naxum.task = Self::NAME,
+            "running task (first ack message)"
+        );
+
         if let Err(err) = self.acker.ack_with(jetstream::AckKind::Progress).await {
             warn!(si.error.message = ?err, "failed initial ack");
         }
@@ -36,11 +39,11 @@ impl MaintainProgressTask {
         loop {
             tokio::select! {
                 _ = self.shutdown_token.cancelled() => {
-                    info!(si.naxum.task = Self::NAME, "received cancellation");
+                    trace!(si.naxum.task = Self::NAME, "received cancellation for maintain progress task");
                     break;
                 }
                 _ = self.interval.tick() => {
-                    debug!(task = Self::NAME, "acking message with progress");
+                    trace!(task = Self::NAME, "acking message with progress");
                     if let Err(err) = self.acker.ack_with(jetstream::AckKind::Progress).await {
                         warn!(si.error.message = ?err, "failed to ack with progress");
                     }
@@ -48,6 +51,9 @@ impl MaintainProgressTask {
             }
         }
 
-        info!(si.naxum.task = Self::NAME, "naxum shutdown complete");
+        trace!(
+            si.naxum.task = Self::NAME,
+            "naxum maintain progress shutdown complete"
+        );
     }
 }
--- a/lib/naxum/src/serve.rs
+++ b/lib/naxum/src/serve.rs
@@ -163,6 +163,8 @@ where
             tokio::pin!(stream);
 
             metric!(counter.naxum.next_message.processing = 0);
+            metric!(counter.naxum.next_message.failed = 0);
+            metric!(counter.naxum.next_message.none = 0);
 
             loop {
                 let (msg, permit) = tokio::select! {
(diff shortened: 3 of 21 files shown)


lib/naxum/src/middleware/ack/maintain_progress.rs
```diff
@@ -27,27 +27,33 @@ impl MaintainProgressTask {
...
                     }
                 }
             }
         }
 
-        info!(si.naxum.task = Self::NAME, "naxum shutdown complete");
+        trace!(
+            si.naxum.task = Self::NAME,
+            "naxum maintain progress shutdown complete"
+        );
```
nickgerace, line 57: These are now all `trace` or `warn` for ease of use. The `trace` ones are very noisy and don't belong in debug runs

lib/naxum/src/serve.rs
```diff
@@ -176,19 +173,16 @@ where
...
-                                metric!(counter.naxum.failed_count = 0);
-
                                 (msg, permit)
                             },
                             Some(Err(err)) => {
                                 // TODO(fnichol): this level might need to be `trace!()`, just
                                 // unclear at the moment
                                 warn!(si.error.message = ?err, "failed to read next message from stream");
-                                failed_count += 1;
+                                metric!(counter.naxum.failed_count = 1);
```
nickgerace, line 185: Move the failed count to a `metric!`

lib/naxum/src/serve.rs
```diff
@@ -274,18 +267,7 @@ where
...
-                    si.error.message = ?err,
-                    "failed to read message in after {} consecutive failures; closing stream",
-                    failed_count,
-                );
-                (None, permit)
-            } else {
-                (Some(Err(err)), permit)
-            }
-        }
+        Err(err) => (Some(Err(err)), permit),
```
nickgerace, line 270: Important: for both the new app and the old app (default), we no longer have the circuit-breaker

lib/veritech-server/src/config.rs
```diff
@@ -557,6 +604,26 @@ fn default_concurrency_limit() -> usize {
     DEFAULT_CONCURRENCY_LIMIT
 }
 
+fn default_enable_pause_resume_stream() -> bool {
+    false
+}
```
nickgerace, line 609: Disabled by default

lib/veritech-server/src/handlers.rs
```diff
@@ -88,6 +89,7 @@ pub async fn process_request(
     msg: Message<InnerMessage>,
 ) -> HandlerResult<()> {
     let span = Span::current();
+    metric!(counter.veritech.handlers_doing_work = 1);
```
nickgerace, line 92: Now track "handlers doing work". Worded this carefully because "in progress" does not mean "doing work" as we understand it. Let's say they are failing to wrap up somewhere deep in naxum or at some middleware layer... not "doing work" but still in progress, ya know?

lib/veritech-server/src/handlers.rs
```diff
@@ -155,6 +160,9 @@ pub async fn process_request(
         }
     }
 
+    info!(%execution_kind, %execution_id, "finished execution");
+    metric!(counter.veritech.handlers_doing_work = -1);
```
nickgerace, line 164: Now we log when it's done.

lib/veritech-server/src/server/pause_resume_stream.rs
```diff
@@ -0,0 +1,248 @@
...
+    }
+}
+
+impl futures::Stream for PauseResumeStream {
+    type Item = Result<Message, consumer::pull::MessagesError>;
+
+    fn poll_next(
+        self: std::pin::Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Option<Self::Item>> {
```
nickgerace, line 104: This is where the fun begins.

lib/veritech-server/src/server/pause_resume_stream.rs
```diff
@@ -0,0 +1,248 @@
...
+
+#[remain::sorted]
+enum State {
+    ReadyToSubscribe,
+    Subscribed(consumer::pull::Stream),
+    Subscribing(BoxFuture<'static, Result<consumer::pull::Stream, StreamError>>),
+    Unsubscribed(BoxFuture<'static, ()>),
+}
+
+pin_project! {
```
nickgerace, line 59: Used to mutate the fields. You can "project" the fields and then mutate them accordingly.

lib/naxum/src/serve.rs
nickgerace: @fnichol would love your thoughts on this particular. This would affect all naxum apps, but I think we both agree this isn't the place for this kind of breaker unless we change it up a bit

lib/veritech-server/src/server.rs
```diff
@@ -213,6 +243,73 @@ impl Server {
...
+                        telemetry_nats::NatsMakeSpan::builder(connection_metadata).build(),
+                    )
+                    .on_response(telemetry_nats::NatsOnResponse::new()),
+            )
+            .layer(AckLayer::new())
+            .service(handlers::process_request.with_state(state))
+            .map_response(Response::into_response);
+
+        // Reset metrics before creating the naxum server.
+        metric!(counter.veritech.pool_exhausted = 0);
```
nickgerace, line 266: Ah I moved this above and missed a spot to delete it here. Will do later. It's harmless but shouldn't be here since it's now above.

lib/naxum/src/serve.rs
britmyerss: This change isn't behind the startup flag though, yeah? so if we start everything up without the bool passed in, we could spin forever here without the pause?

lib/naxum/src/serve.rs
britmyerss: nit: maybe `metric!(counter.naxum.next_message.failed_count = `);`?

lib/naxum/src/serve.rs
```diff
@@ -176,19 +173,16 @@ where
                         tracker.close();
                         break;
                     }
-                    (msg, permit) = next_message(&mut stream, semaphore.clone(), failed_count) => {
+                    (msg, permit) = next_message(&mut stream, semaphore.clone()) => {
                         match msg {
                             Some(Ok(msg)) => {
-                                failed_count = 0;
-                                metric!(counter.naxum.failed_count = 0);
```
britmyerss, line 183: when should we reset the failed count?

lib/veritech-server/src/handlers.rs
britmyerss: ooooo can we include the execution_id as the label?? (does that make it easier to detect/be able to cancel a long running function???)

lib/veritech-server/src/handlers.rs
britmyerss: might want this to be `si.veritech.execution_kind` and `si.veritech.execution_id` so we can search/filter

britmyerss: (and on the above info event too)

lib/naxum/src/serve.rs
nickgerace: Right here a line below

lib/naxum/src/serve.rs
nickgerace: sure!

lib/veritech-server/src/handlers.rs
nickgerace: yes

lib/naxum/src/serve.rs
nickgerace: Yes you are correct

lib/veritech-server/src/config.rs
```diff
@@ -76,6 +82,15 @@ pub struct Config {
...
     instance_id: String,
+
+    #[builder(default = "default_enable_pause_resume_stream()")]
+    enable_pause_resume_stream: bool,
+
+    #[builder(default = "default_pause_duration()")]
+    pause_duration: Duration,
+
+    #[builder(default = "default_reconnect_backoff_duration()")]
+    reconnect_backoff_duration: Duration,
```
sprutton1, line 93: If this becomes a standard pattern, should this stuff get packed into the NatsConfig?

sprutton1: I guess it's server specific, so maybe not

nickgerace:
Moving to drafts since we are trying WIP commits on this.

bin/veritech/src/args.rs
```diff
@@ -105,6 +105,26 @@ pub(crate) struct Args {
...
+
+    /// The duration for backoff subscribing to a naxum stream with pause-resume semantics.
+    ///
+    /// Only used if the "pause-resume" stream is enabled.
+    #[arg(long)]
+    pub(crate) reconnect_backoff_duration_ms: Option<u64>,
+
+    /// The graceful shutdown timeout in seconds.
+    #[arg(long)]
+    pub(crate) graceful_shutdown_timeout_secs: Option<u64>,
```
nickgerace, line 127: Note: this is explicitly only in the args and not the server config

lib/si-data-nats/src/lib.rs
```diff
@@ -132,15 +142,15 @@ impl Client {
...
                 }
                 Event::ServerError(server_error) => {
                     metric!(counter.nats.event_callback.server_error = 1);
-                    error!(si.error.message = ?server_error, "nats event callback: server error")
+                    trace!(si.error.message = ?server_error, "nats event callback: server error")
                 }
                 Event::ClientError(client_error) => {
                     metric!(counter.nats.event_callback.client_error = 1);
-                    error!(si.error.message = ?client_error, "nats event callback: client error")
+                    trace!(si.error.message = ?client_error, "nats event callback: client error")
```
nickgerace, line 153: These will get LOUD. You don't want them as more than trace. Or even at all.

nickgerace: Doing in #5597

bin/veritech/src/args.rs
```diff
@@ -105,6 +105,27 @@ pub(crate) struct Args {
     /// back to an instance of a Pinga service.
     #[arg(long)]
     pub(crate) instance_id: Option<String>,
+
+    /// Excludes the "pause-resume" stream wrapper around the core stream used for receiving function
+    /// execution requests.
+    #[arg(long, default_value = "false")]
```
nickgerace, line 111: This is probably redundant but just to be sure (damn double negatives)

nickgerace:
There are some great ideas in this PR, but none are actionable at this time.

PR #5571, opened 2025-02-27
bug: lb target needs ec2 instance ids

adamhjk:
The ELBv2 Target needs to accept an instance ID

### DIFF
--- a/bin/clover/src/pipeline-steps/assetSpecificOverrides.ts
+++ b/bin/clover/src/pipeline-steps/assetSpecificOverrides.ts
@@ -241,6 +241,19 @@ const overrides = new Map<string, OverrideFn>([
     setAnnotationOnSocket(groupSocket, { tokens: ["GroupId"] });
     variant.sockets.push(groupSocket);
   }],
+  ["TargetGroup Targets", (spec: ExpandedPkgSpec) => {
+    const variant = spec.schemas[0].variants[0];
+
+    // Add an annotation for the Id output socket to connect to HostedZoneId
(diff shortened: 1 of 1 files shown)


stack72 (approved):

stack72:
we generated this and tested it works!

bin/clover/src/pipeline-steps/assetSpecificOverrides.ts
```diff
@@ -241,6 +241,17 @@ const overrides = new Map<string, OverrideFn>([
...
+  ["TargetGroup Targets", (spec: ExpandedPkgSpec) => {
+    const variant = spec.schemas[0].variants[0];
+
+    // Add an annotation for the Id output socket to connect to HostedZoneId
+    const socket = variant.sockets.find(
+      (s: ExpandedSocketSpec) => s.name === "Id" && s.data.kind === "input",
+    );
+    if (!socket) return;
+
+    setAnnotationOnSocket(socket, { tokens: ["InstanceId"] });
```
sprutton1, line 253: Should we add annotations for other target types?

stack72: Good shout @sprutton1 - we should add ARN as well - as it will allow [other] albs and lambda targets - @adamhjk can you update>

adamhjk: done!

stack72 (approved):

PR #5574, opened 2025-02-28
feat: add AWS::APS::Workspace log group input

johnrwatson:
<div></div>


johnrwatson:
Fixing conflicts, one sec

bin/clover/src/pipeline-steps/assetSpecificOverrides.ts
```diff
@@ -254,6 +254,23 @@ const overrides = new Map<string, OverrideFn>([
...
+      const variant = spec.schemas[0].variants[0];
+      
+      const prop = variant.domain.entries.find((p: ExpandedPropSpec) =>
+        p.name === "LogGroupArn"
+      );
+
+      if (!prop) return;
+      const socket = createInputSocketFromProp(prop);
+      
+      setAnnotationOnSocket(socket, { tokens: ["arn<string>"] });
```
sprutton1, line 269:
```suggestion
      setAnnotationOnSocket(socket, { tokens: ["arn", "string"] });
```

bin/clover/src/pipeline-steps/assetSpecificOverrides.ts
```diff
@@ -254,6 +254,23 @@ const overrides = new Map<string, OverrideFn>([
     setAnnotationOnSocket(socket, { tokens: ["arn<string>"] });
     setAnnotationOnSocket(socket, { tokens: ["arn"] });
   }],
+  [
```
sprutton1, line 257: I just realized we don't recurse into objects when making arn input sockets. This is fine for now, but we should fix that

sprutton1 (changes requested):

bin/clover/src/pipeline-steps/assetSpecificOverrides.ts
johnrwatson: Thanks! I've made the amendment, is this a general issue/mistake as I just copied the example above with a bit of copy pasta

johnrwatson: I wasn't sure if you intended I delete the one below it along with this, as it looks like a dupe to me now

sprutton1 (approved):

PR #5575, opened 2025-02-28
Functional (but not great) source map following, and some vue internal component data

jobelenus:


wendybujalski (approved):
this will definitely be better than the state of things now!

PR #5576, opened 2025-02-28
feat: parallel execution of validations

adamhjk:
This PR runs all validations in parallel, rather than in serial. The result is that initial validation runs feel pretty much instantaneous.

### DIFF
--- a/lib/dal/src/job/definition/compute_validation.rs
+++ b/lib/dal/src/job/definition/compute_validation.rs
@@ -92,6 +92,8 @@ impl JobConsumer for ComputeValidation {
         }
 
         let workspace_snapshot = ctx.workspace_snapshot()?;
+        let mut tasks = Vec::new();
+
         for &av_id in &self.attribute_values {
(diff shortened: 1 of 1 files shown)


lib/dal/src/job/definition/compute_validation.rs
```diff
@@ -107,23 +109,34 @@ impl JobConsumer for ComputeValidation {
                 continue;
             }
 
-            let value = AttributeValue::get_by_id(ctx, av_id)
-                .await?
-                .value(ctx)
+            let ctx_clone = ctx.clone();
```
jkeiser, line 112: Most places I've seen we just use the same variable name (makes the code read the same whether it's in the block or not.)
```suggestion
            let ctx = ctx.clone();
```

wendybujalski:
Do I assume correctly that this change means that we're sure all validations are independent of each other and can run concurrently? Could there possibly be a case where two validations would affect each other when running in parallel?

fnichol:
Whoopsy, we addressed this (finally) in #7450, closing!

PR #5578, opened 2025-02-28
feat(sdf): Keep external connections on paste

jkeiser:
When you paste multiple things, we already restore connections between the components you copied--but we don't copy connections to things *outside* the pasted subgraph. This PR preserves connections to non-copied components when you copy/paste.

### Changes

This PR:

* Copies incoming connections from components not being pasted.
* Copies "managed by" connection from components not being pasted.
* Now sends an event to let the frontend know when the pasted component has a parent--this makes inferred connections show up in the UI.
* Continues to "mirror" incoming and management connections from components being pasted
* Continues to "mirror" parentage for elements where both parent and child are pasted
* Centralizes the copy/paste logic into Component::batch_copy, so it can be tested in the dal; SDF now just sends the WsEvents

### Refactoring Budget

My PRs often try to improve something sort-of-related "while I'm in there." To hopefully move forward while keeping the code review process pleasant, I'm limiting the amount and scope of it. I spent the refactoring budget here on:

* Used Result<()> in component tests, replacing .expect() calls with `?`.
* Added From<Box<E>> for boxed error types in component.rs, and removed `.map_err(Box::new)` everywhere in that file.
* Replaced OutgoingConnections and IncomingConnections with Connections. It is a directionless struct and it clarifies code that wants to handle both outgoing and incoming connections in one place.

## Testing

* Added DAL tests (without WsEvents)
* Manually tested copying with external and internal connections
* Manually tested copying with both external and internal manager connections
* Manually tested copying parent, child, and both

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6299,6 +6299,7 @@ dependencies = [
  "futures-lite",
  "hyper 0.14.32",
  "indoc",
+ "itertools 0.13.0",
  "module-index-client",
  "names",
  "nats-multiplexer",
--- a/lib/dal-test/src/helpers.rs
+++ b/lib/dal-test/src/helpers.rs
@@ -327,6 +327,17 @@ pub async fn get_attribute_value_for_component(
     ctx: &DalContext,
     component_id: ComponentId,
     prop_path: &[&str],
+) -> Result<Value> {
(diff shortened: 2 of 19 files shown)


jkeiser:
Tests are added, and the main scenarios seem to be working!

britmyerss (approved):
great work! and great tests!!

PR #5580, opened 2025-02-28
fix(dal): Exclude root-si-resource_id from generated mgmt templates

stack72:


zacharyhamm (approved):

stack72:
This didn't quite work as expected - we will take another look

britmyerss (dismissed):

britmyerss (approved):

PR #5588, opened 2025-03-03
fix: strip default values from templates

sprutton1:
This ensures that props that contain their default value are not set in mgmt func templates, otherwise updates to the default value will not propagate as the prop will have a set value already.

### DIFF
--- a/lib/dal/src/management/generator.rs
+++ b/lib/dal/src/management/generator.rs
@@ -269,6 +269,13 @@ async fn calculate_paths_to_remove(
             continue;
         }
 
+        // remove the value if it matches the default. This ensures default value changes from prop
+        // updates are propagated
+        if Prop::default_value(ctx, prop_id).await? == Some(current_val.clone()) {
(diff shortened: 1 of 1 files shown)


lib/dal/src/management/generator.rs
```diff
@@ -269,6 +269,13 @@ async fn calculate_paths_to_remove(
             continue;
         }
 
+        // remove the value if it matches the default. This ensures default value changes from prop
+        // updates are propagated
+        if Prop::default_value(ctx, prop_id).await? == Some(current_val.clone()) {
```
zacharyhamm, line 274:
```suggestion
        if Prop::default_value(ctx, prop_id).await?.as_ref() == Some(current_val) {
```

zacharyhamm (commented):
This is a great change, and definitely an oversight in the template generator. I've made a little suggestion to avoid an unnecessary clone

zacharyhamm (approved):

PR #5591, opened 2025-03-03
fix: handle the case where object children do exist, but they are also empty so are not valid refinements

johnrwatson:
For EC2::LaunchTemplate (and some others we've probably not found yet), object children may exist as null and should be skipped during refinement:

```
Skipping import of lt-026952649cf076dbc; it did not match refinement {
  "LaunchTemplateData": {
    "SecurityGroups": []
  }
}
```

Currently, these aren't skipped as a refinement, as we are looking for any nested key or value regardless of value or content. 

This PR now handles the case where an object may have children, but they're also empty.

### DIFF
--- a/bin/clover/src/cloud-control-funcs/management/awsCloudControlDiscover.ts
+++ b/bin/clover/src/cloud-control-funcs/management/awsCloudControlDiscover.ts
@@ -22,6 +22,11 @@ async function main({
   for (const [key, value] of Object.entries(refinement)) {
     if (_.isEmpty(value)) {
       delete refinement[key];
+    } else if (_.isPlainObject(value)) {
+      refinement[key] = _.pickBy(value, (v) => !_.isEmpty(v) || _.isNumber(v) || _.isBoolean(v));
+      if (_.isEmpty(refinement[key])) {
+        delete refinement[key];
+      }
     }
   }
 


bin/clover/src/cloud-control-funcs/management/awsCloudControlDiscover.ts
```diff
@@ -20,8 +20,13 @@ async function main({
   delete refinement["extra"];
   // Remove any empty values, as they are never refinements
   for (const [key, value] of Object.entries(refinement)) {
-    if (_.isEmpty(value)) {
+    if (_.isEmpty(value) && !_.isNumber(value) && !_.isBoolean(value)) {
```
adamhjk, line 23: I don't think you need these?  It can't hurt to have them, but the docs seem to imply they'll never be empty

adamhjk (dismissed):
I left one comment, but it's advisory

bin/clover/src/cloud-control-funcs/management/awsCloudControlDiscover.ts
johnrwatson: I thought it was worth edging on the side of paranoia but happy to adjust.

johnrwatson:
> I left one comment, but it's advisory

Thanks @adamhjk - I'm going to test with a few assets manually that were previously working to check it a bit

johnrwatson:
Tested successfully on VPC / Subnet & NAT Gateway by manually adding the code change to the discovery funcs. All perform the same as before.

johnrwatson:
Rebased and retested VPC and Launch Template without the &&'s to fact check myself

johnrwatson:
@adamhjk if you could kindly re-review that'd be great

stack72 (approved):

PR #5592, opened 2025-03-03
fix(clover): Add specific override for AWS::EC2::VPCEndpoint

stack72:
Fix BUG-803

### DIFF
--- a/bin/clover/src/pipeline-steps/assetSpecificOverrides.ts
+++ b/bin/clover/src/pipeline-steps/assetSpecificOverrides.ts
@@ -110,6 +110,27 @@ const overrides = new Map<string, OverrideFn>([
     setAnnotationOnSocket(socket, { tokens: ["GroupId"] });
     variant.sockets.push(socket);
   }],
+  ["AWS::EC2::NetworkInterface", (spec: ExpandedPkgSpec) => {
+    const variant = spec.schemas[0].variants[0];
+
+    // Add an annotation for the Id output socket to connect to HostedZoneId
+    const socket = variant.sockets.find(
+      (s: ExpandedSocketSpec) => s.name === "Id" && s.data.kind === "output",
+    );
+    if (!socket) return;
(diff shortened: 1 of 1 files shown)


bin/clover/src/pipeline-steps/assetSpecificOverrides.ts
```diff
@@ -110,6 +110,27 @@ const overrides = new Map<string, OverrideFn>([
...
+
+    const prop = variant.domain.entries.find((p: ExpandedPropSpec) =>
+      p.name === "GroupSet"
+    );
+
+    if (!prop) return;
+    const groupSocket = createInputSocketFromProp(prop);
+
+    setAnnotationOnSocket(groupSocket, { tokens: ["GroupId"] });
+    variant.sockets.push(groupSocket);
```
sprutton1, line 132: It feels like we'll get more mileage out of these overrides by going the other way and adding `GroupSet` as an annotation on the relevant output socket, especially since these show up more than once

bin/clover/src/pipeline-steps/assetSpecificOverrides.ts
```diff
@@ -110,6 +110,27 @@ const overrides = new Map<string, OverrideFn>([
...
+  ["AWS::EC2::NetworkInterface", (spec: ExpandedPkgSpec) => {
+    const variant = spec.schemas[0].variants[0];
+
+    // Add an annotation for the Id output socket to connect to HostedZoneId
+    const socket = variant.sockets.find(
+      (s: ExpandedSocketSpec) => s.name === "Id" && s.data.kind === "output",
+    );
+    if (!socket) return;
+
+    setAnnotationOnSocket(socket, { tokens: ["NetworkInterfaceId"] });
```
sprutton1, line 122: I'm going to cover this case in a PR shortly

sprutton1 (approved):

PR #5593, opened 2025-03-04
feat: allow users to connect up NameServers to ResourceRecords in R53

johnrwatson:
Currently if you try and do cross-account domain delegation you're unable to connect up the NameServers from the child account into the Parent Accounts delegated zone. This fixes that.

### DIFF
--- a/bin/clover/src/pipeline-steps/assetSpecificOverrides.ts
+++ b/bin/clover/src/pipeline-steps/assetSpecificOverrides.ts
@@ -2,6 +2,7 @@ import _ from "npm:lodash";
 import _logger from "../logger.ts";
 import {
   createInputSocketFromProp,
+  createOutputSocketFromProp,
   ExpandedSocketSpec,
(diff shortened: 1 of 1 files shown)


bin/clover/src/pipeline-steps/assetSpecificOverrides.ts
```diff
@@ -125,6 +126,31 @@ const overrides = new Map<string, OverrideFn>([
...
+    const prop = variant.domain.entries.find((p: ExpandedPropSpec) =>
+      p.name === "ResourceRecords"
+    );
+
+    if (!prop) return;
+    const socket = createInputSocketFromProp(prop);
+    setAnnotationOnSocket(socket, { tokens: ["NameServers"] });
+
+    variant.sockets.push(socket);
+  }],
```
sprutton1, line 153: I just opened [this](https://github.com/systeminit/si/pull/5595) that covers off on this case.

image

PR #5594, opened 2025-03-04
BUG-778: upgrade component wsevent missing parent, b/c we send it too soon

jobelenus:


zacharyhamm (approved):
Looks right. Message was definitely too early before

PR #5598, opened 2025-03-04
fix(dal): When autoconnecting components, ensure we respect the output socket arity, so we don't try to autoconnect to an output socket with single arity that already has a connection

britmyerss:
<div><br/>via <a href="https://giphy.com/gifs/eKrgVyZ7zLvJrgZNZn">GIPHY</a></div>

### DIFF
--- a/lib/dal/src/component.rs
+++ b/lib/dal/src/component.rs
@@ -2843,42 +2843,64 @@ impl Component {
             }
             // build a map of component output sockets and values
             let output_sockets = component.output_socket_attribute_values(ctx).await?;
+            let outgoing_connections_to_check = component
+                .outgoing_connections(ctx)
+                .await?
+                .iter()
+                .map(|outgoing| outgoing.from_output_socket_id)
+                .collect_vec();
             for output_socket_av in output_sockets {
                 if let Some(output_socket_id) =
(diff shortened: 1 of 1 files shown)


lib/dal/src/component.rs
```diff
@@ -2843,17 +2843,32 @@ impl Component {
...
                     let av = AttributeValue::get_by_id(ctx, output_socket_av).await?;
                     let output_value = av.view(ctx).await?;
-
                     // Check against the list of available input sockets for compatibility
                     for (input_socket_id, info) in &available_input_sockets_to_connect {
-                        // Does this output socket fit this input socket? (including annotations!)
-                        if OutputSocket::fits_input_by_id(ctx, *input_socket_id, output_socket_id)
+                        // First ensure that if the socket arity is single, there isn't an existing connection from this output socket
+                        // Then, check that this output socket fits this input socket? (including annotations!)
+                        if (OutputSocket::get_by_id(ctx, output_socket_id)
```
zacharyhamm, line 2862: @britmyerss if I'm reading this right, we used to match on every output socket that fit. But now, we will only match on output sockets with SocketArity::One (that match). Is that what we want?

britmyerss: Yep you are right - fixed this! doh!

stack72 (approved):

PR #5600, opened 2025-03-04
feat: add discovery exception process to ensure we discover maximum information about the resource

johnrwatson:
Create a basic exception strategy to allow us for specific types to run a more data fruitful discovery strategy. This allows us to absolutely maximise the amount of relevant/important information returned onto the graph using an exception mapping.

E.g. for `AWS::EC2::LaunchTemplate` [as implemented in this pr] by running `cloudcontrol list-resources --type-name awsResourceType` we get extremely limited data back about the launch template, but querying the CLI directly using `aws ec2 describe-launch-template-versions --launch-template-id` we get much more full data back. To that end, I added an exception strategy to the discovery functions for at time of discovery the func looks to whether it's an exception and runs it's slightly different information discovery mechanism instead of via cloud control.

If there is discrepancy between the cloud control API surface and the alternative discovery mechanism, it allows you map the keys individually, so that you can achieve the maximum discovery surface area. In the `AWS::EC2::LaunchTemplate` it was fortunate all the keys exactly matched. However, there were `too many keys` in the response from the altApi so using this method those were filtered out.

```
const getImportExceptions = (resourceId: string, region: string) => ({
  "AWS::EC2::LaunchTemplate": {
      "altCommand": [
          "ec2",
          "describe-launch-template-versions",
          "--launch-template-id",
          resourceId,
          "--versions",
          "$Latest",
          "--region",
          region,
      ],
      "resourceExtractor": "LaunchTemplateVersions.0",
      "fieldMappings": [{
          sourceField: "LaunchTemplateData",
          mappedField: "LaunchTemplateData",
      }, {
          sourceField: "LaunchTemplateName",
          mappedField: "LaunchTemplateName",
      }]
  }
})
```

### DIFF
--- a/bin/clover/src/cloud-control-funcs/management/awsCloudControlDiscover.ts
+++ b/bin/clover/src/cloud-control-funcs/management/awsCloudControlDiscover.ts
@@ -1,12 +1,35 @@
+const getImportExceptions = (resourceId: string, region: string) => ({
+  "AWS::EC2::LaunchTemplate": {
+      "altCommand": [
+          "ec2",
+          "describe-launch-template-versions",
+          "--launch-template-id",
+          resourceId,
+          "--versions",
+          "$Latest",
+          "--region",
+          region,
+      ],
+      "resourceExtractor": "LaunchTemplateVersions.0",
+      "fieldMappings": [{
+          sourceField: "LaunchTemplateData",
+          mappedField: "LaunchTemplateData",
+      }, {
+          sourceField: "LaunchTemplateName",
+          mappedField: "LaunchTemplateName",
+      }]
+  }
+})
+
 async function main({
   thisComponent,
-}: Input): Promise<Output> {
+}: Input): Promise < Output > {
   const component = thisComponent.properties;
   const region = _.get(component, ["domain", "extra", "Region"], "");
   const awsResourceType = _.get(component, [
-    "domain",
-    "extra",
-    "AwsResourceType",
+      "domain",
+      "extra",
+      "AwsResourceType",
   ], "");
 
   let resourceList = [];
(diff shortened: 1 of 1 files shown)


johnrwatson:
@adamhjk This is the kind of thing I was thinking. This works (as tested in prod) but I'd be curious your thoughts whether it's the right way to approach it. It seems a similar strategy to how we're handling the rest of the clover exceptions so curious what you think

adamhjk:
interesting! I hadn't thought about handling the exceptions direclty in the function - intead, I would've just overridden the discovery function for that particular asset (so you won't wind up with a huge number of exceptions in a function that largely doesn't need them.)

Be curious to see what @sprutton1 and @vbustamante @jkeiser have to say

johnrwatson:
Oh absolutely let's move it into the generation code so we're not duping it for every asset.

I'm more thinking in terms of `altCommand` / `fieldMappings` should give us the flexibility here I think

bin/clover/src/cloud-control-funcs/management/awsCloudControlDiscover.ts
```diff
@@ -1,12 +1,35 @@
+const getImportExceptions = (resourceId: string, region: string) => ({
```
jkeiser, line 1: I don't believe you can put anything outside of the function itself; but you can put `function getImportExceptions(...)` inside the main() body itself--in fact, you can put it all the way at the bottom and still use it at the top!

jkeiser:
Thank you, we really need more exception functionality now that we're really running past what AWS schemas can give us!

Yeah, I think just replacing the discovery functions is what you want--that lets you write actual exceptional code to handle even weirder cases. If the fieldMappings gets us 90% of the way there, then modifying the main function might make sense; but if not, letting you rewrite the whole function gets us the rest of the way! I can imagine using that for things like where the *type* AWS gives us back is wrong--I've seen a few places where one place will do something like SecurityGroupIdentifier = { arn: "arn:aws:..." } or { id: "sg-id" }, and another will have it as two separate optional fields (SecurityGroupArn or SecurityGroupId). Maybe SecurityGroup is the wrong example, but still :)

jkeiser:
I do wish there was a way to have common functions that could be used both by the blanket implementation, and the overrides--but for now we can either duplicate it or insert the same shared functions into both.

jkeiser:
I could also imagine scenarios where an alternate discovery function needs to call multiple aws apis and piece the info together.

jkeiser:
... on the other hand, just encoding these into the function might be a great way to do it in the short term--especially if there only turn out to be a limited number and they don't change too often.

Wouldn't we need to bind to a different function ID each time it changed, though? Like, we'd have to regenerate all the clover assets? It's not the hugest deal, but it's not ideal to put a little upgrade icon on every single clover asset just because we added an exception for one of them. From that POV, having actual different functions might be best.

sprutton1:
I agree that we should do this as a per-asset override of a given function. If we end up with common patterns, we add heuristics to cover off on them, as we do for the structural overrides. I would probably just store the funcs in a sub-directory of where we have them now, organize it by asset, and replace the func in the spec. We would just need to ensure the new func also has a unique id

johnrwatson:
Closing as it's superseeded by 
https://github.com/systeminit/si/pull/5603

PR #5606, opened 2025-03-06
fix(dal): Validate default values if present

jkeiser:
This PR makes sure we correctly validate fields with default values and Joi.required(). If the field was unset (meaning it has a default), we were marking it invalid ("field is required") before.

Fixes BUG-808.

### DIFF
--- a/lib/dal/src/attribute/value.rs
+++ b/lib/dal/src/attribute/value.rs
@@ -2267,16 +2267,36 @@ impl AttributeValue {
         Self::fetch_value_from_store(ctx, self.value).await
     }
 
+    async fn default_value(
+        &self,
+        ctx: &DalContext,
+    ) -> AttributeValueResult<Option<serde_json::Value>> {
+        match AttributeValue::is_for(ctx, self.id).await? {
+            ValueIsFor::Prop(prop_id) => Ok(Prop::default_value(ctx, prop_id).await?),
+            ValueIsFor::InputSocket(_) | ValueIsFor::OutputSocket(_) => Ok(None),
+        }
+    }
+
     pub async fn value_or_default(
+        &self,
+        ctx: &DalContext,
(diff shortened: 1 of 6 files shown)


lib/dal/src/validation.rs
```diff
@@ -275,12 +275,16 @@ impl ValidationOutput {
...
+    /// for that format.
+    pub async fn compute_for_attribute_value(
         ctx: &DalContext,
         attribute_value_id: AttributeValueId,
-        value: Option<serde_json::Value>,
     ) -> ValidationResult<Option<ValidationOutput>> {
+        let value = AttributeValue::get_by_id(ctx, attribute_value_id)
+            .await?
+            .value_or_default(ctx)
+            .await?;
```
nickgerace, line 286: Good idea

lib/dal/src/attribute/value.rs
```diff
@@ -2267,16 +2267,36 @@ impl AttributeValue {
...
+
+    pub async fn value_or_default_or_null(
+        &self,
+        ctx: &DalContext,
         prop_id: PropId,
     ) -> AttributeValueResult<serde_json::Value> {
-        match Self::fetch_value_from_store(ctx, self.value).await {
-            Ok(Some(value)) => Ok(value),
-            _ => Ok(Prop::default_value(ctx, prop_id)
+        match self.value(ctx).await? {
```
nickgerace, line 2295: Should this be `self.value_or_default`?

jkeiser: This one uses an existing prop_id to get the default value rather than looking it up; I didn't take the time to check that the callers were all using the prop id directly associated with the av or not, so I didn't want to change behavior.

zacharyhamm (approved):

nickgerace (approved):

PR #5609, opened 2025-03-06
Perf: fix first paint to only wait on what it needs

jobelenus:
Here is how we tested the changes in `WorkspaceModelAndView`.

First, I put the rest of the endpoint calls into a setTimeout, in order to know the diagram would load without those being called. And it does load.

Second, I removed the setTimeout and instead awaited a sleep to prevent those from calling. I wanted to know if `await` inside `onBeforeMount` would prevent the diagram from drawing. It does not.

Third; watching the network tab, `get_diagram` is the only endpoint waiting to return, and once it returns we get a diagram painted, as other calls fire.

The main culprit was the other calls inside the success handler for LIST_VIEWS, due to the nature of the ApiRequest object awaiting on the success handler. This is why nothing would load until `get_all_components_and_edges` would finish (the opposite of what we want)

### DIFF
--- a/app/web/src/components/Workspace/WorkspaceModelAndView.vue
+++ b/app/web/src/components/Workspace/WorkspaceModelAndView.vue
@@ -241,16 +241,14 @@ const approvalData = computed(
 );
 
 onBeforeMount(async () => {
-  // get to first paint ASAP
-  await Promise.all([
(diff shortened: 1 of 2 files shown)


wendybujalski (approved):
😎

PR #5611, opened 2025-03-06
perf(veritech): Batch logs before sending to veritech

jkeiser:
This PR batches logs and sends them every N milliseconds (but will limit batching to 8MB).

One of our theories is we're bogging the reactors sending a log line every 8 nanoseconds. This should help both reduce the total size (only one message overhead per batch instead of per line) and the number of messages.

batch

### Draft

Flaw: if a log line is emitted, and no more logs are emitted for a long time, it will sit in the buffer until another log line comes in or . This is not ideal--you'd want users to see the last thing that happened sooner rather than later. I've used systems like this and watching logs while not being sure you're seeing all the logs is a problem--particularly, if there was an error or warning just before a program hung, you wouldn't see it until it eventually timed out.

Improvement: currently this will only cache log lines as long as they are the same level/group/etc. This could be better if it held onto all of these the whole time, but it's unclear how often it happens in practice.

### DIFF
--- a/bin/veritech/src/args.rs
+++ b/bin/veritech/src/args.rs
@@ -87,6 +87,10 @@ pub(crate) struct Args {
     #[arg(long)]
     pub(crate) cyclone_pool_size: Option<u32>,
 
+    /// Cyclone create firecracker setup scripts
+    #[arg(long)]
+    pub(crate) cyclone_create_firecracker_setup_scripts: Option<bool>,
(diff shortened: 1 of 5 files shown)


nickgerace:
Spoke to @jkeiser and we are closing this PR and killing the branch. Great ideas here that we may want in some shape or form, but not now.

PR #5614, opened 2025-03-06
feat(sdf): Add ability to set a module as private when contributing

stack72:
This means that when we won't show this module in module listings and it will be available for you on all workspaces you log into

This is feature flagged and will not yet be available to users

### DIFF
--- a/app/web/src/components/AssetContributeModal.vue
+++ b/app/web/src/components/AssetContributeModal.vue
@@ -11,20 +11,39 @@
         v-if="contributeModuleReqStatus.isError"
         :requestStatus="contributeModuleReqStatus"
       />
-      <p>
-        Everything you contribute will receive a code review, and we will reach
-        out if we have any questions or concerns. Assuming things look good, we
-        will then include your asset in a future version of System Initiative!
-      </p>
-      <p>
-        By clicking the 'Contribute to System Initiative' button, you agree to
-        license any code submitted under the terms of the
-        <a
-          class="text-action-500"
-          href="https://www.apache.org/licenses/LICENSE-2.0"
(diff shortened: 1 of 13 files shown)


lib/dal/src/cached_module.rs
```diff
@@ -386,6 +406,154 @@ impl CachedModule {
...
+        for row in rows {
+            result.push(row.try_into()?);
+        }
+
+        Ok(result)
+    }
+
+    // TO-DO: Paul
+    // I will refactor this + insert together - this is to temporarily get something working
+    async fn insert_private_module(
```
stack72, line 446: I have done this on purpose for now - I am testing the viability of this and don't want to break any of the existing routes rigth now - I will refactor this + insert into a single function

app/web/src/components/AssetContributeModal.vue
```diff
@@ -11,20 +11,39 @@
...
-          class="text-action-500"
-          href="https://www.apache.org/licenses/LICENSE-2.0"
-          >Apache License, Version 2.0</a
-        >, and that you intend for System Initiative, Inc. to distribute it.
-      </p>
+      <template
+        v-if="isPrivateModule && featureFlagsStore.PRIVATE_SCOPED_MODULES"
+      >
+        <p>
+          By clicking the 'Contribute to System Initiative' button, you confirm
```
britmyerss, line 18: nit: maybe this should say 'Publish' or something when it's private, vs. keeping the 'contribute' verbiage?

stack72: This is something we can think about after when we get eyes on it in tools prod as it's feature flagged

britmyerss (approved):

PR #5620, opened 2025-03-07
fix: hitbox calculation like a professional

johnrwatson:
Select the stuff you want to select via a drag select. Previously we'd miss or over-select the components, frames or view nodes.


jobelenus (approved):

johnrwatson:
https://jam.dev/c/2f5853e7-6d86-4a23-84a5-a88c8ae44d86

Test output above

PR #5621, opened 2025-03-07
fix: CWE-770 ring vulnerability

johnrwatson:
Patches Ring to meet CWE-770 in github https://github.com/systeminit/si/security/dependabot/386


johnrwatson:
Dismissed Vuln, we're not using overflow mode (even in debug) and can inherit the fix from buck2 when it's available.

PR #5623, opened 2025-03-08
feat: move the removeEmtpy function to a jsr module

adamhjk:
This moves the 'removeEmtpy' function, which we are reusing in lots of places, into a small JSR package. Then we can import that package with the major version number specified, and get automatic upgrades without having to republish all the assets in the universe.

It updates the clover resources to call this function from JSR instead of directly embedding it. That way we can test the function, make sure it works, and publish any changes by bumping the minor or patch version numbers.

See also https://jsr.io/@systeminit/remove-empty


adamhjk:
I refrained from adding a full CI workflow here, as I'm not sure how everyone will feel about this approach. For now you treat these functions a pretty yolo-tastic.

stack72 (approved):

sprutton1 (dismissed):

stack72 (approved):

PR #5631, opened 2025-03-10
fix(billing): Don't alert when people delete expired free trial subscriptions

jkeiser:
We had an alert firing because we deleted a user's free trial subscription when we added their free trial coupon. We maybe shouldn't have, but his makes the alert disappear now that we no longer care about whether therre *is* such a subscription.

### DIFF
--- a/component/lambda/sql/workspace_verifications.sql
+++ b/component/lambda/sql/workspace_verifications.sql
@@ -47,14 +47,14 @@ CREATE OR REPLACE VIEW workspace_verifications.owner_subscriptions_summary AS
 
 -- Owners must:
 -- 1. Have subscriptions (we allow for one day of slop on this)
--- 2. Have a free trial subscription
+-- 2. Have a free trial subscription (we allow old free trials to be deleted if they can't possibly affect billing anymore)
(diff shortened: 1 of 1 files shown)


component/lambda/sql/workspace_verifications.sql
```diff
@@ -54,7 +54,7 @@ CREATE OR REPLACE VIEW workspace_verifications.subscription_count_issues AS
         *,
         CASE
             WHEN subscription_count IS NULL AND first_timestamp < getdate() - INTERVAL '1 day' DAY THEN 'no_subscriptions'
-            WHEN free_trial_count = 0 THEN 'no_free_trial'
+            WHEN free_trial_count = 0 AND first_subscription_start_time >= getdate() - INTERVAL '30 days' THEN 'no_free_trial'
```
britmyerss, line 57: Can you adjust the comment above too to account for this?

jkeiser: Done! Also adjusted it to be 2 months: if a user was created Jan 1 with no free trial, it will still affect billing until invoices are finalized in the first week or two of Feb.

britmyerss (approved):

PR #5632, opened 2025-03-10
feat: add custom discovery for EC2::LaunchTemplate

johnrwatson:
Add more in depth discovery/import for EC2::LaunchTemplate to ensure we pull back in as much data as possible for the objects we find.

I've tested both funcs against existing objects in prod.


sprutton1 (approved):

johnrwatson:
Thanks @sprutton1

PR #5634, opened 2025-03-10
fix: patch Axios to mitigate CVE-2023-45857

johnrwatson:
Bump both auth0 and posthog-node to the latest versions to prevent the vulnerable version of axois coming in as a transitive dependency.

NB: This is quite a wide change because of how deep axois is used so I'll write up explicitly what tests were conducted to prevent a regression

---
**Update:** see test execution/overview in comments


fnichol (commented):
Now *that's* the way to upgrade axios! Better to upgrade the packages that depended on axios than to pin axios where it is used transitively.

johnrwatson:
So far testing has validated that the auth-portal is patched correctly without any obvious regression for the auth0 and posthog updates. Tests included new user sign up, modifying profile and log out.

I'm going to validate the application stack itself handles the axios change correctly.

johnrwatson:
On the app itself, I've tested creating a basic VPC (VPC + two subnets) and then deleting it out afterwards. I didn't spot anything untoward in the application logs of any service so think this is clean.

johnrwatson:
Fixed type issue in the auth0 client for auth api, all other functionality has remained the same

stack72 (approved):

johnrwatson:
@stack72 kindly retested the type work didn't regress the fundamental flow for user login etc

PR #5639, opened 2025-03-10
fix: /diff command requires generating the specs.

sprutton1:


sprutton1:
https://github.com/systeminit/si/actions/runs/13775559649/job/38523922299

Run here. We are extracting the name correctly, just forgot to run the spec gen first

vbustamante (approved):

PR #5640, opened 2025-03-10
fix(web): Set the dropdown search to be active by default

stack72:
When a search was enabled by default on a large dropdown list, it needed a user to click to enable the search, this makes it enabled by default

So when a user clicks a dropdown list, they can immediately type


wendybujalski (approved):
👍

PR #5643, opened 2025-03-11
fix: Tone down some spans, and instrument some others

britmyerss:
Some of these were turned up to info for performance testing and are not necessary every day.  Some are duplicately nested, and we don't need all of them.  

I also added some spans for function execution, as this data would be incredibly helpful in debugging. However, this should still be a net-negative impact on the volume of spans.... famous last words...

image

Here's a week's worth of traces in Production by name, and I've knocked out at least 10 of them completely. That should help!

<div><br/>via <a href="https://giphy.com/tennistv/">Tennis TV</a> on <a href="https://giphy.com/gifs/tennistv-tennis-medvedev-daniil-nS94hkAu4ENQeI4ZnQ">GIPHY</a></div>


lib/dal/src/func/runner.rs
britmyerss, line 1: @fnichol these were the missing span fields I was looking for last week.  I stripped the args out, but do you think we might want them at debug?

lib/rebaser-server/src/change_set_processor_task.rs
britmyerss, line 1: I had these at info so we could monitor the new quiescent shutdown logic for the rebaser.  Didn't really need them after all!

lib/dal/src/func/runner.rs
nickgerace: I think we could `debug!` them!

lib/dal/src/func/runner.rs
```diff
@@ -656,7 +641,7 @@ impl FuncRunner {
         // and in order to time the function's preparation vs. execution timings.
         #[instrument(
             name = "func_runner.run_attribute_value.prepare",
-            level = "info",
+            level = "debug",
```
nickgerace, line 644: I see that the parent function is also at `info` level, which feels like it could be noisy. If it's important though, let's keep it at that level.

lib/rebaser-server/src/change_set_processor_task.rs
```diff
@@ -187,7 +187,7 @@ impl post_process::OnSuccess for DeleteMessageOnSuccess {
         let stream = self.stream.clone();
 
         Box::pin(async move {
-            info!("deleting message on success");
+            debug!("deleting message on success");
```
nickgerace, line 190: Nice!

nickgerace (dismissed):

lib/si-layer-cache/src/db/rebase_batch.rs
```diff
@@ -82,7 +82,7 @@ where
 
     #[instrument(
         name = "rebase_batch.read_wait_for_memory",
-        level = "info",
+        level = "debug",
```
jkeiser, line 85: Does this sometimes take a long time and sometimes not? I see it taking a long time in workspace_snapshot.read_wait_for_memory in some traces (like half a second in the dvu job I was just looking at).

Maybe there's a way to only send this when we actually have to wait?

jkeiser (dismissed):

lib/si-layer-cache/src/db/rebase_batch.rs
britmyerss: Would be neat if we could get that - or if histogram worked (does it??), this could just be a metric we ship and see it over time.

jkeiser: Histogram works! You can just replace count with histogram and set the value. Prometheus does then weird, but they work--you query for metric_sum / metric_count to get the value back out.

jkeiser: But it makes sense to have a span. Debug is fine for now, we can figure out whether otelcol or opentelemetry can do what we need.

lib/dal/src/func/runner.rs
britmyerss: Yeah I'll punt this to when we need it :)

lib/dal/src/func/runner.rs
britmyerss: yeah I kept all of the func_runner.run parent functions, but these inner prepares aren't needed!

britmyerss:
> I think there's one thing to check on this PR (and I hate that I couldn't make it a compiler error): in each of the functions/methods where the `level` changes (in an `#[instrument]` call), we'll want to double check that when we get the `Span` into a variable, that the `let span = current_span_for_instrument_at!("debug");` levels match. That is for an `#[instrument(level = "info", ...]` we'd want a `let span = current_span_for_instrument_at!("info");` line to get the span, and same with `"debug"`/`"debug"`. Otherwise the `span.record(...)` calls might put the data on some other span up the call stack--that was the source of our last epic tracing bug.

Thanks @fnichol - don't need any more epic bugs :) I did have a few mismatches, so fixed those up! Also checked all callsites for `current_span_for_instrument_at` and they looked good!

jobelenus (approved):

PR #5644, opened 2025-03-11
Add veritech heartbeat app, naxum metrics and more

nickgerace:
## Introduction

This PR adds a heartbeat app with force reconnection abilities to veritech, adds naxum metrics and tunes metrics and telemetry everywhere.

## Context

Many of the changes in this PR come from the testing grounds PR (#5566), but not all changes were included. The "pause-resume" stream wrapper and client hot-swapper changes are either not ready for production use or are not tactics we should employ for one reason or another.

## Description

For veritech, the new heartbeat app is disabled in tests (by default) and enabled in all other scenarios (by default). The heartbeat app publishes messages via core NATS on a cadence and works with subject prefixes. It uses the instance ID to determine the destination subject.

Metrics have been added to track pool exhaustion, publisher stats, client stats, handlers doing work and more.

For naxum, metrics have been added for the core service and in middleware. Tracing has been tuned to be as noiseless as possible while making use of metrics where possible. These changes include tracking failures for "double ack failure" and standardizing maintain progress task logging.

The "ciruitbreaker" with 4 max messages has been removed as a result of these changes. Not only is it an anti-pattern for handling NACK'd messages, but it also masks over potential poison pill messages for consumers who allow for more than 4 re-deliveries.

Finally, the graceful shutdown time is now tunable for veritech and its default has moved from 6 hours to 20 minutes.


bin/veritech/src/args.rs
```diff
@@ -105,6 +105,26 @@ pub(crate) struct Args {
...
     #[arg(long)]
     pub(crate) instance_id: Option<String>,
+
+    /// Overrides the default graceful shutdown timeout (in seconds).
+    #[arg(long)]
+    pub(crate) graceful_shutdown_timeout_secs: Option<u64>,
+
+    /// Disables the heartbeat app.
+    #[arg(long, default_value = "false")]
+    pub(crate) disable_heartbeat_app: bool,
```
nickgerace, line 115: Will switch this to not be a double negative, like the healthcheck pool

bin/veritech/src/main.rs
```diff
@@ -8,7 +8,7 @@ mod args;
 const BIN_NAME: &str = env!("CARGO_BIN_NAME");
 const LIB_NAME: &str = concat!(env!("CARGO_BIN_NAME"), "_server");
 
-const GRACEFUL_SHUTDOWN_TIMEOUT: Duration = Duration::from_secs(60 * 60 * 6);
+const DEFAULT_GRACEFUL_SHUTDOWN_TIMEOUT: Duration = Duration::from_secs(60 * 20);
```
nickgerace, line 11: 6h to 20m

bin/veritech/src/main.rs
```diff
@@ -49,13 +49,20 @@ async fn async_main() -> Result<()> {
             .set_verbosity_and_wait(args.verbose.into())
             .await?;
     }
+    let graceful_shutdown_timeout = match args.graceful_shutdown_timeout_secs {
+        Some(provided) => Duration::from_secs(provided),
+        None => DEFAULT_GRACEFUL_SHUTDOWN_TIMEOUT,
+    };
```
nickgerace, line 55: If we don't like the change, we can switch it back to 6h

bin/veritech/src/main.rs
```diff
@@ -49,13 +49,20 @@ async fn async_main() -> Result<()> {
...
 
     let config = Config::try_from(args)?;
     debug!(?config, "computed configuration");
 
-    let server = Server::from_config(config, main_token.clone()).await?;
+    let (server, maybe_heartbeat_app) = Server::from_config(config, main_token.clone()).await?;
 
+    if let Some(mut heartbeat_app) = maybe_heartbeat_app {
+        main_tracker.spawn(async move { heartbeat_app.run().await });
+    }
```
nickgerace, line 65: The run method borrows `self` mutably. Moving the app using `async move` is easy and mimics how the main server runs.

Also... we probably can run the app in the main server itself, but that is for another day.

lib/dal-test/src/lib.rs
```diff
@@ -728,14 +728,15 @@ pub async fn veritech_server_for_uds_cyclone(
         let mut config_file = veritech_server::ConfigFile::default_local_uds();
         config_file.nats = nats_config;
         config_file.cyclone.set_pool_size(4);
+        config_file.disable_heartbeat_app = true;
```
nickgerace, line 731: Disable for tests

lib/naxum/src/middleware/ack/maintain_progress.rs
nickgerace, line 1: Everything here is now `trace` or `error`

lib/naxum/src/serve.rs
```diff
@@ -176,32 +177,30 @@ where
...
+                                metric!(counter.naxum.next_message.failed = 1);
                                 continue;
                             },
                             None => {
-                                info!("stream is closed, breaking out of loop");
-                                tracker.close();
-                                break;
+                                warn!(sleep_duration_ms = ?NEXT_NONE_MESSAGE_SLEEP_DURATION, "next message is 'None' on provided stream: going to sleep");
+                                metric!(counter.naxum.next_message.none = 1);
+                                std::thread::sleep(NEXT_NONE_MESSAGE_SLEEP_DURATION);
```
nickgerace, line 195: Non-async sleep... I'm a bit iffy on this. We have never seen this code exercised. This follows the recommendation to accept `Ready(None)` when the connection is unhealthy and reconnection is happening.

lib/naxum/src/serve.rs
```diff
@@ -236,7 +235,6 @@ where
                     }
                 }
             }
-            metric!(counter.naxum.pending_message_count = 0);
```
nickgerace, line 239: Redundant given that we subtract what's in flight. The `failed` and `none` metrics behave the same.

lib/naxum/src/serve.rs
```diff
@@ -274,18 +271,7 @@ where
...
-                    si.error.message = ?err,
-                    "failed to read message in after {} consecutive failures; closing stream",
-                    failed_count,
-                );
-                (None, permit)
-            } else {
-                (Some(Err(err)), permit)
-            }
-        }
+        Err(err) => (Some(Err(err)), permit),
```
nickgerace, line 274: Gone be the circuitbreaker.

lib/si-service/src/shutdown.rs
```diff
@@ -213,7 +213,7 @@ where
                 if let Err(_elapsed) = time::timeout(timeout, await_groups).await {
                     warn!("graceful shutdown timeout exceeded; completing shutdown anyway");
                     if let Some(telemetry_guard) = telemetry_guard {
-                        debug!("performing graceful shutdown for telemetry guard");
+                        warn!("performing graceful shutdown for telemetry guard");
```
nickgerace, line 216: Match the other log level

lib/veritech-client/tests/integration.rs
```diff
@@ -63,11 +63,13 @@ async fn veritech_server_for_uds_cyclone(
         .cyclone_spec(cyclone_spec)
         .crypto(config_file.crypto)
         .healthcheck_pool(false)
+        .disable_heartbeat_app(true)
```
nickgerace, line 66: Disabled for tests

lib/veritech-server/src/config.rs
```diff
@@ -25,11 +25,23 @@ use thiserror::Error;
 pub use si_settings::{StandardConfig, StandardConfigFile};
 
 const DEFAULT_CONCURRENCY_LIMIT: usize = 1000;
+const DEFAULT_POOL_SIZE: u32 = 50;
```
nickgerace, line 28: Moved the pool default up here

lib/veritech-server/src/handlers.rs
```diff
@@ -86,6 +87,18 @@ pub async fn process_request(
     subject: Subject,
     Headers(maybe_headers): Headers,
     msg: Message<InnerMessage>,
+) -> HandlerResult<()> {
+    metric!(counter.veritech.handlers_doing_work = 1);
+    let result = process_request_inner(state, subject, maybe_headers, msg).await;
+    metric!(counter.veritech.handlers_doing_work = -1);
+    result
```
nickgerace, line 94: Now we capture the result and wait for the very possible first and last moment to increment and decrement the metric

lib/veritech-server/src/handlers.rs
```diff
@@ -154,6 +168,7 @@ pub async fn process_request(
             return Err(HandlerError::InvalidIncomingSubject(subject));
         }
     }
+    info!(%execution_kind, %execution_id, "finished execution");
```
nickgerace, line 171: Now we log post dispatch

lib/veritech-server/src/handlers.rs
```diff
@@ -169,11 +184,15 @@ where
...
-    let mut client = state
-        .cyclone_pool
-        .get()
-        .await
-        .map_err(|err| span.record_err(HandlerError::CyclonePool(Box::new(err))))?;
+    let mut client = match state.cyclone_pool.get().await {
+        Ok(client) => client,
+        Err(err) => {
+            if let PoolNoodleError::ExecutionPoolStarved = err {
+                metric!(counter.veritech.pool_exhausted = 1);
```
nickgerace, line 191: Track pool exhaustion from veritech's perspective

lib/veritech-server/src/heartbeat.rs
```diff
@@ -0,0 +1,189 @@
...
+                    metric!(counter.veritech.heartbeat.force_reconnect.success = 1);
+                    self.needs_force_reconnect = false;
+                }
+                Err(err) => {
+                    error!(si.error.message = ?err, "heartbeat: force reconnect error");
+                    metric!(counter.veritech.heartbeat.force_reconnect.error = 1);
+                }
+            },
+            Err(err) => {
+                error!(si.error.message = ?err, "heartbeat: force reconnect timeout");
```
nickgerace, line 184: This is one of those weird elapsed errors. We can toss it if need be

lib/veritech-server/src/heartbeat.rs
```diff
@@ -0,0 +1,189 @@
...
+                    Ok(()) => {
+                        metric!(counter.veritech.heartbeat.publish.success = 1);
+                    }
+                    Err(err) => {
+                        error!(si.error.message = ?err, "heartbeat: publish error");
+                        metric!(counter.veritech.heartbeat.publish.error = 1);
+                    }
+                },
+                Err(err) => {
+                    error!(si.error.message = ?err, "heartbeat: publish timeout");
```
nickgerace, line 113: This is one of those weird elapsed errors. We can toss it if need be

lib/veritech-server/src/heartbeat.rs
```diff
@@ -0,0 +1,189 @@
...
+            publish_timeout_duration,
+            force_reconnect_timeout_duration,
+            heartbeat_subject,
+            heartbeat_payload: Vec::new(),
+            needs_force_reconnect: false,
+        }
+    }
+
+    /// Runs the [`HeartbeatApp`]. This method explicitly does not return anything.
+    pub async fn run(&mut self) {
```
nickgerace, line 55: This function is the heart of the heartbeat app

lib/naxum/src/serve.rs
```diff
@@ -176,32 +177,30 @@ where
...
-                                failed_count += 1;
+                                error!(si.error.message = ?err, "failed to read next message from stream");
+                                metric!(counter.naxum.next_message.failed = 1);
                                 continue;
                             },
                             None => {
-                                info!("stream is closed, breaking out of loop");
-                                tracker.close();
-                                break;
+                                warn!(sleep_duration_ms = ?NEXT_NONE_MESSAGE_SLEEP_DURATION, "next message is 'None' on provided stream: going to sleep");
```
nickgerace, line 193: We should remove this since the metric is doing the same thing

lib/naxum/src/serve.rs
fnichol: We can't do this--the Tokio worker thread will be blocked here for this duration and nothing on this worker will make progress, including any shutdown logic.

lib/naxum/src/serve.rs
```diff
@@ -176,32 +177,30 @@ where
...
+                                metric!(counter.naxum.next_message.none = 1);
+                                std::thread::sleep(NEXT_NONE_MESSAGE_SLEEP_DURATION);
+                                continue;
```
fnichol, line 196: I'm not 100% sure on this, I need to check the rest of the service flow. Reason being, this isn't what an HTTP server would do and while I know that's not what this is, we have no way to know whether this incoming `futures::Stream` would ever yield another `Some(_)` value, usually the streams are fused and thus permanently closed.

lib/naxum/src/serve.rs
fnichol: Agreed!

lib/si-data-nats/src/lib.rs
```diff
@@ -127,6 +127,21 @@ impl Client {
...
+        options = options.reconnect_delay_callback(|attempts| {
+            if attempts <= 1 {
+                Duration::from_millis(0)
+            } else {
+                metric!(counter.nats.reconnect_delay_callback.multiple_attempts = 1);
+                let exp: u32 = (attempts - 1).try_into().unwrap_or(u32::MAX);
+                let max = Duration::from_secs(4);
+                cmp::min(Duration::from_millis(2_u64.saturating_pow(exp)), max)
+            }
+        });
```
fnichol, line 143: Got it, then we have metrics, makes sense, thanks for that comment!

lib/si-service/src/shutdown.rs
fnichol: Makes sense

lib/veritech-server/src/config.rs
```diff
@@ -534,7 +608,7 @@ fn default_runtime_strategy() -> LocalUdsRuntimeStrategy {
 }
 
 fn default_pool_size() -> u32 {
-    50
+    DEFAULT_POOL_SIZE
```
fnichol, line 611: Yeah, this is better

lib/veritech-server/src/handlers.rs
fnichol: Long term I'm not sure we want too `info` level logs in these handlers (as above). It becomes way too noisy given how often these fire and I'm sensing these are here as a lack of trust on the trace spans (I know these lines have been here for a while, but today these are striking me as potential excessive noise as soon as we have our bug tracked down).

fnichol (commented):
Looks great! I want to double check the incoming `futures::Stream` behavior when `None` is passed into a naxum app first because I don't think we want to continue trying to process requests--there is almost certainly no chance the app will ever see another one.

bin/veritech/src/args.rs
nickgerace: Done

lib/naxum/src/serve.rs
nickgerace: Removed!

lib/veritech-server/src/handlers.rs
nickgerace: Removed since the `handlers_doing_work` metric and the log above the dispatch tell the right story

lib/naxum/src/serve.rs
nickgerace: We spoke the folks managing the client and were told reconnection could cause a `Ready(None)` to be followed by an eventual `Ready(Some(_))`.

Despite that, you and I just caught up and believe that we would rather follow convention and shutdown here, otherwise many naxum apps would be following anti-patterns in shutdown procedures.

lib/veritech-client/tests/integration.rs
```diff
@@ -63,11 +63,13 @@ async fn veritech_server_for_uds_cyclone(
         .cyclone_spec(cyclone_spec)
         .crypto(config_file.crypto)
         .healthcheck_pool(false)
+        .heartbeat_app(false)
```
nickgerace, line 66: Disabled for tests

nickgerace:
Addressed your changes (with you on the call 😄) @fnichol!

fnichol (approved):

PR #5657, opened 2025-03-12
build(deps): pin `aws-config`/`aws-sdk-firehose` to prevent native code

fnichol:
This change pins the AWS-related crates, only used in the `lib/data-warehouse-stream-client` crate, used by Forklift. The next versions of `aws-config` (i.e. [1.6.0][aws-config-1.6.0]) and `aws-sdk-firehose` (i.e. [1.69.0][aws-sdk-firehose-1.69.0]) pull in HTTP clients that use the [aws-lc-rs] crate which contains native C code which would need to be built via Buck2.

Porting the 900-line [build script] is going to take some effort and won't be a quick solve...

[aws-config-1.6.0]: https://crates.io/crates/aws-config/1.6.0
[aws-sdk-firehose-1.69.0]: https://crates.io/crates/aws-sdk-firehose/1.69.0
[aws-lc-rs]: https://github.com/aws/aws-lc-rs
[build script]: https://github.com/aws/aws-lc-rs/blob/main/aws-lc-sys/builder/main.rs

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -501,9 +501,9 @@ dependencies = [
 
 [[package]]
 name = "aws-credential-types"
-version = "1.2.1"
+version = "1.2.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "60e8f6b615cb5fc60a98132268508ad104310f0cfb25a1c22eee76efdf9154da"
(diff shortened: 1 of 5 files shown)


britmyerss (approved):
Built and ran the stack locally!

PR #5663, opened 2025-03-13
feat(web) - diagram drag layer

wendybujalski:
Diagram components (nodes, groups, and views) are now rendered in two separate groups - static components on one layer, and components that are being dragged around by the user on another layer.


jobelenus (approved):
🤘🏻🚀

PR #5673, opened 2025-03-13
feat(sdf): Add inferred connections to incomingConnections in management funcs

jkeiser:
This adds inferred connections to `thisComponent.incomingConnections` when running a management func. This means that if your management component lives inside an AWS Credential and Region (and has inferred connections to them because of parentage), your component will get those connections and be able to pass them on to components it creates or manages.

Refactor budget: updated paste.rs and management.rs tests to use `Result<>`

## Testing

* Tested manually with components in the UI
* Created management function in Subnet and verified it received AWS Credential and Region inferred connections
* Integration tests

### DIFF
--- a/lib/dal/src/management/prototype.rs
+++ b/lib/dal/src/management/prototype.rs
@@ -7,6 +7,7 @@ use std::{
 
 use serde::{Deserialize, Serialize};
 use si_events::FuncRunId;
+use si_id::OutputSocketId;
 use telemetry::prelude::*;
 use thiserror::Error;
 use veritech_client::{ManagementFuncStatus, ManagementResultSuccess};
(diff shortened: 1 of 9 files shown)


lib/dal/src/management/prototype.rs
```diff
@@ -171,45 +172,64 @@ async fn build_incoming_connections(
...
-            });
+            socket_connections
+                .push(build_connection(ctx, from_component_id, from_socket_id).await?);
         }
 
-        // If there are multiple sockets, this will only keep the last one.
-        // This situation is supposed to be prevented elsewhere.
+        // Collect inferred connections for this input socket
+        for from in input_socket.find_inferred_connections(ctx).await? {
+            println!("from: {:?}", from);
```
zacharyhamm, line 186: Did you want to leave this println! in?

zacharyhamm (commented):
This looks correct. Going to click test it now. But you probably didn't want that `println!` left in

lib/dal/src/management/prototype.rs
jkeiser: Nope! Removed :)

zacharyhamm (approved):

PR #5675, opened 2025-03-14
chore(sdf): Give Schema::get_by_id/default_variant_id the good name

jkeiser:
This splits a small bit of refactoring so it doesn't pollute a different upcoming PR.

In the continuing quest to make get_by_id more consistent across node types, push errors like NoDefaultSchemaVariantFound down into a single place, and encourage not grabbing entire objects when the ID is all you need, this PR:

* Renames Schema::get_default_schema_variant_by_id[_or_error] -> default_variant_id[_opt]
* Removes Schema::get_default_schema_variant_id[_or_error](&self) since all they did was call _by_id 
* Renames Schema::get_by_id[_or_error] -> get_by_id[_opt]
* Adjusts callers to the new names
* Moves [_opt] callers who immediately error, to get the non-opt version instead
* Calls Schema::exists_locally() instead of Schema::get_by_id_opt().is_some() in a few places
* Moves the "install_component" endpoint to use Schema::get_or_install_default_variant instead of installing the module itself (the code was identical except the error enum being emitted)

## Testing

* Ran integration tests
* Dragged uninstalled component onto the canvas, let it install

### DIFF
--- a/lib/dal-test/src/expected.rs
+++ b/lib/dal-test/src/expected.rs
@@ -230,18 +230,15 @@ impl ExpectSchema {
     }
 
     pub async fn schema(self, ctx: &DalContext) -> Schema {
-        Schema::get_by_id_or_error(ctx, self.0)
+        Schema::get_by_id(ctx, self.0)
             .await
             .expect("get schema by id")
     }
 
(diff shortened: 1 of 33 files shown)


zacharyhamm (approved):
Tested component creation and schema variant listing (the big code changes here). Everything else looks fine to me

PR #5677, opened 2025-03-14
feat: add optional support for tokio-console in all Rust services

fnichol:
This change adds a new CLI flag of `--tokio-console` which will enable collection and transmission of Tokio runtime telemetry for use with the [tokio-console] debugging tool.

By default, the responsible subscriber will not be active as there is a potential for performance impact.

To enable this mode in any of the Rust services there are 2 ways to do this:

- Add a `--tokio-console` flag to the CLI when starting
- Export `SI_TOKIO_CONSOLE=true` when running a service

[tokio-console]: https://github.com/tokio-rs/console

### DIFF
--- a/lib/telemetry-application-rs/src/lib.rs
+++ b/lib/telemetry-application-rs/src/lib.rs
@@ -19,6 +19,7 @@ use std::{
     time::{Duration, Instant},
 };
 
+use console_subscriber::ConsoleLayer;
 use derive_builder::Builder;
 use opentelemetry_sdk::{
     metrics::SdkMeterProvider,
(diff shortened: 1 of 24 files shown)


lib/telemetry-application-rs/src/lib.rs
```diff
@@ -260,6 +264,10 @@ pub fn init(
         "telemetry configuration"
     );
 
+    if config.tokio_console {
+        warn!("tokio-console support is enabled; this could impact production performance");
+    }
```
nickgerace, line 269: Nice.

nickgerace (approved):

PR #5679, opened 2025-03-14
chore(dal): remaining get_by_id_or_error -> get_by_id

jkeiser:
Turns out there's only a few get_by_id_or_errors left in the codebase. Took a few minutes to just finish that off.

These were fixable entirely by rote rename. The only places unnecessarily using the _opt version were in tests (I fixed those).

## Testing

* Integration tests
* Made sure I could create a component in the UI (that it wasn't completely broken)

### DIFF
--- a/lib/dal/src/approval_requirement.rs
+++ b/lib/dal/src/approval_requirement.rs
@@ -187,12 +187,9 @@ impl ApprovalRequirementDefinition {
             approvers: content.approvers,
         }
     }
-    pub async fn get_by_id_or_error(
-        ctx: &DalContext,
-        id: ApprovalRequirementDefinitionId,
-    ) -> Result<Self> {
+    pub async fn get_by_id(ctx: &DalContext, id: ApprovalRequirementDefinitionId) -> Result<Self> {
         ctx.workspace_snapshot()?
-            .get_by_id_or_error(ctx, id)
+            .get_approval_requirement_definition_by_id(ctx, id)
             .await
             .map_err(Into::into)
     }
--- a/lib/dal/src/func/runner.rs
+++ b/lib/dal/src/func/runner.rs
@@ -97,8 +97,6 @@ pub enum FuncRunnerError {
     Func(#[from] FuncError),
     #[error("function backend error: {0}")]
     FuncBackend(#[from] FuncBackendError),
-    #[error("validation function intrinsic func is missing -- bug!")]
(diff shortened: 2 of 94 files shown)


lib/dal/src/approval_requirement.rs
```diff
@@ -187,12 +187,9 @@ impl ApprovalRequirementDefinition {
...
         }
     }
-    pub async fn get_by_id_or_error(
-        ctx: &DalContext,
-        id: ApprovalRequirementDefinitionId,
-    ) -> Result<Self> {
+    pub async fn get_by_id(ctx: &DalContext, id: ApprovalRequirementDefinitionId) -> Result<Self> {
         ctx.workspace_snapshot()?
-            .get_by_id_or_error(ctx, id)
+            .get_by_id(ctx, id)
```
zacharyhamm, line 192: While we're renaming this, we should give it a better name in general. This comes from the ApprovalRequirementExt trait but nothing about the function indicates that. It confused me a lot just now staring at it (i assumed it was getting a workspace snapshot by id or erroring?)

jkeiser: You are completely right! I assumed the same. This is because `ApprovalRequirementExt` is implemented by WorkspaceSnapshot, letting you treat it as a collection of approval requirements.

The situation is even more confusing, because there are two `ApprovalRequirementExt` traits, each of which is impl'd by exactly one struct. They look sort of similar. I'm not sure what the right turn is here. I'm willing to rename in the short term, though! `get_approval_requirement_by_id()` sounds about right.

jkeiser: Fixed.

lib/dal/src/func/runner.rs
```diff
@@ -487,9 +485,7 @@ impl FuncRunner {
         ) -> FuncRunnerResult<FuncRunner> {
             let func_id =
                 Func::find_intrinsic(ctx, super::intrinsics::IntrinsicFunc::Validation).await?;
-            let func = Func::get_by_id(ctx, func_id)
-                .await?
-                .ok_or(FuncRunnerError::FuncIntrinsicValidationMissing)?;
```
zacharyhamm, line 492: We're not checking for this anywhere, and I think it might technically be unreachable since the find_intrinsic would have failed above if the func was not on the graph

zacharyhamm (approved):

PR #5680, opened 2025-03-14
chore(dal): User/Workspace::get_by_pk_or_error -> get_by_pk

jkeiser:
This does the same thing as get_by_id, for the two things with _pk. I did not rename to _id because I didn't want to take on the work of renaming the type (WorkspacePk -> WorkspaceId) here.

This removes a few unnecessary Workspace/UserNotFound error types and relies on WorkspaceError::WorkspaceNotFound/UserError::UsernotFound instead, but there are no real logic changes, just rename / use underlying error.

## Tests

* Integration tests
* Dragged out a component in the diagram

### DIFF
--- a/lib/sdf-server/src/service/session.rs
+++ b/lib/sdf-server/src/service/session.rs
@@ -6,7 +6,7 @@ use axum::{
 };
 use dal::{
     workspace_integrations::WorkspaceIntegrationsError, KeyPairError, StandardModelError,
-    TransactionsError, UserError, UserPk, WorkspaceError, WorkspacePk,
+    TransactionsError, UserError, WorkspaceError, WorkspacePk,
 };
 use serde::{Deserialize, Serialize};
 use si_data_spicedb::SpiceDbError;
@@ -28,10 +28,6 @@ pub enum SessionError {
     AuthApiError(String),
     #[error(transparent)]
     ContextTransactions(#[from] TransactionsError),
-    #[error("Invalid user: {0}")]
-    InvalidUser(UserPk),
-    #[error("Invalid workspace: {0}")]
-    InvalidWorkspace(WorkspacePk),
     #[error("json serialize failed")]
     JSONSerialize(#[from] serde_json::Error),
     #[error(transparent)]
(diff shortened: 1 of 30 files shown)


lib/sdf-server/src/service/v2/workspace/install_workspace.rs
```diff
@@ -42,9 +42,7 @@ pub async fn install_workspace(
             .tenancy()
             .workspace_pk_opt()
             .ok_or(WorkspaceAPIError::RootTenancyInstallAttempt)?;
-        Workspace::get_by_pk(&ctx, &workspace_pk)
-            .await?
-            .ok_or(WorkspaceAPIError::WorkspaceNotFound(workspace_pk))?
```
zacharyhamm, line 47: These WorkspaceAPI errors are interpreted differently by the into_response, so dropping them will break existing behavior

lib/sdf-server/src/service/session.rs
```diff
@@ -73,10 +69,12 @@ pub type SessionResult<T> = std::result::Result<T, SessionError>;
 impl IntoResponse for SessionError {
     fn into_response(self) -> Response {
         let (status_code, error_message) = match self {
-            SessionError::LoginFailed => (StatusCode::CONFLICT, self.to_string()),
-            SessionError::InvalidWorkspace(_) => (StatusCode::CONFLICT, self.to_string()),
-            SessionError::WorkspacePermission(_) => (StatusCode::UNAUTHORIZED, self.to_string()),
-            SessionError::AuthApiError(_) => (StatusCode::UNAUTHORIZED, self.to_string()),
+            Self::LoginFailed => (StatusCode::CONFLICT, self.to_string()),
+            Self::Workspace(WorkspaceError::WorkspaceNotFound(_)) => {
```
zacharyhamm, line 73: Is a StatusCode::CONFLICT what we want here?

jkeiser: NOT_FOUND sounds better to me.

lib/sdf-server/src/service/v2/workspace/install_workspace.rs
jkeiser: I modified the into_response() to look at WorkspaceError::WorkspaceNotFound instead. I also removed the error variant so we can be sure that if anything else interprets that error, it would not compile (everything is clear).

jkeiser: I'm definitely willing to keep using WorkspaceAPIError here; I can see us maybe wanting a rule "if we're interpreting the error differently we should have our own local variant." We don't really follow that in other places right now, but that's not a reason not to have the rule :)

Doesn't feel like a decision with a clear answer to me, but I'm on board if you have stronger opinions.

lib/sdf-server/src/service/session.rs
zacharyhamm: i think that makes more sense!

jkeiser: Especially since this is for session endpoints, I'm worried we have callers that are sensitive to the error codes; unless you feel strongly, I don't think this is a good PR to change this behavior in.

zacharyhamm (approved):

zacharyhamm (approved):

PR #5684, opened 2025-03-14
feat(sdf): allow management functions to create any asset

jkeiser:
This allows management functions to create any asset without having to preselect them in Managed Assets.

It still shows as a type error in the editor, but when you run the management function, the component is created.

**Draft**: because no integration tests yet

image
image

### Tests

* Created management function that creates an AWS::EC2::VPC, watched it get installed and created
* Dragged out VPC Template, dependent assets still install
* TODO Integration tests

### Out of Scope

Planned work that is not in this PR:
* fix typescript type hints/errors (you still see an error in the editor if a schema is not in the Managed Assets list)
* remove the Managed Assets tab


zacharyhamm (commented):
Should probably also update [dal/src/component.rs ](https://github.com/systeminit/si/blob/edf75d67719d6e1a7283ab724954123c06737ebf/lib/dal/src/component.rs#L3847) to no longer require the component-to-be-managed to be in the set of schemas. There is likely some frontend code related to this as well

zacharyhamm (approved):
Works! It's a little weird that you can't reconnect the management edge if you create a component that is not a "managed schema", but that can be fixed in a follow up

jkeiser:
That is an excellent point! I'm happy to either merge this or update it, but it's really good to find that edge. I hadn't thought of that in my plans!

zacharyhamm (dismissed):

jkeiser:
@zacharyhamm turns out I should have re-run the paste tests after I removed the managed assets from them :) To fix it, I fixed Component::manage_component to allow any connection. The changes are localized to three spots where I removed the `ComponentNotManagedSchema` error entirely.

jkeiser:
Try succeeded except the (now fixed) lint error; this is ready again. No changes except the above, still.

zacharyhamm (approved):

PR #5688, opened 2025-03-17
feat: remove "managed assets" list from mgmt funcs

jkeiser:
This PR removes the concept of managed assets from management functions. #5684 already allowed you to use assets that aren't in that list, and this cleans up remaining behavior and removes managed_schemas from the system.

* Editor: management function create/update ops get full types for all installed assets, plus the names of non-installed assets with generic asset types
* Management sockets can be manually connected to any other management socket regardless of schema
* Managed schemas panel removed
* Installing a manager asset will no longer auto-install managed assets. (If they are not installed, they will be auto-installed when running the management function that creates them.)

**Code Review:** I have included comments to highlight places in the code where behavior changes happened. Everything else is a direct result of not passing managed_schemas around in types.

## Risk Mitigation

The first big risk is we fix up some things that depend on managed assets, but later find bugs because some corner of the system we hadn't thought about/tested depends on them. This PR removes that risk by removing managed_schemas from all Rust and JS types, causing everything that uses them not to compile. All callers are adjusted in this PR.

### Serialization/Deserialization

Removing this field from all structs, however, introduces a risk that old changesets, and old uploaded packages, will not work because they do not deserialize correctly. (All of these are optional in any serialized format, so we don't have an issue with serialization.) The strategy:

* Remove from serde-deserialized structs: Serde already allows extra fields to exist in the JSON that we don't deserialize. If the data is there, we will skip it when reading.
* Remove from typescript types: JS can receive extra fields just fine; types are just assertions about the fields the app cares about.
* Remove from graph node struct: Our graph deserializer now reads the prop if it exists, and discards the value (otherwise it would error out because the package contains an unknown field). This was an optional prop before, so not emitting it is not an isssue.
* managed_assets is never written out in any Rust serialization format, even if it existed initially, so round-trip edits of variants (for example) will clean them out of existing management functions.

###  Testing

* [X] Integration tests
  - [X] Remove managed schemas from all existing tests; they should still work.
  - [X] Existing tests to DAL side of connecting management sockets

#### Manual Tests

* [X] Typescript help creating management function
  - [X] no typescript errors creating/updating non-installed assets and their props/sockets
  - [X] no typescript errors creating/updating installed assets and their props/sockets
  - [X] typescript errors misspelling component kind
  - [X] typescript errors misspelling props/sockets or creating wrong type when creating/updating components with installed assets
* [X] Management function functionality:
  - [X] Management functions still create and update both installed and non-installed assets
  - [X] Auto-install assets on management function run
* [X] Package install backcompat: install and run existing VPC Template asset in new changeset. Typescript shows errors appropriately, components are created.
* [X] Contribute: Contribute management function and install in new changeset; should still work.
* [X] si-fs: make sure management functions (from old and new servers) show up.


app/web/src/components/ModelingDiagram/ModelingDiagram.vue
```diff
@@ -2538,15 +2538,9 @@ const drawEdgePossibleTargetSocketKeys = computed(() => {
     if (fromSocket.def.direction === possibleToSocket.def.direction)
       return false;
 
-    const isManagedSchema =
```
jkeiser, line 2541: This logic decides which management sockets we should show as "open" while you are dragging an edge. The logic says "you may not drag from a management socket to another management socket unless it's in managed schemas." It then goes on to disallow them if the connection annotations do not match, which is probably how we implicitly disallow management <-> non-management connections since most (all?) sockets have annotations.

In the absence of management schemas, this logic just says "management sockets only connect to other management sockets."

app/web/src/components/FuncEditor/FuncDetails.vue
```diff
@@ -147,24 +147,6 @@
               </Stack>
             </TreeNode>
 
-            <TreeNode
```
jkeiser, line 150: Removed "Managed Assets" from management function right side panel.

app/web/src/store/components.store.ts
```diff
@@ -326,30 +326,12 @@ export function getPossibleAndExistingPeerSockets(
       (c) =>
         c.def.sockets
           ?.filter((peerSocket) => {
-            // Management inputs can only be connected to management outputs
```
jkeiser, line 329: This logic determines possible connections on the Connections Panel. Before this change, it denied connections between management and non-management sockets, and between management sockets with non-matching schema. Now it denies connections between management and non-management sockets.

lib/dal/src/cached_module.rs
```diff
@@ -430,10 +433,11 @@ impl CachedModule {
         Ok(result)
     }
 
-    pub async fn get_user_scoped_modules(
```
jkeiser, line 433: Before, get_user_scoped_modules() got you *only* user-scoped modules, and you appended that to non-scoped modules. Now, it's just `latest_modules()` and gets both (by retrieving the user from the context).

latest_modules() used to get only non-scoped modules; now all callers of that use latest_user_independent_modules(). It feels like most callers should be using the new latest_modules() and including user-scoped modules, but I don't know enough to be sure they're not excluding them on purpose.

lib/dal/src/func/binding/management.rs
```diff
@@ -70,93 +69,134 @@ impl ManagementBinding {
             component_create_type,
             component_input_type,
         ) = match ManagementPrototype::prototype_id_for_func_id(ctx, func_id).await? {
-            Some(prototype_id) => match ManagementPrototype::get_by_id(ctx, prototype_id).await? {
```
jkeiser, line 73: I suggest using the "hide whitespace" option to review this; removing managed_schemas caused us not to need the ManagementPrototype anymore, which caused a reindent.

lib/dal/src/func/binding/management.rs
```diff
@@ -70,93 +69,134 @@ impl ManagementBinding {
...
+                    component_create_types.push(component_create_type);
+                    component_input_types.push(component_input_type);
+                }
 
-                        let root_prop = Prop::get_by_id(
-                            ctx,
-                            SchemaVariant::get_root_prop_id(ctx, variant_id).await?,
-                        )
-                        .await?;
+                // Add a generic type for any uninstalled modules, but with the names so that
```
jkeiser, line 137: After producing types for installed schemas, we produce a single create and input type that covers all uninstalled schemas; `kind` becomes e.g. `kind: "AWS::EC2::VPC" | "AWS::EC2::Subnet" | ...,`

lib/sdf-server/src/service/diagram/create_component.rs
```diff
@@ -82,12 +82,6 @@ pub async fn create_component(
 
             let variant_id = Schema::get_or_install_default_variant(&ctx, schema_id).await?;
             let variant = SchemaVariant::get_by_id(&ctx, variant_id).await?;
-            let managed_schema_ids = SchemaVariant::all_managed_schemas(&ctx, variant_id).await?;
```
jkeiser, line 85: This is where we would auto-install managed schemas when dragging an asset onto the canvas. Now this happens when the management function runs and creates a component whose asset is not yet installed.

lib/sdf-server/src/service/v2/func/binding/update_binding.rs
```diff
@@ -188,9 +158,6 @@ pub async fn update_binding(
         FuncKind::CodeGeneration | FuncKind::Qualification => {
             update_leaf_func_bindings(&ctx, request.bindings).await?;
         }
-        FuncKind::Management => {
```
jkeiser, line 191: Updates should not be happening anymore, as there is nothing updateable in a management binding.

lib/sdf-server/src/service/v2/variant/list_variants.rs
```diff
@@ -39,21 +38,13 @@ pub async fn list_variants(
         ));
     }
 
-    let mut user_scoped_modules: Vec<CachedModule> = vec![];
```
jkeiser, line 42: This is the place we were using latest_modules and concatenating with user-scoped modules.

lib/si-frontend-types-rs/src/fs.rs
```diff
@@ -196,9 +196,9 @@ pub enum Binding {
     CodeGeneration {
         inputs: Vec<LeafInputLocation>,
     },
-    Management {
-        managed_schemas: Option<Vec<String>>,
-    },
+    // TODO can we remove this? Don't want to remove directories if it's necessary, and I don't
```
jkeiser, line 199: Question for si-fs expert.

lib/si-pkg/src/node/management_func.rs
```diff
@@ -53,17 +44,13 @@ impl ReadBytes for ManagementFuncNode {
         let func_unique_id = read_key_value_line(reader, KEY_FUNC_UNIQUE_ID_STR)?;
         let name = read_key_value_line(reader, KEY_NAME_STR)?;
         let description = read_key_value_line_opt(reader, KEY_DESCRIPTION_STR)?;
-        let managed_schemas_str = read_key_value_line_opt(reader, KEY_MANAGED_SCHEMAS_STR)?;
-        let managed_schemas = match managed_schemas_str {
-            None => None,
-            Some(managed_schemas_str) => Some(serde_json::from_str(&managed_schemas_str)?),
-        };
+        // Ignore managed schemas now that we no longer support it
```
jkeiser, line 47: This is how we make sure we can still read old changesets and/or packages.

lib/sdf-server/src/service/v2/fs.rs
```diff
@@ -187,7 +187,7 @@ pub async fn hydrate(
 
     tracker.track(ctx, "fs/hydrate", serde_json::json!({}));
 
-    let cached_modules = Arc::new(CachedModule::latest_modules(ctx).await?);
+    let cached_modules = Arc::new(CachedModule::latest_user_independent_modules(ctx).await?);
```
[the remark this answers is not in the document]
zacharyhamm: yes, absolutely it should

lib/si-pkg/src/node/management_func.rs
```diff
@@ -53,17 +44,13 @@ impl ReadBytes for ManagementFuncNode {
...
+        let _ = read_key_value_line_opt(reader, KEY_MANAGED_SCHEMAS_STR)?;
```
zacharyhamm, line 48: ✅ went right here to make sure old packages would stay compatible

lib/si-frontend-types-rs/src/fs.rs
zacharyhamm: We can remove this later, right now it's harmless to keep it in

lib/sdf-server/src/service/v2/fs.rs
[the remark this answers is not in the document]
zacharyhamm: We definitely don't want to hide user-scoped modules from the si-fs, since users will want to edit / look at them there

zacharyhamm (commented):
I don't think we want to restrict the schemas visible in the si-fs to "user independent ones". Otherwise this looks good. Doing some more testing

lib/sdf-server/src/service/v2/fs.rs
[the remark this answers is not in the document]
jkeiser: I'm going to leave this as a task for a fast follow, since we're already arrived at the end of this one, I think! This doesn't change behavior anyway.

lib/si-frontend-types-rs/src/fs.rs
jkeiser: Removed the comment!

zacharyhamm (approved):
Was unable to test management edges since the sockets are gone. But everything else works! Tested with a few management func creations and runs

PR #5690, opened 2025-03-17
chore(ci): Add workflow dispatch for jsr publish

stack72:
Takes the name of a library based on lib/jsr-systeminit and then runs a publish inside that library

### DIFF
--- a/.github/workflows/jsr-publish.yml
+++ b/.github/workflows/jsr-publish.yml
@@ -1,0 +1,68 @@
+name: Publish to JSR
+
+on:
+  workflow_dispatch:
+    inputs:
+      library:
+        description: "Select the library to publish"
+        required: true
+        type: choice
+        options:
+          - remove-empty
+
+jobs:
+  check_branch:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check branch is main
+        if: github.ref != 'refs/heads/main'
+        run: |
+          echo "This workflow can only be run on the main branch."
+          exit 1
+  publish:
+    needs: check_branch
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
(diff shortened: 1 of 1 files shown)


.github/workflows/jsr-publish.yml
```diff
@@ -0,0 +1,59 @@
+name: Publish to JSR
+
```
johnrwatson, line 2: Should restrict this to only run on main

stack72: oh good shout!! I will update that

stack72: You mean do a check like this?

```
jobs:
  check_branch:
    runs-on: ubuntu-latest
    steps:
      - name: Check branch is main
        if: github.ref != 'refs/heads/main'
        run: |
          echo "This workflow can only be run on the main branch."
          exit 1
```

johnrwatson: Yeah that works, you could run the whole thing apart from the publish, at least you can then check it's linted correctly

.github/workflows/jsr-publish.yml
```diff
@@ -0,0 +1,59 @@
...
+        working-directory: ${{ steps.set-path.outputs.library_path }}
+        run: deno lint
+
+      - name: Authenticate with JSR
+        run: deno add @jsr/npm
+
+      - name: Configure JSR token
+        env:
+          JSR_TOKEN: ${{ secrets.JSR_TOKEN }}
+        run: deno jsr login "$JSR_TOKEN"
```
johnrwatson, line 55: Please put this token in the tech ops vault / make it a token from systeminit bot if possible

johnrwatson (approved):

PR #5691, opened 2025-03-17
feat(dal,web): Expose socket values for Components

britmyerss:
This PR includes the Value for a given Component's Sockets when fetching Component Data.  The Front end change is behind a feature flag (`socket-value`), which when enabled, exposes this information (and a bit more) in the Connections Panel. We might never want to enable this for users, but shipping this so we can use the data in other places! 

Notably awful about this PR is I'm not handling secrets, so the value that is displayed for a secret socket is the secret ID. Not very helpful. 

Before:
image

After: 
image

<div><br/>via <a href="https://giphy.com/outlander/">Outlander</a> on <a href="https://giphy.com/gifs/outlander-season-7-starz-704-qE9AFvF4QCPeiP4bh8">GIPHY</a></div>

Tested with and without the flag on, all works as expected!

### DIFF
--- a/lib/dal/src/component.rs
+++ b/lib/dal/src/component.rs
@@ -8,6 +8,7 @@ use si_pkg::KeyOrIndex;
 use socket::{ComponentInputSocket, ComponentOutputSocket};
 use std::collections::{hash_map, HashMap, HashSet, VecDeque};
 use std::num::{ParseFloatError, ParseIntError};
+use std::str::FromStr;
 use std::sync::Arc;
 use telemetry::prelude::*;
 use thiserror::Error;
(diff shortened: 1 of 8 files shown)


lib/dal/src/component.rs
```diff
@@ -4013,13 +4017,43 @@ impl Component {
...
+                    continue;
+                }
+            }
+            let socket_value = match comp_socket.direction {
+                DiagramSocketDirection::Bidirectional => None,
+                DiagramSocketDirection::Input => {
+                    ComponentInputSocket::value_for_input_socket_id_for_component_id_opt(
+                        ctx,
+                        self.id(),
+                        InputSocketId::from_str(&comp_socket.id)?,
```
britmyerss, line 4041: this is a bit dirty, but I didn't want to refactor this whole thing right now.

johnrwatson (approved):

stack72 (approved):

PR #5696, opened 2025-03-18
Initial work for frontend data flow rearchitecture

jobelenus:
This outlines an initial structure for defining materialized views (atoms) that are generated on every change.

These atoms are kept in a KV store (`frigg`), along side an "index" of all of the atoms that exist for the current state of each change set.

The materialized view definition that generates each atom is able to specify that changes to a specific `EntityKind` should cause the atom to be (re)generated, and are also able to specify that it "references" other materialized view definitions. This "references" information is used to build a dependency graph to ensure that no materialized view that references another will be built before all of the atoms for the referenced materialized view are built.

If the frontend requests the index of atoms for a change set from SDF, and the change set is valid, but SDF/frigg is unable to find the index for it, SDF will build all materialized views so that it can provide the front end with the resulting index of atoms.

This also creates a new websocket (bifrost) that is used for broadcasting JSON Patch information about atom changes whenever materialized views are built, and there is a pre-existing index for that change set, in order to help minimize the amount of data going over the wire on every change for the front end to become aware the change.

A new pinia store (heimdall) controls a web worker that has access to wasm sqlite where we persist data. We have database tables and an update algorithm that keeps a single unique reference to an Atom based on its EntityKind, id, and, checksum. This prevents duplicative data. And we store the Atom's relationship to change sets through the snapshot address. We accept both JSON patch data and "cold start" (mjolnir) events over the bifrost web socket.

Heimdall exposes the bifrost querying interface to tan stack, which is used within individual components. When we write to an Atom in sqlite, we also fire a callback to Heimdall to bust the tan stack query cache, which re-fires the queryFn, and re-renders component data.

We prune data from sqlite as old data becomes obsolete from being patched, as well as change sets that are applied, abandoned or closed.


app/web/src/store/apis.ts
```diff
@@ -30,13 +19,11 @@ export const sdfApiInstance = Axios.create({
...
+export function injectBearerTokenAuth(config: InternalAxiosRequestConfig) {
   // inject auth token from the store as a custom header
-  const authStore = useAuthStore();
   config.headers = config.headers || {};
 
-  const token = authStore.selectedOrDefaultAuthToken;
+  const token = `eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VyX3BrIjoiMDFIUkZFVjBSTVdNSDVTR0JHREFSSDNHNDgiLCJ3b3Jrc3BhY2VfcGsiOiIwMUhSRkVWMFMyM1IxRzIzUlA3NVFRRENBNyIsImlhdCI6MTc0MTAxMjUxMiwic3ViIjoiMDFIUkZFVjBSTVdNSDVTR0JHREFSSDNHNDgifQ.Om96vxeFVbeTebsc5ECkbPvtzt3tg_6YeiDM6QZ9H5Kzp4-IWK0vKXbF6hPCbornp-8IuUZBUowPg6BzXgqsvePzqoud8x6S01kzO2AmDCHTPdSprfT1hF34KkBDBVnCAS-hbz5w7iwWoVyfX5goRkODgfWXhjxSgsmHmjEwoYFGC7dUbZvl7h2M3wyMcr8Ls6CTi3FAuOmPZ0ld1mKg3323C8pPEkCt4e15AZ0dHAyWxVu8MioJjpf7XB_pQt_GT3MTznGOtb7OWnV3jmhGyVKhgM95hycRImWmT39YZrnwroNkosZFFChsm8fLQDs9u1b2Ap227RWd1rXWtMXIREaH6Qp_mWQURbC0HJs8CtS8z05CvhtLjqbQQ1S6P1ybjYPMD-bToPeFzNJd2UbNHR1eGRF0Yl4OZfnOf0O97r2H5CrvtjDYlA8PKma1GcBlk24NStfnPoiZmfzhPNZuYuYL6E2DEwM6GXUajTPQmvHavujYwV1ziy9-H8CiVHpa4TBW4UxVDyBI85k4RBXqvYBRJb1Sf6bS09DVWXMS5u01BHIsX9GgFl3t0L_0BkElseOWQL8s-5ARBXVi1o-O7qZ1Eph7mRkuxbl1d4Asym1j8PbrsQFfw3waUL_UOrhvDcKDMkG2Cxmj1cZr6QJSK668AdFDkEWe4lx84TCkKx8`;
```
britmyerss, line 26: probably don't want this :)

app/web/src/store/feature_flags.store.ts
```diff
@@ -81,16 +82,19 @@ export function useFeatureFlagsStore() {
...
+            ([storeFlagKey, phFlag]) => {
+              this[storeFlagKey] = result.featureFlags[phFlag] ?? false;
+            },
+          );
+        }
 
         // You can override feature flags while working on a feature by setting them to true/false here
         // for example:
         // this.FEATURE_FLAG_NAME = false;
+        this.FRONTEND_ARCH_VIEWS = true;
```
britmyerss, line 97: or this one :)

lib/dal/src/materialized_view.rs
```diff
@@ -0,0 +1,351 @@
...
+        .await?;
+
+    let (frontend_objects, patches) = build_mv_inner(
+        ctx,
+        frigg,
+        ctx.workspace_pk()?,
+        ctx.change_set_id(),
+        &changes,
+    )
+    .instrument(tracing::info_span!("Building MVs"))
```
britmyerss, line 81: @jhelwig I don't think I've used  instrumentation this way before! What's the delta between having it here like this vs. instrumenting the `build_mv_inner`?

jhelwig: It's mostly useful if you want to have spans for only some usages of an async fn, or want it to have different log levels in different places. While the wrappers around `build_mv_inner` have instrumentation, it's at the `debug` level. I figured (at least at the beginning) we _definitely_ wanted to know how this was affecting the rebase process,  which is why these are at `info`.

lib/dal/src/data_cache.rs
```diff
@@ -0,0 +1,32 @@
...
+
+impl DataCache {
+    pub async fn publish_patch_batch(
+        ctx: &DalContext,
+        patch_batch: PatchBatch,
+    ) -> DataCacheResult<()> {
+        ctx.txns()
+            .await?
+            .nats()
+            .publish(patch_batch.publish_subject(), &patch_batch)
```
britmyerss, line 27: This doesn't do any batching of the patch-batches yet, right?

jhelwig: `patch_batch` itself is the batching. It contains all of the JSON Patches for every atom affected by the change.

lib/dal/src/materialized_view.rs
```diff
@@ -0,0 +1,351 @@
...
+        // instead of attempting to process the remaining MVs.
+        for mv_kind in &independent_mvs {
+            mv_dependency_graph.remove_id(*mv_kind);
+        }
+
+        for change in changes {
+            for &mv_kind in &independent_mvs {
+                match mv_kind {
+                    ReferenceKind::View => {
+                        if mv_kind != ReferenceKind::View {
```
britmyerss, line 242: when would this be true? `mv_kind != ReferenceKind::View` but we're matching on `mv_kind` above?

jhelwig: It shouldn't be. That's a hold-over from a previous iteration of that code, and should go away.

lib/rebaser-server/src/rebase.rs
```diff
@@ -144,6 +159,37 @@ pub async fn perform_rebase(
     );
     info!("rebase performed: {:?}", start.elapsed());
 
+    if frigg
```
britmyerss, line 162: Should we wrap this in a feature flag? either passed via startup config, or hitting posthog right here? (that way once this lands, if we see a performance hit in the short term we can 'turn it off')

lib/rebaser-server/src/rebase.rs
```diff
@@ -144,6 +159,37 @@ pub async fn perform_rebase(
...
+        .get_index(ctx.workspace_pk()?, ctx.change_set_id())
+        .await?
+        .is_some()
+    {
+        info!("Frigg index found, triggering rebuild of materialized views");
+        let changes = original_workspace_snapshot
+            .detect_changes(&to_rebase_workspace_snapshot)
```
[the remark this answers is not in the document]
jhelwig: That and `Change` isn't the same information as `Update`. `Change` is just the ID, kind, and merkle tree hash. For rebuilding the MVs/atoms we don't need the full information in an `Update` (`Change` is a lot faster to calculate.), and as you said, we want to operate on what it looks like after the corrected operational transforms have been applied.

app/web/src/store/apis.ts
```diff
@@ -1,22 +1,11 @@
...
-import UnscheduledDowntime from "@/components/toasts/UnscheduledDowntime.vue";
+import Axios, { InternalAxiosRequestConfig } from "axios";
 
 // api base url - can use a proxy or set a full url
 let apiUrl: string;
-if (import.meta.env.VITE_API_PROXY_PATH)
-  apiUrl = `${window.location.origin}${import.meta.env.VITE_API_PROXY_PATH}`;
-else if (import.meta.env.VITE_API_URL) apiUrl = import.meta.env.VITE_API_URL;
+if (import.meta.env.VITE_API_PROXY_PATH) {
+  apiUrl = `http://localhost:8080${import.meta.env.VITE_API_PROXY_PATH}`;
```
britmyerss, line 7: probably don't want localhost here either?

lib/dal/src/data_cache.rs
britmyerss: Ok got it. Just thinking this *could* eventually be too big for a single nats message but it's fine for now.

lib/dal/src/materialized_view.rs
britmyerss: Brilliant! Stashing this one away for the future :)

britmyerss (approved):
🔥 GREAT WORK!

fnichol (dismissed):

stack72 (dismissed):

stack72 (approved):

PR #5699, opened 2025-03-18
Allow firecracker creation skip and fix config mismatch in veritech

nickgerace:
This PR allows firecracker setup script creation to be skipped. This is useful for a variety of reasons: getting a new rootfs, changing the script itself, and more.

This PR also fixes a config mismatch in veritech where pinga's concurrency limit will be derived for all global merged configurations. Until we fix the "pool exhausted == NACK counting towards max failed message count" problem, we cannot afford the concurrency limit to be higher than the pool size. Now, we will be able to individual customize.

Why not pursue a dynamic solution where the concurrency limit is equal to the pool size by default? A bit of magic lies in there as well as a potential side effect to the other runtime strategies (local process and docker). We are really only trying to target the use of concurrency limit in production settings using firecracker, so renaming the variable will suffice.

### DIFF
--- a/bin/veritech/src/args.rs
+++ b/bin/veritech/src/args.rs
@@ -98,6 +98,10 @@ pub(crate) struct Args {
     #[arg(long)]
     pub(crate) cyclone_pool_size: Option<u32>,
 
+    /// Cyclone create firecracker setup scripts
+    #[arg(long)]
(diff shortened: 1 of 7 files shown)


fnichol (dismissed):
All right!

fnichol (approved):

PR #5701, opened 2025-03-18
feat: si-split-graph

zacharyhamm:
A new crate, lib/si-split-graph, containing an initial, generic implementation of a merkle tree hashed graph which can be split into multiple "subgraphs" but allows edges between nodes in different subgraphs. Calculates updates, performs them, and handles the "ordering" concept internally.

This is not yet integrated with the si "dal", and will likely require more changes before that is possible.

### DIFF
--- a/lib/si-split-graph/src/lib.rs
+++ b/lib/si-split-graph/src/lib.rs
@@ -179,12 +179,11 @@ where
 
     pub fn node_hash(&self) -> ContentHash {
         let mut hasher = ContentHash::hasher();
-        if let SplitGraphNodeWeight::Custom(n) = self {
-            return n.node_hash();
-        }
 
         match self {
-            SplitGraphNodeWeight::Custom(_) => {}
+            SplitGraphNodeWeight::Custom(c) => {
+                return c.node_hash();
+            }
             SplitGraphNodeWeight::ExternalTarget {
                 id,
                 subgraph,
--- a/lib/si-split-graph/src/lib.rs
+++ b/lib/si-split-graph/src/lib.rs
@@ -1,10 +1,14 @@
-use std::collections::HashSet;
+use std::collections::{HashSet, VecDeque};
 
 use async_trait::async_trait;
 use opt_zip::OptZip;
 use petgraph::{prelude::*, stable_graph};
 use serde::{Deserialize, Serialize};
-use si_events::{merkle_tree_hash::MerkleTreeHash, ContentHash};
+use si_events::{
+    merkle_tree_hash::MerkleTreeHash,
+    workspace_snapshot::{Change, EntityKind},
+    ContentHash,
+};
 use si_id::ulid::Ulid;
 use thiserror::Error;
 
@@ -124,6 +128,16 @@ where
         match self {
             SplitGraphNodeWeight::Custom(n) => n.lineage_id(),
             other => other.id(),
+        }
+    }
+
+    pub fn entity_kind(&self) -> EntityKind {
+        match self {
+            SplitGraphNodeWeight::Custom(c) => c.entity_kind(),
+            SplitGraphNodeWeight::ExternalTarget { .. } => EntityKind::ExternalTarget,
+            SplitGraphNodeWeight::Ordering { .. } => EntityKind::Ordering,
+            SplitGraphNodeWeight::GraphRoot { .. } => EntityKind::Root,
+            SplitGraphNodeWeight::SubGraphRoot { .. } => EntityKind::SubGraphRoot,
         }
     }
 
@@ -336,6 +350,7 @@ pub trait EdgeKind: std::hash::Hash + PartialEq + Eq + Copy + Clone + std::fmt::
 pub trait CustomNodeWeight: PartialEq + Eq + Clone + std::fmt::Debug {
     fn id(&self) -> SplitGraphNodeId;
     fn lineage_id(&self) -> SplitGraphNodeId;
+    fn entity_kind(&self) -> EntityKind;
 
     fn set_merkle_tree_hash(&mut self, hash: MerkleTreeHash);
     fn merkle_tree_hash(&self) -> MerkleTreeHash;
@@ -807,7 +822,7 @@ where
         let subgraph = self.subgraphs.get(split_graph_index.subgraph as usize)?;
 
         subgraph
-            .ordered_children_for_node(split_graph_index.index)
+            .ordered_children(split_graph_index.index)
             .map(|node_indexes| {
                 node_indexes
                     .into_iter()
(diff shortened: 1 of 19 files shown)


zacharyhamm:
Want to merge this and get eyes on it before I begin "integration" work.

One thing to note is that the public interface here will never expose NodeIndexes, but only node ids. Will have to rework our dal code to do the same before integration

lib/si-split-graph/src/lib.rs
```diff
@@ -0,0 +1,1178 @@
...
+        id: SplitGraphNodeId,
+        order: Vec<SplitGraphNodeId>,
+        merkle_tree_hash: MerkleTreeHash,
+    },
+    /// The root node for the entire graph
+    GraphRoot {
+        id: SplitGraphNodeId,
+        merkle_tree_hash: MerkleTreeHash,
+    },
+    /// A ordering node, which cont
```
jhelwig, line 102: Stale comment fragment?

lib/si-split-graph/src/lib.rs
```diff
@@ -0,0 +1,1178 @@
...
+        self.subgraphs
+            .iter_mut()
+            .for_each(|subgraph| subgraph.recalculate_merkle_tree_hash_based_on_touched_nodes());
+    }
+
+    pub fn recalculate_entire_merkle_tree_hashes(&mut self) {
+        self.subgraphs
+            .iter_mut()
+            .for_each(|subgraph| subgraph.recalculate_entire_merkle_tree_hash());
+    }
```
jhelwig, line 495: These don't appear to propagate the merkle tree hash across subgraphs? The subgraph recalculations appear to do a dfs post-order, but cross subgraph edges don't appear to affect the other subgraph?

zacharyhamm: correct. The merkle tree hashes are restricted to each subgraph, and updates apply only to the subgraph. If a child of a parent in another subgraph changes, it doesn't matter. The merkle tree for the subgraph that changed will be updated, and we will apply the update to that hash. The change will be propagated up to the "supergraph" because the hash for that subgraph will change, so we will get a new graph in the end.

lib/si-split-graph/src/lib.rs
zacharyhamm: oops

zacharyhamm:
@jhelwig @nickgerace I've taken a stab at implementing `detect_changes` here, supporting detection across subgraphs without having to merge the merkle tree hash

lib/si-split-graph/src/updates.rs
```diff
@@ -0,0 +1,499 @@
...
+    pub fn detect_changes(&self) -> Vec<Change> {
+        let mut changes = vec![];
+
+        petgraph::visit::depth_first_search(
+            &self.updated_graph.graph,
+            Some(self.updated_graph.root_index),
+            |event| {
+                if let DfsEvent::Discover(updated_graph_index, _) = event {
+                    let Some(updated_node_weight) =
+                        self.updated_graph.graph.node_weight(updated_graph_index)
```
zacharyhamm, line 127: This should be a safe unwrap, really

lib/si-split-graph/src/lib.rs
```diff
@@ -0,0 +1,1343 @@
...
+mod opt_zip;
+pub mod subgraph;
+pub mod subgraph_address;
+pub mod updates;
+
+use subgraph::{SubGraph, SubGraphEdgeIndex, SubGraphNodeIndex};
+pub use subgraph_address::SubGraphAddress;
+use updates::Update;
+
+pub const MAX_NODES: usize = ((u16::MAX / 2) - 1) as usize;
```
nickgerace, line 24: How did we get to this value?

zacharyhamm: arbitrary. however i think we're going to move towards providing a closure to the graph that decides when to split

lib/si-split-graph/src/lib.rs
```diff
@@ -0,0 +1,1343 @@
...
+        self.subgraphs.push(subgraph);
+
+        subgraph_index
+    }
+
+    fn new_empty_subgraph(&mut self) -> u16 {
+        self.supergraph.addresses.push(SubGraphAddress::nil());
+
+        let subgraph = SubGraph::new();
+        let subgraph_index = self.subgraphs.len() as u16;
```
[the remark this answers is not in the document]
zacharyhamm: i've been thinking about removing subgraphs. will have a pr about that later. otherwise node indexes are always stable so long as we never make it possible to move a node to another subgraph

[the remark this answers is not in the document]
zacharyhamm: i've been thinking about removing subgraphs. will have a pr about that later. otherwise node indexes are always stable so long as we never make it possible to move a node to another subgraph

lib/si-split-graph/src/lib.rs
```diff
@@ -0,0 +1,1343 @@
...
+            SplitGraphNodeWeight::Custom(_) => {}
+            SplitGraphNodeWeight::ExternalTarget {
+                id,
+                subgraph,
+                target,
+                ..
+            } => {
+                hasher.update(&id.inner().to_bytes());
+                hasher.update(&subgraph.to_le_bytes());
+                hasher.update(&target.inner().to_bytes());
```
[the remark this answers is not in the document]
zacharyhamm: not sure i understand what you mean by compressing it

lib/si-split-graph/src/lib.rs
```diff
@@ -0,0 +1,1343 @@
...
+                }
+            }
+        }
+    }
+
+    pub fn tiny_dot_to_file(&self, prefix: &str) {
+        for (idx, subgraph) in self.subgraphs.iter().enumerate() {
+            subgraph.tiny_dot_to_file(&format!("{prefix}-subgraph-{}", idx + 1));
+        }
+    }
```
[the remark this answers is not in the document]
zacharyhamm: it was definitely useful when building the split graph, probably less useful once things work

lib/si-split-graph/src/lib.rs
```diff
@@ -0,0 +1,1343 @@
...
+{
+    type Item = SplitGraphEdgeReference<'a, E, K>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.edges.next().and_then(|edge_ref| match self.direction {
+            Outgoing => {
+                if matches!(edge_ref.weight(), SplitGraphEdgeWeight::Ordering) {
+                    // Ordering nodes are hidden
+                    self.next()
+                } else {
```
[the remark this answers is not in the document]
zacharyhamm: It's just an implementation decision here, that ordering nodes are an internal part of the graph, and not something we need to worry about when traversing the graph in practice

nickgerace (approved):

PR #5709, opened 2025-03-19
chore: remove NodeIndex from all public WorkspaceSnapshot methods

zacharyhamm:
In order to prepare for the split-graph to integrate with the DAL, we need to remove all of the public methods on WorkspaceSnapshot that make use of the petgraph NodeIndex type, since the SplitGraph will not expose the NodeIndex type externally.  This is also a quality of life refactor, which simplifies some "warty" usage patterns (where we were converting between Ulid and NodeIndex, book-keeping work that the snapshot is capable of doing internally).

### DIFF
--- a/lib/dal/src/func.rs
+++ b/lib/dal/src/func.rs
@@ -21,7 +21,6 @@ use crate::func::argument::FuncArgumentId;
 use crate::func::intrinsics::IntrinsicFunc;
 use crate::layer_db_types::{FuncContent, FuncContentV2};
 use crate::workspace_snapshot::edge_weight::{EdgeWeightKind, EdgeWeightKindDiscriminants};
-use crate::workspace_snapshot::graph::WorkspaceSnapshotGraphError;
 use crate::workspace_snapshot::node_weight::category_node_weight::CategoryNodeKind;
 use crate::workspace_snapshot::node_weight::{FuncNodeWeight, NodeWeight, NodeWeightError};
 use crate::workspace_snapshot::WorkspaceSnapshotError;
@@ -468,15 +467,10 @@ impl Func {
         func_id: FuncId,
     ) -> FuncResult<Option<(FuncNodeWeight, ContentHash)>> {
         let workspace_snapshot = ctx.workspace_snapshot()?;
-        let id: Ulid = func_id.into();
-        let node_index = match workspace_snapshot.get_node_index_by_id(id).await {
-            Ok(node_index) => node_index,
-            Err(WorkspaceSnapshotError::WorkspaceSnapshotGraph(
-                WorkspaceSnapshotGraphError::NodeWithIdNotFound(_),
-            )) => return Ok(None),
-            Err(err) => return Err(err.into()),
+
+        let Some(node_weight) = workspace_snapshot.get_node_weight_opt(func_id).await else {
+            return Ok(None);
         };
-        let node_weight = workspace_snapshot.get_node_weight(node_index).await?;
 
         let hash = node_weight.content_hash();
         let func_node_weight = node_weight.get_func_node_weight()?;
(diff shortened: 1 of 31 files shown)


lib/dal/src/workspace_snapshot.rs
```diff
@@ -843,39 +855,12 @@ impl WorkspaceSnapshot {
...
-            .working_copy()
-            .await
-            .find_equivalent_node(id, lineage_id)?)
+    pub async fn get_node_weight_opt(&self, id: impl Into<Ulid>) -> Option<NodeWeight> {
+        let working_copy = self.working_copy().await;
+        working_copy
+            .get_node_index_by_id_opt(id)
+            .and_then(|node_index| working_copy.get_node_weight_opt(node_index))
+            .cloned()
     }
```
zacharyhamm, line 864: This method is never used (the graph version is used internally by the internal graph, but it's never used by external consumers of the snapshot)

lib/dal/src/func.rs
```diff
@@ -468,15 +467,10 @@ impl Func {
...
         let workspace_snapshot = ctx.workspace_snapshot()?;
-        let id: Ulid = func_id.into();
-        let node_index = match workspace_snapshot.get_node_index_by_id(id).await {
-            Ok(node_index) => node_index,
-            Err(WorkspaceSnapshotError::WorkspaceSnapshotGraph(
-                WorkspaceSnapshotGraphError::NodeWithIdNotFound(_),
-            )) => return Ok(None),
-            Err(err) => return Err(err.into()),
+
+        let Some(node_weight) = workspace_snapshot.get_node_weight_opt(func_id).await else {
```
jkeiser, line 471: Very nice :)

lib/dal/src/prop.rs
```diff
@@ -45,8 +46,8 @@ pub enum PropError {
     AttributePrototypeArgument(#[from] AttributePrototypeArgumentError),
     #[error("change set error: {0}")]
     ChangeSet(#[from] ChangeSetError),
-    #[error("child prop of {0:?} not found by name: {1}")]
-    ChildPropNotFoundByName(NodeIndex, String),
+    #[error("child prop of {0} not found by name: {1}")]
+    ChildPropNotFoundByName(Ulid, String),
```
jkeiser, line 50: Should this be PropId? Or is it possible for these not to be props?

lib/dal/src/prop.rs
```diff
@@ -751,31 +748,28 @@ impl Prop {
             .ok_or(PropError::MapOrArrayMissingElementProp(prop_id))
     }
 
-    pub async fn find_child_prop_index_by_name(
+    pub async fn find_child_prop_id_by_name(
         ctx: &DalContext,
-        node_index: NodeIndex,
+        parent_node_id: Ulid,
```
jkeiser, line 753: Same question: seems like this should be PropId

jkeiser, line 753: I had some comments earlier that this should be PropId, but reading further into the code I see it can be a SchemaVariantId as well. Oy!

lib/dal/src/schema/variant.rs
```diff
@@ -1136,16 +1121,15 @@ impl SchemaVariant {
...
     ) -> SchemaVariantResult<PropNodeWeight> {
         let workspace_snapshot = ctx.workspace_snapshot()?;
-        let edge_targets: Vec<NodeIndex> = workspace_snapshot
+        let edge_targets: Vec<_> = workspace_snapshot
             .edges_directed(schema_variant_id, Direction::Outgoing)
             .await?
             .into_iter()
-            .map(|(_, _, target_idx)| target_idx)
+            .map(|(_, _, target_id)| target_id)
             .collect();
```
jkeiser, line 1129: Just noticing while I'm in here: we don't need to do a collect() here (and shouldn't). Not something you need to fix.

jkeiser (approved):
This is an awesome simplification and makes a bunch of callers easier to read! Thank you for doing this.

PR #5712, opened 2025-03-19
perf(sdf): Make funcs endpoint faster

jkeiser:
This PR speeds up the /funcs endpoint by 4x on my 33-asset workspace locally (from 2.4s -> 0.6s warm, and 5.6s -> 2.9s cold). It does this by speeding up `prop.ts_type()`, which is frequently called. The changes, and their rough impact percentage-wise (the above numbers are for main, and the 17s numbers below relate to my other PR, which had slowed things down by calling ts_type() even more):

* 17s -> 4.7s: Eliminating CAS loads by traversing the graph using `PropId` instead of loading a full `Prop` for each prop in the tree brought a 3-4x improvement.
* 4.7s -> 2.7s: Eliminating #[async_recursion] and await boundaries in the prop tree by holding a lock (SnapshotReadGuard) and traversing the prop tree synchronously bought us another 1.5-2x improvement.
* 2.7s -> 2.6s: Decreasing allocation by building up the type in a String buffer (as opposed to building and concatenating lots of little Strings) bought us a consistent 4-5% or so.
* 2.6s: Moving from ulid/PropId to NodeIndex has negligible performance impact; this PR does it simply because our synchronous graph interface runs largely in terms of NodeIndexes.

One minor change is, it also turns `prop.ts_type()` -> `Prop::ts_type(PropId)`, as it turns out callers only load the root prop in order to call ts_type().

## Testing

* [X] Existing tests: this is a perf-only change, designed to yield the same results as before
* [X] Manual tests: make sure TS types show correctly in editor. Pay particular attention to resource.status and resource.payload, since they are special cases.

### DIFF
--- a/lib/dal/src/func/binding/action.rs
+++ b/lib/dal/src/func/binding/action.rs
@@ -152,7 +152,7 @@ impl ActionBinding {
                 .map_err(|_| {
                     SchemaVariantError::PropNotFoundAtPath(variant_id, path.to_string())
                 })?;
-            ts_types.push(prop.ts_type(ctx).await?)
+            ts_types.push(Prop::ts_type(ctx, prop.id()).await?)
         }
         Ok(format!(
             "type Input {{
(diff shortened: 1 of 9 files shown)


zacharyhamm (commented):
This is a huge improvement! Everything looks good to me but the pr reviewer is unhappy.

zacharyhamm (dismissed):

zacharyhamm (approved):

PR #5716, opened 2025-03-19
Remove force reconnect from heartbeat app

nickgerace:
## Description

This PR removes force reconnect logic and metrics from the veritech heartbeat app. This was useful during earlier verideath investigation efforts, but having a side effect force reconnection in the heartbeat app is unnecessary nor a great idea in a long term.


fnichol (approved):
Nice and surgical!

PR #5724, opened 2025-03-20
bug: fixing the deno lsp with the si-fs

adamhjk:
Removes the typescript config that triggers the native typescript LSP, in favor of only supporting Deno. The TypeScript LSP will always be wrong-er than the Deno LSP.

It also relaxes some settings to reflect the values we have in the browser (which does not use the Deno LSP, so now it will be wrong sometimes, but.. that is a problem for another PR).


zacharyhamm (approved):
Works with vscode, 🚀

PR #5725, opened 2025-03-20
fix(dal): Calculate Action Dependencies using DFS to capture transitive dependencies between components

britmyerss:
Before this change, we were calculating the order for actions based only on the components with actions enqueued and their direct dependencies. This meant if you had Component A (with action) ---edge to --> Component B (no action) -- edge to --> Component C (with action), we would not detect that Action C should depend on Action A.  

With this change we now build a more full tree of component dependencies, and use petgraph's depth_first_search to find the closest dependency for a given component's action, which allows us to capture these transitive dependencies. 

TESTING
I added an integration test covering both forwards and backwards transitive dependencies (forwards applies to create/update/manual actions, backwards applies to destroy actions)

<div><br/>via <a href="https://giphy.com/heyarnold/">Hey Arnold</a> on <a href="https://giphy.com/gifs/heyarnold-nickelodeon-hey-arnold-xUNd9WbEMv0CGDEK3u">GIPHY</a></div>

### DIFF
--- a/lib/dal/src/action/dependency_graph.rs
+++ b/lib/dal/src/action/dependency_graph.rs
@@ -1,5 +1,8 @@
 use itertools::Itertools;
-use petgraph::prelude::*;
+use petgraph::{
+    prelude::*,
+    visit::{Control, DfsEvent},
+};
 use std::collections::{HashMap, HashSet, VecDeque};
 use telemetry::prelude::*;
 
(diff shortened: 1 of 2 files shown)


lib/dal/src/action/dependency_graph.rs
```diff
@@ -122,11 +129,19 @@ impl ActionDependencyGraph {
...
                     component_dependencies.update_edge(source_component_index, component_index, ());
                 }
+                new_components_to_process.insert(inferred_connection.source_component_id);
+            }
+            // only add new components to the queue if they are not already in the queue
+            for component_id in new_components_to_process {
+                if !components_to_process.contains(&component_id) {
+                    components_to_process.push_back(component_id);
+                }
             }
```
jhelwig, line 139: Rather than checking if it's currently in the queue, it would probably be better to keep track of which ones have already been "Seen" (already visited, and when adding to the queue), and only add if it isn't in the seen set. Only checking if it's currently in the queue could potentially end up in an infinite loop if there's a cycle.

jhelwig (approved):

PR #5729, opened 2025-03-20
chore(firecracker): re-enable vsock telemetry forwarding

fnichol:
This change re-enables the stream forwarder in Cyclone and the listener in Veritech, both of which help are used to forward OpenTelemetry data out of a Firecracker VM.

Prior to disabling these tasks, it was unclear whether the spawned tokio tasks would naturally terminate on their own, or whether they would continue to run even as more Firecracker instances were spawing in Pool Noodle's pool. This change attempts to tie the lifetime of these forwarding tasks to the lifetime of the Firecracker instance (that is the termination of the VM should also trigger the shutdown of the forwarding task).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6975,11 +6975,11 @@ dependencies = [
  "krata-loopdev",
  "nix 0.29.0",
  "remain",
+ "telemetry",
  "thiserror 2.0.12",
  "tokio",
  "tokio-util",
  "tokio-vsock",
- "tracing",
 ]
 
 [[package]]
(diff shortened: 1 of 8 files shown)


fnichol:
This may need a careful load test as it re-enables listener sockets on Veritech. Should be fine now, but worthy of a confirmation under load.

Note: I'm not 100% convinced that `copy_bidirectional` will always do the right thing but am hedging this with a cancel-able task now (i.e. that task can be pre-empted and cancelled).

nickgerace:
@fnichol would you like to keep this PR around? No rush, just curious

PR #5731, opened 2025-03-21
chore: Ulid -> SchemaId

jkeiser:
Modified Ulid -> SchemaId / ModuleId in a few places and then fixed the knock-on effects on types (lots fewer into()s). Also used `.parse()` instead of ::from_string() for type system happiness. Separating this out for reviewer joy!

### Testing

Relying on the integration tests in this PR to catch any issues; changes are rote.

### DIFF
--- a/bin/hoist/src/main.rs
+++ b/bin/hoist/src/main.rs
@@ -10,7 +10,6 @@ use std::path::{Path, PathBuf};
 use std::sync::atomic::{AtomicU64, AtomicUsize, Ordering};
 use std::sync::Arc;
 use std::time::Duration;
-use ulid::Ulid;
 
 use crate::diff::patch_list_to_summary;
 use clap::Parser;
@@ -125,7 +124,7 @@ async fn detailed_diff_with_module_index(
         }
     };
 
-    let remote_ulid = Ulid::from_string(&remote_module_id)?;
+    let remote_ulid = remote_module_id.parse()?;
 
     let module_bytes = client.download_module(remote_ulid).await?;
 
@@ -300,7 +299,7 @@ async fn diff_summaries_with_module_index(
                     }
                 };
 
-                let remote_ulid = Ulid::from_string(&remote_module_id).unwrap();
+                let remote_ulid = remote_module_id.parse().unwrap();
 
                 let module_bytes = client.download_module(remote_ulid).await.unwrap();
                 let current_pkg = SiPkg::load_from_bytes(&module_bytes)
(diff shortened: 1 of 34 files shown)


lib/dal/src/cached_module.rs
```diff
@@ -246,14 +245,10 @@ impl CachedModule {
 
                 let client = module_index_client.clone();
                 join_set.spawn(async move {
-                    let module_id = module.id.to_owned();
+                    let module_id = Ulid::from_string(&module.id).unwrap_or_default().into();
```
jkeiser, line 248: I left Ulid::from_string() here because we use unwrap_or_default(), which rightfully does not exist on ModuleId. We arguably should not be doing this, but I really don't want to change behavior in this PR!

lib/sdf-server/src/service/module/install_module.rs
```diff
@@ -147,7 +144,7 @@ pub async fn install_module(
             }
             variants.push(front_end_variant);
         } else {
-            return Err(ModuleError::SchemaNotFoundFromInstall(id));
+            return Err(ModuleError::NoSchemaVariantsInModule(id));
```
jkeiser, line 147: Changed this type because it's used for two different purposes (one for schemas and one for modules).

lib/module-index-client/src/lib.rs
```diff
@@ -232,11 +233,11 @@ impl ModuleIndexClient {
         Ok(upsert_response.json::<bool>().await?)
     }
 
-    pub async fn download_module(&self, module_id: Ulid) -> ModuleIndexClientResult<Vec<u8>> {
+    pub async fn download_module(&self, module_id: ModuleId) -> ModuleIndexClientResult<Vec<u8>> {
```
fnichol, line 236: This does seem to be a nice payoff--get some specific type safety around what kind of ID is expected!

fnichol (approved):
Yeah, that is looking nicer, well done!

bin/hoist/src/main.rs
```diff
@@ -125,7 +124,7 @@ async fn detailed_diff_with_module_index(
         }
     };
 
-    let remote_ulid = Ulid::from_string(&remote_module_id)?;
+    let remote_ulid = remote_module_id.parse()?;
```
nickgerace, line 127: Nice

lib/dal-test/src/test_exclusive_schemas/mod.rs
```diff
@@ -87,153 +86,31 @@ pub const SCHEMA_ID_FAKE_BUTANE: &str = "01JARH2BTA5DK4J9Q4Q0XH46SR";
 // allow expect here for the Ulid conversion. These will never panic.
 #[allow(clippy::expect_used)]
 pub(crate) async fn migrate(ctx: &DalContext) -> BuiltinsResult<()> {
-    migrate_test_exclusive_schema_starfield(
-        ctx,
-        ulid::Ulid::from_str(SCHEMA_ID_STARFIELD)
-            .expect("should convert")
-            .into(),
```
nickgerace, line 94: Great!

lib/dal/src/cached_module.rs
nickgerace: Good call

lib/sdf-server/src/service/module/install_module.rs
nickgerace: Great job! That would've been confusing to get otherwise

nickgerace (approved):
🙌

PR #5741, opened 2025-03-22
feat(sdf): Include socket count for uninstalled modules in variantSocketMap

jkeiser:
Now that management functions can create components for any known asset (including ones that are not yet installed), we need to send the socket count for uninstalled assets in `variantSocketMap`. This PR:

* Caches socket count in sdf cache for uninstalled assets
  - automatically cached when downloading new modules/versions from the module index
  - automatically added when contributing a private module
  - automatically fills in missing socket counts for existing modules when sdf first starts up after upgrade
* Includes socket count for both installed and uninstalled assets in variantSocketMap

**Guide:** this PR has comments throughout pointing out the salient bits.

## Testing

- [X] Test initial sdf migration to make sure new assets include socket count
- [x] Check that socket count is accurate for uninstalled assets with and without management functions
- [x] Test sdf migration from main to branch to ensure socket count is filled in on existing assets
- [x] Test contributing private modules to make sure socket count carries through
- [x] Test sdf "Update module cache" button in Admin Panel to make sure new assets include socket count

### DIFF
--- a/lib/dal/src/cached_module.rs
+++ b/lib/dal/src/cached_module.rs
@@ -1,7 +1,7 @@
 use std::{
     collections::{HashMap, HashSet},
-    str::FromStr,
     sync::Arc,
+    time::Duration,
 };
 
 use chrono::{DateTime, Utc};
@@ -20,7 +20,7 @@ use crate::{
 };
 use module_index_client::{ModuleDetailsResponse, ModuleIndexClient, ModuleIndexClientError};
 use si_data_pg::{PgError, PgRow};
-use si_pkg::{SiPkg, SiPkgError};
+use si_pkg::{SiPkg, SiPkgError, SiPkgSchemaData, SiPkgSchemaVariantData};
 
 pub use si_id::CachedModuleId;
 
@@ -31,6 +31,8 @@ const PLACEHOLDER_OWNER_USER_ID: &str = "-";
 pub enum CachedModuleError {
     #[error("join error: {0}")]
     Join(#[from] tokio::task::JoinError),
+    #[error("json error: {0}")]
+    Json(#[from] serde_json::Error),
     #[error("module index client error: {0}")]
     ModuleIndexClient(#[from] ModuleIndexClientError),
     #[error("No module index url set on the services context")]
@@ -67,10 +69,17 @@ pub struct CachedModule {
     pub color: Option<String>,
     pub description: Option<String>,
     pub component_type: ComponentType,
+    pub package_summary: Option<PackageSummary>,
     pub latest_hash: String,
     pub created_at: DateTime<Utc>,
     pub package_data: Option<Vec<u8>>,
     pub scoped_to_user_pk: Option<UserPk>,
+}
+
+#[derive(Serialize, Deserialize, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct PackageSummary {
+    pub socket_count: u32,
 }
 
 impl From<CachedModule> for si_frontend_types::UninstalledVariant {
@@ -112,7 +121,9 @@ impl TryFrom<PgRow> for CachedModule {
 
     fn try_from(row: PgRow) -> Result<Self, Self::Error> {
         let component_type_string: String = row.try_get("component_type")?;
-        let component_type = ComponentType::from_str(&component_type_string)?;
+        let component_type = component_type_string.parse()?;
+        let package_summary: Option<serde_json::Value> = row.try_get("package_summary")?;
+        let package_summary = package_summary.map(serde_json::from_value).transpose()?;
 
         Ok(Self {
             id: row.try_get("id")?,
@@ -124,6 +135,7 @@ impl TryFrom<PgRow> for CachedModule {
             color: row.try_get("color")?,
             description: row.try_get("description")?,
             component_type,
+            package_summary,
             latest_hash: row.try_get("latest_hash")?,
             created_at: row.try_get("created_at")?,
             package_data: row.try_get("package_data")?,
@@ -131,6 +143,43 @@ impl TryFrom<PgRow> for CachedModule {
         })
     }
 }
+
+const CACHED_MODULE_GET_FIELDS: &str = "
+    id,
+    schema_id,
+    schema_name,
+    display_name,
+    category,
+    link,
+    color,
+    description,
+    component_type,
+    latest_hash,
+    created_at,
+    package_data,
+    scoped_to_user_pk,
+    package_summary
+";
+
+const CACHED_MODULE_LIST_FIELDS: &str = "
+    id,
+    schema_id,
+    schema_name,
+    display_name,
+    category,
+    link,
+    color,
+    description,
+    component_type,
+    latest_hash,
+    created_at,
+    NULL::bytea AS package_data,
+    scoped_to_user_pk,
+    package_summary
+";
+
+const BATCH_SIZE: usize = 10;
+const WAIT_BETWEEN_BATCHES: Duration = Duration::from_millis(100);
 
 impl CachedModule {
     pub async fn si_pkg(&mut self, ctx: &DalContext) -> CachedModuleResult<SiPkg> {
@@ -183,14 +232,10 @@ impl CachedModule {
         );
 
         let rows = ctx.txns().await?.pg().query(&query, &params).await?;
-
-        let mut result = vec![];
-
-        for row in rows {
-            result.push(row.try_get("hash")?);
-        }
-
-        Ok(result)
+        Ok(rows
+            .into_iter()
+            .map(|row| row.try_get("hash"))
+            .try_collect()?)
     }
 
     pub async fn create_private_module(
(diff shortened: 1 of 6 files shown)


lib/dal/src/cached_module.rs
```diff
@@ -132,6 +144,43 @@ impl TryFrom<PgRow> for CachedModule {
     }
 }
 
+const CACHED_MODULE_GET_FIELDS: &str = "
```
jkeiser, line 147: There are an increasing number of places we reference the list of fields, so I centralized them in a const here.

lib/dal/src/cached_module.rs
```diff
@@ -183,23 +232,21 @@ impl CachedModule {
...
     pub async fn create_private_module(
         ctx: &DalContext,
         module_details: ModuleDetailsResponse,
         payload: Vec<u8>,
     ) -> CachedModuleResult<Option<Self>> {
+        let user_pk: UserPk = module_details.owner_user_id.parse()?;
+        dbg!(&user_pk);
         let maybe_module =
-            Self::insert_private_module(ctx, &module_details, Arc::new(payload)).await?;
+            Self::insert(ctx, &module_details, Arc::new(payload), Some(user_pk)).await?;
```
jkeiser, line 249: Insert is now shared, and takes an optional user_pk

lib/dal/src/cached_module.rs
```diff
@@ -230,12 +277,24 @@ impl CachedModule {
         let ctx_clone = ctx.clone();
         ctx_clone.commit_no_rebase().await?;
 
-        let hashes: Vec<_> = modules.keys().map(ToOwned::to_owned).collect();
+        let new_modules = Self::cache_modules(ctx, &modules, module_index_client).await?;
```
jkeiser, line 280: Moved cache_modules to its own method for clarity instead of dumping my new stuff into the method along with it.

lib/dal/src/cached_module.rs
```diff
@@ -230,12 +277,24 @@ impl CachedModule {
...
+
+        // Now check and fix up any missing package summaries
+        Self::update_missing_package_summaries(ctx).await?;
```
jkeiser, line 283: This is the new.

lib/dal/src/cached_module.rs
```diff
@@ -288,271 +349,171 @@ impl CachedModule {
...
-            .iter()
-            .map(|lm| lm.schema_id)
-            .collect();
 
         // Look at all schema IDs in the cache and determine if any of them are no longer builtins.
         // If they aren't, ALL modules corresponding to them get remove.
-        for schema_id in cached_schema_ids {
-            if !builtin_schema_ids.contains(&schema_id) {
-                CachedModule::remove_all_for_schema_id(ctx, schema_id).await?;
+        for lm in CachedModule::latest_user_independent_modules(ctx).await? {
```
jkeiser, line 355: Same as before, just one less collect() and fewer words

lib/dal/src/cached_module.rs
```diff
@@ -288,271 +349,171 @@ impl CachedModule {
...
-              display_name,
-              category,
-              link,
-              color,
-              description,
-              component_type,
-              latest_hash,
-              created_at,
-              NULL::bytea AS package_data,
-              NULL::ident AS scoped_to_user_pk
```
jkeiser, line 515: No longer yielding null here! No one seems to be using the result, but it seems likely you want scoped_to_user_pk to be set to the actual user_pk it was scoped to.

lib/dal/src/cached_module.rs
```diff
@@ -288,271 +349,171 @@ impl CachedModule {
...
-            warn!("builtin module {} has no schema", module_details.id);
-            return Ok(None);
-        };
-
-        let Some(pkg_variant) = pkg_schema.variants()?.first().cloned() else {
-            warn!(
-                "builtin module {} has a schema with no variant",
-                module_details.id
-            );
+        let Some(package) = PackageData::load(&module_details.id, &pkg_bytes).await? else {
```
jkeiser, line 501: This encompasses the "load the package data in slow_rt," the "deserialize and get the name/color/description/etc." It now also includes "get the JSON PackageSummary (with the socket count).

lib/dal/src/cached_module.rs
```diff
@@ -288,271 +349,171 @@ impl CachedModule {
...
 
-        Ok(result)
+        let rows = ctx.txns().await?.pg().query(&query, &[&user_pk]).await?;
+        rows.into_iter().map(TryInto::try_into).try_collect()
     }
 
-    // TO-DO: Paul
-    // I will refactor this + insert together - this is to temporarily get something working
-    async fn insert_private_module(
+    async fn insert(
```
jkeiser, line 465: Literally everything except scoped_to_user_pk was shared between this and insert_private_module, so it's down to a single method that gets passed an optional scoped_to_user_pk.

lib/dal/src/cached_module.rs
```diff
@@ -561,143 +522,199 @@ impl CachedModule {
...
-                    &user_pk,
+                    &scoped_to_user_pk,
                 ],
             )
             .await?;
 
         Ok(Some(row.try_into()?))
     }
 
-    async fn insert(
```
jkeiser, line 585: This method is merged with insert_private_module().

lib/dal/src/cached_module.rs
```diff
@@ -561,143 +522,199 @@ impl CachedModule {
...
+            DELETE FROM cached_modules WHERE schema_id = $1
+            "#;
+
+        ctx.txns().await?.pg().query(query, &[&schema_id]).await?;
+
+        Ok(())
+    }
 
+    // module_id is just for debugging purposes; we only want one record per hash
+    pub async fn update_missing_package_summaries(ctx: &DalContext) -> CachedModuleResult<()> {
```
jkeiser, line 563: This goes through packages that are missing `package_summary`, and fills them in. Used for sdf migrate, but one imagines the loop will be used later to detect and correct incorrect package versions.

lib/dal/src/cached_module.rs
```diff
@@ -561,143 +522,199 @@ impl CachedModule {
...
-        let schema_id: SchemaId = schema_id.into();
+        let hashes_without_summaries = ctx.txns().await?.pg().query(query, &[]).await?;
+        if hashes_without_summaries.is_empty() {
+            return Ok(());
+        }
+        info!(
+            "Updating {} hashes without summaries",
+            hashes_without_summaries.len()
+        );
+        for hash_without_summary in hashes_without_summaries {
```
jkeiser, line 591: I considered batching these, but it turns out they don't take that long.

lib/dal/src/cached_module.rs
```diff
@@ -561,143 +522,199 @@ impl CachedModule {
...
+            return Ok(None);
+        };
+
+        let Some(variant) = schema.variants()?.into_iter().next() else {
+            warn!("builtin module {} has a schema with no variant", module_id);
+            return Ok(None);
+        };
+
+        let package_summary = PackageSummary {
+            socket_count: variant.sockets()?.len() as u32
```
jkeiser, line 677: This is the calculation. It yields the same result as the existing socket count algorithm that runs against installed packages.

lib/dal/src/management/prototype.rs
```diff
@@ -486,25 +486,7 @@ impl ManagementPrototype {
...
-                let manage_count = if has_manage_prototype { 1 } else { 0 };
-
-                // 1 + for the Manager input socket
-                variant_socket_map.insert(
-                    schema.name().to_string(),
-                    1 + manage_count + output_sockets.len() + input_sockets.len(),
-                );
-            }
-        }
+        let variant_socket_map = Self::variant_socket_map(ctx).await?;
```
jkeiser, line 489: execute_by_id() is getting rather large, so I moved variant_socket_map() into its own method rather than make it bigger.

lib/dal/src/management/prototype.rs
```diff
@@ -589,6 +571,43 @@ impl ManagementPrototype {
         })
     }
 
+    async fn variant_socket_map(
+        ctx: &DalContext,
+    ) -> ManagementPrototypeResult<HashMap<String, u32>> {
+        // Count sockets on uninstalled schemas
+        let mut variant_socket_map: HashMap<_, _> = CachedModule::latest_modules(ctx)
```
jkeiser, line 578: This is where we add in the uninstalled schemas to the variantSocketMap. Instead of checking which schemas are installed, we just get them all and then overwrite the installed ones in the next loop.

lib/si-pkg/src/pkg/variant.rs
```diff
@@ -56,7 +56,7 @@ impl SiPkgSchemaVariantData {
 #[derive(Clone, Debug)]
 pub struct SiPkgSchemaVariant<'a> {
     version: String,
-    data: Option<SiPkgSchemaVariantData>,
+    pub data: Option<SiPkgSchemaVariantData>,
```
zacharyhamm, line 59: There *is* a getter, but this is also something I don't think matters to much

zacharyhamm (approved):
Tested by rebooting sdf, seeing it write the new socket counts, and executing management funcs and checking out the variant socket map input

PR #5742, opened 2025-03-22
patch(docs): update aws ha ec2 how to guide with new assets

aaron-dernley:


stack72 (commented):
@aaron-dernley the only change we need to make is to resize the images like we did for the VPC tutorial :)

stack72 (approved):

PR #5743, opened 2025-03-22
road-map-aws-ga

mahirl:


app/docs/src/roadmap/index.md
```diff
@@ -127,19 +127,18 @@ System Initiative.
...
-We have implemented integration with
-[AWS Cloud Control API](https://aws.amazon.com/cloudcontrolapi/), providing full
+We have an integration with the
+[AWS Cloud Control API](https://aws.amazon.com/cloudcontrolapi/), providing complete
 coverage of all AWS resources supported by
 [CloudFormation](https://aws.amazon.com/cloudformation/). This enables our
-platform to create, read, update, and delete any AWS resource that
-CloudFormation supports, expanding our coverage beyond previously supported
-resource types.
+platform to create, read, update, and delete any AWS resource
```
johnrwatson, line 134: Depending how you read this I don’t believe it is technically true but I might be wrong. We support all of the resources of AWS supported by cloudformation but not all of those support CRUD (I think) some are only CRD

nickgerace: I think the next line is doing a lot of lifting here

> This enables our platform to create, read, update, and delete any AWS resource _**CloudFormation supports**_.

stack72 (approved):

PR #5748, opened 2025-03-24
fix(clover): Override TaskDefinition and ensure we can set task def updates

stack72:


bin/clover/src/pipeline-steps/assetSpecificOverrides.ts
```diff
@@ -493,6 +493,58 @@ const overrides = new Map<string, OverrideFn>([
...
+      ];
+
+      const extraProp = propForOverride(variant.domain, "extra");
+      if (!extraProp || extraProp.kind !== "object") return;
+
+      const propUsageMapProp = propForOverride(extraProp, "PropUsageMap");
+      if (!propUsageMapProp || !propUsageMapProp.data?.defaultValue) return;
+
+      const defaultValue = JSON.parse(propUsageMapProp.data.defaultValue as string);
+      console.log(defaultValue);
```
vbustamante, line 526: I think this was left in by accident?

```suggestion
```

vbustamante (approved):

PR #5750, opened 2025-03-24
chore: WorkspaceSnapshotSelector enum for split-graph integration

zacharyhamm:
This enum will make it possible for us to swap in the split graph for
the current and soon-to-be-legacy graph seamlessly without massive
changes to the DAL code. Only has one variant now, for the current
WorkspaceSnapshotGraph.


lib/dal/src/workspace_snapshot.rs
```diff
@@ -1607,92 +1594,6 @@ impl WorkspaceSnapshot {
         Ok(roots)
     }
 
-    /// If this node is associated to a single av, return it
```
zacharyhamm, line 1610: This method is unused

nickgerace (approved):
Excellent work!

PR #5752, opened 2025-03-25
fix(dal): Remove remain:sorted from FuncArgumentKind enum

stack72:
The comment explains it all but we need the ordering to be stable for postcard


lib/dal/src/func/argument.rs
```diff
@@ -62,7 +62,16 @@ pub enum FuncArgumentError {
 
 type FuncArgumentResult<T> = Result<T, FuncArgumentError>;
 
-#[remain::sorted]
+/// This type gathers all the kinds of things we will store in the
```
zacharyhamm, line 65: This comment is a little wrong since it talks about CasValue and the macros. We just want to say: don't reorder, only add to the bottom, this goes into the content store

stack72: fixed!

zacharyhamm (approved):

PR #5753, opened 2025-03-25
chore: add content type deserialization test

zacharyhamm:
This test will ensure we cannot break the serialization of types that go into the content store.


jkeiser (approved):
Straightforward and excellent. This will save us many issues, no doubt :)

PR #5756, opened 2025-03-25
feat: adding jsr ai-agent, centralizing cf-db

adamhjk:
* Adds an ai-agent to our jsr organization
* Adds cf-db to our jsr organization
* Moves cf-db to be a common dependency of clover and ai-agent
* Imports the cloudformation specs as a single, compressed JSON array, rather than as individual files


bin/hoist/diff.md
sprutton1, line 1: Maybe this should be .gitignored?

lib/jsr-systeminit/ai-agent/BUCK
sprutton1, line 1: I think we need to add all new libraries to the workspace in the root BUCK file, but maybe we can glob everything in jsr-systeminit somehow?

This is what I did in the last PR https://github.com/systeminit/si/commit/000d323d0519710e89922fc2fc22a7e8f61c84eb#diff-1d54d55142692add352d32eea043ed6f521ab1fba165af31aa41e2722bc77418R98-R101

sprutton1 (changes requested):

lib/jsr-systeminit/ai-agent/BUCK
adamhjk: nice catch!

bin/hoist/diff.md
adamhjk: 👍🏼

sprutton1 (approved):

stack72 (approved):

jkeiser (approved):

PR #5759, opened 2025-03-26
feat: add optional tokio runtime worker thread core affinity support

fnichol:
This change adds a new optional capability in the `lib/si-runtime` crate to setup a Tokio runtime that spawns its threads with a CPU core affinity assignment. This way it's possible to run the main multi-threaded runtime on specific CPU cores--that is the worker and blocking pool threads.

An additional change is made in the `lib/si-service` crate to expose an optional `rt::block_on_with_core_affinity()` function that takes a list of `CoreId`s which is used when pinning the runtime threads.

Finally, Veritech's CLI args are augmented to include a `--tokio-cpu-cores` option (and corresponding `SI_TOKIO_CPU_CORES` environment variable) that will configure the runtime to pin to the desired CPU cores.

For example running Veritech with:

```
veritech --tokio-cpu-cores 0-3
```

would configure the Tokio runtime to run only on cores 0,1,2,3. Without this option, the default is typically a thread-per-core setup as before.

Future work may enable this capability for all other services.


lib/si-runtime-rs/src/lib.rs
```diff
@@ -18,13 +21,16 @@ const COMPUTE_EXECUTOR_THREAD_PRIORITY: u8 = 25;
 // Tokio runtime shutdown timeout for compute executors
 const COMPUTE_EXECUTOR_TOKIO_RT_SHUTDOWN_TIMEOUT: Duration = Duration::from_secs(60 * 10);
 
+const MAX_WORKER_THREADS: usize = 32;
```
sprutton1, line 24: Why 32?

fnichol: Oh, this came from the blog post I was using and I think it's kind of on the arbitrary side.

johnrwatson (approved):

PR #5767, opened 2025-03-28
Remove code plaintext span metadata

nickgerace:
Undoes (not reverts) changes from d6e343b (https://github.com/systeminit/si/pull/5747) to remove code plaintext span metadata. This wasn't being used at all and could prove to be expensive (with time) to populate.


stack72 (approved):

mahirl:
always forward, never back! Love it.

PR #5768, opened 2025-03-28
Prevent: one million toasts for maintenance or downtime mode

jobelenus:
Using an ID with a toast means a new toast should replace an old toast. Which will appear as a "toast that never goes away"... but not one million toasts that take up the entire screen.

### DIFF
--- a/app/web/src/store/apis.web.ts
+++ b/app/web/src/store/apis.web.ts
@@ -107,6 +107,7 @@ async function handleOutageModes(error: AxiosError) {
     const toast = useToast();
     toast(
       {
+        id: "MAINTENANCE",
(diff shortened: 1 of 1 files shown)


johnrwatson:
@jobelenus does this re-post the toast twice if one doesn't already match it's ID that's active?

nickgerace (approved):

jobelenus:
> @jobelenus does this re-post the toast twice if one doesn't already match it's ID that's active?

If no toast exists, it creates a toast with that ID. If a toast exists with that ID, it replaces it.

johnrwatson (approved):

johnrwatson:
Ship it! ⛵

PR #5770, opened 2025-03-28
feat(web): Add input socket values to connections menu

vbustamante:
image

This may slow down the diagram endpoint! Probably better to test it before merging


britmyerss (commented):
This isn't quite what we need, so I'll be closing this PR.  The issue we've got is we aren't reactive when the values for sockets change, so it looks like they're not there until a reload.  We don't want the value that's at the prop, we want what's at the socket, so I'll be closing this PR and fixing the reactivity bug which should accomplish what we were trying to do here!

PR #5773, opened 2025-03-31
Update index if reference kinds do not match

nickgerace:
## Description

For the new frontend architecture, this PR updates the index if reference kinds found in the existing index do not match what is currently available. This is useful when new types are added to the system, but an existing system needs a new index to know that they exist.

This takes about ~200ms when it happens, which we should keep an eye on when equivalent functionality moves to "edda".

## Known Issue

In the case of a rolling update with multiple SDF instances, it is possible for both SDF instances to update the index. This is unlikely given that a tenant will likely be "speaking" to one SDF _and_ that this is a stop-gap solution until "edda" takes on this behavior. However, it is worth calling out.

### DIFF
--- a/lib/dal/src/materialized_view.rs
+++ b/lib/dal/src/materialized_view.rs
@@ -66,40 +66,7 @@ pub async fn build_all_mv_for_change_set(
     let span = current_span_for_instrument_at!("debug");
     span.record("si.workspace.id", ctx.workspace_pk()?.to_string());
 
-    // Pretend everything has changed, and build all MVs.
-    let changes = ctx
-        .workspace_snapshot()?
-        .map_all_nodes_to_change_objects()
-        .instrument(tracing::info_span!("Generating Changes for all nodes"))
-        .await?;
-
-    let (frontend_objects, patches) = build_mv_inner(
-        ctx,
-        frigg,
-        ctx.workspace_pk()?,
-        ctx.change_set_id(),
-        &changes,
-    )
-    .instrument(tracing::info_span!("Building MVs"))
-    .await?;
-
-    let index_entries = frontend_objects.into_iter().map(Into::into).collect();
-    let mv_index = MvIndex::new(ctx.change_set_id(), index_entries);
-    let mv_index_frontend_object = FrontendObject::try_from(mv_index)?;
-    frigg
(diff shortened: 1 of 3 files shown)


lib/dal/src/materialized_view.rs
```diff
@@ -121,6 +121,78 @@ pub async fn build_all_mv_for_change_set(
...
+            kv_revision_for_existing_index,
+        )
+        .await?;
+
+    DataCache::publish_patch_batch(ctx, patch_batch)
+        .instrument(tracing::info_span!("Publishing patch batch"))
+        .await?;
+
+    Ok(())
+}
```
jhelwig, line 194: This appears to be identical to `build_all_mv_for_change_set` except for `update_index` (and having the kv revision passed in)? Feels like we should try to minimize the duplication between them.

lib/sdf-server/src/service/v2/index.rs
```diff
@@ -85,7 +89,51 @@ pub async fn get_change_set_index(
...
+                .map_err(IndexError::DeserializingMvIndexData)?;
+
+            let mut implemented_kinds = HashSet::new();
+            for index_ref in mv_index.mv_list {
+                let kind = ReferenceKind::try_from(index_ref.kind.as_str())
+                    .map_err(IndexError::InvalidStringForReferenceKind)?;
+                if kind.is_revision_sensitive() {
+                    implemented_kinds.insert(kind);
+                }
+            }
```
jhelwig, line 103: Probably worth adding a span around this to have some information on how costly this bit is over time.

jhelwig (commented):
So, the idea is that this is basically a form of cache-busting? Even though any new MV kinds would be generated/added whenever their triggering entity kind is changed, that would only populate the specific entities that changed, not back-fill for all of that entity kind, so this would ensure that the back-fill happens, assuming that there hasn't been a change that has come in to cause the rebaser to do the generation for the changed-subset of that new MV kind.

Checking on every index request feels excessive, but how to be more selective on when to validate in a way that would also lessen the possibility for partial generation isn't immediately clear and probably isn't the most important place to put effort at the moment.

jhelwig (approved):

PR #5774, opened 2025-03-31
chore(deps): Update deps for auth-portal

stack72:


johnrwatson (approved):

stack72:
> @stack72 is there any way we can test this? Obviously we can't test the netlify-cli one as that's all post-deploy

I am able to use local auth stack and it works as expected

PR #5777, opened 2025-03-31
Asset List on Modeling Diagram now uses the bifrost

jobelenus:

### DIFF
--- a/lib/si-frontend-types-rs/src/view.rs
+++ b/lib/si-frontend-types-rs/src/view.rs
@@ -1,4 +1,5 @@
-use serde::{Deserialize, Serialize};
+use serde::Deserialize;
+use serde::Serialize;
 use si_events::{
     workspace_snapshot::{Checksum, ChecksumHasher, EntityKind},
     Timestamp,
--- a/app/web/src/components/AssetPalette.vue
+++ b/app/web/src/components/AssetPalette.vue
@@ -75,7 +75,7 @@
               schemaVariant.id
             "
             showSelection
-            @mousedown.left.stop="onSelect(schemaVariant.id, $event)"
+            @mousedown.left.stop="onSelect(useCorrectId(schemaVariant), $event)"
             @click.right.prevent
           >
             <template #label>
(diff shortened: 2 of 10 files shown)


lib/si-frontend-types-rs/src/view.rs
```diff
@@ -57,3 +59,82 @@ pub struct ViewList {
...
+)]
+#[serde(rename_all = "camelCase")]
+#[mv(
+    trigger_entity = EntityKind::CategorySchema,
+    reference_kind = ReferenceKind::SchemaVariantCategories,
+)]
+pub struct SchemaVariantCategories {
+    pub id: ChangeSetId,
+    pub categories: Vec<SchemaVariantsByCategory>,
+}
```
jhelwig, line 140: We should probably move these to `lib/si-frontend-types-rs/src/schema_variant.rs`. `view.rs` was meant for things related to the `View` SI entity kind, rather than all things MV related.

jhelwig (commented):
Nothing too unexpected here. It definitely reinforces my desire to come up with a blanket `FrontendChecksum` impl for our Id types, but that was turning out to be a lot more complicated than I had the brainpower for at the time, and it feels like there are more impactful places to put the effort at the moment.

lib/si-frontend-types-rs/src/view.rs
jobelenus: Done

jobelenus:
> Nothing too unexpected here. It definitely reinforces my desire to come up with a blanket `FrontendChecksum` impl for our Id types, but that was turning out to be a lot more complicated than I had the brainpower for at the time, and it feels like there are more impactful places to put the effort at the moment.

I think my big question on this is ... does `impl FrontendChecksum for Prop {` look correct (e.g. there isn't a better way to do it?)

jhelwig (approved):

jhelwig:
> I think my big question on this is ... does `impl FrontendChecksum for Prop {` look correct (e.g. there isn't a better way to do it?)

It's possible to derive the impl like we do for some of the other struct definitions, assuming all of the struct's fields' types also implement `FrontendChecksum`: https://github.com/systeminit/si/blob/2804202a58d7ec10170b98e179e21aab441c319e/lib/si-frontend-types-rs/src/view.rs#L22C1-L22C48

We haven't done the derive outside of `si-frontend-types` yet, and it would need to be part of the `Prop` struct definition in `dal`, but it should be possible to add it to the list of derives there, after adding `si-frontend-types-macros` to the `dal`'s dependencies in the `Cargo.toml`, and `BUCK` files.

jhelwig (approved):

PR #5788, opened 2025-04-01
fix(sdf): Don't truncate logs

jkeiser:
When there are a lot of logs for a function, we don't upload or report them.

This PR fixes this by leaving the output processor running until all logs have been processed.

Fixes BUG-811.

## Testing

- [x] Test management function that prints 10000 log messages, make sure all log messages show up and output task is shut down
- [x] Test multiple DVUs and management functions running in paralllel, make sure all logs show up
- [x] Kill Veritech after 10000 log message function exits but before logs are processed, make sure all logs show up and output task is shut down
- [x] Kill Veritech while 5-minute "print log every second" function is running, make sure job recovers, new log messages show up, and output task is shut down

### DIFF
--- a/lib/veritech-client/src/lib.rs
+++ b/lib/veritech-client/src/lib.rs
@@ -39,6 +39,8 @@ pub enum ClientError {
     RootConnectionClosed,
     #[error("subscriber error: {0}")]
     Subscriber(#[from] SubscriberError),
+    #[error("tokio join error: {0}")]
+    TokioJoinError(#[from] tokio::task::JoinError),
     #[error(transparent)]
     Transport(Box<dyn std::error::Error + Sync + Send + 'static>),
 }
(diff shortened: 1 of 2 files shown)


nickgerace (approved):

lib/veritech-client/src/lib.rs
```diff
@@ -212,31 +214,69 @@ impl Client {
...
+
+        // Forward output messages in the background
+        let cancel_output = CancellationToken::new();
+        if let Some(output_tx) = output_tx {
+            let cancel_output = cancel_output.clone();
+            tokio::spawn(cancel_output.run_until_cancelled_owned(forward_output(
+                output_subscriber,
+                output_tx,
+                Span::current(),
+            )));
```
nickgerace, line 236: Eventually we should use tokens and trackers for spawned tasks, but given that `veritech-client` already had `tokio::spawn` usage and both the dal and sdf don't have the best mechanisms for spawned task management, this should be fine for now!

PR #4546, opened 2024-09-11
Add forklift service

nickgerace:
## Description

This PR adds the forklift service, which "forklifts" data into a data warehouse. Right now, it only "forklifts" one set of data: billing events. That functionality is "opt-in" and the layered load config for the forklift instance must be used to enable data warehouse stream client use.

In addition to these changes, we are publishing fully-formed billing events over NATS Jetstream in four scenarios: the HEAD change set's pointer has changed, the HEAD change set's status has changed, a resource has been created, and a resource has been deleted.

## Additional Changes

- Transform `billing-events-server` into `forklift-server`
- Use and spawn `forklift-server` in anything using the dal test harness
- Move billing events publishing to its own module in the dal
- Move and rename jetstreams streams wrapper to its own module in the dal
- Remove unused and aged dal integration tests
- Create resource metadata and change set status modules with
  corresponding structs and enums in `si-events-rs`
- Adjust relevant `.github` files

PR #5790, opened 2025-04-01
Add edda service

nickgerace:
## Description

This PR adds "edda", a service written by @fnichol that is the "keeper of the materialize views". All materialized view generation has been moved out of the `dal` (and `sdf` and `rebaser` by extension) and into `edda-server`. Using naxum and NATS KV, "edda" will process materialized view generation requests for a given change set and publish them accordingly. This reduces load on our other services while ensuring we send quick updates to materialized views to the frontend. As part of these changes, there is a new LayerDb table for "ChangeBatches", which are used to "send" a batch of changes from the rebaser to edda.

## Deployment Checklist

- [x] No new parameters or configuration needed
- [x] Synced with @sprutton1 @stack72  
- [x] Checked prior art with `forklift`: https://github.com/systeminit/si/pull/4546

## Other PRs

This should be merged when https://github.com/systeminit/si/pull/5794 is merged and vice versa.


lib/si-layer-cache/src/db.rs
```diff
@@ -116,8 +120,22 @@ where
...
+                change_batch::CACHE_NAME,
+                pg_pool.clone(),
+                cache_config.clone(),
+                compute_executor.clone(),
+                tracker.clone(),
+                token.clone(),
+                // TODO(nick,jacob): ask Scott about this.
+                5,
+                // TODO(nick,jacob): ask Scott about this.
+                5
```
nickgerace, line 138: @sprutton1 @zacharyhamm we are taking 5% headroom from CAS. All the tunables added up to 100% before and we felt as though this might be the lowest impact location. Is this the right move?
cc @jhelwig

nickgerace: We validated this was the right call 👍

nickgerace:
Okay, this should be ready for real now!

lib/edda-client/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+            .map_err(Error::CreateStream)?;
+        let _ = nats::edda_requests_jetstream_stream(&context)
+            .await
+            .map_err(Error::CreateStream)?;
+
+        Ok(Self { context })
+    }
+
+    /// Asynchronously request an index update from a workspace past snapshot to the current
+    /// snapshot & return a [`RequetId`].
```
zacharyhamm, line 79: Typo here in the doc comment (Requet)

lib/edda-core/src/api_types/rebuild_request.rs
```diff
@@ -0,0 +1,213 @@
...
+    };
+
+    use serde::{de::DeserializeOwned, Serialize};
+
+    use super::RebuildRequestVersionsDiscriminants;
+
+    /// Tests that a versioned object will always serialize to the same representation, no matter
+    /// what future versions or changes to the object.
+    ///
+    /// NOTE: It is imperative that incrementatl refactorings do not lead to a version-incompatible
```
zacharyhamm, line 108: while we're talking comment typos: incrementatl, the next outkast single

lib/edda-core/src/api_types/rebuild_request.rs
```diff
@@ -0,0 +1,213 @@
...
+    /// or serialize-incompatible change. If a test fails and it's because there is a diff to the
+    /// commiteed `.snap` file, this should be considered a failed refactoring. The remediation is
+    /// to *not* update the `.snap` file but rather to fix the code so that the `.snap` format is
+    /// 100% preserved.
+    pub fn assert_serialize(
+        name: &str,
+        version: RebuildRequestVersionsDiscriminants,
+        serialize: impl Serialize,
+    ) {
+        insta::with_settings!({
```
zacharyhamm, line 118: huh interesting, i wonder if we should use this for the serialization tests in the dal too

lib/rebaser-server/src/rebase.rs
```diff
@@ -163,36 +161,26 @@ pub async fn perform_rebase(
...
-            dal::materialized_view::build_all_mv_for_change_set(ctx, frigg)
-                .instrument(tracing::info_span!(
-                    "Initial build of all materialized views"
-                ))
-                .await?;
-        }
+        span.record(
+            "si.edda.update_request.id",
+            edda_update_request_id.to_string(),
+        );
```
zacharyhamm, line 183: I wonder, should this happen in a spawned task or does the rebaser need to block here (to keep it in serial order, for example)

lib/si-layer-cache/src/db.rs
```diff
@@ -116,8 +120,18 @@ where
                 compute_executor.clone(),
                 tracker.clone(),
                 token.clone(),
-                30,
-                30
+                25,
```
zacharyhamm, line 123: @sprutton1 is there anything we should think about when it comes to resizing these slices in production?

zacharyhamm (commented):
Aside from a couple of typos in comments, I have one major question: should the write of the change batch and dispatch to edda happen in the same task as the rebase or should we spawn a task that we fire and forget to not hold the rebaser up (for example if the change batch is huge and the write is slow)

lib/edda-client/src/lib.rs
nickgerace: Fletcher! 🇨🇦

lib/edda-core/src/api_types/rebuild_request.rs
nickgerace: Fletcher! 🇨🇦

lib/edda-server/src/handlers.rs
```diff
@@ -0,0 +1,308 @@
...
+
+    let subject_str = subject.as_str();
+    let (workspace, change_set) = parse_subject(subject_prefix, subject_str)?;
+
+    let requests_stream_filter_subject =
+        nats::subject::request_for_change_set(subject_prefix, workspace.str, change_set.str);
+
+    let tracker = TaskTracker::new();
+
+    // We want to indendently control the lifecyle of our tasks
```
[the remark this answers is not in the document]
nickgerace: Fletcher! 🇨🇦

lib/edda-core/src/api_types/rebuild_request.rs
nickgerace: yeah it's pretty sweet! Fletcher detailed how this work and I think we absolutely can.

lib/si-layer-cache/src/db.rs
nickgerace: Discussed offline

lib/rebaser-server/src/rebase.rs
jhelwig: Serial order does matter. The chain of to/from snapshot IDs needs to be sent out (and processed on the receiving end) in an order where no snapshot that appears in a `to` can appear after it appears in a `from`. If the front-end starts getting batches out of order, it'll be unable to use the patches generated, and trigger a refetch-the-world scenario.

When we start doing batching on the edda side (drain all queued requests and condense into a single batch), we won't care about the order of the requests within a single condensed batch, but we still have the rule that no snapshot that's in the `to` of a request can appear in a request in a `from` of a subsequent condensed batch (right now the condensed batch sizes are `n = 1`, instead of `n >= 1`).

This could be offloaded to a separate task via a channel, but only if we can guarantee that no snapshot referenced by any of the tasks in the channel (from/to snapshot for calculating the set of changes) is ever evicted before all tasks that reference it have been processed.

zacharyhamm (dismissed):

zacharyhamm (approved):

PR #5804, opened 2025-04-03
feat(clover): Listener DefaultAction set target group arns

sprutton1:
Given that that the target group arn is managed in two places in this asset, this ensures that if you are using the target group arn input socket that we work accordingly.

Autoformatting really made a mess of this.


bin/clover/src/pipeline-steps/assetSpecificOverrides.ts
```diff
@@ -551,22 +590,80 @@ const overrides = new Map<string, OverrideFn>([
...
     const securityGroupsSocket = variant.sockets.find(
-      (s: ExpandedSocketSpec) => s.name === "Security Groups" && s.data.kind === "input",
+      (s: ExpandedSocketSpec) =>
+        s.name === "Security Groups" && s.data.kind === "input",
     );
     if (!securityGroupsSocket) return;
 
     setAnnotationOnSocket(securityGroupsSocket, { tokens: ["GroupId"] });
   }],
+  ["Listener DefaultActions", (spec: ExpandedPkgSpec) => {
```
sprutton1, line 600: This is the relevant part

stack72 (approved):

stack72 (approved):

stack72 (approved):

PR #5811, opened 2025-04-03
Fix: force "teleport" input to always have focus

jobelenus:
Repro steps:
1. select a component
2. hit `c`
3. see the input has focus ✅ 
4. click the modal title bar
5. hit backspace/del ❌ 

the Modal "chrome" clicks were removing focus

I also got rid of the "read only" code editor in the display


britmyerss (approved):
that's exactly what I had to have done, to trigger the delete modal... well done!

PR #5812, opened 2025-04-04
chore: make pg recycle method configurable

sprutton1:
These settings were changed when we added pgbouncer to better handle migrations. Pgbouncer recommends sending a `RECONNECT` after DDL changes to reset the open transactions and `RESET ALL` in order to clear the prepared statement cache. Without doing these, we were seeing open connections serving the incorrect schema.

We have now switched to using RDS Proxy, which, after DDL changes, purports to close the transaction and clear the prepared statement cache. Because of this, I believe it is safe to change back to using Fast recycles here.

https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/rds-proxy.howitworks.html

Unfortunately, the only way to test this is to create a new database migration and see how it handles.

This PR allows us the configure the recycle method. Setting `Custom` uses the const defined alongside the poolconfig. 

```
  manager: Some(
        ManagerConfig {
            recycling_method: Custom(
                "\n    SET CLIENT_MIN_MESSAGES TO ERROR;\n    ROLLBACK;\n    SET CLIENT_MIN_MESSAGES TO WARNING;\n    CLOSE ALL;\n    SET SESSION AUTHORIZATION DEFAULT;\n    RESET ALL;\n    UNLISTEN *;\n    SELECT pg_advisory_unlock_all();\n    DISCARD TEMP;\n    DISCARD SEQUENCES;\n    DISCARD PLANS;\n",
            ),
        },
    ),
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -134,7 +134,7 @@ crossbeam-queue = { version = "0.3.11" }
 darling = "0.20.10"
 dashmap = "6.1.0"
 deadpool = { version = "0.12.1", features = ["rt_tokio_1"] }
-deadpool-postgres = "0.14.0"
+deadpool-postgres = { version = "0.14.0", features = ["serde"] }
 derive_builder = "0.20.2"
 derive_more = "0.99.17"
 devicemapper = { version = "=0.34.1" }
--- a/lib/si-data-pg/src/lib.rs
+++ b/lib/si-data-pg/src/lib.rs
@@ -147,6 +147,7 @@ pub struct PgPoolConfig {
     pub pool_timeout_wait_secs: Option<u64>,
     pub pool_timeout_create_secs: Option<u64>,
     pub pool_timeout_recycle_secs: Option<u64>,
+    pub recycling_method: Option<RecyclingMethodConfig>,
 }
 
 impl Default for PgPoolConfig {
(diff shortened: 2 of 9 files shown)


nickgerace (approved):

lib/si-data-pg/src/lib.rs
```diff
@@ -66,7 +66,7 @@ const TEST_QUERY: &str = "SELECT 1";
 // We set `CLIENT_MIN_MESSAGES` to `ERROR` to silence the possible warning from `ROLLBACK` about
 // there not being any open transaction. We immediately set it back to the default value
 // (`WARNING`) after so we're not hiding any "real" warnings that might happen.
-const CONNECTION_RECYCLING_METHOD: &str = r#"
+pub const CONNECTION_RECYCLING_METHOD: &str = r#"
```
nickgerace, line 69: Approved, but why is this now pub?

Cargo.toml
```diff
@@ -134,7 +134,7 @@ crossbeam-queue = { version = "0.3.11" }
 darling = "0.20.10"
 dashmap = "6.1.0"
 deadpool = { version = "0.12.1", features = ["rt_tokio_1"] }
-deadpool-postgres = "0.14.0"
```
nickgerace, line 137: Also non-blocking, but what is using this feature?

sprutton1: PoolConfig derives ser/de so it can be `try_from`'d the command line args. I needed it so we can use `RecycleMethod` there

lib/si-data-pg/src/lib.rs
nickgerace: I think this actually works well though because you're implying that an external user could use it if they wanted to

sprutton1: Fixed

jhelwig:
Lint problem around an unused constant.
```
✗ Fail: root//lib/si-data-pg:check-lint-rust-unit (0.0s)
--
  | ---- STDOUT ----
  | warning: constant `CONNECTION_RECYCLING_METHOD` is never used
  | --> lib/si-data-pg/src/lib.rs:69:7
  | \|
  | 69 \| const CONNECTION_RECYCLING_METHOD: &str = r#"
  | \|       ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  | \|
  | = note: `#[warn(dead_code)]` on by default
  |  
  |  
  | warning: 1 warning emitted
```

nickgerace (approved):

PR #5813, opened 2025-04-04
fix(sdf): Close output forwarder tasks when callers stop listening

jkeiser:
Integration tests are timing out sometimes right now, because we don't close out the output-forwarding tasks when we try to shut down. This causes output forwarders to shut down when the one receiving the output shuts down (because they will drop the `Receiver`).

## Tests

* [x] Integration tests pass and do not time out in CI

### DIFF
--- a/lib/veritech-client/src/lib.rs
+++ b/lib/veritech-client/src/lib.rs
@@ -358,8 +358,15 @@ async fn forward_output(
     output_tx: mpsc::Sender<OutputStream>,
     request_span: Span,
 ) {
-    // Return the loop where we listen for messages
-    while let Some(msg) = output_subscriber.next().await {
+    // Listen for messages
+    while let Some(msg) = tokio::select! {
+        next = output_subscriber.next() => next,
+        // Exit early if the output channel is closed; no one cares about the output!
+        _ = output_tx.closed() => {
+            warn!("cancelling output forwarder: receiver closed early");
+            None
+        }
+    } {
         match msg {
             Ok(output) => {
                 output.process_span.follows_from(&request_span);


sprutton1 (approved):

lib/veritech-client/src/lib.rs
```diff
@@ -358,8 +358,15 @@ async fn forward_output(
...
-    while let Some(msg) = output_subscriber.next().await {
+    // Listen for messages
+    while let Some(msg) = tokio::select! {
+        next = output_subscriber.next() => next,
+        // Exit early if the output channel is closed; no one cares about the output!
+        _ = output_tx.closed() => {
+            warn!("cancelling output forwarder: receiver closed early");
+            None
+        }
+    } {
```
jhelwig, line 369: So, the thinking is that the hang we were seeing in CI was coming from the `rx` side of `output_tx` going away, and causing usage of the `tx` to block/hang, rather than anything related to the other `tokio::select!`, and its handling of cancellation tokens that we were initially discussing?

jkeiser: Yeah; cancellation tokens would have worked as well, but is unnecessary since closing output_rx is already a signal (and probably a more reliable one) that we don't need to be listening to logs anymore. By fixing it this way, we ensure that if the process putting logs in the database dies for *any* reason, the forwarder dies as well.

jkeiser: ... at least, that's the only explanation for why the PR in question could have caused this--the VeritechClient is the one that had its logic changed. Reading the code, I'm a little confused as to how the *FuncRunner* manages to get killed. But clearly it's making things better, and I think I might be Fridaying myself out here. I'll read through code when I'm less stale, and monitor CI as well.

PR #5818, opened 2025-04-04
Zack/integrate split graph

zacharyhamm:
Integrates the "split graph" with the DAL.

Some functionality that was previously on the workspace snapshot, but
did not rely on internal implementation details and had no need of
`self` was moved off. For example, the DependentValueRoots and the
InputSocket::new code was moved off the snapshot, so that it could be
shared between legacy and split snapshots.

Adds the column `snapshot_kind` to the workpaces table to indicate
whether the snapshot is a "legacy" or "split".

Some functionality is not yet implemented for the split snapshot:
  1. cycle detection is implemented but is not correct.
  2. Approval flow is not yet implemented.
  3. The prop type trait needs to be re-implemented for the split
     snapshot.

Split graph code should only be executed if the workspace is marked as
split. The most important thing to review in this PR is that it does not
regress existing functionality and that we will not execute split
snapshot code on accident.


lib/dal-test/src/signup.rs
```diff
@@ -40,6 +103,16 @@ impl WorkspaceSignup {
         let workspace =
             Workspace::new_from_builtin(ctx, WorkspacePk::generate(), workspace_name, token)
                 .await?;
+
```
britmyerss, line 106: Might be neat to modify the dal test setup to create 2 workspaces for each test (or run the tests both ways), depending on the specifics of the rollout strategy for this and how long we expect to have to support the legacy snapshots and want to protect against regressions.

lib/dal/src/approval_requirement.rs
```diff
@@ -193,6 +193,16 @@ impl ApprovalRequirementDefinition {
         Err(ApprovalRequirementError::EntityNotFound(entity_id))
     }
 
+    /// Used by the incomplete implementation of approval requirements in the split snapshot.
+    /// Delete when approval requirements implemented
+    pub fn fake() -> Self {
```
britmyerss, line 198: great idea!

lib/dal/src/workspace_snapshot/node_weight.rs
```diff
@@ -1221,3 +1258,118 @@ impl CorrectExclusiveOutgoingEdge for NodeWeight {
...
+    }
+
+    fn node_hash(&self) -> ContentHash {
+        self.node_hash()
+    }
+}
+
+impl From<&NodeWeight> for EntityKind {
+    fn from(value: &NodeWeight) -> Self {
+        match value {
```
nickgerace, line 1302: Good idea

lib/dal/src/workspace_snapshot/split_snapshot.rs
```diff
@@ -0,0 +1,1634 @@
...
+                        .await?
+                        .ok_or(
+                            WorkspaceSnapshotError::SplitSnapshotSubGraphMissingAtAddress(
+                                subgraph_address,
+                            ),
+                        )?;
+
+                    // xxx: we have to make the splitgraph constructable from arcs, it will
+                    // xxx: need to handle the copy-on-write behavior internally
+                    subgraphs.push(subgraph.as_ref().clone());
```
nickgerace, line 407: 👍

lib/dal/src/workspace_snapshot/split_snapshot.rs
```diff
@@ -0,0 +1,1634 @@
...
+        })
+    }
+
+    pub async fn find_for_change_set(
+        ctx: &DalContext,
+        change_set_id: ChangeSetId,
+    ) -> WorkspaceSnapshotResult<Self> {
+        // There's a race between finding which address to retrieve and actually retrieving it
+        // where it's possible for the content at the address to be garbage collected, and no
+        // longer be retrievable. We'll re-fetch which snapshot address to use, and will retry,
```
nickgerace, line 440: How often does this happen? Should we trace this at the warn level?

lib/dal/src/workspace_snapshot/split_snapshot.rs
```diff
@@ -0,0 +1,1634 @@
...
+                Err(e) => return Err(e),
+            }
+        }
+
+        error!(
+            "Retries exceeded trying to fetch split snapshot for change set {:?}",
+            change_set_id
+        );
+
+        Err(WorkspaceSnapshotError::WorkspaceSnapshotNotFetched)
```
nickgerace, line 484: Well... I guess we'll know if it happens five times

lib/dal/src/workspace_snapshot/split_snapshot.rs
```diff
@@ -0,0 +1,1634 @@
...
+        let supergraph_address = slow_rt::spawn(async move {
+            let mut working_copy = self_clone.working_copy_mut().await;
+            let start = Instant::now();
+            working_copy.cleanup_and_merkle_tree_hash();
+            warn!("cleaned up working copy in {:?}", start.elapsed());
+
+            warn!(
+                "current addresses: {:?}",
+                working_copy.supergraph().addresses()
+            );
```
nickgerace, line 505: Could this become noisy quickly?

lib/dal/src/workspace_snapshot/split_snapshot.rs
```diff
@@ -0,0 +1,1634 @@
...
+    }
+
+    async fn entity_id_for_approval_requirement_definition_id(
+        &self,
+        _id: ApprovalRequirementDefinitionId,
+    ) -> WorkspaceSnapshotResult<EntityId> {
+        Ok(EntityId::new())
+    }
+
+    async fn get_approval_requirement_definition_by_id(
```
nickgerace, line 1226: Noted given the PR description

lib/si-layer-cache/src/db.rs
```diff
@@ -203,7 +244,37 @@ where
...
+            ),
+            create_layer_cache(
+                split_snapshot_rebase_batch::CACHE_NAME,
+                pg_pool.clone(),
+                cache_config.clone(),
+                compute_executor.clone(),
+                tracker.clone(),
+                token.clone(),
+                1,
+                1
```
nickgerace, line 277: Confirmed these add to 100 each 🤓

lib/si-split-graph/src/tests/mod.rs
```diff
@@ -303,20 +267,486 @@ fn ordered_container() -> SplitGraphResult<()> {
...
+    let updates = split_graph.detect_updates(&updated_graph);
+    split_graph.perform_updates(&updates);
+    split_graph.cleanup_and_merkle_tree_hash();
+    default_edge_assertions(&split_graph, a_id, b_id)?;
+
+    Ok(())
+}
+
+#[test]
+fn cross_graph_node_id_updates() -> SplitGraphResult<()> {
```
nickgerace, line 392: Very great work. I was curious about how this would work and I can see it clearly. Really great tests

nickgerace (approved):
I reviewed with the primary goal of regression hunting. The hunt yielded nothing. This was really well done and this is a banger of a PR 🔥

lib/dal-test/src/signup.rs
zacharyhamm: At the least we want a compile time switch. My big worry about running both is that the tests will take forever

lib/dal/src/workspace_snapshot/split_snapshot.rs
zacharyhamm: you know, i copied this from the existing find code. I don't know if the race is still a problem

lib/dal/src/workspace_snapshot/split_snapshot.rs
zacharyhamm: yes for sure we'll need to tone down the warns here, to debug

lib/rebaser-core/src/api_types/enqueue_updates_response/v1.rs
```diff
@@ -22,7 +22,7 @@ pub struct EnqueueUpdatesResponseV1 {
 #[serde(rename_all = "camelCase")]
 pub enum RebaseStatus {
     Success {
-        updates_performed: RebaseBatchAddress,
+        updates_performed: RebaseBatchAddressKind,
```
zacharyhamm, line 25: @fnichol will this change cause any problems? This type is not versioned but not sure if we postcard serialize it anywhere

lib/dal/src/change_set.rs
```diff
@@ -848,14 +890,40 @@ impl ChangeSet {
             .base_change_set_id
             .ok_or(ChangeSetError::NoBaseChangeSet(self.id))?;
 
-        if let Some(rebase_batch) = self.detect_updates_that_will_be_applied(ctx).await? {
-            let updates_address = ctx.write_rebase_batch(rebase_batch).await?;
+        let snapshot_kind: WorkspaceSnapshotSelectorDiscriminants =
```
britmyerss, line 893: why do you want to expose both split and legacy update detection here? (note haven't pulled down locally yet to trace the callers...)

lib/dal/src/workspace_snapshot.rs
```diff
@@ -1368,35 +1313,6 @@ impl WorkspaceSnapshot {
         .await??)
     }
 
-    /// Mark whether a prop can be used as an input to a function. Props below
-    /// Maps and Arrays are not valid inputs. Must only be used when
-    /// "finalizing" a schema variant!
-    pub async fn mark_prop_as_able_to_be_used_as_prototype_arg(
```
britmyerss, line 1374: can you explain a bit more about the impact/need for moving these out of the snapshot itself? Just wondering how to think about this going forward....

lib/dal/src/workspace_snapshot/split_snapshot.rs
britmyerss: yeah if you search for this in honeycomb you'll see it :)

lib/dal/tests/integration_test/component/delete.rs
```diff
@@ -1470,57 +1458,49 @@ async fn delete_multiple_components(ctx: &mut DalContext) {
...
 
     for (component_id, status) in expected_deletion_statuses {
         assert_eq!(Some(status), deletion_statuses.get(component_id));
     }
 
     assert!(
         Component::try_get_by_id(ctx, component_to_delete.id())
-            .await
-            .expect("should succeed")
+            .await?
```
britmyerss, line 1472: yay!

lib/rebaser-core/src/api_types/enqueue_updates_response/v1.rs
britmyerss: but I don't think we blat this over layer_db anymore, so we might be ok...

lib/si-split-graph/src/lib.rs
```diff
@@ -567,16 +636,26 @@ where
         subgraph_index: SubGraphIndex,
         node: SplitGraphNodeWeight<N>,
     ) -> SplitGraphResult<SplitGraphNodeIndex> {
-        let subgraph = self.get_subgraph_mut(subgraph_index as usize)?;
+        let subgraph = self.get_subgraph_mut(subgraph_index)?;
+        let node_id = node.id();
         let node_index = subgraph.add_node(node);
+        if node_index == NodeIndex::new(2085) {
```
britmyerss, line 642: 2085?

zacharyhamm: whoops! this is leftover debugging code. that node index was important :)

lib/dal/src/workspace_snapshot.rs
zacharyhamm: This is a pretty old piece of code, from the first implementation of the graph, where we didn't have the right patterns in place. There's no need to have this method on the snapshot directly, it's really the responsibility of the Prop code to make these changes. I moved it off the snapshot so I wouldn't have to re-implement it for the split graph.

lib/dal/src/change_set.rs
zacharyhamm: We're applying to head, and it should work for either snapshot kind here. There may be a simpler abstraction

britmyerss: yeah just thinking about the framing of 'don't break existing' and if you want to handle this at a higher level in the short term. but again, I didn't trace the callers so it's probably fine

lib/dal/src/workspace_snapshot/graph/v4.rs
```diff
@@ -1768,29 +1770,6 @@ impl WorkspaceSnapshotGraphV4 {
         Ok(())
     }
 
-    /// Update node weight in place with a lambda. Use with caution. Generally
-    /// we treat node weights as immutable and replace them by creating a new
-    /// node with a new node weight and replacing references to point to the new
-    /// node.
-    pub(crate) fn update_node_weight<L>(
```
britmyerss, line 1775: where'd this go? just wasn't used?

lib/dal/tests/integration_test/component.rs
```diff
@@ -648,3 +648,44 @@ async fn set_the_universe(ctx: &mut DalContext) -> Result<()> {
 
     Ok(())
 }
+
+#[test]
+#[ignore = "this test does not pass on the legacy snapshot because id is not included in the node hash"]
```
britmyerss, line 653: .. should it have been?

lib/dal/src/workspace_snapshot/selector.rs
```diff
@@ -59,45 +66,52 @@ use crate::{
...
-        ComponentResult,
-        Connection,
-    },
+    component::ComponentResult,
     prop::PropResult,
-    socket::connection_annotation::ConnectionAnnotation,
-    workspace_snapshot::traits::approval_requirement::ApprovalRequirementExt,
 };
 
+/// The `WorkspaceSnapshotSelector` enum acts as a dispatcher for workspace snapshot methods
```
britmyerss, line 78: If we add more Workspace kinds down the road - is there anything additional we need to do to account for (thinking postcard/versioning/whatever)

lib/dal-test/src/signup.rs
britmyerss: yeah I almost also typed RIP build times :)

lib/dal-test/src/expand_helpers.rs
```diff
@@ -169,8 +171,11 @@ fn tracing_init_inner(span_events_env_var: &str, log_env_var: &str) {
 }
 
 /// This function is used during macro expansion for setting up the workspace for integration tests.
-pub async fn workspace_signup(ctx: &DalContext) -> crate::Result<(WorkspaceSignup, String)> {
+pub async fn workspace_signup(ctx: &mut DalContext) -> crate::Result<(WorkspaceSignup, String)> {
     use color_eyre::eyre::WrapErr;
+    // Tenancy must be set here or the update_visibility call will fail, since it now needs to check
+    // the workspace snapshot_kind
+    ctx.update_tenancy(Tenancy::new(WorkspacePk::NONE));
```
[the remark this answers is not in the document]
britmyerss: this is a bigger conversation I think here around DalContext and authorization scopes

lib/dal/src/workspace_snapshot.rs
britmyerss: Sorry, poorly worded question. Thinking about your comment here: 

> Some functionality that was previously on the workspace snapshot, but did not rely on internal implementation details and had no need of self was moved off. For example, the DependentValueRoots and the InputSocket::new code was moved off the snapshot, so that it could be shared between legacy and split snapshots.

I am looking to improve my my mental model/vocabulary for what's "on the snapshot" what get's a trait extension.... etc..

lib/dal-test/src/signup.rs
```diff
@@ -28,6 +37,60 @@ pub struct WorkspaceSignup {
...
+        // crate::test_exclusive_schemas::migrate(ctx).await?;
+
+        let workspace_snapshot_address = ctx
+            .workspace_snapshot()?
+            .as_split_snapshot()?
+            .write(ctx)
+            .await?;
+
+        let mut head_change_set =
+            ChangeSet::new(ctx, "HEAD", None, workspace_snapshot_address).await?;
```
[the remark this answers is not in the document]
zacharyhamm: yeah this is basically identical to that except it creates the "split" workspace

lib/dal/src/workspace_snapshot.rs
zacharyhamm: Yeah i think generally it doesn't need to be inside the snapshot implementation unless you expect it to require access to snapshot "internals". One reason for that would be if you want the operation to hold onto the lock for the entire time because that makes it more performant. Otherwise, if it only uses public methods, it should probably  be encapsulated elsewhere

lib/dal/src/workspace_snapshot/graph/v4.rs
zacharyhamm: yep! never used anymore. We do `add_or_replace` and just rewrite the node that way.

lib/dal/tests/integration_test/component.rs
zacharyhamm: yes! it should have. one thing this uncovered was a bug. Our component node id updates only work because we create a new component that gets a new timestamp, and thus has an updated hash.

nickgerace (dismissed):

nickgerace (approved):

PR #5829, opened 2025-04-07
feat: add incoming connections to mgmt functions components

adamhjk:
Adding an `incomingConnections` field to every component, rather than just the input component.

### DIFF
--- a/bin/lang-js/src/function_kinds/management.ts
+++ b/bin/lang-js/src/function_kinds/management.ts
@@ -15,19 +15,11 @@ const debug = Debug("langJs:management");
 
 export interface ManagementFunc extends Func {
   currentView: string;
-  thisComponent: ThisComponent;
+  thisComponent: ComponentWithGeometry;
   components: {
     [key: string]: ComponentWithGeometry;
   };
   variantSocketMap: Record<SchemaName, number>;
-}
-
-export interface ThisComponent extends ComponentWithGeometry {
-  incomingConnections: {
-    [key: SocketName]: SocketRefAndValue[] | SocketRefAndValue | undefined;
-  };
-  // TODO outgoingConnections so we can automatically connect *output* from created components
-  // to existing components
 }
 
 export type ManagementFuncResult =
--- a/lib/cyclone-core/src/component_view.rs
+++ b/lib/cyclone-core/src/component_view.rs
@@ -38,6 +38,7 @@ pub struct ComponentViewWithGeometry {
     pub kind: Option<String>,
     pub properties: Value,
     pub geometry: Value,
+    pub incoming_connections: Value,
 }
 
 impl Default for ComponentViewWithGeometry {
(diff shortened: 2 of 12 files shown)


lib/dal/src/management/prototype.rs
```diff
@@ -251,9 +254,12 @@ impl ManagerComponent {
         kind: &str,
         views: &HashMap<ViewId, String>,
     ) -> ManagementPrototypeResult<ManagerComponent> {
+        let component = ManagedComponent::new(ctx, component_id, kind, views).await?;
```
jkeiser, line 257: AFAICT, ManagerComponent no longer needs to exist--incoming_connections is literally the only difference. If Component turns out to be a bad name, we could rename `ManagedComponent` -> `ManagementComponent` and remove `ManagerComponent` entirely.

lib/cyclone-core/src/component_view.rs
```diff
@@ -56,14 +58,12 @@ impl Default for ComponentViewWithGeometry {
 pub struct ThisComponent {
```
jkeiser, line 58: This can go away, and the caller can just use `ComponentViewWithGeometry` now :)

lib/dal/src/func/binding/management.rs
```diff
@@ -95,6 +95,7 @@ impl ManagementBinding {
                                 kind: {json_name},
                                 properties?: {sv_type},
                                 geometry?: {{ [key: string]: Geometry }},
+                                incomingConnections?: {{ [socket: string]: {INCOMING_CONNECTION_TYPE} | {INCOMING_CONNECTION_TYPE}[] }},
                                 connect?: {{
```
jkeiser, line 99: I don't think this is for your review at all, but ... `connect` is an operation, not an input, right? Like, it seems like this is only used in `type Input`, so it's weird we have this here. I guess it technically passes typecheck since we always pass undefined, but still.

jkeiser (approved):

jkeiser (approved):

bin/lang-js/src/function_kinds/management.ts
```diff
@@ -15,21 +15,13 @@ const debug = Debug("langJs:management");
 
 export interface ManagementFunc extends Func {
   currentView: string;
-  thisComponent: ThisComponent;
+  thisComponent: ComponentWithGeometry;
```
wendybujalski, line 18: This code is so much cleaner/clearer with the removal of `ThisComponent` and putting `incomingConnections` directly on `ComponentWithGeometry` - why wasn't it done that way in the first place? Is there some benefit or past assumption or restriction that I'm not aware of?

stack72 (approved):

bin/lang-js/src/function_kinds/management.ts
jkeiser: Because nobody was sure we wanted to add all that data to every single component :) it's definitely cleaner this way.

PR #5834, opened 2025-04-08
feat(module-index): allow specifying a cloudfront domain

sprutton1:
Allows the module-index to generate urls for either s3 directly or cloudfront. If the cloudfront config is empty (the default), we behave exactly the same as we do now, so this should not cause any behavioral changes until config changes happen.

If the cloudfront domain is set, gets for modules or exports will go through cloudfront, leveraging the cdn. This should allow for cache rebuilds to be faster and generate less overall traffic it to S3. This will likely make the biggest difference in local development, but only if the cloudfront url is configured. We can configure this by default after the production instance is deployed successfully.


fnichol (approved):
Looks good!

PR #5837, opened 2025-04-08
Polish spans for edda change set processor task

nickgerace:
## Description

This PR polishes spans for the edda change set processor task. It ensures that the request ID will always have a home to be recorded and it co-locates request handling into an inner function.

Finding the current span and recording fields are now protected too.

### DIFF
--- a/lib/edda-server/src/change_set_processor_task.rs
+++ b/lib/edda-server/src/change_set_processor_task.rs
@@ -273,6 +273,8 @@ async fn graceful_shutdown_signal(
 mod handlers {
     use std::result;
 
+    use dal::{ChangeSetId, DalContext, WorkspacePk};
+    use frigg::FriggStore;
     use naxum::{
         extract::State,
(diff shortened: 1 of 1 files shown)


lib/edda-server/src/change_set_processor_task.rs
```diff
@@ -342,9 +344,34 @@ mod handlers {
             .build_for_change_set_as_system(workspace_id, change_set_id, None)
             .await?;
 
-        let span = Span::current();
-        span.record("si.workspace.id", workspace_id.to_string());
-        span.record("si.change_set.id", change_set_id.to_string());
+        let span = current_span_for_instrument_at!("info");
```
fnichol, line 347: That's great!

fnichol (approved):
Assuming you're able to clear the lint errors, this all looks good otherwise! I'll try to re-approve if there are re-pushes

stack72 (approved):

PR #5842, opened 2025-04-09
feat(billing): Upload Posthog billing data

jkeiser:
This uploads one event per hour per workspace showing the workspace's resource count. The value is chosen such that SUM(resource_count) for a given owner_pk is the same as the billing resource hours for that user (and thus if it is over 100, the user has crossed to paid tier).

The progress table is particularly nasty to work with, but it seems to be working at this point.

## Testing

- [ ] Test initial upload, progressive upload, and end of upload using a temporary update_progress table
- [ ] Upload events to dev and prod, see how people feel about the data

### DIFF
--- a/component/lambda/functions/upload-posthog-billing-data.py
+++ b/component/lambda/functions/upload-posthog-billing-data.py
@@ -1,0 +1,129 @@
+from typing import Iterable, NotRequired, cast
+from datetime import date, datetime, timedelta, timezone
+import itertools
+import logging
+
+from si_posthog_api import PosthogApi
+from si_lago_api import ExternalSubscriptionId
+from si_lambda import SiLambda, SiLambdaEnv
+from si_types import OwnerPk, SqlTimestamp, WorkspaceId, sql_to_iso_timestamp
+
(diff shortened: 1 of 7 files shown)


component/lambda/functions/upload-posthog-billing-data.py
```diff
@@ -0,0 +1,123 @@
...
+            )
+            response = self.posthog.post("/batch", {
+                "historical_migration": True,
+                "batch": [
+                    {
+                        "event": "test-billing-workspace_resource_hours",
+                        "properties": {
+                            "distinct_id": str(owner_pk),
+                            "workspace_id": str(workspace_id),
+                            "resource_count": str(resource_count),
```
britmyerss, line 107: let's send this as a number instead of a string so posthog picks it up as such

jkeiser: I don't even know what yesterday me was thinking, putting str() here, thanks :)

britmyerss (dismissed):

britmyerss (approved):

PR #5855, opened 2025-04-09
feat(web) - simple socket UI connecting edges on the diagram

wendybujalski:
Components can now have their data sockets connected via the diagram. Dragging an edge between components will open up the connections panel to create a connection. A new connection can also be added to two components which are already connected by selecting the edge between them and clicking the button on the `EdgeDetailsPanel`.

Currently the empty states for the connections panel could be improved and the user cannot add multiple connections at once.

### DIFF
--- a/app/web/src/components/ModelingDiagram/ModelingDiagram.vue
+++ b/app/web/src/components/ModelingDiagram/ModelingDiagram.vue
@@ -363,7 +363,10 @@ import { useRoute } from "vue-router";
 import { useToast } from "vue-toastification";
 import { useCustomFontsLoaded } from "@/utils/useFontLoaded";
 import DiagramGroup from "@/components/ModelingDiagram/DiagramGroup.vue";
-import { useComponentsStore } from "@/store/components.store";
+import {
+  ConnectionDirection,
+  useComponentsStore,
+} from "@/store/components.store";
 import DiagramGroupOverlay from "@/components/ModelingDiagram/DiagramGroupOverlay.vue";
 import { DiagramCursorDef, usePresenceStore } from "@/store/presence.store";
 import { useRealtimeStore } from "@/store/realtime/realtime.store";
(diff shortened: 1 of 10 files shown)


app/web/src/components/ModelingDiagram/ModelingDiagram.vue
```diff
@@ -3156,6 +3198,18 @@ function getElementByKey(key?: DiagramElementUniqueKey) {
   return key ? allElementsByKey.value[key] : undefined;
 }
 
+function getDiagramSocketByKey(key?: DiagramElementUniqueKey) {
+  if (key) {
+    const parent = allElementsByKey.value[key.slice(0, 28)] as
```
jobelenus, line 3203: this seems suspicious?

jobelenus: Ah, I think its all except `-inputsocket`/`-outputsocket`? Maybe just drop a comment there for that

app/web/src/components/ModelingDiagram/ModelingDiagram.vue
```diff
@@ -2481,13 +2497,24 @@ function onResizeMove() {
...
+});
 const drawEdgeToSocketKey = ref<DiagramElementUniqueKey>();
-const drawEdgeToSocket = computed(
-  () => getElementByKey(drawEdgeToSocketKey.value) as DiagramSocketData,
-);
+const drawEdgeToSocket = computed(() => {
+  if (featureFlagsStore.SIMPLE_SOCKET_UI) {
+    return getDiagramSocketByKey(
+      drawEdgeToSocketKey.value,
+    ) as DiagramSocketData;
```
jobelenus, line 2513: I'm slightly worried about the "cast" `as DiagramSocketData` since the fn _can_ return null.

jobelenus (approved):
A nit or two, but otherwise excellent! :shipit:

app/web/src/components/ModelingDiagram/ModelingDiagram.vue
wendybujalski: Ah yeah, it slices out the node/group id from the key. I can add a comment!

app/web/src/components/ModelingDiagram/ModelingDiagram.vue
wendybujalski: Shouldn't ever be null but I can add a check for it!

jobelenus (approved):

PR #5859, opened 2025-04-10
From cypress, load new html page, invoke web worker, run tests, log output, assert.

jobelenus:
And it builds a new webworker html page for use within CI.


sprutton1 (approved):

britmyerss (approved):
WOOT THIS IS GREAT!!!

PR #5862, opened 2025-04-10
Move actions list to norse stack

nickgerace:
## Description

This PR moves the "list_actions" functionality to the norse stack. It also brings in action prototype views into the fold from the "AssetActionsDetails" component.

As a result of these changes, the new "mead-hall" directory contains the beginning of flagged, bifrost-powered components that should be guarded by feature flag. As things move over to the new paradigm, we do not want to infect existing components. Rather, we want to feature flag which component is used solely at the template level, if possible.

One of the materialized views from these changes involves a multi-trigger approach where the entity kind is "root", but additional checks need to be performed to know the minimal set of MVs to build. The "build_mv!" macro has been changed to accomodate this.

### DIFF
--- a/app/web/src/components/ApplyChangeSetButton.vue
+++ b/app/web/src/components/ApplyChangeSetButton.vue
@@ -33,7 +33,20 @@
       />
     </template>
 
-    <ApprovalFlowModal ref="approvalFlowModalRef" votingKind="merge" />
+    <template
+      v-if="
+        featureFlagsStore.FRONTEND_ARCH_VIEWS &&
+        featureFlagsStore.BIFROST_ACTIONS
+      "
+    >
+      <BifrostApprovalFlowModal
+        ref="bifrostApprovalFlowModalRef"
+        votingKind="merge"
+      />
+    </template>
+    <template v-else>
+      <ApprovalFlowModal ref="approvalFlowModalRef" votingKind="merge" />
+    </template>
   </VButton>
 </template>
 
@@ -50,11 +63,14 @@ import clsx from "clsx";
 import { useChangeSetsStore } from "@/store/change_sets.store";
 import { useStatusStore } from "@/store/status.store";
 import { useActionsStore } from "@/store/actions.store";
+import { useFeatureFlagsStore } from "@/store/feature_flags.store";
 import { ChangeSetStatus } from "@/api/sdf/dal/change_set";
+import BifrostApprovalFlowModal from "@/mead-hall/ApprovalFlowModal.vue";
 import ApprovalFlowModal from "./ApprovalFlowModal.vue";
 
 const actionsStore = useActionsStore();
 const changeSetsStore = useChangeSetsStore();
+const featureFlagsStore = useFeatureFlagsStore();
 const statusStore = useStatusStore();
 
 const displayCount = computed(() => actionsStore.proposedActions.length);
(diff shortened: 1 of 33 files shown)


lib/edda-server/src/change_set_processor_task/materialized_view.rs
```diff
@@ -239,74 +250,141 @@ async fn build_mv_inner(
...
+                                    EntityKind::ActionPrototype =>  match ActionPrototype::schema_variant_id(ctx, maybe_action_change.entity_id.into_inner().into()).await {
+                                        Ok(schema_variant_id) => match SchemaVariant::list_component_ids(ctx, schema_variant_id).await {
+                                            Ok(component_ids_for_prototype) => component_ids.extend(component_ids_for_prototype),
+                                            Err(err) => error!(si.error.message = ?err, "error getting components for schema variant")
+                                        }
+                                        Err(err) => error!(si.error.message = ?err, "error getting schema variant for action prototype")
+                                    },
+                                    _ => {}
+                                }
+                            }
```
nickgerace, line 276: @jhelwig @fnichol @britmyerss @jobelenus this is what we discussed about partial completion vs. error and eject. I don't think our work is done here, but is probably done for the purposes of this PR.

app/web/src/components/ApplyChangeSetButton.vue
```diff
@@ -33,7 +33,20 @@
       />
     </template>
 
-    <ApprovalFlowModal ref="approvalFlowModalRef" votingKind="merge" />
+    <template
+      v-if="
+        featureFlagsStore.FRONTEND_ARCH_VIEWS &&
+        featureFlagsStore.BIFROST_ACTIONS
+      "
```
nickgerace, line 40: For context, the `FRONTEND_ARCH_VIEWS` will die soon.

lib/edda-server/src/lib.rs
```diff
@@ -1,3 +1,5 @@
+#![recursion_limit = "256"]
```
nickgerace, line 1: Going deeper in `materialized_view.rs` was not appreciated by Johnrustc

lib/si-frontend-types-macros/src/lib.rs
```diff
@@ -425,12 +412,8 @@ pub fn build_mv(input: proc_macro2::TokenStream) -> syn::Result<proc_macro2::Tok
                         }),
                         Some(frontend_object),
                     ))
-                } else {
-                    Ok((None, None))
                 }
             }
-
-            #build_mv_ident(#ctx, #frigg, #change, #mv_id)
```
nickgerace, line 433: Needed to allow more arguments to be passed into MV functions

app/web/src/components/NoSelectionDetailsPanel.vue
```diff
@@ -59,7 +59,14 @@
         trackingSlug="no-selection-details-panel"
       >
         <TabGroupItem label="Changes" slug="changes">
-          <ChangesPanelProposed />
+          <template
+            v-if="ffStore.FRONTEND_ARCH_VIEWS && ffStore.BIFROST_ACTIONS"
```
wendybujalski, line 63: The drop in component replacement makes this so clean!

wendybujalski (approved):
The `mead-hall` component replacement strategy really pays off in this PR, awesome work!

PR #5866, opened 2025-04-11
fix(sdf): Remove the feature flag from enqueuing update actions

stack72:
We have been running this in internal workspaces for a while now and the correct actions fire when we make changes to the components

If I make a change to a TaskDefinition then it triggers a task definition update and an ECS Service update - those are the expected actions!x


wendybujalski (approved):
👍

PR #5869, opened 2025-04-11
chore(ci): Update Module Cache after deploying modules to module index

stack72:


.github/workflows/asset-push.yml
```diff
@@ -29,3 +29,15 @@ jobs:
...
+
+      - name: Update Module Cache in Tools Prod
+        run: |
+          curl -X POST "https://tools.systeminit.com/api/v2/admin/update_module_cache" -H "Authorization: Bearer ${{ env.MY_BEARER_TOKEN }}"
+        env:
+          MY_BEARER_TOKEN: ${{ secrets.SI_PRODUCTION_SAAS_WORKSPACE_API_TOKEN }}
+          
+      - name: Update Module Cache in Prod
+        run: |
+          curl -X POST "https://app.systeminit.com/api/v2/admin/update_module_cache" -H "Authorization: Bearer ${{ env.MY_BEARER_TOKEN }}"
```
sprutton1, line 41: Does this return after success or fire and let it run in the background? I ask because I imagine we want to avoid pushing to prod if there are issues with the generated modules in tools.

sprutton1 (changes requested):

.github/workflows/asset-push.yml
stack72: This is an Async API - it's a fire and forget - even in the UI

.github/workflows/asset-push.yml
```diff
@@ -29,3 +29,15 @@ jobs:
         run: |
           cd bin/hoist
           cargo run -- upload-all-specs -y --non-interactive --max-concurrent 25 -t ../clover/si-specs
+
+      - name: Update Module Cache in Tools Prod
+        run: |
+          curl -X POST "https://tools.systeminit.com/api/v2/admin/update_module_cache" -H "Authorization: Bearer ${{ env.MY_BEARER_TOKEN }}"
+        env:
+          MY_BEARER_TOKEN: ${{ secrets.SI_PRODUCTION_SAAS_WORKSPACE_API_TOKEN }}
```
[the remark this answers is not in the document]
stack72: Good shout! I will change that

.github/workflows/asset-push.yml
sprutton1: I guess I worry about pushing a bad update to prod then. We don't really have a way to rollback changes here. Let's talk about it

sprutton1 (approved):
We talked about this and are willing to accept the risk of an issue occurring here. Problematic modules can be manually rejected, malformed ones should never upload, and this exactly mimics the manual process we follow today.

PR #5873, opened 2025-04-11
fix(rebaser): fix edda request data & rebaser/edda race condition

fnichol:


nickgerace (approved):

lib/rebaser-server/src/rebase.rs
```diff
@@ -176,6 +176,10 @@ pub async fn perform_rebase(
     );
     debug!("rebase performed: {:?}", start.elapsed());
     span.record("si.rebase.rebase_time", start.elapsed().as_millis());
+
+    // Before replying to the requester or sending the Edda request, we must commit.
+    ctx.commit_no_rebase().await?;
+
     if features.generate_mvs {
```
nickgerace, line 183: This will not affect prod where the flag is off so 👍

PR #5875, opened 2025-04-11
fix(sdf): Change UPGRADE_ALL modules to be async

stack72:


lib/sdf-server/src/service/module/upgrade_modules.rs
```diff
@@ -33,16 +39,52 @@ pub async fn upgrade_modules(
...
-) -> Result<ForceChangeSetResponse<Vec<FrontendVariant>>, ModuleError> {
+) -> ModuleResult<ForceChangeSetResponse<Ulid>> {
     let mut ctx = builder.build(request_ctx.build(request.visibility)).await?;
 
     let force_change_set_id = ChangeSet::force_new(&mut ctx).await?;
 
-    let mut variants = Vec::new();
+    let task_id = Ulid::new();
+
+    tokio::task::spawn(async move {
```
nickgerace, line 49: I can't remember if we have a tracker in the app state but this should be fine for now
cc @fnichol

nickgerace (dismissed):

stack72:
This works but the UI doesn't indicate really well that an upgrade is in flight - I need to add something more indicative

nickgerace (approved):

PR #5876, opened 2025-04-12
chore: clover schema update

systeminitbot:


stack72:
The app crashed:

```
[bin/hoist/src/diff.rs:328:9] operation = Replace(
    ReplaceOperation {
        path: PointerBuf(
            "/description",
        ),
        value: String("Resource Type definition for AWS::Batch::ConsumableResource"),
    },
)
Message:  Unhandled patch operation
Location: bin/hoist/src/diff.rs:329
```

stack72:
Going to close this out and let it work again tonight with the fixed hoist binary

PR #5881, opened 2025-04-14
Use trait kind method when building MV dependency graph

jhelwig:
The `MaterializedView` trait already provided a method to get the `ReferenceKind` for the MV in question. Rather than manually specifying what the `ReferenceKind` is for the MV you want to add the dependencies for, the helper macro now uses the `MaterializedView` trait method generated from the derive macro.


fnichol (approved):
I believe I remember us talking about this last week, very nice

PR #5888, opened 2025-04-15
feat: add lightweight ci image

sprutton1:


stack72 (approved):

component/ci-light/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM rust:1-alpine
+
+ARG BUILDEVENTS_VERSION=v0.17.0
+RUN apk add curl
```
fnichol, line 4: We could consider adding a couple Git-related packages as well, namely `github-cli openssh-client` as we tend to install these most of the time in the `alpine` images. However, we can also do this in followup work

component/ci-light/BUCK
```diff
@@ -0,0 +1,8 @@
+load(
+    "@prelude-si//:macros.bzl",
+    "docker_image",
+)
+
+docker_image(
+    name = "ci-light",
+)
```
fnichol, line 8: Might be worth extracting the version of the `rust` image (that is the `1-alpine` in this case) into a `build_args` entry so we can bump this version here, much like in the NATS [BUCK file](https://github.com/systeminit/si/blob/main/component/nats/BUCK) and [Dockerfile](https://github.com/systeminit/si/blob/main/component/nats/Dockerfile) setup.

component/ci-light/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+RUN set -eux; \
+    ARCH="$(arch)"; \
+    case "${ARCH}" in \
+        x86_64) BUILDEVENTS_ARCH="linux-amd64" ;; \
+        aarch64) BUILDEVENTS_ARCH="linux-arm64" ;; \
```
fnichol, line 9: Good catch!

fnichol (approved):
Looks good, had a couple of thoughts but these could be follow ups if need be

PR #5890, opened 2025-04-15
feat(web) - fuzzy search highlighting and other connections menu improvements

wendybujalski:
Screenshot 2025-04-15 at 5 19 20 PM

### DIFF
--- a/app/web/src/components/ModelingView/ConnectionMenuSocketList.vue
+++ b/app/web/src/components/ModelingView/ConnectionMenuSocketList.vue
@@ -2,7 +2,7 @@
   <div
     :class="
       clsx(
-        'basis-1/2 h-full min-h-0 border-x-2 border-b-2 p-xs ',
+        'h-full min-h-0 border-x-2 border-b-2 p-xs pt-2xs ',
         active && themeClasses('bg-neutral-100', 'bg-neutral-800'),
         showCodeViewer ? 'children:h-1/2' : 'children:h-full',
       )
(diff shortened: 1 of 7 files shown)


lib/vue-lib/src/design-system/general/TruncateWithTooltip.vue
```diff
@@ -93,8 +93,12 @@ const tooltip = computed(() => {
   // we invoke forceRecompute here to force the tooltip to recompute when we need it to
   // eslint-disable-next-line @typescript-eslint/no-unused-expressions
   forceRecompute.value;
+  if (forceRecompute.value > 100) {
+    // eslint-disable-next-line vue/no-side-effects-in-computed-properties
+    forceRecompute.value = 0;
```
jobelenus, line 98: Why do we need to set this back to 0? This will cause `tooltip` to be computed two times (and if we make a logic error in here, infinite loop)

jobelenus (changes requested):

app/web/src/components/ModelingView/ConnectionMenuSocketList.vue
```diff
@@ -284,4 +160,45 @@ watch(
...
+    return;
+
+  const allHighlightRanges = [] as Range[];
+
+  socketListItemsRef.value.forEach((listItem) => {
+    allHighlightRanges.push(...listItem.highlightRanges);
+  });
+
+  const fuzzySearchHighlight = new Highlight(...allHighlightRanges);
+  CSS.highlights.set("fuzzy-search-highlight", fuzzySearchHighlight);
```
vbustamante, line 196: These will throw in browsers without highlight API support  (firefox). We should check for the `Highlight` and `CSS.highlights` not being undefined before using them

wendybujalski: Ah yeah I will make that change

jobelenus (approved):

vbustamante (approved):

PR #5897, opened 2025-04-16
chore: break out some v1 routes into crates

zacharyhamm:
Timeboxing an hour a day to break out new crates for sdf routes or dal functionality.

This moves:
  1. v1/action
  2. v1/attribute
  3. v1/change-set
  4. v1/component
  5. v1/diagram

Into separate crates. This reduces the size of sdf-server by about 5000 lines of code.


nickgerace (approved):
I love it! An hour a day is a good idea

jhelwig (approved):

lib/sdf-v1-routes-actions/src/retry.rs
```diff
@@ -25,6 +22,7 @@ pub async fn retry(
     AccessBuilder(request_ctx): AccessBuilder,
     Json(request): Json<RetryRequest>,
 ) -> ActionResult<()> {
+    println!("test");
```
fnichol, line 25: I don't think we want this?

lib/sdf-v1-routes-actions/BUCK
```diff
@@ -0,0 +1,22 @@
...
+rust_library(
+    name = "sdf-v1-routes-actions",
+    deps = [
+        "//lib/dal:dal",
+        "//lib/sdf-core:sdf-core",
+        "//lib/sdf-extract:sdf-extract",
+        "//lib/si-events-rs:si-events",
+        "//lib/si-layer-cache:si-layer-cache",
+        "//lib/telemetry-rs:telemetry",
+
```
fnichol, line 12: I wouldn't mind if we didn't add the empty lines here--the idea was simply to lexical sort this list of deps, not necessarily to delineate between "our deps" and "other deps". In the Buck2 instance, these happen to sort apart. I may be able to help with a future lint/format rule, but just find it harder to keep this sorting alphabetical when there are blank lines in the middle.

fnichol (approved):
Excellent work!

I only saw one `println!()` which we can pull out in a future commit

PR #5898, opened 2025-04-16
chore(cleanup): Remove Asset Sprayer

stack72:
We no longer use the AssetSprayer work for our CodeGen so we can remove the need to use it in our builds!


zacharyhamm (approved):
RIP asset-sprayer

zacharyhamm (approved):

PR #5901, opened 2025-04-16
feat: add basic retries to the init container

sprutton1:
basic retry in case we hit ssm rate limits. Manually built and tested in tools prod


jhelwig (approved):

fnichol (approved):
I was going to call out a potential problem with `[ $? -eq 0]` but then realized we aren't using `set -e` so I guess we're ok, but given that we're not using `set -e` there's probably a few things that could go wrong with this script that will silently fail.

PR #5902, opened 2025-04-16
fix(sdf): Report errors from management functions correctly

jkeiser:
When management functions have errors, we report their status differently depending on whether it's in History or is the latest management function run. This PR makes them uniform and shows the correct success / error icon depending on what is reported.

image

* Makes manage/history and manage/latest APIs return the same shape, including both execution state and actionResultState.
* Calculates the status icon using a common funcRunStatus() function, treating these cases as errors:
  * Function throws exception
  * Function returns `{ status: "error", message: "message" }`
  * Function returns `{ schtatus: "ok" }`
* Uses appropriate icons (similar to other status indicators). In particular, "running" states now show a spinner, just like other running status indicators.

Fixes BUG-792.

**Factor Budget:** I changed a couple of our enums to string unions, which works a tiny bit better with typescript type checking and is the recommendation of the TS community for string enums, as far as I could find one.

### Out of Scope

These cases either report the wrong status or don't update run status at all, and fall outside the scope of this UI-only fix:

* When `ops` contains invalid data or otherwise cannot be deserialized, we still treat status as Success instead of Failure
* Function timeouts and early terminations don't get updated run status and show perennially as Running
* There are other failure cases in the runner where we still don't update func run status, but these are harder to trigger

Additionally, there are reactivity issues showing long-running management functions n history due to the lack of a FuncRunStarted WsEvent. Finally, when the function fails, management function interface returns 400 and the "last run" does not immediately show failure because the store is listening for success, though it does show up in history. All of these are fixed by clicking away from history and coming back in.

## Testing

- [x] Test that exceptions, bad return payloads, and successful "return error" JS runs all report correct errors
- [x] Test that successful and long-running functions report status as expected
- [x] Integration tests still succeed

### DIFF
--- a/app/web/src/components/ManagementRunPrototype.vue
+++ b/app/web/src/components/ManagementRunPrototype.vue
@@ -60,7 +60,7 @@ import { useFuncStore, MgmtPrototype } from "@/store/func/funcs.store";
 import {
   FuncRun,
   FuncRunId,
-  FuncRunState,
+  funcRunStatus,
   useFuncRunsStore,
 } from "@/store/func_runs.store";
 import { useManagementRunsStore } from "@/store/management_runs.store";
@@ -142,17 +142,7 @@ const lastExecution = computed<FuncRun | null>(() => {
   }
 });
 
-const lastExecutionState = computed<"ok" | "error" | null>(() => {
-  switch (lastExecution.value?.state) {
-    case FuncRunState.Success:
-      return "ok";
-    case FuncRunState.Failure:
-      return "error";
-    default:
-      return null;
-  }
-});
-
+const lastExecutionState = computed(() => funcRunStatus(lastExecution.value));
 const runPrototype = async (viewId: ViewId) => {
   funcStore.RUN_MGMT_PROTOTYPE(
     props.prototype.managementPrototypeId,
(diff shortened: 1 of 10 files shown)


app/web/src/components/ManagementRunPrototype.vue
```diff
@@ -142,17 +142,7 @@ const lastExecution = computed<FuncRun | null>(() => {
   }
 });
 
-const lastExecutionState = computed<"ok" | "error" | null>(() => {
```
jobelenus, line 145: It feels like _this part_ is the main part of the problem (fixed by `funcRunStatus`)? We were only checking 1 state field, when really there are 2 state fields to check?? (The rest of the changes seem to just be settling on the struct/type names and format?)

jobelenus (approved):

wendybujalski (approved):
Excellent work! And I appreciate your thorough comments - some of those things that were out of scope are definitely things we will need to tackle sometime in the future.

app/web/src/components/ManagementRunPrototype.vue
jkeiser: Yep, that's it! One of the places used one state, the other used a different state. It was not strictly necessary to massage the shapes of history and latest into the same shape, but it *was* necessary to make sure both states were sent back in history.

PR #5904, opened 2025-04-17
chore: break out all v1 variants

adamhjk:
* new crate sdf-v1-routes-variant
* new crate sdf-v1-routes-node-debug
* new crate sdf-v1-routes-module
* new crate sdf-v1-routes-graphviz
* new crate sdf-v1-routes-qualification
* new crate sdf-v1-routes-secret
* new crate sdf-v1-routes-session


lib/sdf-core/src/async_route.rs
zacharyhamm, line 1: was this deleted from sdf-server? surprised it's not marked as a file rename

lib/sdf-v1-routes-session/src/lib.rs
```diff
@@ -59,28 +55,14 @@ pub enum SessionError {
...
-            }
-            Self::WorkspacePermission(_) => (StatusCode::UNAUTHORIZED, self.to_string()),
-            Self::AuthApiError(_) => (StatusCode::UNAUTHORIZED, self.to_string()),
-            _ => (StatusCode::INTERNAL_SERVER_ERROR, self.to_string()),
-        };
-
-        ApiError::new(status_code, error_message).into_response()
-    }
-}
+impl_default_error_into_response!(SessionError);
```
zacharyhamm, line 65: We don't want to get rid of our custom error handling for SessionError

zacharyhamm (commented):
That change to SessionError's IntoResponse should definitely be reverted. And async_route should be removed from sdf-server. Otherwise looks solid

lib/sdf-v1-routes-session/src/lib.rs
adamhjk: fixed!

lib/sdf-core/src/async_route.rs
adamhjk: fixed!

zacharyhamm (approved):

PR #5914, opened 2025-04-18
Only build MVs where the MV ID is immediate

nickgerace:
## Description

This PR ensures that we only build MVs where the MV ID is the same as the trigger's change ID or is the change set ID. This ensures that we do not run into situations where deleted nodes cannot be found when "hopping" to the desired MV ID from a change ID.

As a result of these changes, this is the first commit where we have removed an existing MV. It is no longer in use. We are not concerned about backwards compatibility here since it is not deserialized anywhere.

This PR also fixes the issue where the icons for action prototypes were all question marks.

### DIFF
--- a/app/web/src/mead-hall/AssetActionsDetails.vue
+++ b/app/web/src/mead-hall/AssetActionsDetails.vue
@@ -26,6 +26,7 @@
             v-for="actionPrototypeView in actionPrototypeViews"
             :key="actionPrototypeView.id"
             :actionPrototypeView="actionPrototypeView"
+            :actionId="actionByPrototype[actionPrototypeView.id]"
             :component="props.component"
           />
         </div>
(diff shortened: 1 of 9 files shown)


app/web/src/mead-hall/AssetActionsDetails.vue
```diff
@@ -71,34 +71,59 @@ function onTabSelected(newTabSlug?: string) {
...
+
+  const result: Record<ActionPrototypeId, ActionId> = {};
+  for (const action of actionViewList.data.value.actions) {
+    if (action.componentId === props.component.def.id) {
+      // NOTE(nick): this assumes that there can be one action for a given prototype and component.
+      // As of the time of writing, this is true, but multiple actions per prototype and component
+      // aren't disallowed from the underlying graph's perspective. Theorhetically, you could
+      // enqueue two refreshes back-to-back. What then? I don't think we'll expose an interface to
+      // do that for awhile, so this should be sufficient.
+      result[action.prototypeId] = action.id;
```
nickgerace, line 121: @jhelwig added a comment here for what we discussed

jhelwig (approved):

PR #5916, opened 2025-04-18
feat(innit): client and mtls

sprutton1:
* adds a client to interact with innit
  *  includes the ability to pass a client auth cert if one is supplied
  * has basic endpoints for getting and checking health
  * has tests that generate a ca and cert, spins up a server, and makes a single request for health
*  adds a ca_cert config to innit-server and relevant middleware for using it to check against client certs if present
* si-tls extended to support key pairs and pem/der certs
  * some refactoring here for less repetition


lib/innit-server/Cargo.toml
```diff
@@ -13,17 +13,23 @@ innit-core = { path = "../../lib/innit-core" }
 si-data-ssm = { path = "../../lib/si-data-ssm" }
 si-settings = { path = "../../lib/si-settings" }
 si-std = { path = "../../lib/si-std" }
+si-tls = { path = "../../lib/si-tls" }
 telemetry = { path = "../../lib/telemetry-rs" }
 
 axum = { workspace = true }
+axum-server = { workspace = true }
```
fnichol, line 20: I believe this crate isn't being used anywhere as I don't see it in init-server's `BUCK` file, is that right?

Cargo.toml
```diff
@@ -133,6 +134,7 @@ axum = { version = "0.6.20", features = [
     "multipart",
     "ws",
 ] } # todo: upgrade this alongside hyper/http/tokio-tungstenite
+axum-server = { version = "0.7.2", features = ["tls-rustls-no-provider"] }
```
fnichol, line 137: I think we don't need this?

lib/innit-client/src/config.rs
```diff
@@ -0,0 +1,108 @@
...
+type Result<T> = std::result::Result<T, ConfigError>;
+
+/// The config for the forklift server.
+#[derive(Debug, Builder)]
+pub struct Config {
+    #[builder]
+    auth_config: AuthConfig,
+
+    #[builder]
+    base_url: Url,
```
fnichol, line 36: Nice, I love these builders that result in a nice and validated type!

lib/innit-client/tests/integration_test/mod.rs
```diff
@@ -0,0 +1,111 @@
...
+        let server = Server::http(config, token.clone())
+            .await
+            .expect("should create server");
+        (server, token)
+    }
+
+    #[tokio::test]
+    async fn test_client_server_mtls_interaction() {
+        let certs = create_certs();
+        let addr = SocketAddr::from(([0, 0, 0, 0], 5166));
```
fnichol, line 58: I *believe* you can pass a port of `0` in here which tells the kernel to assign a random/unallocated port for this socket. Then you can ask for that port number via [`addr.port()`](https://doc.rust-lang.org/std/net/enum.SocketAddr.html#method.port) which should allow for better test concurrency in the future.

Ok, I'm pretty sure this works, according to [users.rust-lang.org](https://users.rust-lang.org/t/what-does-the-port-0-represent/122583/4)

lib/si-tls/src/lib.rs
```diff
@@ -33,67 +36,117 @@ pub enum TlsError {
...
-            CertificateSource::Path(path) => Self::get_certificate_from_path(path).await,
-            CertificateSource::Base64(data) => {
-                Self::get_certificate_from_base64(data.clone()).await
+    /// Load certs as bytes from any CertificateSource
+    pub async fn load_certificates_as_bytes(&self) -> Result<Vec<u8>> {
+        match self {
+            Self::AsString(string) => Ok(string.as_bytes().to_vec()),
+
+            Self::Base64(data) => Ok(STANDARD.decode(data)?),
+            Self::Path(path) => Ok(std::fs::read(path)?),
```
fnichol, line 57: There is also a `tokio::fs::read(path).await?` call that would flip this to async...

lib/si-tls/src/lib.rs
```diff
@@ -33,67 +36,117 @@ pub enum TlsError {
...
-        path: impl AsRef<Path>,
-    ) -> Result<Vec<CertificateDer<'static>>> {
-        let buf = std::fs::read(path)?;
-        Self::get_certificate_from_bytes(&buf).await
+    /// Load private keys as bytes from any CertificateSource
+    pub async fn load_private_key_as_bytes(&self) -> Result<Vec<u8>> {
+        match self {
+            Self::AsString(string) => Ok(string.as_bytes().to_vec()),
+            Self::Base64(data) => Ok(STANDARD.decode(data)?),
+            Self::Path(path) => Ok(std::fs::read(path)?),
```
fnichol, line 117: Same as above (i.e. there is a Tokio alternative call)

lib/innit-server/src/middleware/client_cert_auth.rs
```diff
@@ -0,0 +1,83 @@
...
+            .ok_or_else(|| ClientCertError::InvalidFormat("No certificate found".into()))?
+            .clone();
+
+        return Ok(cert);
+    }
+
+    Err(ClientCertError::MissingCertificate)
+}
+
+pub async fn verify_client_cert_middleware(
```
fnichol, line 66: Very cool!

fnichol (dismissed):
I had a couple of questions but otherwise this is looking great!!

Cargo.toml
sprutton1: Ah, good catch. I used it earlier on in the build out but didn't end up needing it at the end

stack72 (approved):

lib/innit-server/src/server.rs
```diff
@@ -37,6 +42,12 @@ pub struct Server<I> {
     token: CancellationToken,
 }
 
+impl Server<AddrIncoming> {
+    pub fn bound_port(&self) -> u16 {
+        self.inner.local_addr().port()
+    }
```
fnichol, line 48: Nice, yeah!

fnichol (approved):
Nice!

PR #5917, opened 2025-04-18
Add base contents for job-specific UI

nickgerace:
## Description

First and foremost, I want to be very clear that @jobelenus wrote all of this. Wendy and I mainly polished it up.

This PR adds the "big boom, foundational pieces and solves some problems" for the new UI. In addition, it showcases a README with lines of achievement for the new UI... all written by @jobelenus.

This PR also adds a feature flag for the new UI and uses router re-direction for gating the feature.

### DIFF
--- a/app/web/src/newhotness/Component.vue
+++ b/app/web/src/newhotness/Component.vue
@@ -1,0 +1,64 @@
+<template>
+  <section class="grid gap-md h-[100%]">
+    <div>Name</div>
+    <div class="row-span-2">Qualifications</div>
+    <div class="row-span-4 grid grid-cols-subgrid gap-sm">
+      <div>Documentation</div>
+      <div>Actions</div>
+      <div>Connections</div>
+    </div>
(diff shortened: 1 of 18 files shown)


adamhjk:
My only concern here is dropping tailwind. There is a dramatic change in how non-experts can get things done using tailwind, vs having to understand how CSS interacts (much less how the particular variants of CSS interact). I certainly wouldn't block the merge, but it should not be done without real thought (how many people on the team discovered they could work on teh frontend in no small part because tailwind let them deal with styling? a lot, I suspect).

adamhjk:
that said - merge away. :)

nickgerace:
@adamhjk 100% agree. We will replace those styles. I will block us dropping Tailwind too if it comes to that, but we're all in agreement it should be there. We believe this is mergeable with that in mind. Good call out though :)

stack72 (approved):

PR #5921, opened 2025-04-20
fix(sdf): Change how generate_template works

stack72:
We are removing the complexity of the generate_template function and are now keeping overrides and defaults anymore as well as the views array

We are also removing thesync to views function as well


adamhjk (approved):
looks great. ship it. we should have aaron update the existing templates to, so they get simpler.

PR #5923, opened 2025-04-21
chore(ci): Ensure Edda is part of our weekend restart policy

stack72:


.github/workflows/deploy-service-restart.yml
```diff
@@ -45,6 +45,14 @@ jobs:
       environment: ${{ inputs.environment }}
       service: pinga
     secrets: inherit
+    
+  restart-edda:
+    needs: restart-edda
```
jhelwig, line 50: I don't think it's supposed to depend on itself?

jhelwig (approved):

PR #5927, opened 2025-04-21
chore: upgrade to Rust Edition 2024

fnichol:
See: <https://doc.rust-lang.org/nightly/edition-guide/rust-2024/index.html>


DOCS.md
```diff
@@ -750,7 +750,7 @@ To add a new Rust library, there are a few steps:
 ```toml
 [package]
 name = "MY-LIBRARY"
-edition = "2021"
+edition = "2024"
```
nickgerace, line 753: Thank you for updating the docs!

lib/audit-database/Cargo.toml
```diff
@@ -1,10 +1,10 @@
 [package]
 name = "audit-database"
-edition = "2021"
 version.workspace = true
 authors.workspace = true
 license.workspace = true
 repository.workspace = true
+edition.workspace = true
```
nickgerace, line 7: Nice!!

lib/dal-test/src/helpers.rs
```diff
@@ -37,7 +37,7 @@ pub fn generate_fake_name() -> Result<String> {
 #[allow(clippy::expect_used)]
 #[macro_export]
 macro_rules! connection_annotation_string {
-    ($str:expr) => {
+    ($str:expr_2021) => {
```
nickgerace, line 40: Interesting... we may want to clean these up later

nickgerace (approved):
This is excellent work! To be clear, I am approving this without having reviewed every line of every file knowing that 99% of changes are imports related. I do see some care for the `expr` changes and workspace value setting too. This looks great.

lib/dal-test/src/helpers.rs
fnichol: Yeah, this was the result of running `cargo fix --edition` so is perhaps a backwards compat thing

PR #5932, opened 2025-04-22
feat: add reusable qualifunc override pattern

aaron-dernley:
- Add a qualification function to the AWS::EC2::SecurityGroupIngress component which ensures failure if a user supplies both a GroupId and GroupName (only one is required on creation, two will fail)


bin/clover/src/pipeline-steps/assetSpecificOverrides.ts
```diff
@@ -854,6 +869,41 @@ function attachExtraActionFunction(
...
+
+  const func = createFunc(
+    name,
+    "jsAttribute",
+    "qualification",
+    strippedBase64(funcCode),
+    uniqueId,
+    [
+      {
+        name: "domain",
```
stack72, line 888: nice!

stack72:
Screenshot 2025-04-23 at 00 26 04

Looks great!!

stack72 (approved):

PR #5934, opened 2025-04-22
fix(luminork): Refactor the code to add connections to share between create and update components

stack72:


lib/luminork-server/src/service/v1/components/connections.rs
```diff
@@ -0,0 +1,206 @@
...
+    #[schema(example = json!({"from": {"component": "OtherComponentName", "socketName": "output"}, "to": "ThisComponentInputSocketName"}))]
+    #[schema(example = json!({"from": {"componentId": "01H9ZQD35JPMBGHH69BT0Q79VY", "socketName": "output"}, "to": "ThisComponentInputSocketName"}))]
+    Incoming { from: ConnectionPoint, to: String },
+    #[schema(example = json!({"from": "ThisComponentOutputSocketName", "to": {"component": "OtherComponentName", "socketName": "InputSocketName"}}))]
+    #[schema(example = json!({"from": "ThisComponentOutputSocketName", "to": {"componentId": "01H9ZQD35JPMBGHH69BT0Q79VY", "socketName": "InputSocketName"}}))]
+    Outgoing { from: String, to: ConnectionPoint },
+}
+
+/// Returns the component ID if found, or appropriate error if not found or if duplicate names exist
+pub async fn find_component_id_by_name(
```
britmyerss, line 46: we should move these helpers to the dal, but can be in a follow up pass late.

britmyerss (approved):

PR #5936, opened 2025-04-23
style(fmt): reformat Rust code to enforce consistent group imports

fnichol:
This change uses 3 unstable Rustfmt configuration options to enforce a
more consistent import grouping, sorting, and layout that can be
maintained across all code editors and IDEs. It relies on prior work to
use a "nightly" Rustfmt and augments the `rustfmt.toml` configuration.

The driver for these changes comes from observing that the primary
source of Git merge conflicts (i.e. when rebasing branches) comes from
inconsistent changes in the "imports" section of a Rust module file. In
most cases these conflicts are trivial and not that important, but
clearing each of the merge conflicts can be mentally taxing.

The rationale for the style choices are informed by:

- A commonly agreed upon Rust convention is to group imports into 3
  groups in the following order:

  1. `std` and `core` imports, lexically sorted
  2. external crate imports, lexically sorted
  3. local modules from within the current crate (i.e.
     `crate::*`/`super::*`/`self::*` imports), lexically sorted

  Some more discussion can be found in a [Rustfmt
  issue](https://github.com/rust-lang/rustfmt/issues/2979)
- An import formatting strategy for this repository needs to be as "Git
  friendly" as possible, as it relates to merge conflicts. As more
  developers submit more close and overlapping work in this codebase,
  the chances of "import conflict churn" has risen. A format that allows
  for easier additions and deletions is preferred and so a
  one-import-per-line is highly desirable as Git tends to do better at
  tracking differences across lines rather than within long lines.
- A way of grouping the origin of the imports and the relative "weight"
  of each source can help the developer understand the dependency tree
  making refactoring work easier. In this case, grouping imports by
  Crate means that all imports from, say the DAL crate would be grouped
  and indented together.


jhelwig (dismissed):
Configuring rustfmt like this, and getting the initial pass out of the way will definitely cut down on the merge/rebase conflicts. 👍🏻

nickgerace:
I added this to the queue not seeing it already failed once. I approve though :)

jhelwig (approved):

PR #5940, opened 2025-04-23
fix(web): Restore showing action details when selecting an Action

britmyerss:
This fixes a regression where users could no longer view the logs/arguments/etc. for actions. This is because we were fetching the func runs details using the `actionId` and not the `funcRunId`.

Tested manually by running actions and seeing their details both when the action is still enqueued, and when it's no longer enqueued. 

<div><br/>via <a href="https://giphy.com/gerbert/">Gerbert!</a> on <a href="https://giphy.com/gifs/gerbert-puppet-VL5JGMwlSOFsIuc8z0">GIPHY</a></div>

### DIFF
--- a/app/web/src/components/ChangesPanelHistory.vue
+++ b/app/web/src/components/ChangesPanelHistory.vue
@@ -23,7 +23,7 @@
             :key="detail.changeSetId"
             :actions="actions"
             :changeSet="getChangeSet(detail)"
-            :clickAction="clickActionOrMgmtRun"
+            :clickAction="clickActionRun"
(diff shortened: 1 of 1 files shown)


stack72 (approved):

jkeiser (approved):
Doh! Looks good. Down the road I suspect there's a similar factor coming to what happened with ManagementHistoryItem, since ActionHistoryView is basically a FuncRun; but there may be other type shenanigans with the intersection between that and Action, so this is the right fix until we have time to think all that through.

stack72 (approved):

PR #5941, opened 2025-04-23
Add patch support for component diff

nickgerace:
## Description

This PR adds patch support for component diff. This is needed for the new UI in order to count how many fields changed in a component in a given change set.

In addition to these changes, the original diff code has been refactored for performance and readability. The existence of a component is now covered independently of its other error scenarios, null value creation is cleaner and the concept of multiple diffs has been removed. The original idea (and I'm dusting off cobwebs here) was that you could have multiple diffs corresponding to code generations. We may want that again in the future, but it is clear that an optional diff is the better move for the world as it exists today.


wendybujalski (dismissed):

wendybujalski (dismissed):

wendybujalski (approved):

lib/dal/src/component/diff.rs
```diff
@@ -1,78 +1,45 @@
-//! This module contains [`ComponentDiff`].
+//! This module contains functionalk
```
nickgerace, line 1: Whoops! I'll fix this in post.

PR #5945, opened 2025-04-23
feat(dal): Support AVs subscribing to other AVs via JSON pointers

jkeiser:
This PR allows an attribute to connected to any other attribute of another component via "subscription" (e.g. "I want /domain/foo/bar from component X").

* `AttributeValue::subscribe(av_id, other_root_av_id, "/domain/SubnetId")` will connect that AV to another via subscription. It sets the prototype to si:identity and creates one argument with a new ValueSubscription edge. The edge contains the JSON pointer inside the edge itself, and the edge points to the root attribute value of the other component.
* DependentValuesUpdate flows values through subscriptions just like any other connection
* resolve_value_subscription() uses the `jsonptr` crate to parse the pointer and look up the target AV
* Subscriptions to AVs that don't currenlty exist are considered a "missing value," not an error, similar to what would be yielded by `si:unset`

## Out of Scope

Out of scope features/bugs here will not affect users, since subscriptions cannot be created through the API or UI yet.

- UI/UX/API to do this is for the future
- Arity=many: multiple subscriptions from one value are currently unsupported (multiple subscriptions *to* a single value are supported).
- Action dependencies do not respect value subscriptions. We know the fix here, but are putting it off to the next PR.
- Component Upgrade / Regenerate Variant drops subscriptions
- Copy/Paste does not copy subscriptions
- bug: Removing an array element does not currently trigger DVU. We'll need to fix this, but believe we know the fix.

## Risk Management

* We have explored the Out of Scope bits enough to believe the base design can support their implementation
* We placed the new subscription path on the new Edge into an Enum, so it can be versioned if we want to make new kinds of path to new places
* Subscriptions cannot be made in real workspaces, so if we absolutely have to make fundamental changes existing workspaces will not be affected

## Testing

- [X] New tests covering subscription via object, array, and map props, with update and unsetting of values, flowing through DVU
- [X] Existing tests work
- [ ] Manual tests are important and have not been done!
- [ ] Test default value resolution


lib/dal/src/attribute/prototype/debug.rs
```diff
@@ -162,21 +163,43 @@ impl AttributePrototypeDebugView {
...
-                        .await?
-                        .ok_or(
-                            AttributeValueError::AttributePrototypeArgumentMissingValueSource(
-                                attribute_prototype_arg_id,
-                            ),
-                        )?;
+                    AttributePrototypeArgument::value_source(ctx, attribute_prototype_arg_id)
+                        .await?;
                 let values_for_arg = match value_source {
+                    ValueSource::ValueSubscription(ref subscription) => {
```
zacharyhamm, line 169: great to see this added to the debug

lib/dal/src/workspace_snapshot/edge_weight.rs
```diff
@@ -93,7 +100,7 @@ impl EdgeWeightKind {
 
 #[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq)]
 pub struct EdgeWeight {
-    kind: EdgeWeightKind,
+    pub kind: EdgeWeightKind,
```
zacharyhamm, line 103: Beginning to wonder why EdgeWeightKind isn't just EdgeWeight. but it's too late for that :)

lib/dal/src/workspace_snapshot/graph/v4/attribute_prototype/argument.rs
```diff
@@ -0,0 +1,100 @@
...
+        },
+    },
+};
+
+impl AttributePrototypeArgumentExt for WorkspaceSnapshotGraphV4 {
+    fn value_source_opt(
+        &self,
+        apa_id: AttributePrototypeArgumentId,
+    ) -> WorkspaceSnapshotGraphResult<Option<ValueSource>> {
+        let apa_index = self.get_node_index_by_id(apa_id)?;
```
zacharyhamm, line 35: I'm curious. Why implement this as trait extension instead of as a method on AttributePrototypeArgument ?

adamhjk:
I love this design.

lib/dal/src/workspace_snapshot/graph/v4/component.rs
```diff
@@ -102,3 +107,20 @@ impl WorkspaceSnapshotGraphV4 {
         Ok(results)
     }
 }
+
+impl ComponentExt for WorkspaceSnapshotGraphV4 {
+    fn root_attribute_value(
+        &self,
+        component_id: ComponentId,
+    ) -> WorkspaceSnapshotGraphResult<AttributeValueId> {
+        let component_index = self.get_node_index_by_id(component_id)?;
```
zacharyhamm, line 116: doesn't this functionality already exist? just wondering why an extension trait. My question here is because this means these will have to be implemented twice now (one for this, one for the split graph)

zacharyhamm (commented):
I think the use of the json pointer here for value sources is straightforward and even elegant, really like this. Is it possible to create these kinds of connections via the frontend yet?

lib/dal/src/attribute/value/dependent_value_graph.rs
```diff
@@ -186,6 +191,58 @@ impl DependentValueGraph {
             let current_component_id =
                 AttributeValue::component_id(ctx, current_attribute_value.id()).await?;
 
+            // Look for subscription edges to this AV, and resolve the paths to figure out what
```
britmyerss, line 194: super helpful comments here - thank you!

lib/dal/src/attribute/value/subscription.rs
```diff
@@ -0,0 +1,36 @@
...
+/// A subscription to an attribute value: the root value and path relative to that value
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct ValueSubscription {
+    // The root attribute value
+    pub attribute_value_id: AttributeValueId,
+    // The path to the actual attribute value, relative to the root
+    pub path: ValueSubscriptionPath,
+}
+
+/// A path to an attribute value, relative to its root value
```
britmyerss, line 18: love leaving this door open given the discussions we've had

lib/dal/src/component.rs
```diff
@@ -1656,19 +1660,17 @@ impl Component {
...
 
         Ok(())
     }
 
     pub async fn clear_resource(&self, ctx: &DalContext) -> ComponentResult<()> {
-        let av_for_resource = self
-            .attribute_value_for_prop(ctx, &["root", "resource"])
-            .await?;
+        let av_for_resource =
+            Component::attribute_value_for_prop(ctx, self.id(), &["root", "resource"]).await?;
```
britmyerss, line 1673: nit: `Self::attribute_value_for_prop`

lib/dal/src/workspace_snapshot/edge_weight.rs
jkeiser: Yeah. Next graph migration, maybe :)

lib/dal/src/workspace_snapshot/graph/v4/attribute_prototype/argument.rs
jkeiser: The rubric in my head is, common things that make multiple graph moves and/or need detailed graph knowledge, should live in graph land where it can be centralized and written a little more easily. Right now, extension traits are the blessed way to move something there. I'm not married to that rubric :)

lib/dal/src/workspace_snapshot/graph/v4/component.rs
jkeiser: Oh, we can remove this one I think--we stopped using it when we moved the edge to point to AttributeValue instead of Component.

jkeiser: Removed.

lib/dal/src/workspace_snapshot/graph/v4/attribute_prototype/argument.rs
jkeiser: Moved out of traits for now during the split-graph transition period!

zacharyhamm (approved):
Great work! Extensive testing, didn't cause any regressions in a quick test of DVUs.

lib/dal/src/func/binding/attribute_argument.rs
```diff
@@ -94,46 +94,33 @@ impl AttributeArgumentBinding {
 
     pub async fn assemble(
         ctx: &DalContext,
-        attribute_prototype_argument_id: AttributePrototypeArgumentId,
+        apa_id: AttributePrototypeArgumentId,
```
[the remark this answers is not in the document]
jkeiser: Yeah, I'm guessing we started to mess with this but ultimately plan to move it when we report these via API and UI (since that's what the caller is using it for).

PR #5949, opened 2025-04-24
feat: use innit-client in layered load

sprutton1:
* add generic provider for getting parameters into layered_load
* thread that sucker through to si_settings and up to the bins
* do so using divergent (copy paste) implementations to avoid blast radius
* get from global path `/si/${env}/global` and `/si/${env}/${service}` to layer global and service-specific configs
* tell luminork to use the layered_load that gets from parameter store, passing in innit-client

```bash
❯ curl http://localhost:5166/parameters/si/tools/ | jq
{
  "parameters": [
    {
      "name": "/si/tools/global/pg/user",
      "value": "si"
    },
    {
      "name": "/si/tools/global/pg/hostname",
      "value": "abc"
    },
    {
      "name": "/si/tools/global/pg/dbname",
      "value": "butt"
    },
    {
      "name": "/si/tools/sdf/pg/hostname",
      "value": "derpdoop"
    },
    {
      "name": "/si/tools/luminork/pg/hostname",
      "value": "def"
    }
  ]
}
```

For luminork, it will generate the following pg config

```json
{
    "user": "si",
    "password": "",
    "dbname": "butt",
    "application_name": "luminork",
    "hostname": "def"
}
```

### DIFF
--- a/bin/luminork/src/args.rs
+++ b/bin/luminork/src/args.rs
@@ -12,7 +12,9 @@ use luminork_server::{
     Config,
     ConfigError,
     ConfigFile,
+    ConfigMap,
     FeatureFlag,
+    ParameterProvider,
     StandardConfigFile,
     WorkspacePermissions,
     WorkspacePermissionsMode,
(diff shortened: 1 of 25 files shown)


bin/luminork/src/args.rs
```diff
@@ -262,149 +253,168 @@ impl Args {
...
+where
+    P: ParameterProvider + 'static,
+{
+    ConfigFile::layered_load_with_provider::<_, P>(NAME, provider, move |config_map| {
+        build_config_map(args, config_map);
+    })
+    .await?
+    .try_into()
+}
+
```
sprutton1, line 417: This divergent behavior won't be repeated with the other clients once we verify this pattern works with luminork. We'll just replace the `try` calls entirely.

bin/luminork/src/main.rs
```diff
@@ -101,7 +91,13 @@ async fn async_main() -> Result<()> {
         )
         .await
     } else {
-        let config = Config::try_from(args)?;
+        debug!("creating innit-client...");
+        let provider = Some((
+            InnitClient::new(InnitConfig::builder().build()?).await?,
+            get_host_environment(),
+        ));
```
sprutton1, line 98: I don't super love doing this here and it may move as we'll need to pass some actual config to innit or have it discover it sing env vars

lib/config-file/src/layered_load.rs
```diff
@@ -155,6 +145,121 @@ where
...
+                .set_override(key, value)
+                .expect("tried to set programmatic value that was impossible! bug!");
+        }
+    }
+
+    // Parameter provider (if provided)
+    if let Some((provider, environment)) = parameter_provider {
+        let param_source = ParameterSource::new(provider, environment, app_name.to_string());
+        builder = param_source.load(builder).await?;
+    }
```
sprutton1, line 254: This is the actual change

lib/config-file/src/parameter_provider.rs
```diff
@@ -0,0 +1,105 @@
...
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Parameter {
+    pub name: String,
+    pub value: Option<String>,
+}
+
+#[async_trait]
+pub trait ParameterProvider: Send + Sync {
+    async fn get_parameters_by_path(&self, path: String) -> Result<Vec<Parameter>, ParameterError>;
+}
```
sprutton1, line 24: Providing a trait here to get around cyclic dependencies as basically everything depends on this crate eventually. This will let us swap out clients later, too, in case we ever need to do that. Neat.

lib/dal/examples/rebase/main.rs
```diff
@@ -36,7 +36,7 @@ fn write_snapshot_graph(path: &str, graph: &WorkspaceSnapshotGraph) -> Result<()
 }
 
 fn main() -> Result<()> {
```
sprutton1, line 38: my ide would not stop squawking about this. unrelated.

lib/innit-client/src/config.rs
```diff
@@ -104,7 +113,33 @@ pub fn detect_and_configure_development(config: &mut ConfigFile) -> Result<()> {
     }
 }
 
-fn buck2_development(_config: &mut ConfigFile) -> Result<()> {
+fn buck2_development(config: &mut ConfigFile) -> Result<()> {
```
sprutton1, line 116: Technically still unused until I generate certs

lib/innit-client/src/lib.rs
```diff
@@ -85,18 +85,43 @@ impl InnitClient {
...
     pub async fn get_parameters_by_path(&self, path: String) -> Result<ListParametersResponse> {
-        let url = self.base_url.join("parameters")?.join(&path)?;
+        let url = self.join_path("parameters", &path)?;
         let resp = self.client.get(url).send().await?.error_for_status()?;
         let parameters = resp.json::<ListParametersResponse>().await?;
 
         Ok(parameters)
     }
+
+    fn join_path(&self, base_segment: &str, path: &str) -> Result<Url> {
```
sprutton1, line 103: This is really stupid. Subsequent `.join` calls were smashing over each other due to `/` placement in the path, so this gets around that

lib/innit-server/src/routes/list_parameters.rs
```diff
@@ -17,6 +17,12 @@ pub async fn list_parameters_route(
         ..
     }): State<AppState>,
 ) -> Result<Json<ListParametersResponse>, AppError> {
+    let path = if !path.starts_with('/') {
+        format!("/{}", path)
+    } else {
+        path
+    };
+
```
sprutton1, line 25: axum strips the `/` off, but we need it

lib/si-settings/src/lib.rs
```diff
@@ -50,4 +49,32 @@ pub trait StandardConfigFile:
...
+            parameter_provider,
+            set_func,
+        )
+        .await
+        .map_err(SettingsError::ConfigFile)
+        .map_err(Into::into)
+    }
+}
+
+#[allow(clippy::disallowed_methods)]
```
sprutton1, line 77: Maybe there's a better place for this? I need to know a few things when innit-client does its work and I figured pulling out of env vars would be best. I suppose I could let innit-client run layered-load without the parameter bits to do this?

stack72 (approved):

lib/innit-server/src/routes/list_parameters.rs
nickgerace: sheesh

lib/si-settings/src/lib.rs
nickgerace: I think this is the right call for now. I'd consider moving `local` to a const though with the `DEFAULT_` prefix

nickgerace (approved):

PR #5952, opened 2025-04-24
fix(web): Better UX for the connections menu with a lot of data

vbustamante:
- Cache possible peer sockets calculations
- Show loading status when data from non active views hasn't loaded yet
- Don't freeze UI on long running peer calculations

image


stack72 (approved):
Nice!!

wendybujalski (approved):

PR #5953, opened 2025-04-24
feat(innit): in-memory cache

sprutton1:
Adds an in-memory cache for parameter values to manage backend pressure. There are separate caches for individual params and lists just for simplicity. By default, cache items expire after 24 hours. 

Also adds a create endpoint to facilitate testing.

### DIFF
--- a/lib/innit-client/tests/integration_test/mod.rs
+++ b/lib/innit-client/tests/integration_test/mod.rs
@@ -2,7 +2,11 @@ mod certs;
 #[cfg(test)]
 mod integration_tests {
 
-    use std::net::SocketAddr;
+    use std::{
+        env,
+        net::SocketAddr,
+        sync::Once,
+    };
 
     use base64::{
         Engine as _,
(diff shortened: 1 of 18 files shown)


lib/innit-client/tests/integration_test/mod.rs
```diff
@@ -31,10 +35,20 @@ mod integration_tests {
...
+    const ENV_VAR_LOCALSTACK_URL: &str = "SI_TEST_LOCALSTACK_URL";
+    static INIT: Once = Once::new();
+
+    fn initialize_crypto() {
+        INIT.call_once(|| {
+            rustls::crypto::ring::default_provider()
+                .install_default()
+                .expect("Failed to install rustls crypto provider");
+        });
+    }
```
nickgerace, line 47: What's going on here?

lib/innit-client/tests/integration_test/mod.rs
```diff
@@ -50,9 +64,18 @@ mod integration_tests {
         ca_cert: String,
     ) -> (Server<AddrIncoming>, CancellationToken) {
         let token = CancellationToken::new();
+
+        let mut endpoint = "http://localhost:4566".to_string();
+        #[allow(clippy::disallowed_methods)] // Used only in tests & so prefixed with `SI_TEST_`
+        if let Ok(value) = env::var(ENV_VAR_LOCALSTACK_URL) {
+            endpoint = value;
+        }
+        dbg!("connecting to {}", &endpoint);
```
nickgerace, line 73: We should remove or replace this

lib/innit-client/tests/integration_test/mod.rs
```diff
@@ -50,9 +64,18 @@ mod integration_tests {
         ca_cert: String,
     ) -> (Server<AddrIncoming>, CancellationToken) {
         let token = CancellationToken::new();
+
+        let mut endpoint = "http://localhost:4566".to_string();
```
nickgerace, line 68: This would be great in a const

lib/innit-client/tests/integration_test/mod.rs
```diff
@@ -125,4 +148,177 @@ mod integration_tests {
         token.cancel();
         server_handle.abort();
     }
+
+    #[tokio::test]
+    async fn create_param() {
```
nickgerace, line 153: Nice!!

lib/innit-client/tests/integration_test/mod.rs
sprutton1: The tests generate certs to ensure the certificate-based auth works. This, apparently, requires a crypto provider. However, you can only init one of these, so we poop into the once here.

nickgerace (approved):

PR #5957, opened 2025-04-24
feat(luminork): Add support for queuing an action or executing a management function

stack72:
2 test cases:

Put AWS Standard VPC on the canvas and connect it to a Region and a component. Set a name on the template. Then based on the componentId, make the following request:

```
{
  "managementFunction": {
    "function": "Component Sync"
  }
}
```

Add a VPC asset to the canvas - manually dequeue the create action. Enqueue it using the post to the actions endpoint for the component:

```
{
  "action": {
    "function": "Create Asset"
  }
}
```


britmyerss (approved):

lib/dal/src/management/prototype.rs
```diff
@@ -592,6 +595,89 @@ impl ManagementPrototype {
...
+            .await?
+            .into_iter()
+            .map(|view| (view.id(), view.name().to_owned()))
+            .collect();
+
+        let view_id = match view_id {
+            Some(view_id) => view_id,
+            None => View::get_id_for_default(ctx).await?,
+        };
+        let current_view = views.get(&view_id).cloned().unwrap_or(view_id.to_string());
```
nickgerace, line 618: I'm a bit confused here. Is there a reason to do this over the following?

```rust
let view: View = match view_id {
    Some(view_id) => View::get_by_id(ctx, view_id).await?,
    None => View::get_for_default(ctx).await?,
};
```

lib/dal/src/management/prototype.rs
```diff
@@ -592,6 +595,89 @@ impl ManagementPrototype {
...
+
+        let management_func_id = ManagementPrototype::func_id(ctx, id).await?;
+        let manager_component = Component::get_by_id(ctx, manager_component_id).await?;
+
+        let mut managed_component_placeholders = HashMap::new();
+        let mut managed_components = HashMap::new();
+        for component_id in manager_component.get_managed(ctx).await? {
+            let schema = Component::schema_for_component_id(ctx, component_id).await?;
+            let managed_component =
+                ManagedComponent::new(ctx, component_id, schema.name(), &views).await?;
```
nickgerace, line 628: Hm... looks like you need it for this.

lib/dal/src/management/prototype.rs
```diff
@@ -592,6 +595,89 @@ impl ManagementPrototype {
...
+
+        let func_runner =
+            FuncRunner::build_management(ctx, id, manager_component_id, management_func_id, args)
+                .await?;
+
+        let func_runner_id = func_runner.id();
+        let ctx_clone = ctx.clone();
+
+        tokio::task::spawn(async move {
+            dbg!("Kicking the func");
```
nickgerace, line 662: We should drop this in a follow-up

lib/luminork-server/src/service/v1/components/add_action.rs
```diff
@@ -0,0 +1,161 @@
...
+    Action::new(ctx, action_prototype_id, component_id.into()).await?;
+
+    ctx.commit().await?;
+
+    Ok(Json(AddActionV1Response {}))
+}
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, ToSchema)]
+#[serde(rename_all = "camelCase")]
+#[serde(untagged)]
```
nickgerace, line 89: Good use of `untagged`

nickgerace (approved):

PR #5961, opened 2025-04-25
fix: add transitgatewayattachmentid prop and output socket to vpnconnection asset

aaron-dernley:
Props:
Screenshot 2025-04-26 at 00 07 37

Connected:
Screenshot 2025-04-25 at 21 17 11

Output Socket:
Screenshot 2025-04-25 at 21 24 22

### DIFF
--- a/bin/clover/src/pipeline-steps/assetSpecificOverrides.ts
+++ b/bin/clover/src/pipeline-steps/assetSpecificOverrides.ts
@@ -935,6 +935,32 @@ const overrides = new Map<string, OverrideFn>([
     setAnnotationOnSocket(idSocket, { tokens: ["TransitGatewayId"] });
     setAnnotationOnSocket(idSocket, { tokens: ["Transit Gateway Id"] });
   }],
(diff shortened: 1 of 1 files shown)


bin/clover/src/pipeline-steps/assetSpecificOverrides.ts
```diff
@@ -935,6 +935,32 @@ const overrides = new Map<string, OverrideFn>([
...
     setAnnotationOnSocket(idSocket, { tokens: ["Transit Gateway Id"] });
   }],
+  ["AWS::EC2::VPNConnection", (spec: ExpandedPkgSpec) => {
+    const variant = spec.schemas[0].variants[0];
+    const domain = variant.domain;
+
+    const transitGatewayAttachmentIdProp = createScalarProp(
+      "TransitGatewayAttachmentId",
+      "string",
+      ["root", "domain"],
```
stack72, line 945: i think this will be a resource prop as the data will be set from the AWS response - not via anything the user will set

PR #5971, opened 2025-04-28
chore(dal,sdf): remove all rust traces of the standard model

zacharyhamm:
The standard model is long dead. Bury it.

NOTE: There is some standard model code in the PL/PGSQL.


jhelwig (approved):
🔥

nickgerace (approved):

PR #5976, opened 2025-04-29
feat(web): Change tooltip in code editor to be over the word rather than the current letter

JakeGinnivan:
It is currently quite hard to get the mouse onto the editor tooltip to view type information etc when you need to scroll.

This changes the tooltip to be on the word rather than just the current letter so it's easier to keep the tooltip up

Before

Kapture 2025-04-29 at 17 55 31

After

Kapture 2025-04-29 at 17 57 21


stack72:
Superceeded by #5980 - I cherry-picked the commit and made the linting fix needed to the typescriptLinter.ts file

PR #5984, opened 2025-04-29
fix(clover): Ensure that we can hand;e the different between objects and arrays in oneOf prop types

stack72:
Array props take precedence over JSON props as well as explicit array props,
because we are assuming that props that can be either object or array are really just a
"one or many", i.e. T or T[]

Thank you @jkeiser for the help to fix this up!


adamhjk:
love it. ship that shit.

adamhjk (approved):

PR #5989, opened 2025-04-29
feat(web) - virtualization for component tiles + improved newhotness UI

wendybujalski:
Screenshot 2025-04-29 at 6 18 44 PM

### DIFF
--- a/app/web/src/newhotness/ComponentGridTile.vue
+++ b/app/web/src/newhotness/ComponentGridTile.vue
@@ -1,20 +1,71 @@
 <template>
-  <div :class="clsx('component tile cursor-pointer')">
-    <header>
+  <div
+    :class="
+      clsx(
+        'component tile',
+        'cursor-pointer border border-neutral-500 rounded overflow-hidden',
+        themeClasses('bg-shade-0', 'bg-neutral-900'),
+      )
+    "
+  >
+    <header
+      :class="clsx('p-xs', themeClasses('bg-neutral-200', 'bg-neutral-800'))"
+    >
       <Icon name="logo-aws" size="lg" />
-      <h2>{{ component.name }}</h2>
-      <h3>{{ component.schemaName }}</h3>
-      <Icon name="loader" size="lg" />
+      <h2>
+        <TruncateWithTooltip class="pb-xs">{{
+          component.name
+        }}</TruncateWithTooltip>
+      </h2>
+      <h3>
+        <TruncateWithTooltip class="pb-xs">{{
+          component.schemaName
+        }}</TruncateWithTooltip>
+      </h3>
+      <Icon name="none" size="lg" />
     </header>
-    <ol>
-      <li>Qualifications: {{ component.qualificationTotals.total }}</li>
-      <li>Resource: {{ component.hasResource }}</li>
-      <li>Diff: {{ component.diffCount }}</li>
-      <hr />
-      <li>Inputs: {{ component.inputCount }}</li>
-      <li>Outputs: {{ component.outputCount }}</li>
+    <ol
+      class="[&>li]:p-xs [&>li]:flex [&>li]:flex-row [&>li]:items-center [&>li]:gap-xs [&_.pillcounter]:w-5 [&_.pillcounter]:h-5"
+    >
+      <li>
+        <StatusIndicatorIcon
+          type="qualification"
+          size="sm"
+          :status="qualificationSummary"
+        />
+        <div>Qualifications</div>
+        <PillCounter
+          :count="component.qualificationTotals.total"
+          size="sm"
+          class="ml-auto"
+        />
+      </li>
+      <li>
+        <Icon name="tilde" class="text-warning-500" size="sm" />
+        <div>Diff</div>
+        <PillCounter :count="component.diffCount" size="sm" class="ml-auto" />
+      </li>
+      <li>
+        <StatusIndicatorIcon
+          type="resource"
+          size="sm"
+          :status="component.hasResource ? 'exists' : 'notexists'"
+        />
+        <div>Resource</div>
+      </li>
+      <hr class="border-neutral-500" />
+      <li>
+        <Icon name="output-connection" size="sm" />
+        <div>Inputs</div>
+        <PillCounter :count="component.inputCount" size="sm" class="ml-auto" />
+      </li>
+      <li>
+        <Icon name="input-connection" size="sm" />
+        <div>Outputs</div>
+        <PillCounter :count="component.outputCount" size="sm" class="ml-auto" />
+      </li>
     </ol>
-    <footer class="grid grid-cols-2">
+    <!-- <footer class="grid grid-cols-2 p-xs">
       <div class="place-self-start">
         <VButton label="📌" size="sm" tone="neutral" variant="ghost" />
         <VButton label="Upgrade" size="sm" tone="action" />
@@ -23,17 +74,30 @@
         <VButton label="Delete" size="sm" tone="destructive" />
         <VButton label="Edit" size="sm" tone="action" />
       </div>
-    </footer>
+    </footer> -->
   </div>
 </template>
 
 <script lang="ts" setup>
-import { Icon, VButton } from "@si/vue-lib/design-system";
+import {
+  Icon,
+  PillCounter,
+  themeClasses,
+  TruncateWithTooltip,
+} from "@si/vue-lib/design-system";
 import clsx from "clsx";
-import { PropType } from "vue";
+import { computed, PropType } from "vue";
 import { BifrostComponent } from "@/workers/types/dbinterface";
+import StatusIndicatorIcon from "@/components/StatusIndicatorIcon.vue";
 
-defineProps({
+const props = defineProps({
   component: { type: Object as PropType<BifrostComponent>, required: true },
+});
+
+const qualificationSummary = computed(() => {
+  if (props.component.qualificationTotals.failed > 0) return "failure";
+  if (props.component.qualificationTotals.running > 0) return "running";
+  if (props.component.qualificationTotals.warned > 0) return "warning";
+  return "success";
 });
 </script>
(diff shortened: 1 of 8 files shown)


app/web/src/newhotness/ComponentGridTile.vue
```diff
@@ -1,20 +1,71 @@
...
     </header>
-    <ol>
-      <li>Qualifications: {{ component.qualificationTotals.total }}</li>
-      <li>Resource: {{ component.hasResource }}</li>
-      <li>Diff: {{ component.diffCount }}</li>
-      <hr />
-      <li>Inputs: {{ component.inputCount }}</li>
-      <li>Outputs: {{ component.outputCount }}</li>
+    <ol
+      class="[&>li]:p-xs [&>li]:flex [&>li]:flex-row [&>li]:items-center [&>li]:gap-xs [&_.pillcounter]:w-5 [&_.pillcounter]:h-5"
```
nickgerace, line 28: Cool!

app/web/src/newhotness/ComponentGridTile.vue
```diff
@@ -1,20 +1,71 @@
...
+        <PillCounter :count="component.inputCount" size="sm" class="ml-auto" />
+      </li>
+      <li>
+        <Icon name="input-connection" size="sm" />
+        <div>Outputs</div>
+        <PillCounter :count="component.outputCount" size="sm" class="ml-auto" />
+      </li>
     </ol>
-    <footer class="grid grid-cols-2">
+    <!-- <footer class="grid grid-cols-2 p-xs">
```
nickgerace, line 68: LEF'S GO

app/web/src/newhotness/Explore.vue
```diff
@@ -124,14 +172,39 @@ const actionViewList = computed(
...
+const scrollRef = ref<HTMLDivElement>();
+
+const virtualizerOptions = computed(() => {
+  return {
+    count: componentList.value.length,
+    // `lanes` gives virtualizer a "second-dimension" (aka columns for vertical lists and rows for horizontal lists)
+    // https://tanstack.com/virtual/latest/docs/api/virtualizer#lanes
+    // Our grid is based on a minimum 250px width tile... so how many tiles can we fit?
+    // thats the value of `lanes`
+    lanes: Math.floor(scrollRef.value?.offsetWidth ?? 0 / 250),
```
nickgerace, line 193: Clever

nickgerace, line 193: Clever

nickgerace (approved):

app/web/src/newhotness/ComponentGridTile.vue
```diff
@@ -1,20 +1,71 @@
...
+        <StatusIndicatorIcon
+          type="qualification"
+          size="sm"
+          :status="qualificationSummary"
+        />
+        <div>Qualifications</div>
+        <PillCounter
+          :count="component.qualificationTotals.total"
+          size="sm"
+          class="ml-auto"
```
jobelenus, line 40: I'm internally torn whether `ml-auto` or `justify-between` results in a more understandable rule... I feel like we know what `ml-auto` does, but, its not very intuitive for others. Where `justify-between` on the container can set a very clear signal for "one on the left, one on the right"... which of course fails the moment you have more than 2 things and want N on the left and M on the right...

app/web/src/newhotness/Explore.vue
```diff
@@ -1,78 +1,118 @@
 <template>
-  <section class="grid gap-md h-[100%]">
-    <div>
-      <!-- filter / top header -->
-      <InstructiveVormInput
-        classes="py-2"
-        :activeClasses="
-          clsx(themeClasses('border-action-500', 'border-action-300'))
+  <section class="grid h-full">
```
jobelenus, line 2: i don't believe `h-full` is needed, thats a default with grid as far as I can tell?

app/web/src/newhotness/Explore.vue
```diff
@@ -1,78 +1,118 @@
...
+    <div
+      :class="
+        clsx(
+          'flex flex-col border-l border-neutral-500',
+          themeClasses('bg-neutral-100', 'bg-neutral-800'),
+        )
+      "
+    >
+      <!-- TODO(Wendy) - this section UI is still rough, see Figma -->
+      <div class="grow grid grid-rows-subgrid" :style="collapsingStyles">
```
jobelenus, line 70: `subgrid` is only needed as a child of a `.grid` (unless I'm missing something here?)

app/web/src/newhotness/Workspace.vue
```diff
@@ -215,21 +192,14 @@ onBeforeUnmount(() => {
   scrollbar-width: thin;
 }
 body.dark .scrollable {
-  scrollbar-color: rgb(33, 33, 33) rgba(0, 0, 0, 0);
+  scrollbar-color: @colors-neutral-800 @colors-black;
```
jobelenus, line 195: 🙌

app/web/src/newhotness/Workspace.vue
```diff
@@ -257,6 +226,7 @@ body.light .scrollable {
 @supports (width: min(250px, 100%)) {
   .tilegrid {
     grid-template-columns: repeat(auto-fit, minmax(min(250px, 100%), 1fr));
+    grid-auto-rows: max-content;
```
jobelenus, line 229: what does this get us?

app/web/src/newhotness/util.ts
```diff
@@ -6,7 +6,7 @@ import { Toggle } from "./logic_composables/toggle_containers";
  * When its open use the fractional unit so it grows to the available size, sharing the remaining space with other open grid items
  */
 export const gridCollapseStyle = (open: boolean | Ref<boolean, boolean>) =>
-  unref(open) ? "1fr" : "2em";
+  unref(open) ? "1fr" : "28px";
```
jobelenus, line 9: I'm curious and I want to understand better, I was using `em` because the `sm`, `md`, `lg`, use `em`, not `px`... what leads you to `px` here?

app/web/src/store/feature_flags.store.ts
```diff
@@ -97,6 +97,7 @@ export function useFeatureFlagsStore() {
         // You can override feature flags while working on a feature by setting them to true/false here
         // for example:
         // this.FEATURE_FLAG_NAME = false;
+        this.NEW_HOTNESS = true;
```
jobelenus, line 100: gotta take this out

app/web/src/newhotness/ComponentGridTile.vue
jobelenus: Yes, very. Unfortunately tailwind is not at all interested in offering support (based on github issues) for collapsing multiple style applications like this into a single `[&>li]` statement... so... duplication it is!

app/web/src/newhotness/Explore.vue
jobelenus: we're very glad they have this `lanes` parameter... otherwise it would have gotten a lot uglier

app/web/src/store/feature_flags.store.ts
britmyerss: would be great if we can move this to local config, so we don't have to worry about accidentally committing this type of thing.

app/web/src/newhotness/Explore.vue
wendybujalski: Yeah idk I don't think I touched those classes at all after moving that section. Probably a dangling class from the previous way things were laid out!

app/web/src/newhotness/ComponentGridTile.vue
wendybujalski: Yeah, neither is ideal because both break when you adjust the surrounding contents. I tend to stick with `ml-auto` because then it puts the breakable behavior of "go to the right side" on the element I want to do that instead of obscuring that behavior on the parent element.

app/web/src/newhotness/Workspace.vue
```diff
@@ -47,11 +47,9 @@
       <NavbarPanelRight />
     </nav>
 
-    <!-- grow the main body to fit all the space in between the nav and footer
-     by default flexbox containers grow at minimum to the size of their content (e.g. autp)
-     min-h-0 prevents the main container from being *larger* than the max it can grow, no matter its contents
-     also, i chose 12 here and not a standard unit because 12 puts the left border of contents exactly on the SI logo in the top left -->
-    <main class="grow p-[12px] min-h-0">
```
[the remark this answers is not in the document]
wendybujalski: No, the problem is the footer - until it moves back into `Workspace` (currently it is inside `Explore` since its contents are specific to that screen) we can't move the padding back to this level because the footer needs to not have that padding.

app/web/src/newhotness/Workspace.vue
wendybujalski: This prevents any of the grid rows from growing taller than its contents. It might be redundant with another approach at this point but I would have to poke around to find out.

wendybujalski: Though actually this might be a mistake and should be `min-content`? I'll take a look.

app/web/src/newhotness/Workspace.vue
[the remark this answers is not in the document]
jobelenus: Mmmm, interesting... I'll have to think on this a bit. Typically I'd smash a negative margin on a thing to pull it back... but that's the same category of "human needs to remember"

app/web/src/newhotness/util.ts
wendybujalski: 2em is equal to 32px - the headers were not collapsing quite right, a tiny sliver of the content was still visible! If we wanna convert 28px to 1.75em that's fine by me.

app/web/src/store/feature_flags.store.ts
wendybujalski: ah whoops of course i missed that in the kurfuffle, thanks will do

app/web/src/newhotness/Workspace.vue
jobelenus: Ok, let's play with it a bit to get a full understanding of how it behaves.

jobelenus (dismissed):

jobelenus (approved):

PR #5991, opened 2025-04-30
Add new UI survival guide and relieve macro builds

nickgerace:
## Description

This PR adds a "newhotness" survival guide for those working on the new UI, building MVs, working with edda, and more.

## Rendered Survival Guide

You can view the rendered survival guide [here](https://github.com/systeminit/si/blob/nick/11d229d/app/web/src/newhotness/SURVIVAL_GUIDE.md)!

## Improving the build graph

This PR also moves the "build_mv!" macro to a new crate called "dal-materialized-views-macros" that only "edda-server" uses. This ensures that changes to the "build_mv!" macro do not result in a rebuild of the "dal". Not only that, but it ensures that the only crate that can build MVs is "edda-server".

## Misc

Finally, this PR renames "dal-newhotness" to "dal-materialized-views" for better clarity.


britmyerss (approved):

app/web/src/newhotness/SURVIVAL_GUIDE.md
```diff
@@ -0,0 +1,112 @@
...
+
+If you follow all aforementioned steps, you should be in a good spot for this one.
+However, you may notice that in [#5914](https://github.com/systeminit/si/pull/5914), we have a `computed` usage that partially "links" two MVs.
+That decision not only predates formal, weak references, but it was required to "obey the merkle tree hash gods".
+
+Every decision in authoring an MV will be nuanced.
+We want to prevent the cross-referential objects store palooza that exists in the current UI.
+However, there may be times where, because of the graph structure, you may need more than one MV to solve your problem.
+In fact, multiple MVs is totally fine!
+It's the cross-referential-big-ass-computed types that can get hairy.
```
jobelenus, line 64: I think one additional mention would help drive the point home: The ultimate goal here is to _prevent_ this leaking into the front end. We have the webworker layer where we handle references (and soon weak references). There is no "logic" involved in the connection–its just the "graph" after all. In ye olden days we would have branching logic for "walk these related things and if its X do Y and if its A do B". That way lies madness.

britmyerss: @jobelenus let's include this change in your func run pr!

PR #5992, opened 2025-04-30
feat(sdf): fetch and set cas data in admin panel

zacharyhamm:
Combined with snapshot fetching, this will allow us to debug remote workspaces better locally.


britmyerss:
@zacharyhamm as you use this for testing, would be great to extend the examples for how to use this alongside the snapshot to interrogate.

britmyerss (approved):

PR #6003, opened 2025-04-30
feat(sdf): API endpoint for attribute subscriptions

jkeiser:
This adds the ability to set up attribute subscriptions via the API. The endpoint does a little more than that, allowing you to set, subscribe or unset as many attributes as you want.

## The Endpoint

`PUT /v2/.../components/:componentId/attributes` can create subscriptions to other values as well as set static values (including arrays and maps):

```json
{
  "/domain/SubnetId": { "$kind": "subscription", "component": "My New Subnet", "path": "/domain/SubnetId" },
  "/domain/IpAddresses/0": "10.0.0.1"
}
```

* Can set multiple attributes per call, with a hash of PATH: VALUE pairs
  - Specify the attributes you want to set as a JSON pointer just like normal subscriptions. /domain/IpAddresses/0. This should allow us 
* Subscribe is `{ "$kind": "subscription", component: "<id or name>", path: "/domain/IpAddresses/0" }`.
  - You may subscribe to paths that *do not yet exist* (i.e. `/resource/IpAddresses/0` which is created by the create action)
* Setting a value is just `name: value`
  - If you are setting an object value that might have "$kind" as a key, you can set the value using `{ "$kind": "value", "value": ... }` and the system will correctly set the value. This is a sort of escape.
* Unset is `{ "$kind": "value", value: undefined }` (which is the same as not specifying the attribute: setting to `null` will set the value to null rather than unset it)

## Out of Scope

The shape seems reasonable at this point, but there are fixes and edges here that could make this nicer:

* Support append via JSON pointer `-`: `"/domain/IpAddresses/-": "10.0.0.2"` should add 10.0.0.2 as a new element
* Support AttributeValueId as well as path when specifying the attribute to set (and maybe even the one to subscribe to)
* Make sure it's properly constructing entire trees (when you set /domain/Foo/Bar, if Foo is not yet set, we should set it)
* The API should yell at you if you try to create a subscription that can *never* work (like, if you misspell props)
* Setting an entire object or array all at once isn't working as expected
* Consider allowing the user to set subscriptions in nested values (e.g. `"/domain/SubnetMappings": [ { "Name": "A", "SubnetId": { "$kind": "subscription", ... } ]`

Additionally, we still haven't implemented multiple-arity subscriptions, which will need to be supported in the API as well.

## Testing

- [X] Integration Tests Pass
- [X] Setting values and subscriptions via API works
- [X] Manual testing of existing connections


lib/dal/src/attribute/value.rs
```diff
@@ -2210,12 +2206,10 @@ impl AttributeValue {
...
             ctx,
             prototype_arg.id(),
-            ValueSource::ValueSubscription(ValueSubscription {
-                attribute_value_id: subscribed_to_av_id,
-                path,
-            }),
+            ValueSource::ValueSubscription(subscription),
         )
         .await?;
+        // TODO arity (this blows away any existing prototype and replaces it with the single subscription!)
```
britmyerss, line 2212: noted!

lib/dal/src/component.rs
```diff
@@ -1581,6 +1581,18 @@ impl Component {
         Ok(None)
     }
 
+    pub async fn find_by_name(
```
britmyerss, line 1584: I'm surprised we don't already have something like this....

lib/sdf-server/src/service/v2/component/attributes.rs
```diff
@@ -0,0 +1,197 @@
...
+    Error,
+    Result,
+};
+use crate::app_state::AppState;
+
+pub fn v2_routes() -> Router<AppState> {
+    Router::new().route("/", put(update_attributes))
+}
+
+// {
```
britmyerss, line 40: great examples!

lib/dal/src/component.rs
jkeiser: So was I! Could be I didn't do the right search, but I couldn't find it in component.rs.

britmyerss (approved):

britmyerss (dismissed):

jobelenus (approved):

PR #6009, opened 2025-05-01
Parallel MV builds in edda

jhelwig:
We have already been providing dependency ordering information between classes of MVs, and been using that for the serialized ordering of builds. We now spawn each MV build into a new tokio task to allow multiple MVs to build at the same time whenever possible.

This also removes the `build_mv_task!` procedural macro in `si-frontend-types-macros` as a lot of the code duplication it was handling is no longer necessary after much of the duplicated code was moved into the `join_next` handling of the `JoinSet` the build tasks are spawned into. The `spawn_mv_task!` macro is able to be a much simplified declarative macro, instead of the previous procedural macro.

Because the build tasks are now spawned off into their own tokio task, they must now take an owned `DalMacro`.


lib/edda-server/src/change_set_processor_task/materialized_view.rs
```diff
@@ -275,6 +284,33 @@ pub async fn build_mv_for_changes_in_change_set(
...
+        // tasks for any given MV kind have finished.
+        $mv_task_ids
+            .entry(kind)
+            .and_modify(|task_ids| {
+                task_ids.insert(task_id);
+            })
+            .or_insert_with(|| [task_id].iter().copied().collect());
+        // Record the currently running number of MV build tasks so we can see just how parallel this actually
+        // ends up being.
+        metric!(counter.edda.mv_build = 1);
```
fnichol, line 301: That's a great idea

lib/edda-server/src/change_set_processor_task/materialized_view.rs
```diff
@@ -294,227 +330,257 @@ async fn build_mv_inner(
...
-                                    if let Some(object) = maybe_frontend_object {
-                                        frontend_objects.push(object);
-                                    }
-                                }
+                                frigg,
+                                change,
+                                mv_id,
+                                si_frontend_types::action::prototype::ActionPrototypeViewList,
+                                dal::action::prototype::ActionPrototype::as_frontend_list_type(
+                                    ctx.clone(),
```
fnichol, line 374: Yeah, this is where the owned `DalContext` change pays off, makes sense

fnichol (dismissed):
Looks great and I like the task setup functions--it feels like it cleans things up

fnichol (approved):

PR #6017, opened 2025-05-02
feat(dal,sdf): migrate to split graph

zacharyhamm:
Adds new migration for moving from the latest version of the current graph to the split graph.

Some included changes:
 1. New column on the workspaces table to accurately describe all the versioning info about the snapshots: subgraph_version. The version column is the version of the legacy graph, or of the supergraph for the split graph (which could change separately from the subgraph version).
 2. This meant reworking our queries. They are no longer "standard model" queries that construct a json blob from the row, but instead we fetch the row's columns and use the try_from to construct the object from the database.
 3. I've removed the double write of the snapshot on ChangeSet::new. It's a holdover from when we had to update the vector clocks on the graph for a new change set. We no longer use vector clocks, so it's unnecessary (and it would've had to have been reworked to support split graph change sets).


lib/dal/src/context.rs
```diff
@@ -979,16 +979,12 @@ impl DalContext {
         new
     }
 
-    async fn workspace(&self) -> TransactionsResult<Workspace> {
```
jkeiser, line 982: One semantic difference here: when you have no workspace in your DalContext, get_workspace() will query for WorkspacePk::NONE and hopefully yield `WorkspaceError::WorkspaceNotFound`, while workspace() will return `TenancyError::NoWorkspace`. Is this change intentionally doing this differently or is it just that we know it'll probably eventually error in either case (assuming NONE is not in the db)?

zacharyhamm: ok i've kept the two functions and made their behavior more clear.

lib/dal/src/workspace_snapshot/migrator.rs
```diff
@@ -77,95 +83,105 @@ impl SnapshotGraphMigrator {
...
+                )
                 .await?;
+            workspace_count += 1;
         }
 
-        info!("Migration finished, marking all workspaces as migrated to latest version");
-
-        Workspace::set_snapshot_version_for_all_workspaces(
-            ctx,
-            WorkspaceSnapshotGraph::current_discriminant(),
```
jkeiser, line 166: Oh man, did we fail to set the version on some workspaces before if it failed halfway? Or was a transaction protecting us here?

lib/dal/src/workspace_snapshot/migrator.rs
```diff
@@ -77,95 +83,105 @@ impl SnapshotGraphMigrator {
...
+
+                change_set
+                    .update_pointer(&ctx_after_migration, new_snapshot_address)
+                    .await?;
+
+                migration_map.insert(snapshot_address, new_snapshot_address);
+                change_set_count += 1;
+            }
+
+            workspace
```
jkeiser, line 174: I'm a little lost as to where the transaction is here--I was expecting to see a ctx.commit() at some point, but maybe we just don't start a transaction? If we're not starting a transaction, does that mean that if we fail to migrate one of the changesets in a workspace, we'll update their pointer but not update the workspace's snapshot version? Would that cause failures to read the snapshots later?

lib/dal/src/workspace_snapshot/migrator/split_v1.rs
```diff
@@ -0,0 +1,105 @@
...
+            continue;
+        }
+
+        node_id_by_index.insert(node_index, node_id);
+        split_graph.add_or_replace_node(node)?;
+
+        if let Some(ordering_node) = v4_graph.ordering_node_for_container(node_index)? {
+            let order = ordering_node.order().clone();
+            ordering_for_node.insert(node_id, order);
+            // Ordering nodes are an internal implementation detail for the split graph,
```
jkeiser, line 57: Oh nice, I am stoked that you solved this in the split graph impl :)

jkeiser (approved):
Except maybe the transaction thing (which probably has an answer) I think this will do the right thing, scary as migrations are :)

PR #6019, opened 2025-05-02
feat(luminork): Add support for secrets

stack72:


lib/luminork-server/src/service/v1/secrets/mod.rs
```diff
@@ -0,0 +1,117 @@
...
+            _ => (StatusCode::INTERNAL_SERVER_ERROR, self.to_string()),
+        }
+    }
+}
+
+pub fn routes() -> Router<AppState> {
+    Router::new()
+        .route("/", get(get_secrets::get_secrets))
+        .route("/", post(create_secret::create_secret))
+        .route("/", patch(update_secret::update_secret))
```
britmyerss, line 99: This isn't hanging off /:secret_id because you also want to accept name?

britmyerss (approved):
Left a question about update, but no need to hold it up for it. Great work!

lib/luminork-server/src/service/v1/secrets/mod.rs
stack72: There was an issue withe handler - i will follow up with a change

PR #6021, opened 2025-05-03
feat(sdf): Make PUT /attributes API more concise/consistent

jkeiser:
This makes the /attributes API for setting an attribute's source a little more consistent, with explicit sources being `{ $source: {<source>} }`. Unsetting or removing a value becomes much clearer, being `{ $source: null }` instead of `{ $source: "value" }`.

```js
{
  // Subscription
  "/domain/SubnetId": { $source: { component: "My Subnet", path: "/resource/SubnetId" } },

  // Set value
  "/domain/IpAddresses/0": "10.0.0.0",

  // Unset or remove value
  "/domain/Tags/Environment": { $source: null },
  "/domain/Tags/Name": { $source: { value: undefined } }, // This works too, making some JS code easier

  // Explicitly set value to static source (longhand version of "set value")
  "/domain/IpAddresses/1": { $source: { value: "10.0.0.1" } },
}
```

It also moves the typescript API call into the components store, because the component attributes store is more tightly tied to the attributes panel.

## Testing

- [x] Existing tests pass
- [x] Manual API tests

sources


app/web/src/store/components.store.ts
```diff
@@ -1168,6 +1286,19 @@ export const useComponentsStore = (forceChangeSetId?: ChangeSetId) => {
             });
           },
 
+          async UPDATE_COMPONENT_ATTRIBUTES(
+            componentId: ComponentId,
+            payload: UpdateComponentAttributesArgs,
+          ) {
+            return new ApiRequest({
+              method: "put",
+              url: `v2/workspaces/${workspaceId}/change-sets/${changeSetId}components/${componentId}/attributes`,
```
vbustamante, line 1295: This is missing  a /:
```suggestion
              url:
                `v2/workspaces/${workspaceId}/change-sets/${changeSetId}/components/${componentId}/attributes`,
```

PR #6022, opened 2025-05-03
Component Page Editing Attributes (not map or array)

jobelenus:
image

# Foundation

This PR introduces many foundational pieces for the front end architecture:

- Making tanStack Form re-usable & more informative for our circumstances.
- Defining the HTTP API interface
- Introducing a pattern for non-graph/non-bifrost data querying and invalidating.
- There are examples of using `logic_composables` and `layout_components` together, and keeping API calls and business logic out of `layout_components`. 

And there are some techniques that can be learned from in future circumstances, potentially:

- Emitter for coordinating events between focusing disparate elements (we don't need a store, b/c its not state, its just events).
- Filtering a tree structure by clearing out children and repopulating them (i can re-imagine the debug panel implemented this way, since searching it has been difficult to implement)

## TanStack Form via `useWatchedForm`

 **WHEN?** You want to use this form when the data that displays initially on the form is populated via the bifrost.
 
 **WHY?** form.reset(data) is a crucial aspect of this operation.

 1. TanStack forms are designed to be submitted just one time.
 2. So if you don't reset after a submission you can't re-submit.
 3. `defaultValues` is only set 1 time. Which means, if you were querying for the data for the form, the values will be blank because the form was created prior to having the data

**BONUS!** you can `bifrosting` to indicate to the user that the form is not capable of being re-submitted until its underlying data has successfully mutated. Otherwise the UI experience can be:

1. I submit a form
2. I still see the previous value elsewhere in the UI
3. I wonder... did my form submit work?!
4. Once the patched data lands in the browser, you see a re-render
5. Ah! It worked :)

With `bifrosting` you can show a loader/spinner/disable a form, so the user know their form submission worked and we're waiting for updated data

If you're just making a standard "create", or signup form, where there is no underlying data — you don't need this.

## Making API calls via `useApi`

The API endpoint definitions and URLs are defined and enforced here. This also removes the need to populate the `workspacePk` and `changeSetId` from the calling component. We have also assumed that all the API calls in SDF are nested within the `v2/workspaces/${workspacePk}/change-sets/${changeSetId}` locations.

This example is updating a component's name:
```
const api = useApi();
const id = component.value?.id;
const call = api.endpoint(routes.UpdateComponentName, { id });
const name = "foo";
await call.put<UpdateComponentNameArgs>({ name });
```
To get a better sense of how this is working, let's look at some type defs in `api_composables/index.ts`
```
export enum routes
const endpoint = <Response>(key: routes, args?: Record<string, string>)
async put<D = Record<string, unknown>>(data: D, params?: URLSearchParams)
```

- The `routes` are an enum which defines all possible routes. And the paths are defined in a record where the keys are an elements of the enum, and the values are the url path.
- The `endpoint` fn accepts a one of the enum values and a record to replace url paths with (e.g. `{ id }` does a replace of `<id>` in the path string.
- The `put` fn takes a type to define the shape of the payload data (you don't need to set it, we have a default)

We also include an `ok` fn to check the http status, because Axios doesn't have a helper fn that determines general success.

In the near-term future we will handle "making an edit from HEAD". Right now, this is explicitly forbidden. The front end will control the "force new change set". Also—there are POST/PUTS that _are_ allowed on HEAD, we'll get that worked in there too.

## Non-Bifrost Data

We have data that does not exist on the graph. In this PR we work with the func run and log data. In general the approach is:

1. Use tanStack `useQuery` or `useInfiniteQuery`
2. `useApi` within the tan stack `query` fn.
3. invalidate the query key when the related `realtimeStore` WsEvent fires

For a specific example see `FuncRunList.vue`

## Logic Composables & Layout Components

- The show/hide behavior of the component attribute tree is fully encapsulated within `AttributeChildLayout`
- The `AttributeInput` component has the form, but does not perform the the API call. We emit that back up the component chain.
- `CollapsingFlexItem` and `CollapsingGridItem` only handle the HTML and CSS elements, while encapsulating the toggle logic in a common `logic_composable` named `useToggle`. The `Explore` page to use the toggle to create the `collapsingStyles`.
- The `attributeEmitter` tells each `AttributeInput` which input is focused. Only when another attribute input is focused is the open menu closed. We need this because we can't simply close the open menu on `input:blur` because users need to interact with the open menu which blurs the input and closes the menu. That means we couldn't implement filter boxes on dropdown menus. And it also meant that we were double submitting, both on `input:blur` _and_ clicking the option.

### Rules For Stores

---

The top-level `Workspace` component is using stores to gather the basic data from the URL bar and authStore. We have a `Context` object (think of this like `DalContext`) via `provide/inject` that contains the global & immutable data. Workspace and Change Set aren't "reactive" like other data its based purely on the URL. Same goes for the logged in user, and whether the given change set is head.

**Do not re-introduce stores at any level of the application**, with one caveat—the `realtimeStore`. The action store found its way in there. I'll get it ripped out now that we have API interfaces. Everything that is in the old stores has a home elsewhere.

🚀 As always please reach out if its not clear how to handle re-building the application 🚀 

#### The Attribute Functionality

The basic input & option editing, as well as editing component name are functioning. We don't yet have the map or array editing working. We are re-working the `AttributeTree` `MvIndex` so we don't have to rely on weak references for this data and a pair of fixes to the AV paths. But, we can definitely ship this working for now while we get that work done.


app/web/package.json
```diff
@@ -58,11 +58,12 @@
     "@si/ts-lib": "workspace:*",
     "@si/vue-lib": "workspace:*",
     "@sqlite.org/sqlite-wasm": "3.48.0-build4",
+    "@tanstack/vue-form": "^1.9.0",
     "@tanstack/vue-query": "^5.67.3",
     "@tanstack/vue-table": "^8.20.5",
     "@tanstack/vue-virtual": "^3.13.6",
     "@types/async": "^3.2.15",
-    "@typescript/vfs": "^1.4.0",
+    "@typescript/vfs": "^1.5.3",
```
wendybujalski, line 66: UPGRADED! 🎉

app/web/src/store/feature_flags.store.ts
```diff
@@ -100,6 +100,7 @@ export function useFeatureFlagsStore() {
         // events (e.g. change set abandoned). For now, let's play it safe and make sure that
         // everything in the new frontend arch comes in together. Both of these flags should
         // eventually die though.
+        this.NEW_HOTNESS = true;
```
wendybujalski, line 103: ah whoops, make sure to pull this out before you merge!

wendybujalski (changes requested):
Looks great - just remove the forced feature flag and I'll hit approve!

app/web/src/store/feature_flags.store.ts
jobelenus: Thank you!

wendybujalski (approved):
👍

PR #6032, opened 2025-05-05
feat(web): Show props on the floating connections menu

vbustamante:
- Show props of the selected component on the connections menu list
- Add an option to skip the requestUlid on the payload

image

### DIFF
--- a/app/web/src/components/ModelingView/ConnectionMenuCandidateList.vue
+++ b/app/web/src/components/ModelingView/ConnectionMenuCandidateList.vue
@@ -30,18 +30,18 @@
           position: 'relative',
         }"
       >
-        <ConnectionMenuSocketListItem
+        <ConnectionMenuCandidateListItem
           v-for="item in socketList"
-          ref="socketListItemsRef"
           :key="item.index"
-          :data-index="item.index"
-          :item="item"
-          :entry="listItems[item.index]!"
-          :highlighted="item.index === localHighlightedIndex"
+          ref="socketListItemsRef"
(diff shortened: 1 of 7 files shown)


vbustamante:
closes ENG-2997

app/web/src/components/ModelingView/ConnectionMenuCandidateList.vue
```diff
@@ -202,6 +181,60 @@ watchEffect(() => {
...
+): boolean {
+  if (!selectedSocket) return false;
+
+  if (!candidateIsSocket(selectedEntry)) return false;
+
+  return selectedEntry.socket?.def.id === selectedSocket?.def.id;
+}
+</script>
+
+<script lang="ts">
```
jkeiser, line 197: Why two <script> blocks?

app/web/src/components/ModelingView/ConnectionMenuCandidateListItem.vue
```diff
@@ -123,6 +126,24 @@ const props = defineProps({
   controlScheme: { type: String, default: "arrows" },
 });
 
+const typeText = computed(() => {
+  if (candidateIsProp(props.entry)) return "Prop";
+
+  if (!candidateIsSocket(props.entry)) return "?";
```
jkeiser, line 132: Just confirming: this is impossible, right? Maybe this could be a `candidateType()` method that returns "prop" | "socket" and throws if it's neither? Then you could just have a simple switch () statement here

PR #6035, opened 2025-05-05
Core reactivity fixes

jobelenus:
1. First we made the context we're providing into a computed structure.
2. Then Heimdall returned a computed structure out of the `useMakeKey` fn, with the standard `useXXX` pattern.
3. That key needs to be reactive, so that when the changeset changes, the tan stack query re-fires.


app/web/src/store/feature_flags.store.ts
```diff
@@ -100,6 +100,7 @@ export function useFeatureFlagsStore() {
         // events (e.g. change set abandoned). For now, let's play it safe and make sure that
         // everything in the new frontend arch comes in together. Both of these flags should
         // eventually die though.
+        this.NEW_HOTNESS = true;
```
britmyerss, line 103: don't want this one :)

britmyerss (dismissed):

app/web/src/newhotness/types.ts
```diff
@@ -1,9 +1,9 @@
-import { Ref } from "vue";
+import { ComputedRef } from "vue";
 import { User } from "@/api/sdf/dal/user";
 
 export interface Context {
```
nickgerace, line 4: Ah CONTEXT. Got it.

nickgerace (approved):

PR #6037, opened 2025-05-05
Centralize materialized view generation

nickgerace:
## Description

This PR splits materialized view generation methods into individual modules that are flattened in "dal-materialized-views". This change helps authors not have to keep track of the crate structure, and keeps it flattened to mimic edda's flattened usage of materialized views.

## Improving Build Times

Another benefit of this change is that all MVs assembled are now in "dal-materialized-views" and NOT in "dal" or "sdf-server". This should help build times for the latter two crates and is now solely-scoped to the "edda" build chain.

## Domain Layout and Tracing

Domain nesting (e.g. "component" contains all component methods and "aciton" contains its methods and nests the "prototype" methods in a sub-module) causes needless segmentation of the crate. In addition to that change, every MV method is called "assemble" and is instrumented at the debug level. This standardization should help with local and spot-debugging in limited production scenarios.

## Bonus

Getting the reference ID for a geometry now returns an option rather than an unused error.


jobelenus (dismissed):

wendybujalski (approved):
👍

jobelenus (approved):

PR #6039, opened 2025-05-06
feat(docs): Add the ability to autogenerate the OpenAPI docs for the external API

stack72:
To generate the openapi.json file required for the docs:

```
buck2 run //bin/openapi-extractor:openapi-extractor
```

Then you can run the generator:

```
buck2 run :update-api-docs
```

This is a deterministic command and can be run as many times as needed


app/docs/package.json
```diff
@@ -6,7 +6,8 @@
   "scripts": {
     "dev": "pnpm run start",
     "start": "vitepress dev src",
-    "build": "vitepress build src"
+    "build": "vitepress build src",
+    "generate-api-docs": "npx widdershins --omitHeader --search false --language_tabs \"typescript:TypeScript\" --summary ../../bin/openapi-extractor/data/openapi.json -o src/reference/external-api.md -u ./templates"
```
nickgerace, line 10: Does the `generate-api-docs` work from any directory in the repository? I see the `../../bin/<etc>` path and it made me curious.

I'm not going to be hard on the specifics here because seriously nice work getting this into buck2. We can make changes later if need be.

stack72: I will be following up with changes for this - I need to understand how path traversal will work better here

stack72:
This PR specifically doesn't add the link to the left hand menu for these docs - I am going to be adding more detail to them as the week goes on to ensure we have descriptions and examples

johnrwatson (approved):

PR #6041, opened 2025-05-06
Remove attribute tree MV

nickgerace:


app/web/src/store/feature_flags.store.ts
```diff
@@ -95,6 +95,7 @@ export function useFeatureFlagsStore() {
         // You can override feature flags while working on a feature by setting them to true/false here
         // for example:
         // this.FEATURE_FLAG_NAME = false;
+        this.NEW_HOTNESS = true;
```
jobelenus, line 98: remove

jhelwig (dismissed):

jobelenus (changes requested):
Don't comment out the front end code. It's going to cause massive conflicts with work we've got going. Let it run—it will 404 on mjolnir and return empty objects to the UI, which doesn't matter.

app/web/src/store/feature_flags.store.ts
nickgerace: UGH I thought I didn't have this! Thanks. This is my bad.

jobelenus (approved):
Thank you! 🙇

PR #6043, opened 2025-05-06
Limit concurrency of parallel MV building in edda

jhelwig:
This PR requires #6041.

Rather than allowing unlimited parallelism (until we overwhelm the machine the code is running on), we now have a queue of MV build triggers that _would_ have spawned new tasks, but didn't because it would have exceeded the max concurrency we've defined.

Once MV build triggers have been queued, we skip processing the changes with the ready to start MV kinds until we have spawned tasks for every queued trigger.

When determining whether we have finished all of the work necessary, we also now check to make sure that there aren't any queued triggers, in addition to making sure that there are not any ready to start MV kinds, and no currently outstanding build MV tasks that need to finish.

### DIFF
--- a/lib/edda-server/src/change_set_processor_task/materialized_view.rs
+++ b/lib/edda-server/src/change_set_processor_task/materialized_view.rs
@@ -1,6 +1,7 @@
 use std::collections::{
     HashMap,
     HashSet,
+    VecDeque,
 };
 
 use dal::{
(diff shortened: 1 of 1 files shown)


nickgerace (dismissed):
Reviewed the second commit!

fnichol (dismissed):

fnichol (approved):

PR #6044, opened 2025-05-06
feat(web) - newhotness functional qualifications panel

wendybujalski:
Screenshot 2025-05-06 at 6 34 50 PM


app/web/src/store/feature_flags.store.ts
```diff
@@ -95,6 +95,7 @@ export function useFeatureFlagsStore() {
         // You can override feature flags while working on a feature by setting them to true/false here
         // for example:
         // this.FEATURE_FLAG_NAME = false;
+        this.NEW_HOTNESS = true;
```
britmyerss, line 98: we need a new solution here so this doesn't get committed accidentally.

jobelenus: true story.. i'll get a permanent solution with my morning time 🫡

lib/sdf-server/src/service/v2/func/get_func_run_by_av.rs
```diff
@@ -0,0 +1,57 @@
...
+    },
+};
+
+#[derive(Deserialize, Serialize, Debug, Eq, PartialEq, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct GetFuncByAvIdResponse {
+    pub qual_output: Vec<QualificationOutputStreamView>,
+}
+
+pub async fn get_func_run_by_av(
```
britmyerss, line 33: I'm a little confused by this route. If we want to fetch the latest func run for an Attribute Value Id, we should probably return a consistent shape that's agnostic to what that func kind was. This is building a QualificationView and returning it - but the shape of that object hasn't changed in some time and it looks like we're only doing this to get the logs? Do we need this to be more specific to qualifications? There's already a separate route for fetching only the logs, but I suppose the issue is we don't know when there's a new func run id for a qualification in particular?

app/web/src/mead-hall/QualificationView.vue
britmyerss, line 1: why meadhall for this one?

app/web/src/store/feature_flags.store.ts
wendybujalski: yeah a custom lint rule is the way, we just gotta add it

app/web/src/mead-hall/QualificationView.vue
wendybujalski: It was based on a component in the old UI, but diverged to the point where it should no longer be in meadhall

lib/sdf-server/src/service/v2/func/get_func_run_by_av.rs
wendybujalski: Yeah, perhaps this should just be more specific to qualifications. And yes, we can't know when there's a new func run without checking, so the UI just checks when you click the "More Details" link.

jobelenus: Right. The prop data has the AV, so that's what we're querying by. And we're only returning the output. Just re-using a DAL fn.

britmyerss: Yeah so I think you shift this route to look like this: `/runs/latest/:attribute_value_id/logs` or something like that. Then instead of building a qualification view, write a new helper to get the latest func run by attribute value id, then get the logs for that func run and return them (similar to what's happening within `QualificationView::new()` only without all of the other stuff you don't need and it'll work for all attribute values.

britmyerss (dismissed):
Let's fix up the route before we ship this one

lib/sdf-server/src/service/v2/func/get_func_run_by_av.rs
jobelenus: Updated!

jobelenus (approved):

PR #6046, opened 2025-05-07
Add "diff" content to the component mv. Also, fix the feature flag sponge we keep messing up

jobelenus:
I'm not making a separate Mv to keep a lid on total number of mvs, given everything that is going on. Maybe in the future we have a better place for this to live. But for now, it functions and stays up to date.

Also—please give me the more correct rust way to do the if-let-some nonsense I did hear 😇 ... it works, but obviously its not pretty.

### DIFF
--- a/app/web/src/store/feature_flags.store.ts
+++ b/app/web/src/store/feature_flags.store.ts
@@ -100,6 +100,10 @@ export function useFeatureFlagsStore() {
         // events (e.g. change set abandoned). For now, let's play it safe and make sure that
         // everything in the new frontend arch comes in together. Both of these flags should
         // eventually die though.
+
+        // turning this on for local development
+        if (import.meta.env.VITE_SI_ENV === "local") this.NEW_HOTNESS = true;
+
         if (this.NEW_HOTNESS) {
           // eslint-disable-next-line no-console
           console.log(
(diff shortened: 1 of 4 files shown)


app/web/src/store/feature_flags.store.ts
```diff
@@ -100,6 +100,10 @@ export function useFeatureFlagsStore() {
         // events (e.g. change set abandoned). For now, let's play it safe and make sure that
         // everything in the new frontend arch comes in together. Both of these flags should
         // eventually die though.
+
+        // turning this on for local development
+        if (import.meta.env.VITE_SI_ENV === "local") this.NEW_HOTNESS = true;
```
stack72, line 105: 🔥

nickgerace: very clever

lib/dal-materialized-views/src/component.rs
```diff
@@ -77,6 +80,14 @@ pub async fn assemble(ctx: DalContext, component_id: ComponentId) -> crate::Resu
     let resource_value_attribute_value_id =
         Component::attribute_value_for_prop(ctx, component_id, &["root", "resource_value"]).await?;
 
+    let dal_component_diff = Component::get_diff(ctx, component_id).await?;
+    let mut diff = None;
+    if let Some(diff_code) = dal_component_diff.diff {
+        diff = diff_code.code;
+    }
```
nickgerace, line 87: Only because you asked 😉

```suggestion
    let diff = dal_component_diff.diff.map(|diff_code| diff_code.diff);
```

nickgerace (approved):

nickgerace (dismissed):

stack72 (approved):

nickgerace (approved):

PR #6048, opened 2025-05-07
Add MVs for incoming socket-to-socket connections

nickgerace:
## Description

This change adds MVs for incoming socket-to-socket connections. The MVs have a "Beta" suffix for easy invalidations once we test all four kinds of connections needed in the new UI.

As a consequence of these changes, the macro(s) used for MVs have been adjusted to handle enums with struct variants.

For the record, @fnichol architected the vast majority of this. I just wired it up to the frontend and drove it home. Go Fletcher!

### DIFF
--- a/lib/si-frontend-types-macros/src/lib.rs
+++ b/lib/si-frontend-types-macros/src/lib.rs
@@ -36,7 +36,7 @@ fn derive_frontend_checksum(
 
     match &type_data {
         Data::Struct(struct_data) => derive_frontend_checksum_struct(ident, struct_data, errors),
-        Data::Enum(_) => derive_frontend_checksum_enum(ident),
+        Data::Enum(enum_data) => derive_frontend_checksum_enum(ident, enum_data, errors),
         _ => bail!("FrontendChecksum can only be derived for structs and enums"),
     }
 }
@@ -55,9 +55,11 @@ fn derive_frontend_checksum_struct(
             );
             continue;
         };
-        field_update_parts.push(
-            quote! { hasher.update(FrontendChecksum::checksum(&self.#field_ident).as_bytes()); },
-        )
+        field_update_parts.push(quote! {
+            hasher.update(
+                crate::checksum::FrontendChecksum::checksum(&self.#field_ident).as_bytes()
+            );
+        })
     }
     errors.into_result()?;
 
@@ -65,15 +67,15 @@ fn derive_frontend_checksum_struct(
     field_updates.extend(field_update_parts);
 
     let checksum_fn = quote! {
-        fn checksum(&self) -> Checksum {
-            let mut hasher = ChecksumHasher::new();
+        fn checksum(&self) -> ::si_events::workspace_snapshot::Checksum {
+            let mut hasher = ::si_events::workspace_snapshot::ChecksumHasher::new();
             #field_updates
             hasher.finalize()
         }
     };
 
     let output = quote! {
-        impl FrontendChecksum for #ident {
+        impl crate::checksum::FrontendChecksum for #ident {
             #checksum_fn
         }
     };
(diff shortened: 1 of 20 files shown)


wendybujalski (approved):

wendybujalski (dismissed):

lib/si-frontend-types-rs/src/newhotness/component_connections.rs
```diff
@@ -0,0 +1,84 @@
...
+};
+
+#[derive(
+    Clone, Debug, Deserialize, Eq, PartialEq, Ord, PartialOrd, Serialize, FrontendChecksum,
+)]
+#[remain::sorted]
+#[serde(rename_all = "camelCase", tag = "kind")]
+pub enum Connection {
+    Prop {
+        from_component_id: ComponentId,
```
jobelenus, line 33: When we were talking about the idea of "weak references" being a thing, I was talking about it just like `Reference`, only without the `checksum`. My expectation was:
```
from_component_id: WeakReference<Component>
```
I want to explicitly call out that this ID correlated to the `EntityId` of a different `EntityKind`. I want the entity kind _in_ the `WeakReference` data.

We can ship this for now.. but when we drop "Beta" from this. Let's also create a `WeakReference`. That way, in `webworker.ts` we can actually write re-usable code that checks _every_ MV for `WeakReference` and knows which `EntityKind` to look up. That way we don't have to keep writing `if` blocks every time.

jobelenus (dismissed):

lib/si-frontend-types-rs/src/newhotness/component_connections.rs
nickgerace: 100%. That will be a fast follow. Good call out.

lib/si-frontend-types-macros/src/lib.rs
```diff
@@ -55,43 +55,127 @@ fn derive_frontend_checksum_struct(
...
+                    }
+                });
+                let checksum_fields_stream = TokenStream::from_iter(checksum_fields);
+
+                quote! {
+                    #ident::#variant_ident(#fields_named) => {
+                        #checksum_fields_stream
+                    }
+                }
+            }
```
jhelwig, line 147: Not something I think needs to block these changes, but I think it's possible to get hash collisions with these derivations. I think there is a form of this problem with the existing derivation for structs, too.

With:
```rust
enum Foo {
    Bar(Vec<u8>),
    Baar(Vec<u8>, Vec<u8>),
    Baz {
        vec_field: Vec<u8>,
    },
    Qux {
        also_a_vec: Vec<u8>,
    },
    Quux {
        first_vec: Vec<u8>,
        second_vec: Vec<u8>,
    },
}
```

I believe these would all have the same checksum:
```rust
Foo::Bar(vec![0x00, 0x01, 0x02])

Foo::Baar(vec![0x00, 0x01, 0x02], vec![])
Foo::Baar(vec![0x00], vec![0x01, 0x02])

Foo::Baz { vec_field: vec![0x00, 0x01, 0x02] }

Foo::Qux { also_a_vec: vec![0x00, 0x01, 0x02] }

Foo::Quux {
    first_vec: vec![0x00, 0x01, 0x02],
    second_vec: vec![],
}
Foo::Quux {
    first_vec: vec![0x00, 0x01],
    second_vec: vec![0x02],
}
...
```

It should be pretty unusual for this to happen, but if the same _sequence_ of bytes happens, regardless of whether the bytes are still parts of the same field then I believe we'd end up with the same checksum for each variant. If we include the variant name as part of the checksum, then `Bar`/`Baar`/`Baz`/`Qux`/`Quux` shouldn't end up with the same checksum, but that still leaves the two `Baar`, and the two `Quux` with the same checksum as the other instance of the same variant.

To handle the two `Baar`, and the two `Quux` having the same checksum as the other instance of the same variant, I think we'd need to update the hasher with `hasher.update(index_of_unnamed_field_or_name_of_field_as_bytes); hasher.update(field_data_as_bytes)`.

jhelwig (approved):
Left a comment about the frontend checksum impl derivation for enums, but I don't think it should block this PR.

lib/si-frontend-types-macros/src/lib.rs
fnichol: Oh yeah, that's a really good idea, agreed! The variant is at least a bit more data to ensure a little more uniqueness

nickgerace: I've already made strides at these changes, so let's do it in this PR. Thanks for pointing this out @jhelwig!

jobelenus (approved):

lib/si-frontend-types-macros/src/lib.rs
nickgerace: Okay, I believe the latest commit fixes the issue and there is now a unit test for it.

jhelwig (approved):

PR #6052, opened 2025-05-07
fix(si-split-graph): cycle detection

zacharyhamm:
Cycle detection was failing because the cleanup of the graph did not
fully remove nodes when they were the children of an external source
which was removed.

Removal of "orphaned" nodes (nodes with no incoming edges) must proceed
in tandem with the removal of external source edges whose "tail" is the
orphaned node. That will orphan new nodes, so the process has to be
repeated until there are no more orphaned nodes left. Previously the
split graph cleanup would only remove the first layer if there were
children of the orphaned node pointed to by an external source edge.
The result would be weird walks of the graph during the cross-graph dfs
searching for cycles, reporting cycles on acyclic graphs.

### DIFF
--- a/lib/dal-test/src/signup.rs
+++ b/lib/dal-test/src/signup.rs
@@ -49,7 +49,7 @@ impl WorkspaceSignup {
         ctx.set_workspace_split_snapshot(workspace_snapshot);
 
         migrate_intrinsics_no_commit(ctx).await.map_err(Box::new)?;
-        // crate::test_exclusive_schemas::migrate(ctx).await?;
+        crate::test_exclusive_schemas::migrate(ctx).await?;
 
         let workspace_snapshot_address = ctx
             .workspace_snapshot()?
@@ -106,18 +106,20 @@ impl WorkspaceSignup {
         user_email: impl AsRef<str>,
         token: impl AsRef<str>,
     ) -> color_eyre::Result<Self> {
-        let workspace =
-            Workspace::new_from_builtin(ctx, WorkspacePk::generate(), workspace_name, token)
-                .await?;
+        let use_split = false;
 
-        // let workspace = Self::new_split_graph_workspace(
-        //     ctx,
-        //     WorkspacePk::generate(),
-        //     workspace_name,
-        //     token,
-        //     500,
-        // )
-        // .await?;
+        let workspace = if use_split {
+            Self::new_split_graph_workspace(
+                ctx,
+                WorkspacePk::generate(),
+                workspace_name,
+                token,
+                500,
+            )
+            .await?
+        } else {
+            Workspace::new_from_builtin(ctx, WorkspacePk::generate(), workspace_name, token).await?
+        };
 
         let key_pair = KeyPair::new(ctx, "default").await?;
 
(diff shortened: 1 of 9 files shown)


lib/dal/examples/rebase/main.rs
```diff
@@ -39,39 +62,149 @@ fn main() -> Result<()> {
...
+        assert!(toposort(&read_only_graph).is_ok());
+        assert!(toposort(&working_copy).is_ok());
+
+        let updates = read_only_graph.detect_updates(&working_copy);
+        let mut graph_to_update = read_only_graph.clone();
+        graph_to_update.perform_updates(&updates);
+
+        if let Err(Cycle(node_id)) = toposort(&graph_to_update) {
+            println!("Cycle detected at node ID: {}", node_id);
+            // dbg!(
```
britmyerss, line 81: Do you want to keep this but only do it if in debug? (i.e. do you expect it to be helpful to have if we get this error?)

zacharyhamm: that's a good question. I could revert the rebase example so it's quicker to use it for the current graph format

zacharyhamm: (to be clear this is just the rebase example code, this is not a production code path)

lib/si-split-graph/src/subgraph.rs
```diff
@@ -103,42 +103,37 @@ where
...
-            for node_index in orphaned_node_indexes {
-                self.graph.remove_node(node_index);
-            }
+        for index in indexes_to_remove {
+            self.graph.remove_node(index);
         }
 
-        let mut removed_ids = vec![];
+        removed_ids
+    }
```
jhelwig, line 125: Won't this need to loop until it stops finding externals (that aren't the root node)? It looks like it would only remove the first "layer" of externals, and not anything that becomes an external when that first set is removed.

lib/dal-test/src/signup.rs
```diff
@@ -106,18 +106,20 @@ impl WorkspaceSignup {
...
+                ctx,
+                WorkspacePk::generate(),
+                workspace_name,
+                token,
+                500,
+            )
+            .await?
+        } else {
+            Workspace::new_from_builtin(ctx, WorkspacePk::generate(), workspace_name, token).await?
+        };
```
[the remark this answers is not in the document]
zacharyhamm: This is just so we can turn on the split graph tests in development more easily. It's not feasible right now to run all the tests under both graphs, since it would double the test time. I might hook it up to a compiler cfg and thread that through buck at some point.

lib/si-split-graph/src/subgraph.rs
zacharyhamm: Yes, that happens at the next level up, in the cleanup function for the splitgraph itself. It loops over all the subgraphs, removing externals, then removing the external source edges, and repeating.

jhelwig: Yeah, I saw that `SplitGraph::cleanup` looped. Even though this is `pub(crate)`, if it's not going to handle its own looping, it should probably at least have a doc comment basically saying "Don't use this, unless you know exactly what you're doing. Go use `SplitGraph::cleanup` instead.".

zacharyhamm: @jhelwig i've added a doc comment to that effect

jhelwig (approved):

PR #6067, opened 2025-05-08
fix: don't name NATS Ordered Consumers

fnichol:
This change removes explicit names when creating [Ordered Consumers]. Ordered consumers are ephemeral and the NATS server may re-create these consumers often server-side. If a name is provided in the consumer config, it's likely that the NATS server attempts to re-use the user-provided name which results in a consumer create error.

The fix here is to not explicitly name the consumers and allow the client and server to coordinate when re-creating consumers is appropriate. Note that in the Rust client, that field type for `OrderedConfig.name` is `Option<String>` with the default value being `None`. As a result we can omit setting this field and rely on the defaults to get us what we need.

[Ordered Consumers]: https://docs.nats.io/using-nats/developer/develop_jetstream/consumers#ordered-consumers

### DIFF
--- a/lib/edda-server/src/handlers.rs
+++ b/lib/edda-server/src/handlers.rs
@@ -34,7 +34,6 @@ use tokio_util::{
     sync::CancellationToken,
     task::TaskTracker,
 };
-use ulid::Ulid;
 
 use crate::{
     app_state::AppState,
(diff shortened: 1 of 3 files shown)


nickgerace (approved):
Yes!!

jhelwig:
@fnichol Do you know if there's ever a valid reason to want to name the ordered consumers? Feels like the kind of thing that should return an error when configuring the client, instead of allowing such a broken config.

fnichol:
> @fnichol Do you know if there's ever a valid reason to want to name the ordered consumers? Feels like the kind of thing that should return an error when configuring the client, instead of allowing such a broken config.

I believe there is not ever a valid reason and that could be considered an error--either handled by clients or the NATS server itself.

PR #6071, opened 2025-05-08
feat: use-or-create kv store FRIGG

johnrwatson:
This change prevents a panic in the implementation of Frigg where the remote nats server nats kv configuration for the given name doesn't match the one given to the client. This also matches the behaviour we have with other NATS Jetstream streams if for example the replication count doesn't match what the client was expecting (Which is ok!)

Now the `frigg_kv` function will use the existing stream if it finds a stream with the correct name, otherwise it'll create it with the given client configuration.

<div></div>

### DIFF
--- a/lib/frigg/src/lib.rs
+++ b/lib/frigg/src/lib.rs
@@ -18,7 +18,10 @@ use si_data_nats::{
                 StreamError,
                 push::OrderedConfig,
             },
-            context::RequestError,
+            context::{
+                KeyValueErrorKind,
+                RequestError,
+            },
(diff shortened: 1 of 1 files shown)


fnichol (dismissed):
We'll see about upstreaming this afterwards--it would be a `jetstream::Context::get_or_create_key_value()` method.

johnrwatson:
I've tested this locally, works all golden. Only realistic way to test if it works with R3 for example is by shipping to tools.

fnichol (approved):

PR #6097, opened 2025-05-12
fix(dal): correctly delete components with incoming ValueSubscription edges

zacharyhamm:
Ensures we remove the root attribute value from a component when deleting. Additionally adds a correction which deletes all edges incoming to the root attribute value if the component will be deleted, ensuring deletion holds across change sets, even if a new value subscription has been added in the change set.

### DIFF
--- a/lib/dal/src/workspace_snapshot/node_weight/component_node_weight.rs
+++ b/lib/dal/src/workspace_snapshot/node_weight/component_node_weight.rs
@@ -423,7 +423,7 @@ impl CorrectTransforms for ComponentNodeWeight {
     ) -> CorrectTransformsResult<Vec<Update>> {
         // New components don't have any corrective transforms.
         let component_id: ComponentId = self.id.into();
-        let Some(component) = graph.get_node_index_by_id_opt(component_id) else {
+        let Some(component_node_idx) = graph.get_node_index_by_id_opt(component_id) else {
             return Ok(updates);
         };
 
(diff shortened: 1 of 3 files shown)


lib/dal/tests/integration_test/attribute_value/subscription.rs
```diff
@@ -304,6 +304,89 @@ async fn subscribe_to_two_values(ctx: &mut DalContext) -> Result<()> {
     Ok(())
 }
 
+#[test]
```
zacharyhamm, line 307: This test fails without the correction

jhelwig (approved):

lib/dal/src/workspace_snapshot/node_weight/component_node_weight.rs
```diff
@@ -517,12 +520,27 @@ impl CorrectTransforms for ComponentNodeWeight {
...
+            if let Some((_, _, root_av_idx)) = graph
+                .edges_directed_for_edge_weight_kind(
+                    component_node_idx,
+                    Outgoing,
+                    EdgeWeightKindDiscriminants::Root,
+                )
+                .next()
+            {
+                remove_edges.extend(graph.edges_directed(root_av_idx, Incoming));
+            }
```
fnichol, line 539: Ahhhhh!

lib/dal/tests/integration_test/attribute_value/subscription.rs
fnichol: Props to you for a regression test, A++!

fnichol (approved):
Nice catch, to my eyes this looks great!

PR #6102, opened 2025-05-13
feat(dal): Respect subscriptions when calculating action dependencies

jkeiser:
This causes actions to run in the proper order when components depend on each other via subscription.

Fixes ENG-3019.

## Tests

- [X] New integration tests
- [ ] Manual test: create VPC and subnet with socket connections and make sure VPC goes first
- [ ] Manual test: create VPC and subnet with subscriptions and make sure VPC goes first

### DIFF
--- a/lib/dal/src/component.rs
+++ b/lib/dal/src/component.rs
@@ -96,6 +96,7 @@ use crate::{
         ActorViewError,
     },
     attribute::{
+        path::AttributePath,
         prototype::{
             AttributePrototypeError,
             AttributePrototypeSource,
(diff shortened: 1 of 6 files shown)


lib/dal/src/component.rs
```diff
@@ -1321,6 +1322,14 @@ impl Component {
         Ok(input_socket_ids)
     }
 
+    pub async fn subscribers(
```
zacharyhamm, line 1325: A doc comment here would be great, explaining what a Subscriber is

jkeiser:
Thanks!

I didn't directly test delete actions because I was reusing the dependency graph's reasoning for it; I should have. I've got some tests written for it, and will put them up shortly once I've checked that they, like, pass :)

zacharyhamm (approved):

jhelwig (approved):

PR #6103, opened 2025-05-13
feat(luminork): Add support for autogenerating SDKs and docs

stack72:
The first iteration of this used direct path structures. Now to generate the SDKs, we do the following:

```
buck2 run //generated-sdks:update_sdks
```

This generates the openapi.json file and then uses that file as inputs for generating the SDKs

To generate the docs:

```
buck2 run //app/docs:generate_api_docs
```

This will ensure that the correct openapi.json file has been generated and then updates the sources for the public-api.md file


johnrwatson (approved):
Looks good to me

britmyerss (approved):

PR #6104, opened 2025-05-13
Connections panel, outgoing connections, weak_references & naming scheme

jobelenus:
We now have a codified pattern: `get` -> `getReferences` -> `getComputed`. This is how we string together strong and weak references. However, it comes with a limitation; circular references/infinite loops/cycles, call it what you will, but its bad 😈 .

This implication of this is that, we can do all the things we want—except compute the outgoing connections when looking up input connections. Thats the cycle. (There is probably a clever way around this... but my brain has been bashing on this its starting to turn into mush... so, we can figure that out another day).

What does this mean for you?! When you are looking up connections and want outgoing connections—you need to make a separate call. Thats it, for now. You'll see this example in the `ConnectionsPanel`. When you're asking for a `Component` or `ComponentList` or `ViewComponentList`, you'll get the `outputCount` populated (and it stays reactive thanks to weak references!).

**_Note_**: the outgoing connections cache is a tad slow right now. We can take a deeper dive into speeding it up, potentially. But for now, the way it works is that all the atom build, regardless of if they have a ready cache of connections. Then once the connections are built—that cache busts, and things re-query, re-compute, and the UI updates.

The best way to observe this effect is by going to the Map page, selecting a component, and hitting refresh. The map will refresh, the component tile will show. Connections won't show. Until the list is generated. And then the connections show.

### Outgoing Connections Cache
I've moved around the cache busting bits a little. This achieves a fully up to date list of outgoing connections is ready and waiting for each computed operation that requires it.
1. When patching atoms, note which kinds are being patched
2. Mark the connections dirty based on the kind (`IncomingConnections`, `IncomingConnectionsList`)
3. Patch the changes
4. Hold the atoms that must be busted
5. Recreate the connections
6. Bust the caches, based on the weak references, et al.

### Naming Scheme
This is taken from my code comment in the `dbinterface.ts` file:
```
/**
 * NAMING RULES
 * 1. If the type is not mutated at all
 * (e.g. it does not have an if-block in `getReferences` or `getComputed`)
 * THOU SHALT name it according the entity kind--EXACT MATCH PLEASE
 * 
 * 2. If the type is mutated, you will prefix the EXACT entity kind...
 * THOU SHALT prefix the type that comes over the wire with `Edda`
 * THOU SHALT prefix the type that returns from `bifrost` fn with `Bifrost`
 * 
 * 3. `EddaXXX` types SHALL NOT have `BifrostXXX` types on them
 * 4. `BifrostXXX` types SHALL NOT have `EddaXXX` types on them
 * 5. Vue components SHALL NEVER use `EddaXXX` types
 * 6. `getReferences` SHALL NEVER return an `EddaXXX` type
 *    (e.g. perform the full translation from Edda to Bifrost)
 * 7. `getReferences` SHALL set default/warning data that `getComputed` will write over
 */
```
I was confusing the hell out of myself without a stricter naming scheme, "what type went with what entity kind?! is this before the reference lookup or not?!"... Happy to take other suggestions, just felt the need to put some pattern down to our typing schemes

### DIFF
--- a/app/web/src/newhotness/ComponentGridTile.vue
+++ b/app/web/src/newhotness/ComponentGridTile.vue
@@ -52,17 +52,28 @@
         />
         <div>Resource</div>
       </li>
-      <hr class="border-neutral-500" />
-      <li>
-        <Icon name="output-connection" size="sm" />
-        <div>Inputs</div>
-        <PillCounter :count="component.inputCount" size="sm" class="ml-auto" />
-      </li>
-      <li>
-        <Icon name="input-connection" size="sm" />
-        <div>Outputs</div>
-        <PillCounter :count="0" size="sm" class="ml-auto" />
-      </li>
+      <!-- NOTE: when coming from the Map page we don't have accurate outputCount, hiding this -->
+      <template v-if="!props.hideConnections">
+        <hr class="border-neutral-500" />
+        <li>
+          <Icon name="output-connection" size="sm" />
+          <div>Inputs</div>
+          <PillCounter
+            :count="component.inputCount"
+            size="sm"
+            class="ml-auto"
+          />
+        </li>
+        <li>
+          <Icon name="input-connection" size="sm" />
+          <div>Outputs</div>
+          <PillCounter
+            :count="component.outputCount"
+            size="sm"
+            class="ml-auto"
+          />
+        </li>
+      </template>
     </ol>
     <!-- <footer class="grid grid-cols-2 p-xs">
       <div class="place-self-start">
@@ -86,12 +97,13 @@ import {
 } from "@si/vue-lib/design-system";
 import clsx from "clsx";
 import { computed } from "vue";
-import { BifrostComponent } from "@/workers/types/dbinterface";
+import { Component } from "@/workers/types/dbinterface";
 import StatusIndicatorIcon from "@/components/StatusIndicatorIcon.vue";
 import { getAssetIcon } from "./util";
 
 const props = defineProps<{
-  component: BifrostComponent;
+  component: Component;
+  hideConnections?: boolean;
 }>();
 
 const qualificationSummary = computed(() => {
--- a/app/web/src/newhotness/Map.vue
+++ b/app/web/src/newhotness/Map.vue
@@ -4,12 +4,16 @@
     <div
       v-if="selectedComponent"
       id="selection"
-      class="absolute top-[11vh] max-w-[350px] min-w-[300px] right-3 p-xs bg-action-800 border-2 border-action-700"
+      class="absolute top-[110px] max-w-[350px] min-w-[300px] right-3 p-xs bg-action-800 border-2 border-action-700 scrollable"
+      style="max-height: calc(100vh - 115px)"
     >
+      <!-- NOTE: 110 is the fixed size (nothing expands/dynamic) above, extra 5px on the bottom, unforuntately tailwind doesn't support that calc -->
       <ComponentGridTile
         :component="selectedComponent"
+        hideConnections
         @dblclick="componentNavigate(selectedComponent.id)"
       />
+      <ConnectionsPanel :component="selectedComponent" />
     </div>
 
     <div
@@ -104,8 +108,8 @@ import * as _ from "lodash-es";
 import { Fzf } from "fzf";
 import { ComponentId } from "@/api/sdf/dal/component";
 import {
-  BifrostComponent,
-  BifrostIncomingConnections,
+  Component,
+  BifrostComponentConnections,
   BifrostIncomingConnectionsList,
 } from "@/workers/types/dbinterface";
 import { bifrost, useMakeArgs, useMakeKey } from "@/store/realtime/heimdall";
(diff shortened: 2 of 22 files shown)


nickgerace (approved):

app/web/src/newhotness/ComponentGridTile.vue
```diff
@@ -52,16 +52,17 @@
         />
         <div>Resource</div>
       </li>
-      <hr class="border-neutral-500" />
-      <li>
+      <!-- NOTE: when coming from the Map page we don't have accurate outputCount, hiding this -->
+      <hr v-if="!props.hideConnections" class="border-neutral-500" />
```
wendybujalski, line 56: would be better to wrap this whole section in a `<template>` instead of repeating the conditional 3 times

app/web/src/newhotness/Map.vue
```diff
@@ -4,12 +4,14 @@
     <div
       v-if="selectedComponent"
       id="selection"
-      class="absolute top-[11vh] max-w-[350px] min-w-[300px] right-3 p-xs bg-action-800 border-2 border-action-700"
+      class="absolute top-[11vh] max-w-[350px] min-w-[300px] right-3 p-xs bg-action-800 border-2 border-action-700 max-h-[85vh] scrollable"
```
wendybujalski, line 7: Basing the maximum height on `vh` like this can be risky for bugs if the other elements taking up that remaining `15vh` are not also set based on `vh` and I know that the `Navbar` has a fixed pixel height. There is probably a better way to get the intended result of this style.

app/web/src/workers/webworker.ts
```diff
@@ -388,6 +399,49 @@ const bustCacheAndReferences: BustCacheFn = async (
...
+  const connectionsById = await _getOutgoingConnectionsByComponentId(
+    workspaceId,
+    changeSetId,
+  );
+  cachedConnections[changeSetId] = connectionsById;
+
+  delete dirtyConnections[`${workspaceId}-${changeSetId}`];
+
+  // whenever these change, the outgoing connections must bust as well
+  bustCacheFn(workspaceId, changeSetId, "OutgoingConnections", changeSetId);
```
nickgerace, line 424: 👍

app/web/src/newhotness/Map.vue
wendybujalski: Assuming that all of the other elements in that `15vh` are a fixed height (I think they are? Should probably double check) then the better style here would be `max-height: calc(100vh - ?px)` with the ? being the amount of pixels of the other elements. To do that CSS in Tailwind, you would put `max-h-[calc(100vh - ?px)]`

nickgerace (approved):
Excellent work!!

app/web/src/newhotness/layout_components/ConnectionLayout.vue
```diff
@@ -0,0 +1,51 @@
+<!-- eslint-disable vue/no-multiple-template-root -->
+<template>
+  <h3 class="m-xs">{{ props.label }}</h3>
+  <ul class="flex flex-col gap-xs mx-xs">
+    <li
+      v-for="conn in props.connections"
+      :key="`${conn.key}`"
+      class="p-xs border-neutral-600 border"
```
wendybujalski, line 8: Dark/light mode styles here?

app/web/src/newhotness/logic_composables/navigation_stack.ts
wendybujalski, line 1: Clever!

app/web/src/workers/types/dbinterface.ts
```diff
@@ -223,7 +230,25 @@ interface WeakReference {
...
+ * THOU SHALT prefix the type that comes over the wire with `Edda`
+ * THOU SHALT prefix the type that returns from `bifrost` fn with `Bifrost`
+ *
+ * 3. `EddaXXX` types SHALL NOT have `BifrostXXX` types on them
+ * 4. `BifrostXXX` types SHALL NOT have `EddaXXX` types on them
+ * 5. Vue components SHALL NEVER use `EddaXXX` types
+ * 6. `getReferences` SHALL NEVER return an `EddaXXX` type
+ *    (e.g. perform the full translation from Edda to Bifrost)
+ * 7. `getReferences` SHALL set default/warning data that `getComputed` will write over
+ */
```
wendybujalski, line 251: This is great and very clear :)

wendybujalski (changes requested):
Awesome work, love the comments which add clarity and guidance, just a few small fixes I'd like to see before I wang the approve button 👍

app/web/src/newhotness/Map.vue
jobelenus: ahh rats... tailwind doesn't seem to support it `calc(X-Ypx)`... I tried, it wasn't working, went looking to see if I did it wrong... found this: https://github.com/tailwindlabs/tailwindcss/discussions/11324. But! I got it in there anyhow

app/web/src/newhotness/Map.vue
```diff
@@ -4,12 +4,15 @@
     <div
       v-if="selectedComponent"
       id="selection"
-      class="absolute top-[11vh] max-w-[350px] min-w-[300px] right-3 p-xs bg-action-800 border-2 border-action-700"
-    >
+      class="absolute top-[110px] max-w-[350px] min-w-[300px] right-3 p-xs bg-action-800 border-2 border-action-700 scrollable"
+      style="max-height: calc(100vh - 115px);"
+    > <!-- NOTE: 110 is the fixed size (nothing expands/dynamic) above, extra 5px on the bottom, unforuntately tailwind doesn't support that calc -->
```
wendybujalski, line 9: Good note, will save us a lot of headache if/when we change the height of those other elements!

wendybujalski (dismissed):

wendybujalski (approved):
👍

PR #6106, opened 2025-05-13
Fix duplicate incoming connections and wrong prop

nickgerace:
## Description

This change fixes two bugs with incoming connections:

1) Fix duplicate incoming connections by ensuring we do not accidentally process outgoing connections during MV generation.
2) Fix incorrect prop IDs for prop-to-prop connections on the "from" side by using the correct ID.

This change was tested with an included integration test and includes the following misc changes:

- Rename the attribute value ID for socket functions to be shorthand
- Move the "make_subscription" helper to the helpers module

### DIFF
--- a/lib/dal-materialized-views/src/incoming_connections.rs
+++ b/lib/dal-materialized-views/src/incoming_connections.rs
@@ -52,12 +52,7 @@ async fn socket_to_socket(ctx: &DalContext, component_id: ComponentId) -> Result
     for input_socket in input_sockets {
         let input_socket_id = input_socket.id();
         let input_socket_attribute_value_id =
-            InputSocket::component_attribute_value_for_input_socket_id(
-                ctx,
-                input_socket_id,
-                component_id,
-            )
-            .await?;
+            InputSocket::component_attribute_value_id(ctx, input_socket_id, component_id).await?;
         let attribute_prototype_id =
             AttributeValue::prototype_id(ctx, input_socket_attribute_value_id).await?;
         let attribute_prototype_argument_ids =
@@ -68,12 +63,8 @@ async fn socket_to_socket(ctx: &DalContext, component_id: ComponentId) -> Result
             continue;
         }
 
-        let input_socket_attribute_value_path =
-            AttributeValue::get_path_for_id(ctx, input_socket_attribute_value_id)
-                .await?
-                .ok_or(Error::EmptyPathForAttributeValue(
-                    input_socket_attribute_value_id,
-                ))?;
+        let (_, input_socket_attribute_value_path) =
+            AttributeValue::path_from_root(ctx, input_socket_attribute_value_id).await?;
 
         for attribute_prototype_argument_id in attribute_prototype_argument_ids {
             let attribute_prototype_argument =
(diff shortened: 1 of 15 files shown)


lib/dal-materialized-views/src/incoming_connections.rs
```diff
@@ -150,7 +144,8 @@ async fn prop_to_prop(ctx: &DalContext, component_id: ComponentId) -> Result<Vec
                         AttributeValue::path_from_root(ctx, from_attribute_value_id).await?;
                     let from_component_id =
                         AttributeValue::component_id(ctx, from_attribute_value_id).await?;
-                    let from_prop_id = AttributeValue::prop_id(ctx, attribute_value_id).await?;
```
jobelenus, line 153: was this just an oopsie wrong variable? or .. am i missing something more subtle here?

lib/dal/tests/integration_test/materialized_views.rs
```diff
@@ -216,16 +229,224 @@ async fn component(ctx: &DalContext) -> Result<()> {
...
+                    from_socket_id: beta_output_socket.id(),
+                    from_socket_name: beta_output_socket.name().to_owned(),
+                    to_component_id: charlie.id().into(),
+                    to_socket_id: charlie_input_socket.id(),
+                    to_socket_name: charlie_input_socket.name().to_owned(),
+                    to_attribute_value_id: charlie_input_socket_attribute_value_id,
+                    to_attribute_value_path: charlie_input_socket_attribute_value_path,
+                }
+            ], // expected
+            charlie_mv.connections // actual
```
nickgerace, line 447: These are sorted on the way out because MV generation requires stable ordering, so we do not need to sort the connections again.

lib/dal/tests/integration_test/materialized_views.rs
```diff
@@ -216,16 +229,224 @@ async fn component(ctx: &DalContext) -> Result<()> {
...
+            make_subscription(
+                ctx,
+                beta.id(),
+                beta_domain_name_attribute_value_path.as_str(),
+            )
+            .await?,
+        )
+        .await?;
+        ChangeSetTestHelpers::commit_and_update_snapshot_to_visibility(ctx).await?;
+    }
```
nickgerace, line 331: This is the "guts" of the test

lib/dal-materialized-views/src/incoming_connections.rs
nickgerace: I was just about to comment. This was the second bug!

lib/dal-materialized-views/src/incoming_connections.rs
```diff
@@ -150,7 +144,8 @@ async fn prop_to_prop(ctx: &DalContext, component_id: ComponentId) -> Result<Vec
...
+                    let from_prop_id =
```
nickgerace, line 147: Bug number 2

lib/dal-materialized-views/src/incoming_connections.rs
```diff
@@ -86,34 +77,37 @@ async fn socket_to_socket(ctx: &DalContext, component_id: ComponentId) -> Result
...
-            }
+            // We've found the connection if the attribute prototype argument has targets whose
+            // destination component ID for the current component. We also need to ensure that we
+            // do not accidentally collect an outgoing connection.
+            let source_component_id = match attribute_prototype_argument.targets() {
+                Some(targets) if targets.destination_component_id == component_id => {
+                    targets.source_component_id
+                }
+                Some(_) | None => continue,
+            };
```
nickgerace, line 88: Bug number 1

jobelenus (approved):

PR #6107, opened 2025-05-13
fix(bifrost): Make components draggable to canvas again

jkeiser:
This makes it possible to drag components onto the canvas again in environments using bifrost for the old UI (localdev).

## Tests

- [ ] Regression tests pass
- [ ] Drag uninstalled and installed components onto the canvas


vbustamante (approved):
yeah this should go in ASAP

PR #6109, opened 2025-05-13
Show prop-to-prop connections on diagram and edge details panel

vbustamante:
image

### DIFF
--- a/app/web/src/components/Connection.vue
+++ b/app/web/src/components/Connection.vue
@@ -6,18 +6,27 @@
     >
       ID = {{ connection.id }}
     </div>
-    <SocketCard
-      :socket="connection.fromSocket"
-      outputSocket
+    <ConnectionEndCard
+      v-if="isSocketConnection(connection)"
       :changeStatus="connection.changeStatus"
+      :componentName="componentNames.from"
+      :subjectLabel="connection.fromSocket.def.label"
+      type="input-socket"
+    />
+    <ConnectionEndCard
+      v-else
+      :changeStatus="connection.changeStatus"
+      :componentName="componentNames.from"
+      :subjectLabel="connection.fromAttributePath"
+      type="prop"
     />
     <div :class="clsx('_connection-label border-l-2', statusColors)">
       <div class="flex flex-row items-center">
         <DetailsPanelTimestamps
-          noMargin
           :changeStatus="connection.changeStatus"
           :created="connection.createdInfo"
           :deleted="connection.deletedInfo"
+          noMargin
         />
         <template v-if="showMenu">
           <DetailsPanelMenuIcon @click="openMenu" />
@@ -25,9 +34,19 @@
         </template>
       </div>
     </div>
-    <SocketCard
-      :socket="connection.toSocket"
+    <ConnectionEndCard
+      v-if="isSocketConnection(connection)"
       :changeStatus="connection.changeStatus"
+      :componentName="componentNames.to"
+      :subjectLabel="connection.toSocket.def.label"
+      type="input-socket"
+    />
+    <ConnectionEndCard
+      v-else
+      :changeStatus="connection.changeStatus"
+      :componentName="componentNames.to"
+      :subjectLabel="connection.toAttributePath"
+      type="prop"
     />
   </div>
 </template>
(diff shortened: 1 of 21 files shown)


app/web/src/components/Connection.vue
```diff
@@ -6,28 +6,47 @@
...
-      outputSocket
+    <ConnectionEndCard
+      v-if="isSocketConnection(connection)"
       :changeStatus="connection.changeStatus"
+      :componentName="componentNames.from"
+      :subjectLabel="connection.fromSocket.def.label"
+      type="input-socket"
+    />
+    <ConnectionEndCard
+      v-else-if="isSubscriptionConnection(connection)"
```
jkeiser, line 17: You should be able to do v-else here (typescript will use inference). That will help if we add more edge types in the future, because this will stop compiling instead of having a dangling else :)

jkeiser: (Same with other similar places)

app/web/src/components/Connection.vue
```diff
@@ -73,6 +118,30 @@ const componentsStore = useComponentsStore();
...
+    const toComponent =
+      componentsStore.allComponentsById[props.connection.toComponentId];
+    const fromComponent =
+      componentsStore.allComponentsById[props.connection.fromComponentId];
+    return {
+      to: toComponent?.def.displayName ?? "?",
+      from: fromComponent?.def.displayName ?? "?",
+    };
+  }
+  if (isSocketConnection(props.connection)) {
```
jkeiser, line 132: Also a very good place to use if / else, so that typescript will essentially do exhaustive matching!

app/web/src/components/ModelingDiagram/diagram_types.ts
```diff
@@ -408,6 +411,9 @@ export class DiagramEdgeData extends DiagramElementData {
   }
 
   get toSocketKey() {
+    if (!this.def.toSocketId) {
```
jkeiser, line 414: Oh, we probably don't need these exceptions anymore now that DiagramSocketEdgeData is a subclass with socket-speific `def`--we should get rid of them to avoid confusion.

app/web/src/newhotness/Component.vue
```diff
@@ -6,10 +6,10 @@
     <VButton
       class="border-0 mr-2em"
       icon="arrow--left"
+      label="Back"
```
jkeiser, line 9: Did any newhotness files actually need to change? I think there's only formatter sorting changes here. Probably best not to commit them for now, to avoid unnecessary conflicts.

app/web/src/store/components.store.ts
```diff
@@ -1093,6 +1149,7 @@ export const useComponentsStore = (forceChangeSetId?: ChangeSetId) => {
               idsToDelete.forEach((id) => {
                 delete this.rawEdgesById[id];
               });
+              // TODO do this for subscriptions
```
jkeiser, line 1152: We should probably heed this ... I think it's probably easier now that we finished the type changes!

lib/sdf-server/src/service/v2/component/attributes.rs
```diff
@@ -355,19 +351,79 @@ impl TryFrom<ValueOrSourceSpec> for Option<Source> {
...
-            if ctx.workspace_snapshot()?.node_exists(component_id).await {
-                return Ok(Some(component_id));
-            }
+        if let Some(id) = self.resolve_as_id(ctx).await? {
+            return Ok(Some(id));
         }
         // Otherwise, try to find it by name
         Ok(Component::find_by_name(ctx, &self.0).await?)
     }
+
```
jkeiser, line 366: Oh, that's where all this went! I was wondering :)

jkeiser (approved):
A few changes probably worth making; but could decide on followup too, none of them are in the can't-live-with-it category.

app/web/src/newhotness/Component.vue
vbustamante: I'm conflicted about this because all the changes were made by the buck2 linter target - but I agree. I'll remove the changes to files that weren't functionally edited

jkeiser: oh, that's totally fine then imo! I wasn't sure if it was official linting or just something your editor did because you were in there.

app/web/src/store/components.store.ts
vbustamante: Gonna leave this one for the next PR

jkeiser (approved):

jkeiser (approved):

PR #6111, opened 2025-05-14
feat(dal): Allow arrays to subscribe to multiple values

jkeiser:
This PR allows you to subscribe to multiple values from a single array, and will put them together into a single array. It has two specific changes:

1. Allows multiple subscriptions to array values
   - Changes `AttributeValue::subscribe()` to `AttributeValue::set_to_subscriptions()`, setting multiple subscriptions at once
   - Adds `keepExistingSubscriptions` option to `PUT /component/:componentId/attributes` to make it easier for the "add subscription" button to be built for arrays
2. Uses `si:normalizeToArray` instead of `si:identity` when the subscriber is an array, to uplevel single values

It will return an error if you attempt to assign multiple subscriptions to a non-array value.

## Factor Budget

* Made the tests super ergonomic with some attribute::value::set/subscribe/get helpers, letting you pass component names and attribute paths to reference things
* AttributeValue::list_arguments_for_id -> list_arguments (there is no other variant)

## Testing

- [X] Test that subscribing from an array to an array flows the values through
- [X] Test that subscribing to a single value from an array uplevels it to an array
- [X] Test that subscribing to multiple scalars flows the values through as an array
- [x] Manually test that subscriptions still work for current values

### DIFF
--- a/lib/dal/src/attribute/value.rs
+++ b/lib/dal/src/attribute/value.rs
@@ -261,6 +261,10 @@ pub enum AttributeValueError {
     Secret(#[from] Box<SecretError>),
     #[error("serde_json: {0}")]
     SerdeJson(#[from] serde_json::Error),
+    #[error(
+        "attribute value {0} with type {1} must be set to 1 subscription, attempted to include {2} subscriptions"
+    )]
+    SingleValueMustHaveOneSubscription(AttributeValueId, PropKind, usize),
     #[error("transactions error: {0}")]
     Transactions(#[from] TransactionsError),
     #[error("try lock error: {0}")]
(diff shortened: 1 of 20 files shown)


jhelwig:
It's a bit difficult to find in the wall of output, but there's a failing test: `integration_test::materialized_views::incoming_connections`

The formatting doesn't copy well out of BuildKite...

```
Message:  assertion failed: `(left == right)`
--
 
Diff < left / right > :
[
Prop {
from_component_id: WeakReference {
kind: Component,
id: ReferenceId(
ComponentId(
<                    "01JV7P0XAWY6NBJ5JVSMDAH7Z8",
<                ),
<            ),
<            _kind: PhantomData<si_frontend_mv_types::reference::weak::markers::Component>,
<        },
<        from_attribute_value_id: AttributeValueId(
<            "01JV7P0XGC86E6QGGK95BBNWKQ",
<        ),
<        from_attribute_value_path: "/si/name",
<        from_prop_id: PropId(
<            "01JV7NRZDMESFM62VHD1KCXW9H",
<        ),
<        from_prop_path: "/root/si/name",
<        to_component_id: WeakReference {
<            kind: Component,
<            id: ReferenceId(
<                ComponentId(
<                    "01JV7P0YMJJB7VSJD6GT94Y9P0",
<                ),
<            ),
<            _kind: PhantomData<si_frontend_mv_types::reference::weak::markers::Component>,
<        },
<        to_prop_id: PropId(
<            "01JV7NRZDMESFM62VHD1KCXW9H",
<        ),
<        to_prop_path: "/root/si/name",
<        to_attribute_value_id: AttributeValueId(
<            "01JV7P0YMPCAZRRPWKD5KAJH5D",
<        ),
<        to_attribute_value_path: "/si/name",
<    },
<    Prop {
<        from_component_id: WeakReference {
<            kind: Component,
<            id: ReferenceId(
<                ComponentId(
"01JV7P0YHDBC81YQRJ2250XCJ0",
),
),
_kind: PhantomData<si_frontend_mv_types::reference::weak::markers::Component>,
},
from_attribute_value_id: AttributeValueId(
"01JV7P0YJ933ZR5TNBYW4KTAWH",
),
from_attribute_value_path: "/domain/name",
from_prop_id: PropId(
"01JV7NRZKFKTB1PRJGJ5T2X241",
),
from_prop_path: "/root/domain/name",
to_component_id: WeakReference {
kind: Component,
id: ReferenceId(
ComponentId(
"01JV7P0YMJJB7VSJD6GT94Y9P0",
),
),
_kind: PhantomData<si_frontend_mv_types::reference::weak::markers::Component>,
},
to_prop_id: PropId(
<            "01JV7NRZFFR8WG3AWQVH998Y72",
>            "01JV7NRZDMESFM62VHD1KCXW9H",
),
<        to_prop_path: "/root/domain/name",
>        to_prop_path: "/root/si/name",
to_attribute_value_id: AttributeValueId(
<            "01JV7P0YMR5EKPGJ2SHW7VHZA0",
>            "01JV7P0YMPCAZRRPWKD5KAJH5D",
),
<        to_attribute_value_path: "/domain/name",
>        to_attribute_value_path: "/si/name",
},
Socket {
from_component_id: WeakReference {
kind: Component,
id: ReferenceId(
ComponentId(
"01JV7P0YHDBC81YQRJ2250XCJ0",
),
),
_kind: PhantomData<si_frontend_mv_types::reference::weak::markers::Component>,
},
from_attribute_value_id: AttributeValueId(
"01JV7P0YHDBC81YQRJ2250XCJ4",
),
from_attribute_value_path: "",
from_socket_id: OutputSocketId(
"01JV7NRZKWRXYJWF5E72013J0W",
),
from_socket_name: "one",
to_component_id: WeakReference {
kind: Component,
id: ReferenceId(
ComponentId(
"01JV7P0YMJJB7VSJD6GT94Y9P0",
),
),
_kind: PhantomData<si_frontend_mv_types::reference::weak::markers::Component>,
},
to_socket_id: InputSocketId(
"01JV7NRZG2GEM9A3ZWEG587GQV",
),
to_socket_name: "one",
to_attribute_value_id: AttributeValueId(
"01JV7P0YMK36A8TQX34T0YNKRY",
),
to_attribute_value_path: "",
},
]
 
 
Location: lib/dal/tests/integration_test/materialized_views.rs:403
```

lib/dal/src/attribute/value.rs
```diff
@@ -2252,34 +2256,77 @@ impl AttributeValue {
         Ok(())
     }
 
-    pub async fn subscribe(
+    pub async fn set_to_subscriptions(
```
jhelwig, line 2259: It would be good to have a doc comment here emphasizing that it will overwrite any existing `AttributePrototype` and `AttributePrototypeArguments`, so any function choice, and any existing arguments will be lost. So, if the function was customized, or arguments were added that weren't subscriptions there is work to do before and after calling this function to preserve any of those if desired. And that there's still work to do before calling if the intent is to append new subscriptions, instead of wholesale replacement (like the work that the SDF endpoint does if `keep_existing_subscriptions` is `true`).

lib/dal/src/attribute/value/subscription.rs
```diff
@@ -23,4 +27,15 @@ impl ValueSubscription {
     ) -> AttributeValueResult<Option<AttributeValueId>> {
         self.path.resolve(ctx, self.attribute_value_id).await
     }
+
+    /// Find or create the attribute value a subscription points to
+    pub async fn vivify(&self, ctx: &DalContext) -> AttributeValueResult<AttributeValueId> {
+        self.path.vivify(ctx, self.attribute_value_id).await
+    }
```
jhelwig, line 34: From the doc comment, I'm not sure why this would ever be called?

If `Component A: /domain/some/destination/field` is subscribing to `Component B: /domain/data/source/field`, then `Component B: /domain/data/source/field` should only exist as an AV if something about `Component B` has determined that it should (manually set, function output from B, etc.), the fact that `Component A` is interested in the data in `Component B` should not affect the structure of `B` in any way.

jhelwig (changes requested):

lib/sdf-server/src/service/v2/component/attributes.rs
```diff
@@ -171,27 +184,33 @@ async fn update_attributes(
...
-                            AttributeValue::prop_id(ctx, source_root_id).await?;
-                        source_path.validate(ctx, source_root_prop_id).await?;
+                        subscription.validate(ctx).await?;
+
+                        let mut subscriptions = if keep_existing_subscriptions.unwrap_or(false) {
+                            AttributeValue::subscriptions(ctx, target_av_id).await?
+                        } else {
+                            None
+                        }
+                        .unwrap_or(vec![]);
```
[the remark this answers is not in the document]
jkeiser: Mainly because `subscriptions()` can also return `None`, so I would have had to put the `unwrap_or()` there anyway. I could see just putting it in there, too, neither way looks honestly all that great :)

lib/dal/src/attribute/value.rs
jkeiser: :+1:

lib/dal/src/attribute/value/subscription.rs
jkeiser: It was mainly so the tests could call it, but they really don't mean subscription here, they are using an (AV_ID, path) pair as a *reference* to something they want to set. I'll remove this and just have the caller do that directly, so there isn't confusion!

jhelwig (approved):

PR #6112, opened 2025-05-14
fix(rebaser): 4(+?) error-handling behavioral changes

fnichol:
fix(rebaser): copy & delete errored Rebaser request messages on error
-----------------------------------------------------------------------------------------------------

This change changes the error behavior of the `ChangeSetProcessorTask`
in the Rebaser--that is the task that consumes and processes Rebaser
requests for a change set. Prior to this change, when an early
`Result::Err` is returned from a message processing handler function,
this message was not deleted from the stream and therefore could be
re-processed multiple times.

With this change, when a handler returns with an error a new
`MoveMessageOnFailure` callback in the `PostProcessLayer` will make a
copy of the message into a `rebaser.errored.requests.$wk_id.$cs_id`
subject and then will delete the original message from the requests
stream.

There is little value in re-processing a request, especially if that
request successfully ran through the `perform_rebase()` function. As
before, an `error!` message will be displayed so that we can monitor any
aberrant behavior.

fix(rebaser): don't error out task when Pinga returns error from DVU
---------------------------------------------------------------------------------------------------

This change updates the behavior of the `SerialDvuTask` when an error is
returned from Pinga when executing a dependent values update (i.e. DVU)
job. Prior to this change any errors in setting up the job request to
Pinga and job results reported back would cause the `SerialDvuTask` to
return a `Result::Err`, causing the Rebaser change set task to initiate
a shutdown and reschedule of the task work.

This change treats the `SerialDvuTask` as a trigger point to run DVU
jobs and a coodinator, but it does not propagate the errors back from
Pinga and into the Rebaser's change set task. Now the `run_dvu()`
internal function does not return an error but instead will log an
`error!` when something may have gone wrong when triggering a DVU job.
The return of the `ctx.blocking_commit_no_rebase()` method is checked to
see if an execution error has occurred, and in that case it logs a
`warn!` message and continues on.

The principle here is that by prematurely ending the `SerialDvuTask`
when a Pinga error is encountered, the Rebaser is simply propagating
errors across services for circumstances out of its direct control.
These changes attempt to make the Rebaser less responsible for the
problems of others ;)

chore: add `dlq.*.> subjects` to `DEAD_LETTER_QUEUES` stream
-------------------------------------------------------------------------------------------

This change adds a Jetstream destination for service-specific errored
messages which may be full message copies, only metadata, or whatever
the service deems useful.

The first use case is for Rebaser requests which encounter an error
while processing. In this case the Rebaser will copy the full request
message into `dlq.rebaser.requests.$wk_id.$cs_id` for later inspection.

fix(rebaser): ensure Rebaser reply sent even if error occurs post-rebase
--------------------------------------------------------------------------------------------------------

This change attempts to prevent some of the "early error return" logic
in the `ChangeSetProcessorTask`'s handler function. Once the
`perform_rebase()` function has been called, and assuming it was
successful, 2 main activities are then executed:

1. signal the `SerialDvuTask` to dispatch and track a DVU job
2. check for actions on the graph which can be dispatched to run

Both of these activities involve the status of the graph, the behavior
of the layer cache dbs, database connections, etc. which can fail
post-sucessful rebase (we recently saw an example of this where the
state of the graph led to `Action::dispatch_actions` failing
consistently).

This change invokes these post-rebase activities and captures the return
`Result` without early returning on error. This allows the optional
rebase reply message to be sent (often times this is an SDF handler or
the end of a Pinga/DVU job). Prior to this change, any error post-rebase
that errored would short circuit and skip the remaining logic,
importantly including the rebase reply check. Skipping these replies
lead to errors in SDF such as:

```
"transactions error: rebaser reply deadline elapsed; waited=60s, request_id=01JV852XXCAM95AN2GMZ5CKJ7T"
```


fnichol:
I'm patching this branch with the following to simulate a bad graph in a certain change set:

```diff
diff --git a/lib/dal/src/job/definition/dependent_values_update.rs b/lib/dal/src/job/definition/dependent_values_update.rs
index a439ec415..18d775010 100644
--- a/lib/dal/src/job/definition/dependent_values_update.rs
+++ b/lib/dal/src/job/definition/dependent_values_update.rs
@@ -82,6 +82,8 @@ use crate::{
 pub enum DependentValueUpdateError {
     #[error("attribute value error: {0}")]
     AttributeValue(#[from] AttributeValueError),
+    #[error("FIXME: DVU BORKED ERROR: {0}")]
+    Borked(&'static str),
     #[error("change set error: {0}")]
     ChangeSet(#[from] ChangeSetError),
     #[error("component error: {0}")]
@@ -286,6 +288,25 @@ impl DependentValuesUpdate {
         &self,
         ctx: &mut DalContext,
     ) -> DependentValueUpdateResult<JobCompletionState> {
+        #[allow(clippy::disallowed_methods)]
+        let maybe_borked_change_set_id = std::env::var("SI_BORKED_CHANGE_SET_ID")
+            .map(|s| {
+                crate::ChangeSetId::from(
+                    ulid::Ulid::from_string(&s).expect("fnichol: ulid should parse"),
+                )
+            })
+            .ok();
+        if let Some(borked_change_set_id) = maybe_borked_change_set_id {
+            if borked_change_set_id == ctx.visibility().change_set_id {
+                // Let's simluate some work happening before failing
+                tokio::time::sleep(std::time::Duration::from_secs(2)).await;
+                // Fail the DVU job
+                return Err(DependentValueUpdateError::Borked(
+                    "im a bork this dvu--on purpose!",
+                ));
+            }
+        }
+
         let start = tokio::time::Instant::now();
         let span = Span::current();
         let roots = DependentValueRoot::take_dependent_values(ctx).await?;
diff --git a/lib/rebaser-server/src/change_set_processor_task.rs b/lib/rebaser-server/src/change_set_processor_task.rs
index 9d0056257..671becfbf 100644
--- a/lib/rebaser-server/src/change_set_processor_task.rs
+++ b/lib/rebaser-server/src/change_set_processor_task.rs
@@ -438,6 +438,8 @@ mod handlers {
         /// Failures related to Actions
         #[error("action error: {0}")]
         Action(#[from] ActionError),
+        #[error("FIXME: BORKED ERROR: {0}")]
+        Borked(&'static str),
         /// Failures related to ChangeSets
         #[error("Change set error: {0}")]
         ChangeSet(#[from] dal::ChangeSetError),
@@ -569,6 +571,15 @@ mod handlers {
         // Dispatch eligible actions if the change set is the default for the workspace.
         // Actions are **ONLY** ever dispatched from the default change set for a workspace.
         if let Some(workspace) = Workspace::get_by_pk_opt(ctx, workspace_id).await? {
+            #[allow(clippy::disallowed_methods)]
+            let maybe_borked_change_set_id = std::env::var("SI_BORKED_CHANGE_SET_ID")
+                .map(|s| {
+                    dal::ChangeSetId::from(
+                        ulid::Ulid::from_string(&s).expect("fnichol: ulid should parse"),
+                    )
+                })
+                .ok();
+
             if workspace.default_change_set_id() == ctx.visibility().change_set_id {
                 let mut change_set =
                     ChangeSet::get_by_id(ctx, ctx.visibility().change_set_id).await?;
@@ -594,6 +605,14 @@ mod handlers {
                     // No need to send the request over to the rebaser as we are the rebaser.
                     ctx.commit_no_rebase().await?;
                 }
+            } else if let Some(borked_change_set_id) = maybe_borked_change_set_id {
+                if borked_change_set_id == ctx.visibility().change_set_id {
+                    // Simulate an error from dispatching actions or something else in the
+                    // above if check
+                    return Err(HandlerError::Borked(
+                        "this change set is borked--on purpose!",
+                    ));
+                }
             }
         }
```

Here's how you can use it:
- Check out the branch
- Apply the above patch
- Probably bring your stack down with: `buck2 run dev:down`
- Run the dev stack with: `buck2 run dev:up`
- Create a change set, add some content (doesn't matter what), apply it so that `HEAD` has some content
- Create a `borked` change set, or use whatever name you want
- Copy the change set id out of the address bar
- Bring down the dev stack, usually with `Ctrl+c`
- Run the stack again, but this time with an environment variable, setting up your "borked" change set graph with: `env SI_BORKED_CHANGE_SET_ID=01JV8DW0K2C6ZY7ECTK3KPZ3J1 buck2 run dev:up` (sub in your change set id)
- Change over to your borked change set
- Attempt to add something to the graph
- Observe the errors in the logs for `pinga` in Tilt UI
  - You'll see an `error` message with: `pinga_server::handlers: job execution failed si.error.message=JobConsumer(DependentValueUpdate(Borked("im a bork this dvu--on purpose!"))) job.invocation_id=01JV8ECPTHQGY7H4BHJCQ3PNK3 job.instance="01JV8EAQNP3NM76VHTP04RZV5D"`, showing the DVU has simulated an error (and didn't complete successfully)
- Observe the errors in the logs for `rebaser` in Tilt UI
  - You'll see an `error` message with: `rebaser_server::change_set_processor_task::handlers: failed to process message si.error.message=Borked("this change set is borked--on purpose!")`, showing a rebase request for this change set had simulated an error post-successfully performing a rebase
  - Another `error` message with: `rebaser_server::change_set_processor_task: error encoutered when processing message; moving to errored messages subject subject="rebaser.requests.01GWE92TVKXWVDN2F9TQKK3DHR.01JV8DW0K2C6ZY7ECTK3KPZ3J1" stream_sequence=12 subject="rebaser.requests.01GWE92TVKXWVDN2F9TQKK3DHR.01JV8DW0K2C6ZY7ECTK3KPZ3J1"`, showing that the message will be copied to `dlq.rebaser.requests.01GWE92TVKXWVDN2F9TQKK3DHR.01JV8DW0K2C6ZY7ECTK3KPZ3J1` in the `DEAD_LETTER_QUEUES` stream
  - A `warn` message with: `rebaser_server::serial_dvu_task: pinga job(s) reported an error during execution; continuing job_execution_error_report="Error during job execution: job consumer error: dependent value update error: FIXME: DVU BORKED ERROR: im a bork this dvu--on purpose!"`, showing that the Rebaser received the error from the Pinga job, but it won't cause the `SerialDvuTask` to fail

lib/nats-dead-letter-queue/src/lib.rs
```diff
@@ -1,50 +1,114 @@
...
+    // `$JS.EVENT.ADVISORY.CONSUMER.MAX_DELIVERIES.<STREAM>.<CONSUMER>`
+    //
+    // See: https://docs.nats.io/running-a-nats-service/nats_admin/monitoring/monitoring_jetstream
+    "$JS.EVENT.ADVISORY.CONSUMER.MAX_DELIVERIES.*.*",
+    // Create a service-namespaced subject space for services to place full message copies if they
+    // need to clear their own streams while retaining these messages for later forensics.
+    //
+    // The expected subject prefix is of the form: `dlq.<SERVICE_NAME>...`, for example:
+    // `dlq.rebaser.requests.$wk_id.$cs_id`
+    "dlq.*.>",
```
britmyerss, line 33: I definitely have a list forming in my head of areas where we will want to use this, brilliant!!

lib/rebaser-server/src/change_set_processor_task.rs
```diff
@@ -164,7 +166,11 @@ impl ChangeSetProcessorTask {
                     )
                     .on_response(telemetry_nats::NatsOnResponse::new()),
             )
-            .layer(PostProcessLayer::new().on_success(DeleteMessageOnSuccess::new(stream)))
+            .layer(
+                PostProcessLayer::new()
+                    .on_success(DeleteMessageOnSuccess::new(stream.clone()))
+                    .on_failure(MoveMessageOnFailure::new(stream, dead_letter_queue)),
```
britmyerss, line 172: leaving notes for later: there IS a bit of coupling here between `pending_events` and `rebase_requests` in that we might want to also DLQ the pending events related to the failed rebase request

britmyerss (commented):
👏 👏 👏 

This is a great improvement! 

 I'll reserve the Approval button for someone else as I've been with you through this journey and it'll be great to get another set of eyes here. 

@nickgerace is probably the right person

lib/rebaser-server/src/change_set_processor_task.rs
fnichol: That could be a great idea as well, it's all down to whether it's useful to hold on to some of these messages, or as a way to trigger other monitoring as a signal that something fishy just happened.

nickgerace (approved):
Excellent work!! I love the DLQ publishing pattern.

PR #6114, opened 2025-05-14
chore(dal): Single si_db::Error type,use si_db::*  types from dal

jkeiser:
This is a sort of "cleanup PR" after the work to move DB types from dal into the si-db crate:

* Stops re-exporting si_db types from dal, make callers `use si_db::User` instead
  - This helps us tell easily whether we really depend on the dal (should make it easier to tell when we can wean a caller or endpoint off the DAL)
  - Having done this, I'm no longer sure this needs to be done for everything and might not do it in the future; but this one's done
* Moves all the si-db error types to a single si_db::Error type, to reduce future churn
  - except SiDbTransactionsError, which is still separate because of some weirdness around dal::TransactionsError that I haven't got around to thinking through yet

## Testing

- [ ] Integration tests pass


lib/dal/src/change_set.rs
```diff
@@ -1089,13 +1087,7 @@ impl ChangeSet {
...
     pub async fn extract_userid_from_context_or_error(ctx: &DalContext) -> ChangeSetResult<UserPk> {
         let user_id = match ctx.history_actor() {
-            HistoryActor::User(user_pk) => {
-                let maybe_user = User::get_by_pk(ctx, *user_pk).await;
-                match maybe_user {
-                    Ok(user) => user.pk(),
-                    Err(err) => return Err(ChangeSetError::User(err)),
-                }
-            }
+            HistoryActor::User(user_pk) => User::get_by_pk(ctx, *user_pk).await?.pk(),
```
nickgerace, line 1090: Nice!

lib/si-db/src/context.rs
```diff
@@ -18,22 +16,7 @@ pub trait SiDbContext {
...
-
-pub trait SiDbTransactions {
-    fn pg(&self) -> &PgTxn;
-    fn nats(&self) -> &NatsTxn;
-}
-
-// TODO TransactionsError really needs to be moved to a common place accessible here and in dal
-#[remain::sorted]
-#[derive(Debug, Error, strum::EnumDiscriminants)]
-pub enum BaseTransactionsError {
```
nickgerace, line 34: Nice!

lib/si-db/src/lib.rs
```diff
@@ -1,20 +1,59 @@
...
+
 mod embedded {
     use refinery::embed_migrations;
 
     embed_migrations!("./src/migrations");
 }
+
+#[remain::sorted]
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
```
nickgerace, line 42: Yay one error!

lib/si-db/src/transactions.rs
```diff
@@ -0,0 +1,21 @@
...
+
+pub trait SiDbTransactions {
+    fn pg(&self) -> &PgTxn;
+    fn nats(&self) -> &NatsTxn;
+}
+
+// TODO TransactionsError really needs to be moved to a common place accessible here and in dal
+#[remain::sorted]
+#[derive(Debug, Error, strum::EnumDiscriminants)]
+pub enum SiDbTransactionsError {
```
nickgerace, line 16: Why keep this alive?

jkeiser: Mainly because one thing at a time, honestly :) I think the right thing to do in the end is to call this `TransactionsError` and remove dal::TransactionsError, because most of the things returning that error aren't actually transactions ... but that will be better done by itself.

nickgerace (approved):
Amazing work!! One error type is the way to go too.

lib/si-db/src/transactions.rs
nickgerace: Sounds great to me!

PR #6115, opened 2025-05-14
feat(buck2): remote build cache

sprutton1:
This PR includes settings for enabling remote build cache options in buck2. Note that remote build execution is another topic and will be addressed with different work.


component/bb-storage/config.yml
nickgerace, line 1: dear god

nickgerace (dismissed):
This is amazing

PR #6118, opened 2025-05-14
feat(auth-stack): Ensure all members of a workspace can create a token

stack72:
A workspace token is more like a PAT so we are going to allow all users of a workspace to be able to create a token. This means that individual users of a workspace can create a workspace token and use it with the public API for that workspace


vbustamante (approved):
Looks good to me

PR #6119, opened 2025-05-14
chore(dal): simplify APA creation

jkeiser:
Breaking out this factoring from a feature PR I'm working on, so it and the original can be reviewed easily and quickly. This PR simplifies the creation of AttributePrototypeArguments in two ways:

1. Adds the ValueSource directly as an argument to new(), since in 10 out of 11 places, we set the ValueSource immediately after .
   - The existing source-less function is now new_without_source(), and is called in two places in si-fs.
2. Makes set_value_source() take impl Into<ValueSource>, and removing the named variants like set_value_source_from_input_socket() (so you can just call `set_value_source(<input socket id>)`)
   - Also uses derive_more::From to derive the ValueSource::from() impls, since the variants are (currently) fully discriminated by type

## Testing

- [x] Integration tests pass
- [x] Inspection of code to ensure we're doing the same thing as before

### DIFF
--- a/lib/dal/src/attribute/prototype.rs
+++ b/lib/dal/src/attribute/prototype.rs
@@ -96,6 +96,8 @@ pub enum AttributePrototypeError {
     ChangeSet(#[from] ChangeSetError),
     #[error("func error: {0}")]
     Func(#[from] FuncError),
+    #[error("func error: {0}")]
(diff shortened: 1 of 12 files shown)


nickgerace (approved):
Yep! Kill unneeded functions that can cause footguns. Love it.

nickgerace (approved):

PR #6123, opened 2025-05-15
chore: flake update

sprutton1:
Specifically to get the latest buck2. Needed to pull in a later nodejs as 18 is deprecated. I will confirm this locally to ensure we have no breakages.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -20,11 +20,11 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1744536153,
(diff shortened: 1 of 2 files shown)


adamhjk (approved):
long live node 20! :)

lgtm

fnichol (approved):
Assuming everything builds (primarily `app/web`), then I'd expect we're good.

The only catch is that nothing in our current merge queue pipeline builds `app/web` so it'll be hard to tell without checking seperately

PR #6126, opened 2025-05-15
feat(auth-stack): Add the ability to opt-in / out of approvals

stack72:
This is currently flagged off but is a requirement to get deployed BEFORE we release this to the web app. The TL;DR is, that by default, a workspace will be opt-ed out of approvals. When I turn the feature flag on, we can enable those approvals

Screenshot 2025-05-15 at 20 00 07

Screenshot 2025-05-15 at 19 59 58


bin/auth-api/prisma/migrations/migration_lock.toml
```diff
@@ -1,3 +1,3 @@
 # Please do not edit this file manually
-# It should be added in your version-control system (e.g., Git)
+# It should be added in your version-control system (i.e. Git)
```
vbustamante, line 2: Although the comma should indeed be removed, e.g. is the right term here since we're listing  examples. i.e. is used for clarifications

This is a side note though, no real need to change it 😄

vbustamante (approved):
looks good

vbustamante (approved):

PR #6128, opened 2025-05-15
Add mjolnir smoke test for tools tests

nickgerace:
## Description

This change adds a mjolnir smoke test for tools crons and general testing. This will be in the prod tests too.

Why use diagram calls for the test if it is likely going away in favor of the new map view? The test that was copied (the "2" API test) uses the same routes.

The README for the sdf API tests has been adjusted too for present day usage.

### DIFF
--- a/bin/si-sdf-api-test/tests/8-check_mjolnir.ts
+++ b/bin/si-sdf-api-test/tests/8-check_mjolnir.ts
@@ -1,0 +1,187 @@
+import assert from "node:assert";
+import { SdfApiClient } from "../sdf_api_client.ts";
+import { runWithTemporaryChangeset } from "../test_helpers.ts";
+
+// ==========================
+// Tunables for testing here!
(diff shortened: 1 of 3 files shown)


bin/si-sdf-api-test/tests/8-check_mjolnir.ts
```diff
@@ -0,0 +1,187 @@
...
+      if (maybeComponentName) {
+        break;
+      }
+    } else if (response?.status === 404) {
+      // Retry on 404
+      console.log(`Received 404 error (retrying): ${await response.text()}`);
+    } else {
+      // Fail on non-200 and non-404 errors
+      throw new Error(`Error ${response.status}: ${await response.text()}`);
+    }
```
nickgerace, line 150: This is a complex scenario but based on the result of the index, we want varying levels of logging and error handling. This likely changes as we harden this layer.

stack72 (approved):

PR #6129, opened 2025-05-16
chore: clover schema update

systeminitbot:


aaron-dernley:
Not sure about all the removed stuff on the diff for AWS::DataSync::StorageSystem @stack72 - what do you think?

PR #6144, opened 2025-05-17
Component Props Array & Map. Stabilize reference fetching

jobelenus:
First commit: Array and Map "add/remove" is now working! PSA for @vbustamante and @jkeiser when I delete one element of an array, they all get deleted, seems to be an API bug?
Second commit: Big stabilization of web worker reference checking and returning

### DIFF
--- a/app/web/src/main.ts
+++ b/app/web/src/main.ts
@@ -132,6 +132,10 @@ app.config.errorHandler = (
       (mappedStack) => {
         const stack = mappedStack.join("\n");
         observeError(err.message.toString(), stack, componentPath);
+        if (import.meta.env.VITE_SI_ENV === "local") {
+          // eslint-disable-next-line no-console
+          console.error(err);
+        }
       },
       { cacheGlobally: true },
     );
--- a/app/web/src/newhotness/AttributePanel.vue
+++ b/app/web/src/newhotness/AttributePanel.vue
@@ -11,8 +11,10 @@
       <ComponentAttribute
         v-for="child in filtered.tree.children"
         :key="child.id"
+        :component="component"
         :attributeTree="child"
         @save="save"
+        @delete="remove"
       />
     </ul>
     <div v-else>Oh no, no attributes!</div>
@@ -20,10 +22,8 @@
 </template>
 
 <script lang="ts" setup>
-import { useQuery } from "@tanstack/vue-query";
 import { computed, reactive, ref, watch } from "vue";
 import { Fzf } from "fzf";
-import { bifrost, useMakeArgs, useMakeKey } from "@/store/realtime/heimdall";
 import {
   AttributeTree,
   BifrostComponent,
@@ -43,43 +43,36 @@ const props = defineProps<{
   component: BifrostComponent;
 }>();
 
-const componentId = computed(() => props.component.id);
-
-const attributeTreeMakeKey = useMakeKey();
-const attributeTreeMakeArgs = useMakeArgs();
-const attributeTreeQuery = useQuery<AttributeTree | null>({
-  queryKey: attributeTreeMakeKey("AttributeTree", componentId),
-  queryFn: async () => {
-    const args = attributeTreeMakeArgs("AttributeTree", componentId.value);
-    return await bifrost<AttributeTree>(args);
-  },
-});
-
 export interface AttrTree {
   id: string;
   children: AttrTree[];
   parent?: string;
   prop?: Prop;
   attributeValue: AttributeValue;
+  isBuildable: boolean; // is my parent an array or map?
 }
 
 const makeAvTree = (
   data: AttributeTree,
   avId: string,
+  isBuildable: boolean,
   parent?: string,
 ): AttrTree => {
   // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-  const childrenIds = data.treeInfo[avId]!.children;
-  const children = childrenIds.map((id) => makeAvTree(data, id, avId));
-  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
   const av = data.attributeValues[avId]!;
   const prop = av.propId ? data.props[av.propId] : undefined;
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+  const childrenIds = data.treeInfo[avId]!.children;
+  const children = childrenIds.map((id) =>
+    makeAvTree(data, id, ["array", "map"].includes(prop?.kind ?? ""), avId),
+  );
   const tree: AttrTree = {
     id: avId,
     children,
     parent,
     attributeValue: av,
     prop,
+    isBuildable,
   };
   return tree;
 };
@@ -89,8 +82,9 @@ const root = computed<AttrTree>(() => {
     id: "",
     children: [] as AttrTree[],
     attributeValue: {} as AttributeValue,
+    isBuildable: false,
   };
-  const raw = attributeTreeQuery.data.value;
+  const raw = props.component.attributeTree;
   if (!raw) return empty;
 
   // find the root node in the tree, the only one with parent null
@@ -102,7 +96,7 @@ const root = computed<AttrTree>(() => {
   });
   if (!rootId) return empty;
 
-  const tree = makeAvTree(raw, rootId);
+  const tree = makeAvTree(raw, rootId, false);
   return tree;
 });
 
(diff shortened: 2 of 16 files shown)


app/web/src/main.ts
```diff
@@ -132,6 +132,10 @@ app.config.errorHandler = (
       (mappedStack) => {
         const stack = mappedStack.join("\n");
         observeError(err.message.toString(), stack, componentPath);
+        if (import.meta.env.VITE_SI_ENV === "local") {
```
britmyerss, line 135: what's this for?

app/web/src/newhotness/AttributePanel.vue
```diff
@@ -43,43 +43,36 @@ const props = defineProps<{
...
   parent?: string,
 ): AttrTree => {
-  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-  const childrenIds = data.treeInfo[avId]!.children;
-  const children = childrenIds.map((id) => makeAvTree(data, id, avId));
   // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
   const av = data.attributeValues[avId]!;
   const prop = av.propId ? data.props[av.propId] : undefined;
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+  const childrenIds = data.treeInfo[avId]!.children;
```
britmyerss, line 65: ok - seeing this, I haven't looked more closely, but I could smash the prop and av together in a single struct with only what we need here. I assume documentation, link, widget, etc, could live on the AV itself.  Let me know what you think!

app/web/src/newhotness/layout_components/ComponentAttribute.vue
```diff
@@ -4,64 +4,210 @@
...
     </template>
     <template v-else>
       <AttributeInput
         :displayName="displayName"
-        :attributeTree="props.attributeTree"
+        :attributeValueId="props.attributeTree.attributeValue.id"
+        :path="props.attributeTree.attributeValue.path ?? ''"
+        :kind="props.attributeTree.prop?.widgetKind"
+        :value="props.attributeTree.attributeValue.value?.toString() ?? ''"
+        :canDelete="props.attributeTree.isBuildable"
```
britmyerss, line 70: is this metadata about the av something you might want in the MV itself?

app/web/src/newhotness/layout_components/ComponentAttribute.vue
```diff
@@ -4,64 +4,210 @@
...
+import { useWatchedForm } from "../logic_composables/watched_form";
 
 const props = defineProps<{
+  component: BifrostComponent;
   attributeTree: AttrTree;
 }>();
 
-// this handles objects and arrays but not empty arrays
-const hasChildren = computed(() => props.attributeTree.children.length > 0);
+const hasChildren = computed(() => {
```
britmyerss, line 97: same thing here - this could be in the MV

app/web/src/newhotness/layout_components/ComponentAttribute.vue
```diff
@@ -4,64 +4,210 @@
...
+    props.attributeTree.prop?.path
+      .replace("root", "")
+      .replaceAll("\u000b", "/") ?? ""; // endpoint doesn't want it
+
+  // Do I send `{}` for array of map/object or "" for array of string?
+  // Answer by looking at my prop child
+  const propTree = props.component.schemaVariant.propTree;
+  const childProp =
+    propTree.props[
+      propTree.treeInfo[props.attributeTree.prop?.id ?? ""]?.children.pop() ??
```
britmyerss, line 143: do you actually want this to `pop()`?

app/web/src/newhotness/layout_components/ComponentAttribute.vue
```diff
@@ -4,64 +4,210 @@
...
+        ""
+    ];
+  if (childProp?.kind === "object") payload[`${path}/-`] = {};
+  else payload[`${path}/-`] = "";
+  bifrostingAdd.value = true;
+  await call.put<UpdateComponentAttributesArgs>(payload);
+  watch(
+    () => props.attributeTree.children.length,
+    () => {
+      // once the children count updates, we can stop spinning
```
britmyerss, line 153: is this where we'd ultimately want to leverage strategies for optimistic using tanstack so the state is saved? or is it already?

Side note, this might be a phenomenal place to try and ship some otel traces around this timing, because it's truly the full round trip from the user's perspective. How long is that spinner spinning?!

app/web/src/workers/webworker.ts
```diff
@@ -586,7 +590,9 @@ const handlePatchMessage = async (data: PatchBatch, span?: Span) => {
       };
       return atom;
     })
-    .filter((rawAtom): rawAtom is Required<Atom> => !!rawAtom.fromChecksum);
+    .filter((rawAtom): rawAtom is Required<Atom> => !!rawAtom.fromChecksum)
+    // turning this off for just a bit!
```
britmyerss, line 594: but why? :)

app/web/src/workers/webworker.ts
```diff
@@ -1291,21 +1375,32 @@ const get = async (
...
+      workspaceId,
+      changeSetId,
+      kind,
+      id,
+      followComputed,
+    );
+    // this is a choice, we could send through objects that don't match the types
+    // and potentially have something drawn on the screen—but that seems worse
+    // for the possible side-effects
+    if (hasReferenceError) return -1;
```
britmyerss, line 1390: would love to know more about how this error happens (or could happen) in the first place.

britmyerss (approved):
This is great work! I'll follow up next week to learn more about the stabilization and game out some scenarios together.

app/web/src/main.ts
jobelenus: This is the global error handler for Vue. If an exception is unhandled and bubbles up, it gets caught here. In prod, we dont want to console.error errors. However, its quite annoying to have an exception when developing that you don't see and the thing you're doing isn't working. So, locally, send errors to the console.

app/web/src/newhotness/AttributePanel.vue
jobelenus: Nope, this is fine I think. FWIW the "Tree" you see here isn't the prop tree you made. And its not _exactly_ the av tree Jacob made either. I essentially took the tree info and populated it with the objects, and children objects, all the way down. Its useful to have it this way for rendering and fuzzy searching

app/web/src/newhotness/layout_components/ComponentAttribute.vue
jobelenus: nope you're totally right, there's only one child and i dont want to mutate

app/web/src/newhotness/layout_components/ComponentAttribute.vue
jobelenus: Yea.. this is a tricky one.

- Removal optimistic, easy, remove it from the dom... but keep a spinner somewhere(?)
- Add optimistic, more complicated. As we were discussing I wanted the prop tree so i could look and see what the structure is. When you're adding to an array of strings, easy just add the input box. But, when you're adding `BlockDeviceMappings` (array of complex objects)... its less simple.
  - But yes–it can be done by walking the prop tree, and creating the AV tree thats used in rendering.

I'll see if I can get timing in there in a reasonable way

app/web/src/workers/webworker.ts
jobelenus: heh, didn't mean to commit this. But, right now (even with jacob's changes) the MVs `ComponentList` and `SchemaVariantCategories` are not being deterministically sorted. So, on every page load, most of the time, I actually get a patch for the whole of either list, reordered. And I was trying to figure out some perf stuff and it was getting in my way...

app/web/src/workers/webworker.ts
jobelenus: In terms of the "model of the program", its pretty straightforward why this could happen. A bundle of patches are not guaranteed to land on the client. For example:
- player A makes a new schema variant
- player B missed that patch (networking blip packets never arrive, offline for a moment, etc)
- player A makes a component for that variant
- player B receives a Component patch with a reference to a variant it doesn't have, throw mjolnir to have it returned to you

Why it's happening _to me_, locally, I'm not entirely sure, just that it was. We can still dive into it, this code just handles the error cases—it doesn't remove them. Either I was doing something dumb, and I've fixed it, or the problem remains and we're handling it.

jobelenus: note that, even if player B in that scenario, was offline, we detect it, they come back online, and we cold start which would get them what they missed. Its still a race. Cold start doesn't block. They could be right on the page making the query with the miss while cold start is still happening...

app/web/src/newhotness/layout_components/ComponentAttribute.vue
jobelenus: eh... six of one, half a dozen of the other...

app/web/src/newhotness/AttributePanel.vue
britmyerss: yeah I know - I just mean we can return this exact shape in the MV itself.

app/web/src/workers/webworker.ts
britmyerss: ah ok - something to look into

app/web/src/main.ts
britmyerss: ah, great!

britmyerss (approved):

PR #6146, opened 2025-05-18
Caching possible connections for the fuzzy filter search

jobelenus:
# Possible Connections State Of the Union
- Right now, this only goes over the props within component attributeValue trees.
- It does not include components sockets
- Sockets are going the way of the dodo
- The new prop <-> prop connection API doesn't support sockets at all
- This work does not give us parity with "edges" on the old system
- The docker / butane relationship basically doesn't work in this world
- Right now, we don't yet have secrets, which means you cannot connect an AWS cred to any component that needs it

## How to Call
On attribute input focus use the key `PossibleConnections` and make the call for `getPossibleConnections` for the annotation type you have on your prop, and the direction (when looking to fill a prop, you're looking for output sockets and other props)

## Return Data
There are three pieces of the return:
- `exactMatches`: this is for "non-generic" types (e.g. string)... for example, if your prop takes `VPC ID` this will contain other `VPC ID`... if you're searching for a generic type, this will be empty.
  - right now all the prop types are generic—maybe this goes away if we don't want "stricter" types
- `typeMatches`: these are all the generic type matches (e.g. string = string)
- `nonMatches`: all the generic types that don't match (e.g. if you're looking for a string, you'll find object, map, boolean here)

You have the AV path, component name, prop name, and the value at that path. Entirely up to the UI as to how you want to display these options! And when you want to make prop <-> prop connections  you have the componentId and path for the API call

### DIFF
--- a/app/web/src/api/sdf/dal/schema.ts
+++ b/app/web/src/api/sdf/dal/schema.ts
@@ -27,20 +27,23 @@ export enum ComponentType {
   ConfigurationFrameUp = "configurationFrameUp",
   AggregationFrame = "aggregationFrame",
 }
+interface Socket {
+  name: string;
+  annotations: Array<{ tokens: [] }>;
+  arity: "many" | "one";
+}
 
 export type OutputSocketId = string;
 
-export interface OutputSocket {
+export interface OutputSocket extends Socket {
   id: OutputSocketId;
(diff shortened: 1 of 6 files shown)


britmyerss (approved):
Nice!!

PR #6162, opened 2025-05-20
Fix: send the right option value, and display the label of the chosen option

jobelenus:
Steps to repro:
1. go to an AWS Region component
2. open the `Diff` panel
3. change the region dropdown to a different value by picking from the list
4. see the full label value in the text box
5. see the value of region change in the diff panel (you may need to rebuild the index if patches aren't working)


britmyerss (approved):
Tested locally - looks good!

PR #6163, opened 2025-05-20
wip(edda): Various fixes to Edda and MVs

britmyerss:
This includes the following: 
1. Ensure that within an Action MV, the list of cross action dependencies is stably ordered
2. Return 404 on the other custom error kind I added (I just missed this yesterday)
3. (main fix!) After we loop through the ready to process MV kinds, we need to see if any of those kinds actually had build tasks spawned/queued, and if not, remove it from the dep graph. Otherwise, downstream MV Kinds will never become ready because we were only removing the kind from the graph when the build tasks finished, so if one was never spawned we'd wait forever

### DIFF
--- a/lib/dal-materialized-views/src/action_view_list.rs
+++ b/lib/dal-materialized-views/src/action_view_list.rs
@@ -48,7 +48,13 @@ pub async fn assemble(ctx: DalContext) -> super::Result<ActionViewListMv> {
             }
             None => (None, None),
         };
-
+        let mut my_dependencies = action_graph.get_all_dependencies(action_id);
+        my_dependencies.sort();
+        let mut dependent_on = action_graph.direct_dependencies_of(action_id);
+        dependent_on.sort();
+        let mut hold_status_influenced_by =
+            Action::get_hold_status_influenced_by(ctx, &action_graph, action_id).await?;
+        hold_status_influenced_by.sort();
         views.push(ActionView {
             id: action_id,
             prototype_id: prototype.id(),
@@ -61,17 +67,11 @@ pub async fn assemble(ctx: DalContext) -> super::Result<ActionViewListMv> {
             state: action.state(),
             func_run_id,
             originating_change_set_id: action.originating_changeset_id(),
-            my_dependencies: action_graph.get_all_dependencies(action_id),
-            dependent_on: action_graph.direct_dependencies_of(action_id),
-            hold_status_influenced_by: Action::get_hold_status_influenced_by(
-                ctx,
-                &action_graph,
-                action_id,
-            )
-            .await?,
+            my_dependencies,
+            dependent_on,
+            hold_status_influenced_by,
         })
     }
-
     Ok(ActionViewListMv {
         id: ctx.change_set_id(),
         actions: views,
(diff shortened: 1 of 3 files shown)


lib/dal-materialized-views/src/action_view_list.rs
```diff
@@ -48,7 +48,13 @@ pub async fn assemble(ctx: DalContext) -> super::Result<ActionViewListMv> {
             }
             None => (None, None),
         };
-
+        let mut my_dependencies = action_graph.get_all_dependencies(action_id);
```
britmyerss, line 51: `action_graph.get_all_dependencies` is not guaranteed to return the dependencies in the same order! same with below

lib/edda-server/src/change_set_processor_task/materialized_view.rs
```diff
@@ -429,8 +429,27 @@ async fn build_mv_inner(
             // as MV build tasks finish, and we also don't want to free up the things that depend on them until _all_
             // of the started tasks for that MV kind have finished. Once the build tasks have finished, we'll remove the
             // MV kind from the graph, which will let the downstream MVs start.
+
+            // We need to track which MV kinds actually started, as if an MV kind was ready, but no changes came in that
+            // the MV kind cares about, we should remove it from the graph now.
             for &running_mv_kind in &ready_to_start_mv_kinds {
-                mv_dependency_graph.cycle_on_self(running_mv_kind);
+                // if the task has been kicked off, or it's queued, let's cycle on self.
```
britmyerss, line 436: @jhelwig would love your thoughts on the fix

jhelwig: Looks like it should work. I have a nagging feeling that there is another way to handle this, but it's not immediately apparent, so I wouldn't block on that.

jhelwig (approved):

jhelwig (changes requested):
`info!` -> `debug!`

lib/edda-server/src/change_set_processor_task/materialized_view.rs
```diff
@@ -429,8 +429,27 @@ async fn build_mv_inner(
...
+                    || queued_mv_builds
+                        .iter()
+                        .any(|build| build.mv_kind == running_mv_kind)
+                {
+                    info!("cyle on self: {:?}", running_mv_kind);
+                    mv_dependency_graph.cycle_on_self(running_mv_kind);
+                }
+                // otherwise, no work to do here, let's remove it and free up the downstream MV kinds
+                else {
+                    info!("removing: {:?}", running_mv_kind);
```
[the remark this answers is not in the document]
britmyerss: for sure good catch! holdover from me testing. Patching now!

lib/edda-server/src/change_set_processor_task/materialized_view.rs
britmyerss: I had started thinking about building the graph of build tasks, then pulling them off from there so we can action any task across all build kinds as they become ready, but this was faster

jhelwig (approved):

PR #6166, opened 2025-05-20
Do not collect before functions when running intrinsic functions

jhelwig:
Intrinsic functions are never dispatched to veritech/cyclone which means that before functions will never run, as they are only run by veritech/cyclone in order to modify the environment that the function will run in.

While it is very unlikely that there will be intrinsic functions in all the places where `FuncRunner::before_funcs` is being called, there's much less of a chance for needlessly looking up the before funcs if the short-circuit logic is centralized there.


zacharyhamm (approved):

zacharyhamm:
Very straightforward

PR #6167, opened 2025-05-20
feat(dal,si-split-graph): view removal corrections

zacharyhamm:
Implements the correction that prevents view deletion if deletion would leave a component on another change set without a view.

Adds data about the kind of node represented by an external source edge and external target and adds that data to the updates to make porting corrections more straightforward.

Adds some quality of life methods to make writing corrections easier.

Also moves away from using subgraph indexes in the updates, using the root id of the subgraph instead, so that interleaved new subgraph updates between change sets will not cause us to add edges to the wrong subgraph.

### DIFF
--- a/lib/dal/src/workspace_snapshot/node_weight/geometry_node_weight/v1.rs
+++ b/lib/dal/src/workspace_snapshot/node_weight/geometry_node_weight/v1.rs
@@ -21,6 +21,7 @@ use super::{
     NodeWeightDiscriminants,
 };
 use crate::{
+    EdgeWeight,
     EdgeWeightKindDiscriminants,
     WorkspaceSnapshotGraphVCurrent,
     workspace_snapshot::{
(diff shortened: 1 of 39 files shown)


lib/dal/src/workspace_snapshot/node_weight/geometry_node_weight/v1.rs
```diff
@@ -268,3 +278,30 @@ impl CorrectExclusiveOutgoingEdge for GeometryNodeWeightV1 {
...
+    ) -> split_snapshot::corrections::CorrectTransformsResult<
+        Vec<si_split_graph::Update<NodeWeight, EdgeWeight, EdgeWeightKindDiscriminants>>,
+    > {
+        // find the new geometry node in the list of updates (or a matching external target)
+        //
+        // find the new diagram object (if any)
+        //
+        // find the edge from the view to the diagram object
+        //
+        // find all new edge updates to or from the diagram object (or its external target)
```
zacharyhamm, line 303: This is just stubbed out for implementing it in the future

britmyerss (approved):

PR #6169, opened 2025-05-20
feat(dal): Mirror connections between copy/pasted components

jkeiser:
Right now, when you copy/paste two components with subscriptions to one another (e.g. a VPC and a Subnet), the subscription is not mirrored the same way as connections: Subnet' is connected to VPC (the original). Socket connections are "mirrored" so that VPC' and Subnet' are connected; subscriptions should do the same!

Fixes ENG-3026.

## Testing

- [x] Existing tests pass
- [x] Integration tests for copy/paste subscriptions
- [x] Manual: copy/paste sockets still mirrors connections
- [x] Manual: copy/paste subscriptions mirrors subscriptions

## Factor Budget

- Removed AttributePrototypeArgument::set_value_from_xxx() methods, most of which are now unused.
- Reduced usage of AttributePrototypeArgument::set_value_source() as well in places where new() can be used instead.

giphy

### DIFF
--- a/lib/dal/src/attribute/path.rs
+++ b/lib/dal/src/attribute/path.rs
@@ -27,6 +27,14 @@ pub enum AttributePath {
     JsonPointer(String),
 }
 
+impl std::fmt::Display for AttributePath {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            AttributePath::JsonPointer(path) => write!(f, "{}", path),
+        }
+    }
+}
+
 impl AttributePath {
     pub fn from_json_pointer(path: impl Into<String>) -> Self {
         AttributePath::JsonPointer(path.into())
--- a/lib/dal/src/attribute/prototype/argument.rs
+++ b/lib/dal/src/attribute/prototype/argument.rs
@@ -19,10 +19,7 @@ use si_id::{
     AttributeValueId,
     ComponentId,
     FuncArgumentId,
-    InputSocketId,
     OutputSocketId,
-    PropId,
-    SecretId,
     StaticArgumentValueId,
 };
 use static_value::StaticArgumentValue;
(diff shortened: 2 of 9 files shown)


lib/dal/src/func/authoring/save.rs
```diff
@@ -411,27 +441,25 @@ pub(crate) async fn save_attr_func_proto_arguments(
             AttributePrototypeArgument::remove_or_no_op(ctx, arg.id).await?;
         }
 
-        let attribute_prototype_argument =
```
jkeiser, line 414: I don't believe this code is actually called right now, because this form of APA::new() already doesn't exist. Fixed the code anyway.

lib/dal/src/attribute/path.rs
```diff
@@ -27,6 +27,14 @@ pub enum AttributePath {
     JsonPointer(String),
 }
 
+impl std::fmt::Display for AttributePath {
```
jkeiser, line 30: Was useful for debugging.

lib/dal/src/attribute/prototype/argument.rs
```diff
@@ -234,7 +231,7 @@ impl AttributePrototypeArgument {
         value_source: impl Into<ValueSource>,
     ) -> AttributePrototypeArgumentResult<Self> {
         let argument = Self::new_without_source(ctx, prototype_id, arg_id).await?;
-        Self::set_value_source(ctx, argument.id, value_source).await?;
+        Self::attach_value_source(ctx, argument.id, value_source).await?;
```
jkeiser, line 234: Minor perf thing: in cases where we know there isn't a value source, it's not worth removing the old edges, so I made an internal attach_value_source that doesn't look for existing edges.

lib/dal/src/attribute/prototype/argument.rs
```diff
@@ -478,29 +475,39 @@ impl AttributePrototypeArgument {
...
-            workspace_snapshot
-                .remove_edge(
-                    apa_id,
-                    existing_value_source,
-                    EdgeWeightKindDiscriminants::PrototypeArgumentValue,
-                )
-                .await?;
-        }
+            .await?;
+        workspace_snapshot
```
jkeiser, line 493: There was an issue here where set_value_source() would not remove the old subscription.

PR #6174, opened 2025-05-21
feat(dal): Support chaining before funcs when using attribute value subscriptions

britmyerss:
This PR adapts gathering before functions to work with attribute value subscriptions.  

To test, I added a new test exclusive schema aptly named `double-dummy-secret` which is similar to the `dummy-secret` except that it also accepts an secret itself (from `dummy-secret`) to ensure we chain the before funcs appropriately.  

We didn't have coverage of chaining before funcs already, so this first tests the existing functionality that relies on socket connections, as well as ensures that adding an attribute subscription (prop to prop connection) for secrets works exactly as socket connections do. 

I pulled out some helper functions into a common place for this, so a bit more code churn than I'd like considering the actual change is only a few lines, but so it goes. 

<div><br/>via <a href="https://giphy.com/freeform/">Freeform</a> on <a href="https://giphy.com/gifs/freeform-halloween-spooky-halloweentown-9kwOxInrc1eSdIFHUJ">GIPHY</a></div>

### DIFF
--- a/lib/dal/src/func/runner.rs
+++ b/lib/dal/src/func/runner.rs
@@ -1640,6 +1640,7 @@ impl FuncRunner {
                 match AttributePrototypeArgument::value_source(ctx, attribute_prototype_argument_id)
                     .await?
                 {
+                    // I am not the secret defining component, so find where it comes from
                     ValueSource::InputSocket(input_socket_id) => {
                         let component_input_socket = ComponentInputSocket::get_by_ids_or_error(
                             ctx,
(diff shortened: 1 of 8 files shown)


lib/dal/src/func/runner.rs
```diff
@@ -1654,6 +1655,14 @@ impl FuncRunner {
                             work_queue.push_back(source_component_id);
                         }
                     }
+                    // I am not the secret defining component, but have a subscription, so let's find where it comes from
```
britmyerss, line 1658: Here's the most import bit!

jkeiser (approved):
Nice! Thank you for expanding the tests for connected secrets :)

stack72 (approved):

PR #6179, opened 2025-05-21
fix(sdk): Fixup TypeScript SDK to work for TypeScript and Deno

stack72:


bin/openapi-extractor/sdk-generators/generate-typescript-sdk.sh
```diff
@@ -18,6 +18,10 @@ generate_sdk() {
   mkdir -p "$OUTPUT_DIR"
   echo "Generating TypeScript SDK from $SPEC_PATH to $OUTPUT_DIR..."
 
+  # Download LICENSE file from System Initiative repo
+  echo "Downloading LICENSE file..."
+  curl -s "https://raw.githubusercontent.com/systeminit/si/main/LICENSE" > "$OUTPUT_DIR/LICENSE"
```
johnrwatson, line 23: You should exit here if that download fails, otherwise the license is missing in the resultant build/metadata

generated-sdks/typescript/package.json
```diff
@@ -1,33 +1,53 @@
 {
   "name": "system-initiative-api-client",
-  "version": "1.0.0",
-  "description": "OpenAPI client for system-initiative-api-client",
-  "author": "OpenAPI-Generator Contributors",
+  "version": "0.0.1",
+  "description": "TypeScript/JavaScript SDK for the System Initiative Public API",
+  "author": "System Initiative <technical-operations@systeminit.com>",
```
johnrwatson, line 5: Can we change these generally to support@

That tech ops email DL is meant to be just for tooling we need to support, rather than external libraries

johnrwatson (approved):

generated-sdks/typescript/package.json
stack72: We absolutely can for both! I will ship those changes

PR #6181, opened 2025-05-21
Chore/update release notes

johnrwatson:
Update changelog to be current & re-annotate all existing entries with the relevant tag e.g. `Added` / `Changed` etc.

<hr/>

<div></div>


stack72 (approved):
This is the lost work!! <3

PR #6187, opened 2025-05-22
Add compiling spike of compressing edda messages

nickgerace:
## Description

This change adds a compiling spike of compressing edda messages. It is not functional nor should be merged, but it is a compiling base for processing "compressed" NATS JetStream messages and distributing work to handlers accordingly.

## Details

It contains a "spike within a spike" to see if we can make naxum generic to non-NATS messages. Unfortunately, the juice is likely not worth the squeeze and it may be easier to 1) make a fake NATS message or modify the last processed one or 2) do not use a naxum server for the change set processor task and spawn MV builds via a task tracker.

The good news is that the core logic in "compressor.rs" is mostly close. The idea is that if you see a continguous stream of messages (i.e. the new "from" snapshot was the last "to" snapshot) that are all "update" messages, we should try to compress them. The compressed message contains de-duped changes (WHICH ARE OKAY BECAUSE THEY ARE NOT UPDATES) that can be used for MV builds accordingly.

There's excess code in this commit that we can likely trash. For instance, the API types additions in "edda-core" can likely get thrown out if we decide to not go through NATS because the request type is entirely in-memory.

My 0.02 cents from this spike: we have to decide what is easier and better for the system from two options. The options are 1) get naxum to work with non-NATS messages (or fake it) or 2) do not use naxum for the change set processor task and use traditional streams. Yes, this is written similarly to my message earlier about the "spike within a spike", but you can see that I was contending with this problem. The hacker in me leans towards cutting out naxum for the change set processor task and just double acking every message that was successfully compressed. I fear that the lift to make naxum generic is not worth it when we already have control over the stream.

At any rate, this change should at least help move the needle forward for this effort. Follow the "FIXME" comments in this change for more details.

### DIFF
--- a/lib/edda-core/src/api_types.rs
+++ b/lib/edda-core/src/api_types.rs
@@ -1,3 +1,5 @@
+pub mod compressor_rebuild_request;
+pub mod compressor_update_request;
 pub mod rebuild_request;
 pub mod update_request;
 
--- a/lib/edda-server/src/change_set_processor_task.rs
+++ b/lib/edda-server/src/change_set_processor_task.rs
@@ -42,7 +42,6 @@ use si_data_nats::{
     NatsClient,
     async_nats::jetstream::{
         self,
-        consumer::push,
     },
 };
 use si_events::{
@@ -60,7 +59,10 @@ use tokio_util::{
 };
 
 use self::app_state::AppState;
-use crate::ServerMetadata;
+use crate::{
+    ServerMetadata,
+    compressor::Compressor,
+};
 
 mod materialized_view;
 
@@ -91,7 +93,7 @@ impl ChangeSetProcessorTask {
         metadata: Arc<ServerMetadata>,
         nats: NatsClient,
         stream: jetstream::stream::Stream,
-        incoming: push::Ordered,
+        incoming: Compressor,
         frigg: FriggStore,
         workspace_id: WorkspacePk,
         change_set_id: ChangeSetId,
@@ -148,22 +150,6 @@ impl ChangeSetProcessorTask {
             .filter_map(|maybe_elapsed_item| maybe_elapsed_item.ok());
 
         let app = ServiceBuilder::new()
-            .layer(
-                MatchedSubjectLayer::new()
-                    .for_subject(EddaRequestsForSubject::with_prefix(prefix.as_deref())),
-            )
-            .layer(
-                TraceLayer::new()
-                    .make_span_with(
-                        telemetry_nats::NatsMakeSpan::builder(connection_metadata).build(),
-                    )
-                    .on_response(telemetry_nats::NatsOnResponse::new()),
-            )
-            .layer(
-                PostProcessLayer::new()
-                    .on_success(DeleteMessageOnSuccess::new(stream.clone()))
-                    .on_failure(DeleteMessageOnFailure::new(stream)),
-            )
             .service(handlers::default.with_state(state))
             .map_response(Response::into_response);
 
@@ -354,7 +340,10 @@ mod handlers {
             Response,
         },
     };
-    use si_events::change_batch::ChangeBatchAddress;
+    use si_events::{
+        change_batch::ChangeBatchAddress,
+        workspace_snapshot::Change,
+    };
     use telemetry::prelude::*;
     use telemetry_utils::metric;
     use thiserror::Error;
@@ -421,16 +410,16 @@ mod handlers {
             ctx_builder,
             server_tracker: _,
         } = state;
-        let ctx = ctx_builder
-            .build_for_change_set_as_system(workspace_id, change_set_id, None)
-            .await?;
-
         let span = current_span_for_instrument_at!("info");
 
         if !span.is_disabled() {
             span.record("si.workspace.id", workspace_id.to_string());
             span.record("si.change_set.id", change_set_id.to_string());
         }
+
+        let ctx = ctx_builder
+            .build_for_change_set_as_system(workspace_id, change_set_id, None)
+            .await?;
 
         handle_request(&ctx, &frigg, workspace_id, change_set_id, request).await
     }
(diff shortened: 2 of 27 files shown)


lib/naxum/src/generic/handler.rs
```diff
@@ -0,0 +1,235 @@
...
+
+//                     let $last = match $last::from_message(req, state).await {
+//                         Ok(value) => value,
+//                         Err(rejection) => return rejection.into_response(),
+//                     };
+
+//                     let res = self($($ty,)* $last,).await;
+
+//                     res.into_response()
+//                 })
```
nickgerace, line 117: This is where the pain is when you lose the `Message<R>` where `R: MessageHead` goodies.

lib/naxum/src/generic.rs
```diff
@@ -0,0 +1,5 @@
+// FIXME(nick): here is an experiment in handling generic messages that only require "Send" and not "Message" or "MessageHead".
+// Lot of pain.
+mod handler;
+mod serve;
+mod service_ext;
```
nickgerace, line 5: "Spike within a spike" from the PR description

lib/naxum/src/extract.rs
```diff
@@ -47,6 +54,16 @@ where
...
 
+#[async_trait]
+pub trait FromMessageRaw<R, M = private::ViaMessage>: Sized
+where
+    R: MessageHead,
+{
+    type Rejection: IntoResponse;
+
+    fn from_message_raw(req: Message<R>) -> Result<Self, Self::Rejection>;
+}
```
nickgerace, line 65: Needed for negotiating API types in the compressor

lib/edda-server/src/change_set_processor_task.rs
```diff
@@ -148,22 +150,6 @@ impl ChangeSetProcessorTask {
...
-                    .make_span_with(
-                        telemetry_nats::NatsMakeSpan::builder(connection_metadata).build(),
-                    )
-                    .on_response(telemetry_nats::NatsOnResponse::new()),
-            )
-            .layer(
-                PostProcessLayer::new()
-                    .on_success(DeleteMessageOnSuccess::new(stream.clone()))
-                    .on_failure(DeleteMessageOnFailure::new(stream)),
-            )
```
nickgerace, line 166: If you compress the message and double ack it, none of the NATS stuff is 1) necessary or 2) will work... telemetry is an unknown here as well.

lib/edda-core/src/api_types.rs
```diff
@@ -1,3 +1,5 @@
+pub mod compressor_rebuild_request;
+pub mod compressor_update_request;
```
nickgerace, line 2: This will probably be deleted in the end.

lib/edda-server/src/compressor.rs
```diff
@@ -0,0 +1,422 @@
...
+            .await
+            .map_err(CompressorError::ConsumerCreate)?
+            .messages()
+            .await
+            .map_err(CompressorError::Subscribe)?;
+
+        let ctx = ctx_builder
+            .build_for_change_set_as_system(workspace_id, change_set_id, None)
+            .await
+            .map_err(CompressorError::DalContextBuild)?;
```
nickgerace, line 121: This is really only used for layer db. In theory we should also be using it for checking if the snapshot has moved, but the "is this contiguous" request compression logic should be fine. I think if we don't need a full context, we should consider just giving the compressor its own layer db client.

lib/edda-server/src/compressor.rs
```diff
@@ -0,0 +1,422 @@
...
+        // sets happening first though.
+
+        loop {
+            match project.incoming.poll_next_unpin(cx) {
+                Poll::Ready(Some(result)) => {
+                    // First, get the message off the stream and validate it.
+                    let message = match result {
+                        Ok(message) => message,
+                        Err(err) => {
+                            error!(si.error.message = ?err, "error getting info for jetstream message");
```
nickgerace, line 160: I didn't move errors to the handler because there's a chance naxum isn't the right move after compression. I'm leaning more towards not using naxum for the compressed messages for a few reasons but this is one of them.

lib/edda-server/src/compressor.rs
```diff
@@ -0,0 +1,422 @@
...
+                    let cloned_message = message.clone();
+
+                    // Determine if the stream has been drained.
+                    let drained = match message.info() {
+                        Ok(info) => info.pending == 0,
+                        Err(err) => {
+                            error!(si.error.message = ?err, "error getting info for jetstream message");
+                            continue;
+                        }
+                    };
```
nickgerace, line 175: Subtle, but this is a trick from @fnichol's recommendation to check the stream info's "pending" count to tell if more things are in the pipe because the `Stream` `impl` may not be sufficient on its own. We could be using this more, but this is a start.

lib/edda-server/src/compressor.rs
```diff
@@ -0,0 +1,422 @@
...
+                                // and "R" must implement "MessageHead". I don't think it is worth pursuing that path. Frankly, I am
+                                // not certain naxum is even the right tool for this. We should probably just double ack every message
+                                // here and switch to more primitive Rust stream management semantics.
+                                // ```
+                                // message.payload =
+                                //     Bytes::from(serde_json::to_value(new_message)?);
+                                // ```
+
+                                // Return the compressed message. This is the critical path!
+                                return Poll::Ready(Some(Ok(cloned_message)));
```
nickgerace, line 227: Pain unending. This is where you would return your compressed message.

lib/edda-server/src/compressor.rs
```diff
@@ -0,0 +1,422 @@
...
+                            }
+                            // Everything is not drained so let's update the state and LOOP again!
+                            // This is the critical path!
+                            Poll::Ready(Ok(state)) => {
+                                // FIXME(nick): double ack the message since it has been compressed.
+                                *project.state = state;
+                                break 'inner;
+                            }
```
nickgerace, line 235: Compression done! Loop and poll again.

lib/edda-server/src/compressor.rs
```diff
@@ -0,0 +1,422 @@
...
+                // FIXME(nick): handle shutdown and publishing what we have compressed in flight on the way out.
+                Poll::Ready(None) => todo!(),
+                // FIXME(nick): handle publishing what we have compressed in flight before returning pending.
+                Poll::Pending => todo!(),
+            }
+        }
+    }
+}
+
+async fn compress(ctx: &DalContext, request: EddaRequestKind, state: State) -> Result<State> {
```
nickgerace, line 254: The core business logic.

If the function body looks a bit strange, it is because `state` was originally passed in as a `&mut`. That was great for directly mutating the projection from `pin-project-lite` but WOW did things get confusing and likely error prone. Lot of `(*my_vec).push(*my_id)` code.

lib/edda-server/src/compressor.rs
```diff
@@ -0,0 +1,422 @@
...
+                let mut previous_ids = previous_ids;
+                previous_ids.push(id);
+                let id = update_request.id;
+                Ok(State::Draining(CompressorUpdateRequestVCurrent {
+                    id,
+                    previous_ids,
+                    from_snapshot_address,
+                    to_snapshot_address,
+                    changes,
+                }))
```
nickgerace, line 291: What successful compression looks like without a full rebuild

PR #6196, opened 2025-05-23
feat: add a Pinga client to unlock direct dispatches of jobs

fnichol:
This change adds a new, explicit client library to interact with a Pinga
server. It introduces updated request and response data types that are
versioned and upgradable using the same `lib/naxum-api-types` framework
used in the Edda and Rebaser clients. The Pinga-submitting code that
exists within the `lib/dal` library has been updated to use this client.

While some of the `lib/dal/src/job/*` code remains, it's likely that
over time more of this behavior will migrate into `lib/pinga-client`,
`lib/pinga-server` and other external libraries.

Among some of the other notable changes:

- There are 3 kinds of jobs: actions, dependent value updates, and
  validations
- An Enum covers all 3 possible job types and their concrete arguments
  that are required for job execution, replacing the `kind` String and
  `arg` `serde_json::Value` types that were in the now-removed
  `JobInfo`.
- All id types use their concrete types and raw ULIDs can now be avoided
- The old `JobQueue` has been re-written to acknowledge that there are
  only 3 concrete job types and to remove the shifting nature of jobs
  over time (i.e. a lot of backwards compat code gone).
- The `NatsProcessor` is now exclusively driven by a `PingaClient` (and
  so is now likely poorly named).
- All request and response types for Pinga are frozen in versioned tests
  which should fail if a future refactoring changes a message's
  contract.


lib/pinga-core/src/job/consumer.rs
```diff
@@ -0,0 +1,10 @@
+use si_id::{
```
zacharyhamm, line 1: It looks like this is dead code because of the above file in pinga-core?

fnichol: Correct, fixed

britmyerss (approved):

PR #6214, opened 2025-05-26
Make all the actions API calls v2, no more stores

jobelenus:


app/web/src/newhotness/api_composables/index.ts
```diff
@@ -20,13 +20,22 @@ export enum routes {
...
+  ActionRetry = "ActionRetry",
 }
 
 /**
  * Once we implement the action API calls in here
  * Those routes would also exist in here
  */
-const CAN_MUTATE_ON_HEAD: readonly routes[] = [] as const;
+const CAN_MUTATE_ON_HEAD: readonly routes[] = [
+  routes.ActionAdd,
```
britmyerss, line 34: I don't think we want ActionAdd, just the others!

PR #6219, opened 2025-05-27
Fix si-fs project on macos

JakeGinnivan:
Added libfuse to fuser features
Switch EBADFD to EBADF for cross platform bad file descriptor Pass additional link args on mac so the rust binary can link the the macfuse symbols

```
nix:nix-shell-env) MacBookPro:si jake$ buck2 build //bin/si-fs
Starting new buck2 daemon...
Connected to new buck2 daemon.
Build ID: 0655c259-4946-42e3-b5fb-e813fe62a0f3
Network: Up: 0B  Down: 26MiB
Loading targets.   Remaining     0/25     
Analyzing targets. Remaining     0/740    
Executing actions. Remaining     0/3041   
Command: build.    Finished 905 local                                                                   
Time elapsed: 2:17.2s
BUILD SUCCEEDED
(nix:nix-shell-env) MacBookPro:si jake$ 
```

image
image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3173,6 +3173,7 @@ dependencies = [
  "memchr",
  "nix 0.29.0",
  "page_size",
+ "pkg-config",
  "smallvec",
  "zerocopy",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -179,7 +179,7 @@ fastrace = "0.7.4"
 flate2 = "1.0.35"
 foyer = { version = "0.14.1", features = ["tracing"] }
 fs4 = "0.12.0"
-fuser = { version = "0.15.1", default-features = false }
+fuser = { version = "0.15.1", default-features = false, features = ["libfuse"] }
 futures = "0.3.31"
 futures-lite = "2.5.0"
 glob = "0.3.1"
--- a/bin/si-fs/BUCK
+++ b/bin/si-fs/BUCK
@@ -16,6 +16,10 @@ rust_binary(
     ],
     srcs = glob(["src/**/*.rs"]),
     env = {"CARGO_BIN_NAME": "si-fs"},
+    rustc_flags = select({
+        "config//os:macos": ["-C", "link-arg=-L/usr/local/lib", "-C", "link-arg=-lfuse"],
+        "DEFAULT": [],
+    }),
 )
 
 nix_binary(
--- a/bin/si-fs/README.md
+++ b/bin/si-fs/README.md
@@ -12,7 +12,7 @@ A FUSE filesystem for editing System Initiative assets in your favorite editor.
 2. Mount the filesystem:
 
 ```bash
-buck2 run //bin/si-fs -- --foreground --workspace-id <WORKSPACE_ID> --endpoint <ENDPOINT> /mountpoint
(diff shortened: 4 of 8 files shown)


JakeGinnivan:
Will continue trying to figure out how to make these changes only apply on macos

JakeGinnivan:
@zacharyhamm feedback on the approaches welcome, not used fuse, nix or buck2 before so probably better ways, but looked for existing patterns to solve OS specific ways to solve in buck and seemed it was mostly done through fixups.

zacharyhamm:
@JakeGinnivan sorry for losing track of this one. I have tested it on linux and it builds, so we should be good to go. Can you rebase this with the latest main? Let me know and I'll test it out again on macOS and Linux.

JakeGinnivan:
@zacharyhamm I've rebased.

I've gone through it all again, I think the fixup was the wrong approach, I've removed. Here is some things i've noticed which having someone who is more familiar with the build system setup to validate.

## Cargo.lock

The conditional feature in third-party/rust/Cargo.toml is causing pkg-config to be added to the Cargo.lock file, not sure if this is going to cause an issue.

Also, I've had to duplicate the fuser version number, which will break if the conditional dep and the normal dep don't both get upgraded.

## Reindeer adding to buildscript_run

The above causes reindeer to add the additional platform targets to buildscript_run. But this isn't supported by `_cargo_buildscript_rule` which is called in `prelude/rust/cargo_buildscript.bzl`. I just remove it from the map, not sure if there is a better way.

## platform_fixup.

I removed this in some windows fixup files as that syntax has been removed in reindeer, I was using it previously, but I don't think that's the right approach because it doesn't automatically add the dependencies in, which causes issues downstream.

zacharyhamm:
> @zacharyhamm I've rebased.
> 
> I've gone through it all again, I think the fixup was the wrong approach, I've removed. Here is some things i've noticed which having someone who is more familiar with the build system setup to validate.
> 
> ## Cargo.lock
> The conditional feature in third-party/rust/Cargo.toml is causing pkg-config to be added to the Cargo.lock file, not sure if this is going to cause an issue.
> 
> Also, I've had to duplicate the fuser version number, which will break if the conditional dep and the normal dep don't both get upgraded.
> 
> ## Reindeer adding to buildscript_run
> The above causes reindeer to add the additional platform targets to buildscript_run. But this isn't supported by `_cargo_buildscript_rule` which is called in `prelude/rust/cargo_buildscript.bzl`. I just remove it from the map, not sure if there is a better way.
> 
> ## platform_fixup.
> I removed this in some windows fixup files as that syntax has been removed in reindeer, I was using it previously, but I don't think that's the right approach because it doesn't automatically add the dependencies in, which causes issues downstream.

Great! taking a look

stack72:
hey @JakeGinnivan - so we started work on a new tool - called `si-conduit` that will replace the abilities for si-fs - it is going to be more portable as it's written in deno and can be changed / extended easily

I am going to close this out - sorry we were so bad about responding to this :/ I will make sure this doesn't happen again in the future

JakeGinnivan:
@stack72 all good, was interesting learning the toolchain enough to try fix this. New tool sounds like a nice approach

PR #6232, opened 2025-05-28
Have the secret prop use the AttributeInput

jobelenus:
But the subscription isn't working correctly on the backend


stack72:
Closed via #6244

PR #6234, opened 2025-05-28
Define & verify MV shape checksums for MvIndex validity checking

jhelwig:
If there are new/removed/changed MVs, then an already built MvIndex should no longer be re-used, and we should treat the change set as though it does not currently have an MvIndex at all.

We now have a way for all MVs that derive `si_frontend_mv_types_macros::MV` to automatically register themselves using the `inventory` crate, so that we have an always-up-to-date list of all MV kinds that have been defined. This allows us to do things that would require maintaining a list of all MV kinds, without having to manually do so. For example: Making sure all cross-MV dependencies are accounted for, and making sure that a `MvIndex` we are returning knew about all of the MV kinds that exist currently when it was built.

The structs that derive `si_frontend_mv_types_macros::MV` now also provide a "definition checksum" that consists of making a `Checksum` from the stringified field names & types of the struct (in the order they appear in the code). This now lets us easily detect if the structure of an MV has changed to be able to invalidate the MvIndex more easily.

Because the `MvIndex` now includes a checksum of all of the definition checksums of all of the MVs that it could potentially include, this means that we can now detect if an `MvIndex` was built to include all of the current MVs, even if the workspace snapshot it is indexing would not have included instances of all of the possible MV kinds.

What this DOES NOT currently handle:

* Changes to the shape of types that the MV uses. Because the checksums are built from the field names and the text of the types used, it only detects direct changes to the struct that derives `si_frontend_mv_types_macros::MV`. We will want to eventually be able to detect this kind of "indirect" change, and the way forward to do that will probably be along the lines of a similar `inventory` registration system for types that derive `si_frontend_mv_types_macros::FrontendChecksum` as everything an MV uses _must_ also implement that. `ActionViewList` is an example of an MV that this caveat currently applies to. Changes to `ActionView` would not be detected as it does not derive `si_frontend_mv_types_macros::MV`, only `ActionViewList` does.

* Type aliases. If the field of an MV uses a type alias (`type Foo = Bar;`) and that type alias is changed (`type Foo = u8;`) this _WILL NOT_ be detected as Rust's procedural macros have no ability to do the kind of type inspection necessary to detect this. DO NOT USE TYPE ALIASES IN MVs UNLESS ABSOLUTELY NECESSARY.


lib/si-frontend-mv-types-rs/src/materialized_view.rs
```diff
@@ -1,9 +1,67 @@
...
+    let mut mv_items: Vec<_> = ::inventory::iter::<MaterializedViewInventoryItem>().collect();
+    mv_items.sort_by_key(|item| item.kind());
+
+    let mut hasher = ChecksumHasher::new();
+    for mv in mv_items {
+        hasher.update(mv.definition_checksum().as_bytes());
+    }
+
+    hasher.finalize()
+}
```
jhelwig, line 65: This could probably be memoized via `LazyStatic`, but doing the checksum calculation here should be really fast. Thinking about making that change when addressing the `ActionViewList`/`ActionView` limitation, if it's not handled by then, but I could probably be talked into doing the `LazyStatic` change in this PR without much convincing.

lib/frigg/src/lib.rs
```diff
@@ -402,7 +406,27 @@ impl FriggStore {
...
         };
         let index_pointer_value: IndexPointerValue =
-            serde_json::from_slice(bytes.as_ref()).map_err(Error::Deserialize)?;
+            match serde_json::from_slice(bytes.as_ref()).map_err(Error::Deserialize) {
+                Ok(value) => value,
+                Err(err) => {
+                    warn!(
+                        "Unable to deserialize index pointer value at {}: {}",
+                        index_pointer_key, err
+                    );
```
jhelwig, line 415: Might actually only want this at debug? 🤷🏻‍♂️

fnichol (dismissed):

fnichol (dismissed):

britmyerss (approved):

PR #6237, opened 2025-05-29
feat(edda): Optimize New Change Set Handling 

britmyerss:
The following changes are included: 
## Add a new `EddaRequestKind::NewChangeSet`

-   When Edda recieves this message, first look to see if there's an existing`MvIndex` for a different, active change set  in the workspace with the same snapshot address. 
- If there's an existing `MvIndex`, create a new `IndexPointer` to that`MvIndex` 
- If there isn't, or something goes wrong while looking, fall back to rebuilding from scratch 
- New message kind complete with tests, thanks Fletcher!
- TODO: This is only triggered via new SDF route called exclusively by new-hotness. It doesn't account for any other way of creating a change set aside from this route. Accepting this for now as in the other cases, Edda will fall back to full rebuild we see a missing index. 

## Add a new bifrost message `IndexUpdate`
-  Every time an index is updated or modified, we will send this message 
- This will be most helpful when we've either triggered a full rebuild, or created a new change set, and building the index that takes longer than we'd like to wait, we don't have to block in SDF, have a way to tell the front end that something is happening, using this message as the way to communicate the index has been modified 
- TODO: Actually use this message!

## Update Change Set Creation V2 Route
-  Remove the now unecessary index fetch attempt/full rebuild trigger and use our new request kind 
- If we aren't able to reuse an existing index, fall back to rebuild, and if it takes longer than 30 secs, return the newly created Change Set and a `202` to let the caller know that the change set was successfully created,   
- TODO: Implement the lobby, and on receiving the `IndexUpdate`, allow the user to proceed as normal 
- TODO: Rip out the other places where we're blocking waiting for a rebuild 

## MV Index Updates
- In order to support the reuse of MvIndicies across Change Sets, the `ChangeSetId` is no longer part of the `MvIndex`. 
- Instead we use the `snapshotAddress` as the id for the `FrontEndObject` (it needs an Id) and thread the `ChangeSetId` through when interacting with the indices via Frigg. 
- As we don't want to fetch and deserialize the whole MvIndex unnecessarily, the `IndexPointer` key for the Change Set now points at a new `IndexPointerValue` struct, which contains both the key to the actual MvIndex, as well as the snapshot address from whence the Index was generated. 
 -- Previously, the IndexPointer simply referred to the object key for the MvIndex. But this means we'd have to deserialize the full index to check the snapshotAddress each time. 
-  This allows us to easily interrogate the IndexPointers for active change sets to determine if one of the existing MvIndices can be reused for a newly created Change Set.
- Future improvements can likely be made to make this faster/more ergonomic

<div><br/>via <a href="https://giphy.com/gifs/dory-finding-just-keep-swimming-FMQzq0nYWgwFO">GIPHY</a></div>


lib/edda-server/src/change_set_processor_task/materialized_view.rs
```diff
@@ -138,6 +147,132 @@ pub enum MaterializedViewError {
...
+    let change_set_id = ctx.change_set_id();
+    span.record("si.workspace.id", workspace_id.to_string());
+    // first find a change set Id in my workspace that has the same snapshot address that I do
+    // do we care which change set we use? Assuming there's more than one, we could choose Head...
+    let change_sets_using_snapshot = ChangeSet::list_active(ctx).await?;
+    
+
+    for change_set in change_sets_using_snapshot {
+        // found a match, so let's retrieve that MvIndex and put the same object as ours
+        if let Some((pointer, _revision)) = frigg
```
britmyerss, line 180: here's where we find an existing index to use, and create a new IndexPointer for the change set

lib/edda-server/src/change_set_processor_task/materialized_view.rs
```diff
@@ -138,6 +147,132 @@ pub enum MaterializedViewError {
...
+                    .instrument(tracing::info_span!(
+                        "edda.change_set_processor_task.build_all_mv_for_change_set.initial_build"
+                    ))
+                    .await?
+            }
+        }
+        Err(err) => {
+            error!(si.error.message = ?err, "error copying existing index");
+            // we did not copy anything, so we must rebuild from scratch (no from_snapshot_address this time)
+            self::build_all_mv_for_change_set(ctx, frigg, None)
```
britmyerss, line 266: falls back to build the world

lib/edda-server/src/change_set_processor_task/materialized_view.rs
```diff
@@ -155,7 +290,7 @@ pub enum MaterializedViewError {
 pub async fn build_all_mv_for_change_set(
     ctx: &DalContext,
     frigg: &FriggStore,
-    from_snapshot_address: Option<WorkspaceSnapshotAddress>,
+    snapshot_from_address: Option<WorkspaceSnapshotAddress>,
```
britmyerss, line 293: most everything below this is just handling the MvIndex change, and firing the IndexUpdate message

lib/frigg/src/lib.rs
```diff
@@ -239,7 +243,7 @@ impl FriggStore {
         }
 
         let index_object_key = self.insert_object(workspace_id, object).await?;
-        let index_pointer_key = Self::index_key(workspace_id, &object.id);
+        let index_pointer_key = Self::index_key(workspace_id, change_set_id);
```
britmyerss, line 246: need to thread ChangeSetId in now, as it used to be the `FrontEndObject`'s id and now it is not

lib/sdf-server/src/service/v2/change_set/create.rs
britmyerss, line 1: Do we need to fetch the index if we think it's been built? We're not returning it here, so felt unnecessary...

lib/edda-server/src/change_set_processor_task/materialized_view.rs
```diff
@@ -138,6 +147,131 @@ pub enum MaterializedViewError {
...
+                    };
+                    DataCache::publish_index_update(ctx, index_update).await?;
+                    return Ok(true);
+                }
+            }
+        } else {
+            // try the next one
+            // no need error if this index was never built, as it could be a transient change set I suppose
+            continue;
+        }
```
jhelwig, line 217: How do you feel about something like this to try to avoid the nesting conditionals?

```suggestion
        let Some((pointer, _revision)) = frigg.get_index_pointer_value(workspace_id, change_set.id).await? else {
            // try the next one
            // no need error if this index was never built, as it could be a transient change set I suppose
            continue;
        }

        if pointer.snapshot_address == snapshot_address.to_string() {
            // found one, create a new index pointer to it!
            // first let's get the index, so we can tell subscribers about it
            let Some((index, _revision)) =
                frigg.get_index(workspace_id, change_set.id).await? else {
                continue;
            }

            let change_set_mv_id = change_set_id.to_string();
            frigg
                .insert_index_key_for_existing_index(
                    workspace_id,
                    &change_set_mv_id,
                    pointer,
                )
                .await?;
            span.record("si.from_change_set.id", change_set_mv_id);
            let meta = UpdateMeta {
                workspace_id,
                change_set_id: Some(change_set_id),
                snapshot_from_address: None,
                snapshot_to_address: Some(snapshot_address),
            };
            let index_update = IndexUpdate {
                meta,
                kind: INDEX_UPDATE_KIND,
                checksum: index.checksum,
            };
            DataCache::publish_index_update(ctx, index_update).await?;
            return Ok(true);
        }
```

lib/edda-server/src/change_set_processor_task/materialized_view.rs
```diff
@@ -138,6 +147,131 @@ pub enum MaterializedViewError {
...
+                        workspace_id,
+                        change_set_id: Some(change_set_id),
+                        snapshot_from_address: None,
+                        snapshot_to_address: Some(snapshot_address),
+                    };
+                    let index_update = IndexUpdate {
+                        meta,
+                        kind: INDEX_UPDATE_KIND,
+                        checksum: index.checksum,
+                    };
```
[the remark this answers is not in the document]
britmyerss: yeah I was thinking the frontend would hit the "get me the index for this change set" - I don't know that we want to be asking for specific indices.... should likely always be fetching latest? but given the Id is the snapshot address, you've got that in the meta too. So I think we can remove the checksum. On it!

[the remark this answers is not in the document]
britmyerss: @jhelwig one other thought - even if we remove the checksum, would we eventually want to validate the index here with the new logic before we use it?

jhelwig (approved):

PR #6239, opened 2025-05-29
chore: speed up new component search filtering

adamhjk:
This improves the perceived performance of the add component modal by debouncing the fuzzySearchString, so we don't run the expensive computation (and pay the tax on re-rendering) while the user is still typing.


app/web/src/newhotness/AddComponentModal.vue
wendybujalski, line 1: Ah yes - debouncing makes a big difference here!

wendybujalski (approved):

PR #6240, opened 2025-05-29
feat: fix markdown rendering in new hotness

adamhjk:
Adds the tailwind prose plugin, so that we can render our markdown in a way that's easily tweakable with tailwind. Adds the MarkdownRender component, which takes the source to render, wraps it in the prose declaration, and passes it off to VueMarkdown.

Future additions would customize the prose styling from within the markdownrender component itself - for now, it's left as-is.


adamhjk:
image

PR #6241, opened 2025-05-29
fix(web): Hide hidden props from the attributes panel inputs

stack72:
Before:
Screenshot 2025-05-29 at 22 30 31

After:
Screenshot 2025-05-29 at 22 30 13

### DIFF
--- a/app/web/src/newhotness/layout_components/ComponentAttribute.vue
+++ b/app/web/src/newhotness/layout_components/ComponentAttribute.vue
@@ -75,7 +75,7 @@
         </div>
       </AttributeChildLayout>
     </template>
-    <template v-else>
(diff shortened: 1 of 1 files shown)


app/web/src/newhotness/layout_components/ComponentAttribute.vue
```diff
@@ -77,6 +77,7 @@
     </template>
     <template v-else>
       <AttributeInput
+        v-if="!props.attributeTree.prop?.hidden"
```
wendybujalski, line 80: This conditional should move to merge with the `v-else` on line 78 in the `<template>` tag to be `<template v-else-if="!props.attributeTree.prop?.hidden"`

wendybujalski (changes requested):

PR #6242, opened 2025-05-29
feat(web) - newhotness better keyboard controls, UI improvements

wendybujalski:
* Back button keyboard shortcut - Esc
* Revamp the keyboard controls on the Explore page - open a component with just the keyboard by pressing Enter
* Back button bugginess with the history stack, don't use breadcrumbs for that button
* If the domain tree is empty, don't show it
* If a component has no resource, don't show the resource line at all, no red X ever

### DIFF
--- a/app/web/src/newhotness/Component.vue
+++ b/app/web/src/newhotness/Component.vue
@@ -5,9 +5,9 @@
       This component does not exist on this change set
     </h3>
     <VButton
+      v-tooltip="'Back (Esc)'"
       class="border-0 mr-2em"
       icon="arrow--left"
-      label="Back"
       size="sm"
       tone="shade"
       variant="ghost"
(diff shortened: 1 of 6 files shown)


app/web/src/newhotness/Component.vue
```diff
@@ -292,15 +291,11 @@ const editInPlaceRef = ref<typeof EditInPlace>();
...
-  const last = prevPage();
-  if (last) router.push(last);
-  else {
-    const params = router.currentRoute?.value.params ?? {};
-    router.push({
-      name: "new-hotness",
-      params,
-    });
-  }
+  const params = router.currentRoute?.value.params ?? {};
```
britmyerss, line 294: I think we need to rethink this - this will mean that if you're on the map view and click "back" you'll go back to the grid...

britmyerss (approved):

PR #6244, opened 2025-05-30
chore(edda,web): Remove Secret specific MVs and include necessary data alongside props/attributes

britmyerss:
The secrets subsystem originally was designed to manage and create secrets separately from components. The original Secret MVs were designed to support this, but given secrets will now map 1:1 with components, we no longer need any of these independent MVs or lists of secrets/secret definitions. 

This PR makes that change, along with a few front end fixes alongside: 
1. Components Details will now display all secret props (previously it was only showing one)
2. If the prop is the secret definition, show the form to create/replace the secret
3. If the prop is not secret defining, allow the user to create a prop to prop connection and use an existing secret. 
4. For Possible Connections, we don't want to show the `attributeValue.value` for secrets, as it's just the `SecretId`. Instead, grab the name of the secret. 
5. Add description to the Secret Form

Todos: 
1. Ability to remove a connection for a secret prop
2. We likely need a custom `AttributeInput.vue` component for interacting with secret props
3. Account for secrets correctly in the Possible Connections calculation (right now it'll show all string props as available, which is not what we want, but this can be improved outside of this PR)
4. Styling!
5. Show name + description for a secret (but not the encrypted values).

<div><br/>via <a href="https://giphy.com/thenextstep/">THE NEXT STEP</a> on <a href="https://giphy.com/gifs/thenextstep-the-next-step-tns-tns8-SAAkNwNdpOYgUP6GCM">GIPHY</a></div>


app/web/src/newhotness/AttributePanel.vue
```diff
@@ -135,50 +95,11 @@ export interface AttrTree {
...
   attributeValue: AttributeValue;
   isBuildable: boolean; // is my parent an array or map?
 }
 
-const makeArgs = useMakeArgs();
-const makeKey = useMakeKey();
-
-const secretFormData = ref<Record<string, string>>({});
-// NOTE: this is pretty tortured and will change
-const secretQuery = useQuery({
```
britmyerss, line 148: no more torture!

app/web/src/newhotness/layout_components/ComponentSecretAttribute.vue
```diff
@@ -0,0 +1,196 @@
+<template>
```
britmyerss, line 1: inspired by ComponentAttribute.vue. I started trying to shoehorn all of this in there, but this is better.

app/web/src/newhotness/AttributePanel.vue
```diff
@@ -36,59 +36,25 @@
...
+          clsx(
+            'p-xs border-l',
+            themeClasses('bg-neutral-200', 'bg-neutral-800'),
+          )
+        "
+      >
+        secrets
+      </h3>
+      <ul v-if="'children' in secrets" class="border-l">
+        <ComponentSecretAttribute
```
britmyerss, line 51: A component can have 0-n secrets, and only one of them _could_ be actually defined by the component. That logic is handled in ComponentSecretAttribute.vue

app/web/src/newhotness/layout_components/ComponentSecretAttribute.vue
```diff
@@ -0,0 +1,196 @@
...
+  <AttributeChildLayout>
+    <template #header>
+      <div class="flex flex-row items-center gap-2xs">
+        <div>{{ displayName }}</div>
+      </div>
+    </template>
+    <template v-if="props.attributeTree.prop?.isOriginSecret">
```
britmyerss, line 8: if the **prop** represents a secret that needs defined, create and use the form.

app/web/src/newhotness/layout_components/ComponentSecretAttribute.vue
```diff
@@ -0,0 +1,196 @@
...
+            loadingText="Saving Secret"
+            tone="action"
+            :disabled="!secretForm.state.canSubmit"
+            @click="() => secretForm.handleSubmit()"
+          />
+        </ul>
+      </div>
+    </template>
+    <template v-else>
+      <AttributeInput
```
britmyerss, line 49: otherwise you can set connections similar to other prop kinds!

app/web/src/workers/webworker.ts
```diff
@@ -1213,17 +1206,31 @@ const updateComputed = async (
...
-          value: av.value,
-          path: av.path,
-          name: prop.name,
-          componentId: component.id,
-          componentName: component.name,
-          schemaName: component.schemaName,
-          annotation: prop.kind,
-        };
-        conns[av.id] = conn;
+        if (av.secret) {
```
britmyerss, line 1209: if the attribute value is for a secret, we want to grab the secret name and use that as the value.

app/web/src/newhotness/layout_components/ComponentSecretAttribute.vue
wendybujalski: next turn is optimizing the UX of the form

app/web/src/workers/webworker.ts
wendybujalski: could've just put a conditional on that one line, but this works!

wendybujalski (approved):
👍

PR #6245, opened 2025-05-30
feat(sdf): socket -> prop migration report

jkeiser:
This PR implements an admin button to "migrate connections", which gives migrateability status of connections in a changeset:
* Finds all socket connections in your changeset
* Reports the equivalent prop subscription that it will be migrated to.
* Reports if there are any connections that we are unable to safely migrate.

The algorithm to find a corresponding prop is basically, find the prop associated with each socket (if there is exactly one), and if one side or the other has a non-identity transformation, use that as the transformation function instead of si:identity). 

### Unmigrateable Report

We try to be as safe as possible: we will not migrate a socket connection unless the corresponding prop connection has *exactly the same effects*. We refuse to migrate connections (and report them as unmigrateable) if:
- Multiple arity: if there are multiple connections to the socket, we can't replace it with a single prop subscription, so we report it as unmigrateable. In the future we will handle array props by creating array elements and adding subscriptions to each one. 
- Multiple bindings: multiple things can be bound to an input socket. If this happens, we can't replace it with a simple prop connection 
- Complex binding: if the source or destination binding consumes multiple things via multiple arguments, we can't replace it with a straightforward prop connection.
- Non-prop binding: if the source or destination is bound to anything other than a prop, we can't replace it with a prop connection.

## Out of Scope

The actual migration is straightforward and we'll get to it tomorrow; it's the rules that were hard, so getting this report is the first priority.

## Testing

This doesn't intersect any main codepaths; it is new functionality and does not write.

- [x] Regression tests pass
- [x] Report on RMutt and prod workspaces

### DIFF
--- a/lib/dal/src/workspace_snapshot.rs
+++ b/lib/dal/src/workspace_snapshot.rs
@@ -1424,4 +1424,25 @@ impl WorkspaceSnapshot {
         let mut inferred_connection_write_guard = self.inferred_connection_graph.write().await;
         *inferred_connection_write_guard = None;
     }
+
+    /// Validate the snapshot in the given DalContext
+    pub async fn validate(
+        &self,
+    ) -> WorkspaceSnapshotResult<Vec<(graph::validator::ValidationIssue, String)>> {
+        Ok(graph::validator::validate_graph_with_text(
+            &self.working_copy().await,
+        )?)
+    }
+
+    /// Get the connection migrations that are available for this snapshot.
+    pub async fn connection_migrations(
+        &self,
+    ) -> WorkspaceSnapshotResult<Vec<(graph::validator::connections::ConnectionMigration, String)>>
+    {
+        Ok(
+            graph::validator::connections::connection_migrations_with_text(
+                &self.working_copy().await,
+            ),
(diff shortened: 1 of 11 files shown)


lib/dal/src/workspace_snapshot.rs
```diff
@@ -1424,4 +1424,25 @@ impl WorkspaceSnapshot {
...
+    }
+
+    /// Get the connection migrations that are available for this snapshot.
+    pub async fn connection_migrations(
+        &self,
+    ) -> WorkspaceSnapshotResult<Vec<(graph::validator::connections::ConnectionMigration, String)>>
+    {
+        Ok(
+            graph::validator::connections::connection_migrations_with_text(
+                &self.working_copy().await,
```
zacharyhamm, line 1444: If you want these methods to take `&WorkspaceSnapshotGraphVCurrent` instead of the impl Deref you just pass in this as the argument:

`&*self.working_copy().await`

That is, take a reference to the deref

zacharyhamm (approved):
Looks like you need to run `lint:fix` on the web app

PR #6247, opened 2025-05-30
fix(web, dal): various fixes to listing func runs

britmyerss:
Few fixes: 
1. Need to use the computed change set Id so tanstack knows to refetch functions when switching change sets
2. Only show the Action icon if the function was an action
3. For some reason we weren't recording the component name when executing attribute functions. Now that we are, the component name can be displayed correctly going forward

### DIFF
--- a/app/web/src/newhotness/FuncRunCard.vue
+++ b/app/web/src/newhotness/FuncRunCard.vue
@@ -56,13 +56,12 @@
             {{ funcRun.functionKind }}
           </span>
 
-          <!-- FIXME why is component name empty? -->
-          <span v-if="funcRun.componentId">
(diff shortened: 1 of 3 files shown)


app/web/src/newhotness/FuncRunCard.vue
```diff
@@ -56,13 +56,12 @@
...
 
-          <!-- FIXME why is component name empty? -->
-          <span v-if="funcRun.componentId">
-            {{ funcRun.componentName || "Component" }}
+          <span v-if="funcRun.componentId && funcRun.componentName">
+            {{ funcRun.componentName }}
           </span>
 
           <span
-            v-if="funcRun.actionId || true"
```
stack72, line 65: 🤦

PR #6250, opened 2025-05-30
feat: CRUD transformation funcs

vbustamante:
- Adds new argument to allow for creation of transformation attribute funcs
- Don't allow to create normal bindings to transformation funcs
- Subscription API sets transformation funcs when passed a funcId
- Does not allow for exporting and importing of transformation funcs


lib/dal/src/attribute/value.rs
```diff
@@ -2162,8 +2164,14 @@ impl AttributeValue {
...
     ) -> AttributeValueResult<()> {
         let func_id = if let Some(id) = func_id {
+            let func = Func::get_by_id(ctx, id).await?;
+            if !func.is_transformation && !func.builtin {
+                return Err(AttributeValueError::SubscribingWithInvalidFunction);
+            }
+
             id
         } else {
+            // TODO(victor) remove this new ui comes around
```
jkeiser, line 2174: I think we can remove this now along with the associated functionality in the new UI (not in this PR, just saying I think rolling back "subscribe to an array with a scalar" as a function in the old UI).

lib/dal/src/func/authoring.rs
```diff
@@ -437,6 +447,12 @@ impl FuncAuthoringClient {
         let func = Func::get_by_id(ctx, func_id).await?;
         func.error_if_locked()?;
 
+        if func.is_transformation {
```
jkeiser, line 450: Ah, so this is the key requirement. Attribute functions would be fine, except we want a guarantee that they will never have new arguments added.

lib/sdf-server/src/service/v2/component/attributes.rs
```diff
@@ -198,7 +194,7 @@ async fn update_attributes(
                         component: source_component,
                         path: source_path,
                         keep_existing_subscriptions,
-                        func,
+                        func: func_ident,
```
jkeiser, line 197: What does func_ident mean? It's not guaranteed to be identity anymore, right?

jkeiser (approved):

PR #6252, opened 2025-05-30
feat(sdf): Add inferred connection migration to report

jkeiser:
This brings inferred connections into the connection migration reporter, and shows the report in the admin dashboard (it is not pretty at all, but it's there).

### DIFF
--- a/lib/dal/src/workspace_snapshot/graph/validator/connections.rs
+++ b/lib/dal/src/workspace_snapshot/graph/validator/connections.rs
@@ -44,8 +44,9 @@ use crate::workspace_snapshot::{
 
 pub fn connection_migrations_with_text(
     graph: &impl std::ops::Deref<Target = WorkspaceSnapshotGraphVCurrent>,
+    inferred_connections: impl IntoIterator<Item = SocketConnection>,
 ) -> Vec<(ConnectionMigration, String)> {
-    connection_migrations(graph)
+    connection_migrations(graph, inferred_connections)
         .into_iter()
         .map(|migration| {
             let text = format!("{}", WithGraph(graph.deref(), &migration));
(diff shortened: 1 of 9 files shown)


lib/dal/src/workspace_snapshot/graph/validator/connections.rs
```diff
@@ -56,14 +57,34 @@ pub fn connection_migrations_with_text(
 
 pub fn connection_migrations(
     graph: &impl std::ops::Deref<Target = WorkspaceSnapshotGraphVCurrent>,
+    inferred_connections: impl IntoIterator<Item = SocketConnection>,
 ) -> Vec<ConnectionMigration> {
     let graph = graph.deref();
-    let mut migrations = graph
+    let inferred_connection_migrations =
```
zacharyhamm, line 63: We may want to move this onto the "slow_rt" if it is going to be expensive on large graphs. (Otherwise it will tie up the main executor)

zacharyhamm (commented):
No panel shows for me on migrate. I do see this in the console:

image

Looking at the response, it is correctly reporting inferred and other connections

lib/dal/src/workspace_snapshot/graph/validator/connections.rs
jkeiser: Sure :) It mainly depends on the definition of expensive: this will go through all socket connections in the graph, but does a constant-time burst of graph traversal on each one to determine its migrateabiliity. I'd be very surprised if it took more than single-digit milliseconds for 1000 connections.

I don't mind at *all* throwing this on slow_rt, but I'm asking to get a gauge: what's the feeling on how long a piece of sync code should run before we'd like it off the reactor?

zacharyhamm (approved):

zacharyhamm (approved):

PR #6253, opened 2025-05-30
feat(web): Filter out non eligible props from PossibleConnections

stack72:
We filter out color, type + hidden props!


app/web/src/newhotness/AttributePanel.vue
```diff
@@ -36,59 +36,25 @@
...
-          </label>
-        </li>
+    <template v-if="secrets && secrets.children.length > 0">
+      <h3
+        :class="
+          clsx(
+            'p-xs border-l',
+            themeClasses('bg-neutral-200', 'bg-neutral-800'),
+          )
+        "
```
wendybujalski, line 46: excellent use of clsx here

app/web/src/newhotness/AttributePanel.vue
```diff
@@ -36,59 +36,25 @@
...
+      >
+        secrets
+      </h3>
+      <ul v-if="'children' in secrets" class="border-l">
+        <ComponentSecretAttribute
```
wendybujalski, line 51: so glad this is in a component now :)

PR #6254, opened 2025-05-30
feat(sdf): migrate socket connections

jkeiser:
This flips the socket connection migration endpoint to actually perform the migration. In this PR:

* Adds POST endpoint that does the migration: forces changeset, removes socket connections, creates equivalent prop connections and sends WsEvents, then returns a report of what was migrated and what was not
* Anything that *can* be migrated, *will* (even if there were some we couldn't)
* Only does this for migrateable connections where it is safe--where we didn't detect an issue
* Displays already-migrated inferred connections as "already has a value" so they don't continuously get re-migrated
* Adds a "Dry Run" button with a GET request that tells you what would have been migrated, without migrating
* Groups the displayed issues better and makes them a tiny bit easier on the eyes (still not great though)

## What's Left

This gets us around 80-90% of sockets on our prod workspace and one of the other bigger ones. To get the last 10-20%, next we will want to:

- handle array sockets with multiple connections by creating array elements (gets us to 90-95%)
- figure out what to do with IAM double-transforming connections (gets us to 99%)

### DIFF
--- a/app/web/src/components/AdminDashboard/MigrateConnections.vue
+++ b/app/web/src/components/AdminDashboard/MigrateConnections.vue
@@ -11,29 +11,46 @@
         </div>
       </template>
       <template #success>
-        <div class="flex flex-row gap-xs p-xs w-full">
-          <div>
-            <h2>Migrateable Connections: {{ migrateable.length }}</h2>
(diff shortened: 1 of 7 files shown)


vbustamante (approved):
It'd be good to have a simple migration test but doesn't need to be in this Pr, I know more changes are coming

PR #6255, opened 2025-05-30
feat(web) - newhotness redirect on change set abandon, docs empty state

wendybujalski:
Screenshot 2025-05-30 at 7 52 24 PM

### DIFF
--- a/app/web/index.html
+++ b/app/web/index.html
@@ -3,6 +3,7 @@
 
 <head>
   <meta charset="UTF-8" />
+  <meta name="darkreader-lock">
   <link rel="icon" href="/favicon.png" />
   <meta name="viewport" content="width=device-width, initial-scale=1.0" />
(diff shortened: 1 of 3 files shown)


app/web/index.html
```diff
@@ -3,6 +3,7 @@
 
 <head>
   <meta charset="UTF-8" />
+  <meta name="darkreader-lock">
```
wendybujalski, line 6: threw this in here to fix the issue caused by the [dark reader extension](https://darkreader.org/)

stack72:
@wendybujalski the changes to the redirect look good but I don't think we can merge this - the change to the docs is not the experience we hoped it would be - it's more buried than before :/

stack72 (changes requested):
This makes it more difficult to find docs on props

stack72 (approved):

PR #6258, opened 2025-06-01
fix(sdf,web):Decrease risk of race condition, handle removed nodes, and reactive func runs

britmyerss:
This doesn't solve it completely, but by returning the current index's snapshot address, we *slightly* reduce the risk that the front end has a newer snapshot than the patches it is receiving.  I tend to see `Ragnarok` a bit less with this, but more work is needed. 

When nodes are removed, we try to look up the previous object in Frigg to get the last checksum. This was causing patches with deletes to not land as the `fromChecksum` was incorrect.  

Func Run list is now reactive to Func Runs in the grid view!

<div><br/>via <a href="https://giphy.com/tennistv/">Tennis TV</a> on <a href="https://giphy.com/gifs/tennistv-tennis-rublev-andrey-rubleb-YYht2UTV41u1vxHTss">GIPHY</a></div>

### DIFF
--- a/lib/edda-server/src/change_set_processor_task/materialized_view.rs
+++ b/lib/edda-server/src/change_set_processor_task/materialized_view.rs
@@ -309,7 +309,7 @@ pub async fn build_all_mv_for_change_set(
     let meta = UpdateMeta {
         workspace_id,
         change_set_id: Some(change_set_id),
-        snapshot_from_address,
+        snapshot_from_address: snapshot_from_address.or(Some(snapshot_to_address)),
         snapshot_to_address: Some(snapshot_to_address),
     };
     let patch_batch = PatchBatch {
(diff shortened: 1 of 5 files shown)


lib/edda-server/src/change_set_processor_task/materialized_view.rs
```diff
@@ -689,7 +689,16 @@ where
     MaterializedViewError: From<E>,
     MaterializedViewError: From<<T as TryInto<FrontendObject>>::Error>,
 {
-    let mv_kind = mv_kind.to_string();
+    let mv_kind:String = mv_kind.to_string();
+     let (from_checksum, _previous_data) = if let Some(previous_version) = frigg
+            .get_current_object(ctx.workspace_pk()?, ctx.change_set_id(), &mv_kind, &mv_id)
```
jhelwig, line 694: Should be able to remove this from the new entity else case below, and re-use what is fetched here, since the only difference should be the comment.

lib/edda-server/src/change_set_processor_task/materialized_view.rs
```diff
@@ -689,7 +689,16 @@ where
     MaterializedViewError: From<E>,
     MaterializedViewError: From<<T as TryInto<FrontendObject>>::Error>,
 {
-    let mv_kind = mv_kind.to_string();
+    let mv_kind:String = mv_kind.to_string();
```
jhelwig, line 692: Shouldn't need the type annotation?

britmyerss:
Closing this- as it's being combined with other, better fixes!

PR #6260, opened 2025-06-02
chore(luminork): Adding summaries to all of the endpoints for swagger-ui

stack72:


johnrwatson (approved):
I think we should adhere to our protected nouns:

https://docs.systeminit.com/reference/vocabulary

I.e. Change Set not change set in the description

johnrwatson (approved):

PR #6265, opened 2025-06-02
Cannot use snapshot addresses as markers. Using the checksum of the indexMV instead.

britmyerss:
Thats the main fix. This includes update the table structure, updating the PatchBatch and IndexUpdate messages to return the `fromChecksum` and `toChecksum`, updated the `IndexPointer` in Frigg to include the Checksum (so we don't need to fetch the whole `MvIndex` to get this) and updating the responses for mjolnir requests to include the index checksum. 

Other fixes include: 

Fixed Func history reactivity 

Fixed handling deleted MVs by returning the previous checksum for that object

When querying sometimes you want to query the "completed" change set, and sometimes you want the "in progress" checksum we're processing

When we call `updateComputed` within the `applyPatch` process, we need to query data off the indexChecksum. So we need to pipe that through the computed, references, and get() callstacks, and run the correct SQL. When we query from Vue components, we always want to look at the checksum being referenced on the changeset.

### DIFF
--- a/lib/edda-server/src/change_set_processor_task/materialized_view.rs
+++ b/lib/edda-server/src/change_set_processor_task/materialized_view.rs
@@ -28,10 +28,7 @@ use frigg::{
 };
 use si_events::{
     WorkspaceSnapshotAddress,
-    workspace_snapshot::{
-        Change,
-        Checksum,
-    },
+    workspace_snapshot::Change,
 };
 use si_frontend_mv_types::{
     MaterializedView,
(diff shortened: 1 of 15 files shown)


lib/edda-server/src/change_set_processor_task/materialized_view.rs
```diff
@@ -689,7 +693,16 @@ where
     MaterializedViewError: From<E>,
     MaterializedViewError: From<<T as TryInto<FrontendObject>>::Error>,
 {
-    let mv_kind = mv_kind.to_string();
+    let mv_kind: String = mv_kind.to_string();
```
jhelwig, line 696: Shouldn't need the `: String`?

PR #6267, opened 2025-06-02
feat(web): Ensure we show when an external subscription is in place between props

stack72:
Screenshot 2025-06-02 at 20 35 22


wendybujalski (approved):
👍

PR #6268, opened 2025-06-02
fix(web): In the new prop to prop world, allow connecting to props that don't have values yet

stack72:
Screenshot 2025-06-02 at 22 22 15


wendybujalski (approved):
👍

PR #6269, opened 2025-06-03
Fix/CVE 2025 48387 tar-fs

johnrwatson:
Patch the transitive dependency of tar-fs within the pkg node module.

This also adds the discord bot code to the pnpm workspace file. Without this when changing the version within ./bin/si-discord-bot/package.json, the root pnpm-lock.yaml file would not also get patched, leaving lingering dependencies.

**Ref:** Dependency Review - I've made it the highest possible version of pkg, so without a major re-work I can't patch any further. (This dependency failure is already in main/not the one I'm patching for)


stack72:
Closed via #6329

PR #6271, opened 2025-06-03
chore: Allow any host for web, on dev

vbustamante:
When developing from a machine that is not running the system, one needs to add the external host to the Vite config. This makes sense for production systems, but since we only use vite for development, it's kind of a hassle to have to do it for every external host.

This change allows us use any host to access a web system running with vite

### DIFF
--- a/app/web/vite.config.ts
+++ b/app/web/vite.config.ts
@@ -1,6 +1,6 @@
 import path from "path";
-import { existsSync, readFileSync } from "fs";
-import { loadEnv, defineConfig } from "vite";
+import {existsSync, readFileSync} from "fs";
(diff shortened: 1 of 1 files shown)


app/web/vite.config.ts
```diff
@@ -70,11 +70,12 @@ export default (opts: { mode: string }) => {
     css: {
       postcss,
       preprocessorOptions: {
-        less: { additionalData: lessVars },
+        less: {additionalData: lessVars},
       },
     },
     server: {
       host: config.DEV_HOST,
+      allowedHosts: true,
```
vbustamante, line 78: This is the functional change, the rest is formatting

jobelenus (approved):

PR #6274, opened 2025-06-03
feat(sdf): migrate multiple connections to one array

jkeiser:
This handles multiple connections to a single array, which was previously a migration error. It also makes the migrated connections to array sockets look more like what you would get from the new UI (with single -> array subscriptions hooking directly to explicit array elements instead of to the parent array).

* scalar -> array: We make the array a manual array, create an array element, and subscribe to the single from the array element using si:identity.
* array -> array: The subscription uses si:identity instead of si:normalizeToArray (since that's what si:normalizeToArray would do in the first place and is also the kind of subscription we create through the UI).

We refuse to migrate connections if: 
* If there are multiple array -> array connections to the same destination.
* If there are both single -> array and array -> array connections to the same destination.
* If the array socket does not use si:normalizeToArray, and there are multiple connections to the array.
* If the source socket does not use si:identity, we refuse to migrate, because we can't know whether the source produces an array or not (the error produced is "source and destination socket both have functions")

This PR also fixes an issue with where, when an array/object/map was using to a non-empty default value (such as in the case of socket connections), `PUT /attributes { "/domain/MyArray/-": ... }` (assigning directly to a child element) would leave the array itself using the default value, and the child element would get overwritten on the next DVU.

## Testing

- [x] Integration tests pass
- Manual tests:
  - [X] migrate array with multiple scalar connections
  - [ ] migrate array with single array connection
  - [ ] attempt to migrate array with multiple array connections
  - [ ] attempt to migrate array with array and scalar connections

### DIFF
--- a/lib/dal/src/attribute/path.rs
+++ b/lib/dal/src/attribute/path.rs
@@ -46,6 +46,9 @@ impl AttributePath {
         }
     }
 
+    /// Resolves the attribute value at the JSON pointer, if one exists.
+    ///
+    /// Returns None if it cannot be found.
     pub async fn resolve(
         &self,
         ctx: &DalContext,
@@ -59,6 +62,11 @@ impl AttributePath {
         }
     }
 
+    /// Gets the attribute value at the JSON pointer. After this is complete, the AV can be
+    /// set to an explicit value.
+    ///
+    /// Any parents that do not exist will be created. Any parents that are currently set to
+    /// default values will become explicit values.
     pub async fn vivify(
         &self,
         ctx: &DalContext,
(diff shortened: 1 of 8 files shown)


lib/dal/src/attribute/path.rs
```diff
@@ -155,7 +166,16 @@ async fn vivify_json_pointer(
...
+        // If this is currently using its default value, make it explicit so it can
+        // host child values.
         let prop = AttributeValue::prop(ctx, av_id).await?;
+        if AttributeValue::component_prototype_id(ctx, av_id)
+            .await?
+            .is_none()
+        {
+            AttributeValue::update(ctx, av_id, prop.kind.empty_value()).await?;
+        }
+        // Figure out the
```
vbustamante: Incomplete sentence?

vbustamante (approved):

lib/dal/src/attribute/path.rs
jkeiser: Comments fixed, thanks!

PR #6278, opened 2025-06-03
fix(bedrock): allow a generic restore point to be prepared from remote with no associated test

johnrwatson:
Previously, we Err'ed during preparation when pulling from remote if there were no .sequence files. By not erroring we allow preparing in a remote Postgres state only which can be useful for locally testing a specific state or to test a migration in a new branch. Handy!


johnrwatson:
@stack72 this fixes the bug you saw earlier

stack72 (approved):

PR #6279, opened 2025-06-03
feat(admin): Make workspace admin a bit more pleasant

jkeiser:
This makes the Admin page a bit more pleasant when it comes to workspaces and change sets, now that we're using it a bit more:

* Reduces scrolling:
  - Moves all workspace admin into the (currently unused) main panel
  - Puts more information side by side rather than top to bottom
* Defaults to the current workspace when you load it up
* Selects the HEAD changeset for a workspace by default, so something is loaded up on first load
* Sorts change sets so that open and active changesets come first

image

It's definitely not what I'd class as *wonderful*, but it gets the most annoying things out of the way!

### DIFF
--- a/app/web/src/components/AdminDashboard/WorkspaceAdmin.vue
+++ b/app/web/src/components/AdminDashboard/WorkspaceAdmin.vue
@@ -1,112 +1,85 @@
 <template>
-  <Stack class="max-w-xl">
+  <Stack
+    :class="
+      clsx(
+        '[&_dd]:m-xs [&_dd]:pl-xs [&_dd]:break-all [&_dd]:font-mono',
+        '[&_dt]:p-2xs [&_dt]:font-bold',
+        themeClasses(
+          '[&_select]:text-neutral-900 [&_select]:bg-neutral-100',
+          '[&_select]:text-neutral-200 [&_select]:bg-neutral-900',
+        ),
+      )
+    "
+  >
     <h2 class="font-bold text-xl">WORKSPACES</h2>
-    <VormInput
-      v-model="workspacesFilter"
-      label="type here to search for a workspace by id or name, or change set id or name, or user id, name, or email (50 results max). then press ENTER"
-      @keydown.enter="searchWorkspaces(workspacesFilter)"
-    />
-    <LoadingMessage :requestStatus="searchWorkspacesReqStatus" />
-    <p>{{ filteredWorkspaces?.length ?? 0 }} found</p>
-    <select
-      v-model="selectedWorkspaceId"
-      class="text-neutral-900 dark:text-neutral-200 dark:bg-neutral-900 bg-neutral-100"
-    >
-      <option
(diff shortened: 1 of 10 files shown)


app/web/src/components/AdminDashboard/WorkspaceAdmin.vue
```diff
@@ -1,112 +1,73 @@
 <template>
-  <Stack class="max-w-xl">
+  <Stack class="p-10 w-full">
```
wendybujalski: Generally we wanna avoid nonstandard numeric padding sizes in favor of named abstract ones - I would do `p-lg` here instead for a similar result, or honestly I don't see why so much padding is necessary, might as well put less!

app/web/src/components/AdminDashboard/WorkspaceAdmin.vue
```diff
@@ -1,112 +1,73 @@
...
+      <LoadStatus
+        :requestStatus="listChangeSetsReqStatus.value"
+        loadingMessage="Loading change sets ..."
+      >
+        <template #success>
+          <div class="flex flex-row gap-xs p-xs w-full">
+            <select
+              v-if="filteredChangeSets?.length > 0"
+              v-model="selectedChangeSetId"
+              class="text-neutral-900 dark:text-neutral-200 dark:bg-neutral-900 bg-neutral-100"
```
wendybujalski: We favor using the `themeClasses` helper function over the Tailwind `dark:` selector now.

app/web/src/components/Workspace/WorkspaceAdminDashboard.vue
wendybujalski, line 1: With these changes, the page has both horizontal and vertical scrollbars on a smaller screen. Would be good to restrict the flex layout to prevent horizontal scrolling.

wendybujalski (dismissed):
Nothing worth blocking on since this page is internal, but please do read my comments!

app/web/src/components/AdminDashboard/WorkspaceAdmin.vue
jkeiser: I copied it from somewhere else, so if there's a better one I'll use it! p-md?

app/web/src/components/AdminDashboard/WorkspaceAdmin.vue
```diff
@@ -1,112 +1,82 @@
 <template>
-  <Stack class="max-w-xl">
+  <Stack
+    :class="
+      clsx(
+        '[&_dd]:m-xs [&_dd]:pl-xs [&_dd]:break-all [&_dd]:font-mono',
+        '[&_dt]:p-2xs [&_dt]:font-bold',
+        themeClasses('[&_select]:text-neutral-900 [&_select]:bg-neutral-100', '[&_select]:text-neutral-200 [&_select]:bg-neutral-900')
```
wendybujalski: clever!

PR #6289, opened 2025-06-04
Revert "fix: ec2 instance attribute function and qualification"

stack72:
Reverts systeminit/si#6283

This this error:

```
{ "error": { "message": "schema error: pkg error: Package asked for a function with the unique id 5a5b8c9d6e7f8a9b0c1d2e3f4a5b6c7d8e9f0a1b2c3d4e5f6a7b8c9d0e1f2a3b but none could be found (error found while importing attribute func for prop)", "statusCode": 500 }, "level": null }
```


jobelenus (approved):

stack72:
Not needed since the fix is in #6291

PR #6292, opened 2025-06-04
Align qualification func run result to its state

nickgerace:
## Description

In the new UI, the qualification func run state now is reflected by both a factor of its underlying func run state and the result of said function. Part of what made this change tricky is the distinction between "unprocessed result" and "result". The "unprocessed result" is the one we need for this to work. Tangentially, this should probably be the result used in the func run details panel too, or at least it should be accessible to the user somehow.


app/web/src/newhotness/api_composables/func_run.ts
```diff
@@ -19,13 +19,17 @@ export type FuncRunState =
...
   | "Success";
 
-export type FuncKind =
-  | "action"
-  | "attribute"
-  | "authentication"
-  | "codeGeneration"
-  | "intrinsic"
-  | "management";
+export enum FuncKind {
```
nickgerace, line 22: This reflects the real values in use

app/web/src/newhotness/api_composables/func_run.ts
```diff
@@ -115,6 +120,26 @@ export interface FuncRunLog {
...
+  if (!funcRun.state) return null;
+
+  // If the qualification ran successfully, but it resulted in failure, then the state is a failure state.
+  if (
+    funcRun.functionKind === "Qualification" &&
+    funcRun.state === "Success" &&
+    (funcRun.unprocessedResultValue as any)?.result !== "success"
+  ) {
+    return "Failure";
+  }
```
nickgerace: The core change of the PR

PR #6293, opened 2025-06-04
All the hammers at the same time!

jobelenus:
If we are in "true cold start state" where we don't have the data and we need to fetch it, we want to make ONE request. However, while we do a cold start, the app is rendering and making queries, that are missing and firing their own hammers.

So, we need to stick a queue in there, that collects hammers that _would_ fire, and prevent them from firing. Once the cold start is done, if we were successful, drop those hammers and ignore it — we're fully up to date. If something really weird happens, and the bulk fails (honestly, the only thing I could imagine is a timeout, but 🤷 ), let those hammers fly as a backup to get out of a broken situation.

Testing steps:
1. Go to a change set.
2. Clear your network tab
3. Hit the Ragnarok button (which drops only this change set data)
4. Watch your application re-cold start
5. See only `multi_mjolnir` in the network tab. No other hammers

Aside: small fix to the reactive-ness of the rainbow counter. I put reactive on the DefaultMap, which made it stop working entirely. I need to put the reactive on the interior sets, since thats what we're "counting" in the computed (don't need it on the changeSetId key, since that is reactive via injecting context.

### DIFF
--- a/app/web/src/newhotness/logic_composables/rainbow_counter.ts
+++ b/app/web/src/newhotness/logic_composables/rainbow_counter.ts
@@ -2,8 +2,9 @@ import { computed, inject, reactive } from "vue";
 import { DefaultMap } from "@/utils/defaultmap";
 import { assertIsDefined, Context } from "../types";
 
-const q = new DefaultMap<string, Set<string>>(() => new Set<string>());
-const queueByChangeSet = reactive(q);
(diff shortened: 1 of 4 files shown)


britmyerss:
Tested this in the Prod SaaS workspace last night - works a treat!

britmyerss (approved):

PR #6296, opened 2025-06-04
fix(edda, web): Use Workspace Id for MV list items, rather than change set

britmyerss:
This change allows us to actually reuse MV Indexes across change sets. Without this change, when we copy the Index Pointer for a new change set, we never build the list items, and their Ids are pinned to the change set id from which we copied, making them unusable. By keying off of the workspace Id, we can share these objects across change sets, reducing the time to build, time to create change sets, and we get more shared objects.


lib/dal-materialized-views/src/component_list.rs
```diff
@@ -19,9 +19,9 @@ pub async fn assemble(ctx: DalContext) -> super::Result<ComponentListMv> {
     for component_id in component_ids {
         components.push(super::component::assemble(ctx.clone(), component_id).await?);
     }
-
+    let id: dal::WorkspacePk = ctx.workspace_pk()?;
```
jhelwig: Would be nice to make the formatting of how this is added to the MVs consistent with each other.

jhelwig (commented):
Really only a minor comment aside from generally needing to fix the linting problems.

PR #6301, opened 2025-06-04
Fix: when making a new change set we need to change the optimistic query key accordingly

jobelenus:

### DIFF
--- a/app/web/src/newhotness/AddComponentModal.vue
+++ b/app/web/src/newhotness/AddComponentModal.vue
@@ -370,7 +370,13 @@ const onEnter = async () => {
       schemaVariant: req.data.schemaVariantMaterializedView,
     };
 
-    queryClient.setQueryData(componentQueryKey.value, () => bifrostComponent);
+    const keyValue = componentQueryKey.value;
+    // replace old change set id with new one for the query key
+    if (newChangeSetId)
+      keyValue.forEach((v, idx) => {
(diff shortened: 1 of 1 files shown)


app/web/src/newhotness/AddComponentModal.vue
```diff
@@ -370,7 +370,13 @@ const onEnter = async () => {
       schemaVariant: req.data.schemaVariantMaterializedView,
     };
 
-    queryClient.setQueryData(componentQueryKey.value, () => bifrostComponent);
+    const keyValue = componentQueryKey.value;
+    // replace old change set id with new one for the query key
+    if (newChangeSetId)
+      keyValue.forEach((v, idx) => {
+        if (v === route.params.changeSetId) keyValue[idx] = newChangeSetId;
```
zacharyhamm, line 377: couldn't we just ask about `keyValue[1]`?

zacharyhamm (approved):

app/web/src/newhotness/AddComponentModal.vue
jobelenus: yep.. i bet  this is gonna become a generic soon, so.. just covering all bases (though... i just remembered we removed the changeset from "non-id" keys and replaced it with workspace.. so.. maybe [1] is all we'll ever need 🤔 )

PR #6305, opened 2025-06-04
feat(sdf): Migrate connections IAM/ARN sockets

jkeiser:
There are a bunch of custom socket transform functions in IAM resources that do the same thing as si:identity or si:normalizeToArray. This treats those functions the same way during migration, allowing us to migrate a bunch more connections.

It also handles when you connect multiple AWS ARN resources to an array of ARNs--the ARN socket outputs a single value, even though it's not the identity function, so when you use normalizeToArray on the other side, we can create a separate element for each subscription to the ARN socket.

This should finish out migration of Production SaaS Workspace as well as most (if not all) of some of our other big workspaces.

### DIFF
--- a/lib/dal/src/workspace_snapshot/graph/validator.rs
+++ b/lib/dal/src/workspace_snapshot/graph/validator.rs
@@ -4,6 +4,7 @@ use std::collections::{
 };
 
 use itertools::Itertools as _;
+use lazy_static::lazy_static;
 use petgraph::prelude::*;
 use serde::{
     Deserialize,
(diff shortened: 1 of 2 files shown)


lib/dal/src/workspace_snapshot/graph/validator.rs
```diff
@@ -407,7 +407,49 @@ pub fn is_identity_func(
     let func_node = graph
         .get_node_weight_by_id(func_id)?
         .get_func_node_weight()?;
-    Ok(func_node.func_kind() == FuncKind::Intrinsic && func_node.name() == "si:identity")
+    Ok(IDENTITY_FUNCS.contains(func_node.name()))
+}
+
+lazy_static! {
+    static ref IDENTITY_FUNCS: HashSet<&'static str> = [
```
wendybujalski, line 414: using hash sets for this is great, clean and easy to read!

PR #6311, opened 2025-06-04
Ensure Components in ViewComponentList have stable ordering

jhelwig:
The lack of stable ordering of the `Component`s was causing patches to be unnecessarily large, and for there to be unnecessary patches in the first place.


fnichol (approved):
Yep!

PR #6317, opened 2025-06-05
fix(dal): DVU now enqueues relevant update actions if not on head

britmyerss:
This PR shifts some logic from an SDF route handler to the DVU job.  Previously, updating a prop for a component, we would build the DVU graph, and walk it to see if any downstream components might be impacted - all before returning to the caller.  This causes the response time to increase as it becomes more expensive to build the DVU graph. 

Now, this happens in DVU, where we already have all of the data we need. Better yet, this is more accurate than what we had before, as before we would enqueue the update function if a value is impacted. However, at that time, we don't yet know if the value will *change* as a result of the update made, we only know that it *might*. This lets us more precisely detect if a root/domain prop has actually changed before we look to enqueue an update function.

Special casing this in DVU isn't ideal, but will solve the problem for now as we look for a better solution to handle these effects of change.

Updated the integration test to cover off on this!
<div><br/>via <a href="https://giphy.com/Snapon/">Snap-on Tools</a> on <a href="https://giphy.com/gifs/Snapon-update-alert-needed-2x0tJVAL3IqFnZYhYt">GIPHY</a></div>

### DIFF
--- a/lib/dal/src/job/definition/dependent_values_update.rs
+++ b/lib/dal/src/job/definition/dependent_values_update.rs
@@ -43,11 +43,13 @@ use crate::{
     ComponentId,
     DalContext,
     Func,
+    SchemaVariantError,
     TransactionsError,
     WorkspacePk,
     WorkspaceSnapshotError,
     WsEvent,
     WsEventError,
+    action::ActionError,
     attribute::value::{
         AttributeValueError,
         dependent_value_graph::DependentValueGraph,
@@ -73,6 +75,8 @@ use crate::{
 #[remain::sorted]
 #[derive(Debug, Error)]
 pub enum DependentValueUpdateError {
+    #[error("action error: {0}")]
+    Action(#[from] ActionError),
     #[error("attribute value error: {0}")]
     AttributeValue(#[from] AttributeValueError),
     #[error("change set error: {0}")]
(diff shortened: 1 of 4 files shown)


lib/dal/src/job/definition/dependent_values_update.rs
```diff
@@ -419,6 +428,17 @@ impl DependentValuesUpdate {
...
                                         dependency_graph.remove_value(finished_value_id);
                                         drop(write_guard);
+                                        // if we're not on head, and the value is different after processing,
+                                        // let's see if we should enqueue an update function
+                                        if !is_on_head
+                                            && before_value != execution_values.value().cloned()
+                                        {
+                                            Component::enqueue_update_action_if_applicable(
+                                                ctx,
+                                                finished_value_id,
```
zacharyhamm: Let's avoid this clone and the one above. `before_value` and `execution_value` are available in a higher scope, so you can stash the result of the comparison in a boolean

`let value_is_changed = before_value.as_ref() != execution_values.value();`

I think right before the write lock is acquired makes sense (or right after), a few lines up.

Then you can use it here, and we don't have to clone `execution_values`

zacharyhamm (commented):
The new `enqueue_action_if_applicable` method makes sense, and it's a smart way to avoid calculating the dep graph in the update func. Not sure of a good set of components to test this on?

One request related to the clones of execution_value, which can be avoided entirel

lib/dal/src/job/definition/dependent_values_update.rs
britmyerss: ohh yes - thank you will do this!

lib/dal/src/job/definition/dependent_values_update.rs
```diff
@@ -388,6 +397,8 @@ impl DependentValuesUpdate {
                             // Lock the graph for writing inside this job. The
                             // lock will be released when this guard is dropped
                             // at the end of the scope.
+                            let value_is_changed =
```
zacharyhamm, line 400: 🚀

zacharyhamm (approved):
ship it

PR #6321, opened 2025-06-05
feat: Export and import transformation funcs

vbustamante:
- Exported packages now contain all transformation funcs in the changeset
- Transformation funcs get imported with a stable ID and last update timestamp, so we can check for the latest
- Upgrading a func on import triggers dvu for subscriptions that use it
- Updating code for a func now updates the updated_at timestamp
- Listing transformation funcs skips bindings so we don't get errors (they get listed correctly)


lib/dal/src/pkg/export.rs
```diff
@@ -1104,6 +1117,16 @@ impl PkgExporter {
             }
         }
 
+        // Add all transformation funcs to package
```
britmyerss, line 1120: @vbustamante does this all packages will include all transformation funcs?

vbustamante: Yes, but right now they're not user authorable, so the idea is that we won't get a lot of these

lib/dal/src/func/authoring.rs
```diff
@@ -623,6 +624,7 @@ impl FuncAuthoringClient {
         func.error_if_locked()?;
         Func::modify_by_id(ctx, func.id, |func| {
             func.code_base64 = Some(general_purpose::STANDARD_NO_PAD.encode(code.into()));
+            func.timestamp.updated_at = Utc::now();
```
zacharyhamm, line 627: We probably are not updating the updated_at for any nodes we modify elsewhere, interesting

lib/dal/src/func.rs
```diff
@@ -255,9 +260,16 @@ impl Func {
...
         is_transformation: bool,
+        updated_at: Option<DateTime<Utc>>,
     ) -> FuncResult<Self> {
-        let timestamp = Timestamp::now();
-        let _finalized_once = false;
+        let timestamp = {
+            let mut timestamp = Timestamp::now();
+
+            if let Some(updated) = updated_at {
+                timestamp.updated_at = updated;
```
jkeiser, line 269: I don't get this one--I assume you pass an updated_at that's intended to be some past date, so wouldn't this cause `updated_at` to be *before* `created_at` (since that gets set to now()?

zacharyhamm: Ah good catch there

vbustamante: That's a good catch! Created at should be the same as updated at

vbustamante: I'm gonna change this

lib/dal/src/func/binding/attribute.rs
```diff
@@ -291,6 +291,13 @@ impl AttributeBinding {
         func_id: FuncId,
     ) -> FuncBindingResult<Vec<FuncBinding>> {
         let mut bindings = vec![];
+
+        // Note(victor): AttributeArgumentBinding::assemble, fails for ValueSubscription value source.
```
jkeiser, line 295: If a func is used both for transformations and as an attribute func, this will miss those bindings. Maybe it would be better to have AttributeArgumentBinding::assemble() return None if it's not a binding to a schema, and then do `let Some(binding) = AttributeArgumentBinding::assemble() else { continue; }` in this loop?

lib/dal/src/pkg/export.rs
jkeiser: I *think* the requirement is "Paul would like to author some transformation funcs (probably via API) and then manually repackage and export Region and/or AWS Credential so they contain them."

We probably want to make it so Paul's repackaging/export happens via the export API with an extra "include_transformation_funcs=true" parameter, so we don't just automatically slurp up all transformation funcs whenever any user exports any package. But we can make another turn for that if it's needed.

jkeiser: > I _think_ the requirement is "Paul would like to author some transformation funcs (probably via API) and then manually repackage and export Region and/or AWS Credential so they contain them." We probably want to make it so Paul's repackaging happens via an API call with an extra "include_transformation_funcs" too, and we don't just automatically slurp up all transformation funcs whenever any user exports any package. But we can make another turn for that if it's needed.

@stack72 ^^ checking my understanding

stack72: yeah we are going to make another turn here when we land the actual authoring changes needed

lib/dal/src/pkg/import.rs
```diff
@@ -216,36 +214,64 @@ async fn import_change_set(
...
-                );
+            let func_spec_data = func_spec
+                .data()
+                .ok_or(PkgError::DataNotFound(func_spec.name().into()))?;
 
-                if let Some(module) = installed_module.clone() {
-                    module.create_association(ctx, func.id.into()).await?;
-                }
+            // If this func is a transformation, we need to see if it updates an existing func before creating it
+            if func_spec_data.is_transformation() {
```
jkeiser, line 225: Doesn't this violate `is_locked`? Schema variants get a new ID when you unlock them and update, and so do funcs. I'm worried violating this principle will make it harder to reason about the system and have have downstream implications on the rest of the graph, where we very probably already make assumptions that locked FuncIds never change ... for example, when you update a locked transformation func, do you re-run DVU for everything that uses it?

If we don't *need* to support updating existing transformation funcs, I suggest we leave it for now--we need to think it through more comprehensively before we support it. We can still update transformation funcs, we just need to include the new Func (with new FuncId) in an updated Region/AWS Credential package.

lib/dal/src/pkg/export.rs
jkeiser: Based on that @vbustamante I think this is fine for right now, as long as the bool is off by default--we should add a parameter to the export API call in a future turn for @stack72. (Unless I missed it and it's here.)

lib/dal/src/func/authoring.rs
jkeiser: I would have expected `modify_by_id()` to set it all by itself, anyway!

jkeiser (approved):
I'm a little confused why we still have to have the codepath to skip importing transformation funcs if they already exist--I would have thought that applied to all funcs everywhere?

We should still track down https://github.com/systeminit/si/pull/6321/files#r2129887420 but that can happen later.

PR #6327, opened 2025-06-05
fix(web): Don't early-initialize bifrost with random token

jkeiser:
We're getting some failed requests to bifrost that use entirely the wrong token. I can't make this version of the problem repro precisely in localdev, but this fix will make it so that we will error instead of using a random token when we initialize bifrost, so we can debug it.

## Testing

- [x] Old UI still shows action runs (I can't see or create components locally either way)
- [x] New UI can create components
- [x] Account flows (login / logout / Go To Workspace / reload page) work


jkeiser:
Closing in favor of #6362

PR #6330, opened 2025-06-06
Use SAH Pool VFS for sqlite DB in webworker

jhelwig:
The SyncAccessHandle VFS is the highest performance option of the OPFS-backed configurations (currently) described in the SQLite WASM documentation.

https://sqlite.org/wasm/doc/trunk/persistence.md#vfs-opfs-sahpool

In local testing switching to the SAH VFS from the standard OPFS configuration significantly reduces the time from hitting refresh in the browser to having a fully rendered component list in the new UI when all of the necessary data is already loaded into SQLite, as well as reducing the time needed to do the initial import of data into SQLite. In the warm-sqlite-cache case it was reduced from ~60 seconds to ~20 seconds in local testing.

### DIFF
--- a/app/web/src/workers/webworker.ts
+++ b/app/web/src/workers/webworker.ts
@@ -139,15 +139,16 @@ const getDbName = (testing: boolean) => {
 
 const start = async (sqlite3: Sqlite3Static, testing: boolean) => {
   const dbname = getDbName(testing);
-  db =
-    "opfs" in sqlite3
-      ? new sqlite3.oo1.OpfsDb(`/${dbname}`)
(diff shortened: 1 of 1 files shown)


jhelwig:
@zacharyhamm reminded me that this will completely blow up with multiple browser tabs open at the same time, since the SAH Pool VFS mode does not support concurrent connections to the DB.
image
The plan has always been to have a single webworker task accessing the DB at any given time, but the shared worker doing the routing to handle this is not in place yet. So, this will have to wait until that happens.

jhelwig:
@zacharyhamm brought this home via #6399.

PR #6335, opened 2025-06-06
fix(edda): impl a liveness-aware quiescent period for compressing stream

fnichol:
This change alters the "quiescent_period" timeout on
`ChangeSetProcessorTask`s by basing the quiet period on the last time
the `CompressingStream`'s state machine was updated. This way, if there
is still work to be done inside the compressing stream, a quiet period
will *not* be detected until we are waiting for another "first message"
(or if it stalls on another intermediate state).

Additionally, the current value of the stream's `State` is recorded as
a `task.state` attribute on the `compressing_stream.next` span. This
would allow us to alert on the compressing stream closing in unexpected
states.

chore(edda): increase default quiescent period to `10m`
------------------------

Yep, we're upping this for now!

### DIFF
--- a/lib/edda-server/src/change_set_processor_task.rs
+++ b/lib/edda-server/src/change_set_processor_task.rs
@@ -6,7 +6,10 @@ use std::{
     io,
     result,
     sync::Arc,
-    time::Duration,
+    time::{
+        Duration,
+        Instant,
+    },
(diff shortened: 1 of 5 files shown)


britmyerss (dismissed):
:shipit:

PR #6336, opened 2025-06-06
fix: tweak scroll and zoom behaviour when on map view

aaron-dernley:
A suggested change to the default zoom/scroll behaviour when on the Map view. Instead of scroll by default being zoom in and out, make it pan around (using a touch mouse with up, down, left and right scroll this is lovely) and then alter the arrow key movements to match the scroll direction. To zoom in and out cmd/ctrl + scroll up and down. No other behaviour should be affected here, this matches the current UI Model screen.

### DIFF
--- a/app/web/src/newhotness/Map.vue
+++ b/app/web/src/newhotness/Map.vue
@@ -143,7 +143,7 @@ const origCenter = ref(viewBox.value / 2);
 const origTransformMatrix = [1, 0, 0, 1, 0, 0];
 const transformMatrix = reactive<number[]>([...origTransformMatrix]);
 
-const scaleStep = 0.1;
+const scaleStep = 0.13;
 const baseZoom = 1;
 const zoomLevel = ref(baseZoom);
 const lastPos = ref<xy | null>(null);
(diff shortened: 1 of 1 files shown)


app/web/src/newhotness/Map.vue
```diff
@@ -231,8 +231,18 @@ const zoomOut = () => {
...
+    // Zoom behavior when modifier key is held
+    if (event.deltaY > 0) {
+      zoomOut();
+    } else {
+      zoomIn();
+    }
+  } else {
+    // Pan behavior - move map in scroll direction
+    const panSpeed = 40;
+    pan((-event.deltaX * panSpeed) / 100, (-event.deltaY * panSpeed) / 100);
```
wendybujalski, line 244: Have you tested this with a mouse scroll wheel and a trackpad?

wendybujalski (approved):

app/web/src/newhotness/Map.vue
aaron-dernley: I don't have a wheely mouse so I may need a volunteer! I have tested using touch/magic mouse

johnrwatson: I tested the previous iteration which was a bit too slow, I think this is a few ticks faster which will be better

PR #6339, opened 2025-06-06
fix(module-index-client): re-use `reqwest::Client` across calls

fnichol:
This change forces the `ModuleIndexClient` to re-use a once-built `reqwest::Client` to reuse underlying connection pooling, DNS lookups, etc. The bearer auth is replicated using the underlying associated header and is marked as `sensitive` so that any debug logging doesn't accidentally leak credentials (which is what that
`RequestBuilder.bearer_auth` does in its impl).

Prior to this change, every fresh `reqwest::Client` is forced to DNS-resolve its URLs, re-establish any persistent TCP connections, etc.

### DIFF
--- a/bin/hoist/src/main.rs
+++ b/bin/hoist/src/main.rs
@@ -74,7 +74,7 @@ async fn main() -> Result<()> {
 
     let endpoint = &args.endpoint;
     let token = &args.token;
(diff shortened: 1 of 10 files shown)


adamhjk (approved):
lgtm! lets ship it!

PR #6343, opened 2025-06-07
Add ability to remove non-secret subscriptions

nickgerace:
## Description

This change adds the ability to remove non-secret subscriptions by sending a "null" source via PUT request for the given attribute.

Screenshot 2025-06-06 at 9 18 38 PM


stack72:
@nickgerace just tested this! Works well

PR #6344, opened 2025-06-07
feat: Duplicate Components on new UI

vbustamante:
- Changes the old paste endpoint back to paste
- Implements a duplicate sdf endpoint that takes in multiple component ids, clones them and retains subscriptions and mgmt relationships
- Adds and option to the component context menu and a keyboard shortcut (cmd/ctrl+d) to duplicate
- Adds a flag to avoid us creating multiple keydown listeners
- Doesn't implement optimistic creation of copies (and feels somewhat slow, but it's reactive)

image

### DIFF
--- a/app/web/src/newhotness/Explore.vue
+++ b/app/web/src/newhotness/Explore.vue
@@ -121,7 +121,7 @@
           />
         </footer>
       </template>
-      <Map v-else :active="!showGrid" />
+      <Map v-else :viewId="selectedViewOrDefaultId" :active="!showGrid" />
     </div>
     <!-- Right column -->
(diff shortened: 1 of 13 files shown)


app/web/src/newhotness/Explore.vue
```diff
@@ -121,7 +121,7 @@
           />
         </footer>
       </template>
-      <Map v-else :active="!showGrid" />
+      <Map v-else :viewId="selectedViewOrDefaultId" :active="!showGrid" />
```
britmyerss, line 124: Woot! Checks off another one!

PR #6347, opened 2025-06-07
chore(edda): record `src_requests.count` in `handle_task` span

fnichol:
This change adds an `si.edda.src_requests.count` attribute on each `edda.change_set_processor_task.handle_request` span so that it's easier to track how many API requests were compressed.


fnichol:
Here's an example of the attribute in the span:

Screenshot 2025-06-07 at 14-56-15 Explore - Jaeger Jaeger - Grafana

stack72 (dismissed):

PR #6349, opened 2025-06-07
ViewId doesn't have to be a prop everywhere. We can use provide/inject context instead, which can go to all children

jobelenus:

### DIFF
--- a/app/web/src/newhotness/Map.vue
+++ b/app/web/src/newhotness/Map.vue
@@ -78,7 +78,6 @@
 
     <ComponentContextMenu
       ref="componentContextMenuRef"
-      :viewId="viewId"
       :componentIds="selectedComponent ? [selectedComponent.id] : []"
     />
   </section>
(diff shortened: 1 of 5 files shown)


app/web/src/newhotness/Map.vue
```diff
@@ -78,7 +78,7 @@
 
     <ComponentContextMenu
       ref="componentContextMenuRef"
-      :viewId="viewId"
+      :viewId="explore.viewId.value"
```
vbustamante: This pr removes the view Id prop from this component, should this attr passing here just be removed?

vbustamante: Probably should be removed in Explore.vue too since the component is also there

jobelenus: I caught it there! but mindlessly kept it here. Removed! Good catch

vbustamante (approved):
Nice!

PR #6356, opened 2025-06-09
Don't initiate the workspace without the specific token for that workspace

jobelenus:
Without a token you don't have access.

Testing steps:

1. app loads under happy path conditions
2. incognito mode directly accessing a workspace correctly redirects to the auth portal

If there are steps that land us on "Bad token", we'll find them! But we've got to get out of a local environment it would seem.


nickgerace (approved):

jobelenus:
> Are we duplicating tokensByWorkspacePk instead of using authStore.tokens because we want to start bypassing auth store entirely? Or is the auth store not loading them early enough, and needs to be fixed? It's good to stop grabbing the "default" token, and this should tell us when we get into the bad state.

Yes, both and ;)

PR #6358, opened 2025-06-09
Add ability to remove secrets in new UI

nickgerace:
## Description

This change adds the ability to remove secrets in the new UI via prop subscriptions.

### DIFF
--- a/app/web/src/newhotness/layout_components/AttributeInput.vue
+++ b/app/web/src/newhotness/layout_components/AttributeInput.vue
@@ -93,14 +93,13 @@
         tone="action"
       />
       <!-- NOTE(nick): you need "click.stop" here to prevent the outer click -->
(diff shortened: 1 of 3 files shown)


app/web/src/newhotness/layout_components/AttributeInput.vue
```diff
@@ -1052,11 +1031,6 @@ const selectedConnection = computed(
 //     fixWindowPosition();
 //   });
 // });
-
-// TODO(Wendy) - add the ability to clear a value!
-// const clearValue = (e: Event) => {
-//   e.preventDefault();
-// };
```
nickgerace, line 1059: We can re-introduce if need be, but it remains unused for now and is not needed for both remove subscription scenarios.

Source: https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault

PR #6359, opened 2025-06-09
feat(si-db, dal, pinga): management func jobs

zacharyhamm:
Adds a pinga job for executing management funcs. Adds a tracking table to ensure only one job at time per component and prototype and change set.

Jobs retry until the DVU roots are clear.

The v2 run_prototype route has been changed to spawn the job, but I have not yet updated the public api or luminork api routes, since I'm not sure what the expectation is for the data they return (they cannot return the message and state of the job since it is now dispatched by pinga).

### DIFF
--- a/lib/sdf-server/src/service/v2/management.rs
+++ b/lib/sdf-server/src/service/v2/management.rs
@@ -16,27 +16,26 @@ use axum::{
         post,
     },
 };
+use chrono::{
+    Duration,
+    Utc,
+};
 use dal::{
     ChangeSet,
     ChangeSetError,
     ChangeSetId,
     ComponentId,
-    Func,
     FuncError,
     FuncId,
     SchemaVariantError,
     TransactionsError,
     WorkspacePk,
-    WsEvent,
     WsEventError,
     diagram::view::ViewId,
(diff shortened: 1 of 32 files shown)


zacharyhamm:
@fnichol I need to grab you for a few minutes to talk about the request versioning, and the pinga client tests

lib/sdf-server/src/service/v2/management.rs
```diff
@@ -156,11 +125,28 @@ pub async fn run_prototype(
...
 
     let force_change_set_id = ChangeSet::force_new(&mut ctx).await?;
-
-    // TODO check that this is a valid prototypeId
-    let mut execution_result =
-        ManagementPrototype::execute_by_id(&ctx, prototype_id, component_id, view_id.into())
+    let state_ctx = ctx.clone();
+
+    // Is there already a pending run? if so fail it if it's too old, so that we can start a new one
+    if let Some(pending) =
```
zacharyhamm, line 169: I tested this by having a successful run, then in the database flipping it back to pending. Then, 7 minutes later, trying to run the job again, and seeing that it (a) ran and (b) flipped that job row to failure.

PR #6370, opened 2025-06-09
feat(luminork): Add support for setting prop subscriptions

stack72:
This will allow a user to set subscriptions from a prop on a component to another prop. We can do this as follows:

```
{
  "domain": {
    "propId1": "value1",
    "path/to/prop": "value2"
  },
  "name": "MyComponentName",
  "schemaName": "AWS::EC2::Instance",
  "secrets": {
    "secretDefinitionName": "secretName"
  },
  "subscriptions": {
    "/prop/path/on/this/component": {
      "component": "OtherComponentName",
      "propPath": "/prop/path/on/other/component"
    }
  }
}
```

Please Note: I will refactor this to remove the duplication from the new dal constructs - I need this to get out there!

### DIFF
--- a/lib/luminork-server/src/service/v1/components/create_component.rs
+++ b/lib/luminork-server/src/service/v1/components/create_component.rs
@@ -32,6 +32,11 @@ use super::{
         handle_connection,
     },
     resolve_secret_id,
+    subscriptions::{
+        AttributeValueIdent,
+        Subscription,
+        handle_subscription,
+    },
 };
 use crate::{
     extract::{
@@ -130,6 +135,7 @@ pub async fn create_component(
     )
     .await?;
 
+    let component_list = Component::list_ids(ctx).await?;
     if let Some(resource_id) = payload.resource_id {
         let resource_prop_path = ["root", "si", "resourceId"];
         let resource_prop_id =
@@ -153,6 +159,10 @@ pub async fn create_component(
         AttributeValue::update(ctx, attribute_value_id, Some(value.clone())).await?;
     }
 
+    for (av_to_set, sub) in payload.subscriptions.clone().into_iter() {
+        handle_subscription(ctx, av_to_set, &sub, component.id(), &component_list).await?;
+    }
+
     for (key, value) in payload.secrets.clone().into_iter() {
         let prop_id = key.prop_id(ctx, variant_id).await?;
 
@@ -169,12 +179,6 @@ pub async fn create_component(
         .view(ctx)
         .await?;
     let after_value = serde_json::to_value(after_domain_tree)?;
-
-    let component_list = Component::list_ids(ctx).await?;
-
-    let added_connection_summary =
-        super::connections::summarise_connections(ctx, &payload.connections, &component_list)
-            .await?;
 
     if !payload.connections.is_empty() {
         for connection in payload.connections.iter() {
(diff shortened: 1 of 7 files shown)


lib/luminork-server/src/service/v1/components/create_component.rs
```diff
@@ -195,7 +200,7 @@ pub async fn create_component(
             component_name: comp_name.clone(),
             before_domain_tree: None,
             after_domain_tree: Some(after_value),
-            added_connections: Some(added_connection_summary),
+            added_connections: None,
```
stack72, line 203: this is on purpose as it was too expensive to create for what it is

lib/luminork-server/src/service/v1/components/mod.rs
```diff
@@ -347,6 +353,62 @@ impl SecretPropPath {
...
+        if name == component_name {
+            matching_components.push(*component_id);
+        }
+    }
+
+    match matching_components.len() {
+        0 => Err(ComponentsError::ComponentNotFound(
+            component_name.to_string(),
+        )),
+        1 => Ok(matching_components[0]),
```
nickgerace, line 405: I'm a bit surprised this works. Does that return an `Option`?

lib/luminork-server/src/service/v1/components/mod.rs
```diff
@@ -347,6 +353,62 @@ impl SecretPropPath {
...
+        let name = Component::name_by_id(ctx, *component_id).await?;
+        if name == component_name {
+            matching_components.push(*component_id);
+        }
+    }
+
+    match matching_components.len() {
+        0 => Err(ComponentsError::ComponentNotFound(
+            component_name.to_string(),
+        )),
```
nickgerace, line 404: Another idea to do this: create a `let mut matched_component = None;` variable and then set it on the first match. If it is `Some` the second time around, then you can `return Err` the duplicate component name error. That way you don't need a `Vec`.

lib/luminork-server/src/service/v1/components/subscriptions.rs
```diff
@@ -0,0 +1,174 @@
...
+        let Ok(id) = self.0.parse() else {
+            return Ok(None);
+        };
+
+        let Some(NodeWeight::Func(_)) = ctx.workspace_snapshot()?.get_node_weight_opt(id).await
+        else {
+            return Ok(None);
+        };
+
+        Ok(Some(id))
```
nickgerace, line 113: Nice! These are clever

lib/luminork-server/src/service/v1/components/mod.rs
nickgerace: Pseudo-ish code using `bool` instead of `Option`...
```rust
let mut found_matching_component = false;
for component_id in component_list {
    let name = Component::name_by_id(ctx, *component_id).await?;
    if name == component_name {
        if found_matching_component {
            // return duplicate component error
        }
        found_matching_component = true;
    }
}
if !found_matching_component {
    // return component not found error
}
// return happy path
```

lib/luminork-server/src/service/v1/components/create_component.rs
nickgerace: I'd leave a comment indicating that maybe. It's not obvious that's what's going on (though I pieced it together 😄)

nickgerace (approved):
Looks great to me! Nice job 🚀

lib/luminork-server/src/service/v1/components/mod.rs
nickgerace: Discussed offline. This is a port of existing work and is functional as-is.

lib/luminork-server/src/service/v1/components/mod.rs
nickgerace: https://github.com/systeminit/si/pull/6370#discussion_r2150791808

PR #6373, opened 2025-06-10
feat(dal): prop suggestions

jkeiser:
This allows you to create prop suggestions with the asset builder:

```ts
function main() {
  return new AssetBuilder()
    .addProp(new PropBuilder()
      .setName("VpcId")
      .suggestSource({ schema: "AWS::EC2::VPC", prop: "VpcId" })
    )
    .build();
}
```

DRAFT: this is breaking content deserialization tests with the error `Postcard(DeserializeBadOption)`, and I'm not sure why. Making it `Option<Value>` or even `Option<String>` doesn't help. I'll get help in the morning.


lib/dal/src/layer_db_types/content_types.rs
```diff
@@ -564,6 +568,77 @@ pub struct PropContentV1 {
...
+    /// Props can be connected to eachother to signify that they should contain the same value
+    /// This is useful for diffing the resource with the domain, to suggest actions if the real world changes
+    pub refers_to_prop_id: Option<PropId>,
+    /// Connected props may need a custom diff function
+    pub diff_func_id: Option<FuncId>,
+    /// A serialized validation format JSON object for the prop.
+    pub validation_format: Option<String>,
+    /// Optional UI data that isn't needed in the graph
+    /// (Eventually, many other fields in content will be moved here, e.g. documentation/docLink)
+    pub ui_optionals: Option<HashMap<String, CasValue>>,
```
vbustamante, line 597: @jkeiser 
Since we're introducing this I think it'd be good to at least have the comment be a bit more specific about when to add stuff here vs on widget options (even if it means that widget options is not to be used to any more things)

But I see that the secret kind for example is not moving (and I agree that it shouldn't, mostly because I feel widget options should be used for everything - even if I don't care about it strongly) and that raises some questions around when to use what

jkeiser: @vbustamante our feeling here was that widget options should be more restricted because of how you specify it in asset builder: i.e. it seems like it's options for the corresponding WidgetKind--if you change the widget kind, the options are not necessarily valid anymore.

It seemed awkward to have `suggestSource()` be inside `WidgetDefinitionBuilder` rather than `PropBuilder`, basically. "If I'm not specifying a widget why do I have to have a WidgetDefinition?"

You could definitely slice it the other way though, and say "the widget really includes everything the UI displays," which actually means documentation/docLink should go there too ... it just doesn't quite fit the way we currently define it in the asset builder.

lib/dal/src/prop.rs
```diff
@@ -814,9 +855,7 @@ impl Prop {
             .ok_or(WorkspaceSnapshotError::MissingContentFromStore(id.into()))?;
 
         // NOTE(nick,jacob,zack): if we had a v2, then there would be migration logic here.
```
zacharyhamm: Should probably remove this comment since you've encapsulated the upgrade in the assemble

sprutton1 (approved):

lib/dal/src/schema/variant.rs
```diff
@@ -621,7 +621,7 @@ impl SchemaVariant {
 
         // New SchemVariants are not locked by default.
         let is_locked = false;
-        let content = SchemaVariantContentV3 {
```
zacharyhamm, line 624: What's happening here?

lib/dal/src/prop.rs
```diff
@@ -1129,31 +1165,6 @@ impl Prop {
         Ok(props)
     }
 
-    pub async fn modify<L>(self, ctx: &DalContext, lambda: L) -> PropResult<Self>
```
zacharyhamm, line 1132: Is this unused? seems likely

lib/dal/src/schema/variant.rs
jkeiser: Just makes it so we're not typing `SchemaVariantContent::V3` twice below here. Mirrors what we did in PropContentV3. Not necessary to make this change at all though.

lib/dal/src/prop.rs
jkeiser: Yep, it all compiles after removing, so we decided to remove rather than fix it to use PropContentV2.

lib/dal/src/prop.rs
jkeiser: Done, thanks!

PR #6374, opened 2025-06-10
Ability ability to delete a component with ghosting

nickgerace:


app/web/src/newhotness/ComponentContextMenu.vue
```diff
@@ -111,39 +112,70 @@ const actionByPrototype = computed(() => {
...
 
+  // If we are dealing with both ghosted and regular components (which should not be possible),
+  // then return helper text as a failsafe.
+  if (atLeastOneGhostedComponent.value && atLeastOneNormalComponent.value) {
+    items.push({
+      label: "No options available for both ghosted and regular components",
+      disabled: true,
+    });
+    return items;
+  }
```
nickgerace, line 133: This should not be possible, but it is difficult to use this component without covering off on this case. We've decided that you likely won't be able to select both ghosted and regular components at the same time, but this failsafe should help once we get to multi-selection.

app/web/src/newhotness/ComponentContextMenu.vue
```diff
@@ -111,39 +112,70 @@ const actionByPrototype = computed(() => {
...
+  // If everything is ghosted, only add the ability to restore and return.
+  if (atLeastOneGhostedComponent.value) {
+    items.push({
+      label: "Restore",
+      shortcut: "R",
+      icon: "trash-restore",
+      onSelect: () => componentsRestore(components.value.map((c) => c.id)),
+    });
+    return items;
+  }
```
nickgerace, line 144: In a follow-up will we make it the full word "restore". For now, let's keep things consistent.

app/web/src/newhotness/Explore.vue
```diff
@@ -363,6 +357,11 @@ const componentList = computed(
   () => componentListRaw.data.value?.components ?? [],
 );
 
+// FIXME(nick): use the grid data rather than a "by ID" computed dictionary.
+const componentsById = computed(() =>
+  Object.fromEntries(componentList.value.map((c) => [c.id, c])),
+);
+
```
nickgerace: @jobelenus I don't like this due to data duplication. Is there a way for me to do this using the data that is already there?

app/web/src/newhotness/DeleteModal.vue
```diff
@@ -0,0 +1,129 @@
...
+
+defineExpose({ open, close });
+</script>
+
+<script lang="ts">
+export enum DeleteMode {
+  Delete,
+  Remove,
+}
+</script>
```
nickgerace, line 130: Need a separate script to expose this an a separate `.ts` file did not make sense.

app/web/src/newhotness/DeleteModal.vue
```diff
@@ -0,0 +1,129 @@
...
+import { onBeforeUnmount, onMounted, ref } from "vue";
+import { BifrostComponentInList } from "@/workers/types/entity_kind_types";
+import { getAssetIcon, getAssetColor } from "./util";
+
+const components = ref<BifrostComponentInList[]>([]);
+
+const borderStyle = (component: BifrostComponentInList) => {
+  const color = getAssetColor(component.schemaCategory);
+  return `border-color: ${color}`;
+};
```
nickgerace, line 84: Gets the color of the thin line for each component

app/web/src/newhotness/DeleteModal.vue
```diff
@@ -0,0 +1,130 @@
...
+  close();
+  emit("delete", mode.value);
+}
+
+onMounted(() => {
+  window.addEventListener("keydown", onKeyDown);
+});
+onBeforeUnmount(() => {
+  window.removeEventListener("keydown", onKeyDown);
+});
```
nickgerace, line 110: @wendybujalski taking your advice, but leaving a note for the next PR... I think we should replace listeners with emits at your recommendation!

app/web/src/newhotness/ComponentGridTile.vue
```diff
@@ -8,11 +8,18 @@
...
       )
     "
   >
     <header
-      :class="clsx('p-xs', themeClasses('bg-neutral-200', 'bg-neutral-800'))"
+      :class="
+        clsx(
+          'p-xs',
+          !component.toDelete &&
+            themeClasses('bg-neutral-200', 'bg-neutral-800'),
```
nickgerace, line 20: Remove the title bg fill if ghosted

app/web/src/newhotness/ComponentGridTile.vue
```diff
@@ -8,11 +8,18 @@
           'bg-shade-0 border-neutral-400',
           'bg-neutral-900 border-neutral-600',
         ),
+        component.toDelete && 'opacity-70',
```
nickgerace, line 11: Change opacity if ghosted

app/web/src/newhotness/ComponentDetails.vue
```diff
@@ -14,6 +14,21 @@
...
+    <VButton
+      v-tooltip="'Back (Esc)'"
+      class="border-0 mr-2em"
+      icon="arrow--left"
+      size="sm"
+      tone="shade"
+      variant="ghost"
+      @click="back"
+    />
+  </section>
```
nickgerace, line 31: We have a new, as of today, figma-ready review of what the page should do when the component is ghosted. For now, let's punt to the next PR since that will involve making everything read-only.

app/web/src/newhotness/Explore.vue
jobelenus: this doesn't strike me as duplication.. thats using the data that is already there in the client?

jobelenus (dismissed):

app/web/src/newhotness/Explore.vue
wendybujalski: Restructuring data into different organizational schemas in the frontend to meet our various needs is not unnecessary duplication.

app/web/src/newhotness/DeleteModal.vue
wendybujalski: If you made it a `type` instead of an `enum` you could export it in the setup script. Is there a reason you used an `enum` here instead of just a `type`?

app/web/src/newhotness/DeleteModal.vue
wendybujalski: Yeah, we want to avoid `addEventListener` ever being used in any individual Vue component, instead defining mitt emitters in the `emitters.ts` file in `logic_composables` that handle all events of the same type for the whole system.

app/web/src/newhotness/ComponentGridTile.vue
wendybujalski: Normally we want to avoid using opacity as much as possible, but this is a rare case where it is the correct choice to get what we want.

app/web/src/newhotness/DeleteModal.vue
nickgerace: I liked the `enum` more than the `type` I had :) I can change it back in post.

jobelenus (approved):

wendybujalski (approved):

PR #6379, opened 2025-06-10
feat(luminork): Add ability to get the status of a management function run

stack72:
Screenshot 2025-06-10 at 23 21 07


stack72:
> looks good! you might want to remove that stray dbg

tis gone in that last force push before your review landed :)

PR #6380, opened 2025-06-10
chore: improve performance of component create in luminork

adamhjk:
This PR makes some chanes that improve performance of component create specifically in luminork, but likely across the board. It does the following:

* Removes the extra DVU Graph calculation in component.rs, as it is no longer needed
* Updates the input_socket_attribute_values function so that it no  longer is linear with the number of components of a given schema  variant that are present in the graph. 
* Updates the 'connections' code in luminork to no longer error if  a connection is proposed but there are multiple components whose names match. It instead returns the first match. If the user wants to disambiguate, they can use the component id.

### DIFF
--- a/lib/dal/src/component.rs
+++ b/lib/dal/src/component.rs
@@ -610,8 +610,6 @@ impl Component {
         )
         .await?;
 
-        let mut dvu_roots = vec![];
-
         // Create attribute values for all socket corresponding to input and output sockets.
         for input_socket_id in
             InputSocket::list_ids_for_schema_variant(ctx, schema_variant_id).await?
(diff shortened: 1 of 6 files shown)


lib/dal/src/component.rs
```diff
@@ -1997,22 +2014,26 @@ impl Component {
...
-                    // that hasn't been cleaned up yet, we might detect an orphaned value.
-                    // We don't want to error, as the component_id is gone so we know it
-                    // isn't the one we're looking for.
-                    Err(AttributeValueError::OrphanedAttributeValue(_)) => None,
-                    Err(other_err) => Err(other_err)?,
-                }
-            {
-                if value_component_id == component_id {
-                    result.push(attribute_value_id)
+        let all_relevant_prop_ids = Prop::all_parent_prop_ids_from_prop_id(ctx, prop_id).await?;
```
zacharyhamm, line 2017: I like this refactor here

stack72 (approved):

PR #6383, opened 2025-06-11
chore: compress edda nats update messages to sdf

fnichol:
This change updates the encoding of `UpdateRequest` messages to use
Deflate compression before publishing the messages on NATS. In SDF in
the Bifrost web socket code, each NATS message body is checked for
compression encoding, and if present, will decompress the message body
before re-transmitting it as a web socket text message. As `IndexUpdate`
is not currently compressed, SDF will skip the decompression in this
case, using the headers as its hint.

Some other `NatsMultipler` and common NATS code dealing with subjects is
refactored in this change. Some highlights include:

- Rename the `DataCache` struct as `EddaUpdates`
- Turned `EddaUpdates` into an instance type to cache prefix subject and
  max_payload info across calls (added in `AppState`)
- Update the relevant NATS subjects to use `edda.updates.$wk_id.$kind`
  and use the optional NATS subject prefix for testing isolation
- Use helper functions to compute the subjects for "all updates for all
  workspaces" and "all updates for a workspace"
- Rename the `nats_std::headers` module to `nats_std::header`
- Wrap the Edda updates multipler client in an
  `EddaUpdatesMultiplexerClient` type which internally computes the NATS
  subject necessary for the receiver
- Centralize more common NATS extention headers used in the code base
  such as content type, message type, etc.
- Add helper functions for inserting and fetching common NATS header
  values into and out of a `HeaderMap`
- Add a `ContentEncoding` and `ContentType` helper to check for and
  insert common header values
- Use the DAL's compute executor for CPU-bound
  compression/decompression/serializing work on the Edda server and SDF
  server sides
- Add an Axum state extractor the the compute executor:
  `ComputeExecutor`
- Remove NATS-related subject code from `lib/si-frontend-mv-types-rs`
  crate
- Log an `error` when an update's payload size exceeds the `max_payload`
  value of the currently connected NATS server to help make it clearer
  if we are still in trouble here


jhelwig:
> * Rename the DataCache struct as EddaUpdates

👍🏻

lib/edda-server/src/updates.rs
```diff
@@ -0,0 +1,181 @@
...
+            subject::update_for(
+                self.subject_prefix.as_deref(),
+                index_update.meta.workspace_id.array_to_str(&mut id_buf),
+                index_update.kind(),
+            ),
+            index_update,
+            false,
+        )
+        .await
+    }
```
jhelwig, line 101: ~With how small the `IndexUpdate` message is (I think the payload is <1 kB and is pretty constant size), did it still end up being faster end-to-end for this one with the overhead of the compress/decompress?~ Nevermind, missed the `should_compress` flag, so this tiny message isn't being compressed.

fnichol: It ended up being fractions of an `ms` thankfully, although given the `IndexUpdate` messages are fairly constant in size, this is where I ended up opting for not compressing as the NATS headers I had set up already knew what to do on the consumption side. So this "conditionally compressing messages" kind of fell out and was no extra work 😄

fnichol: That internal `serialize_compress_publish()` method's last `bool` argument is whether or not to compress the JSON-serialized bytes

lib/edda-server/src/updates.rs
```diff
@@ -0,0 +1,181 @@
...
+                if should_compress {
+                    Ok::<_, EddaUpdatesError>((
+                        serialized.len(),
+                        deflate::compress_to_vec(&serialized, 6),
+                    ))
+                } else {
+                    Ok::<_, EddaUpdatesError>((serialized.len(), serialized))
+                }
+            })
+            .await??;
```
jhelwig: Worth only spawning if `should_compress` is true, instead of checking the flag inside the spawned task? Definitely get wanting to move the (blocking) serialize + compress over to the other executor, but in the case(s) where we're not concerned with compression, shouldn't the serialize be really fast since the data we're working with is pretty small to start with?

lib/sdf-v1-routes-ws/src/bifrost/proto.rs
```diff
@@ -164,6 +195,46 @@ impl BifrostStarted {
...
+            self.compute_executor
+                .spawn(async move {
+                    let compressed = msg.into_inner().payload;
+                    inflate::decompress_to_vec_zlib(&compressed)
+                })
+                .await?
+                .map_err(|decompress_err| Error::Decompress(decompress_err.to_string()))?
+        } else {
+            msg.into_inner().payload.into()
+        };
```
jhelwig, line 231: Why the option for both deflate & zlib? Looks like only deflate is used?

fnichol: Oh, I started wondering if there was a chance the frontend could decompress the payload directly, but that it might need the additinal zlib wrapper around the deflate body. Correct that at the moment only `deflate` is being used.

lib/edda-server/src/updates.rs
fnichol: I was a little curious if there would be an impact to spawn this work off the main Tokio runtime, but I can take this part out if we aren't compressing

PR #6385, opened 2025-06-11
feat(new ui): Set resourceId and run import funcs

vbustamante:
image

### DIFF
--- a/app/web/src/newhotness/AttributePanel.vue
+++ b/app/web/src/newhotness/AttributePanel.vue
@@ -1,5 +1,48 @@
 <template>
   <div v-if="root">
+    <div
+      v-if="showImportArea"
+      class="mt-xs grid grid-cols-2 pl-xs gap-2xs relative"
+    >
+      <div class="flex flex-row items-center gap-2xs">
+        <TruncateWithTooltip>{{
+          importing ? "Importing Attributes" : "Import"
+        }}</TruncateWithTooltip>
+      </div>
+      <input
+        ref="importInputRef"
+        v-model="resourceIdFormValue"
+        :class="
+          clsx(
+            'block w-full h-lg p-xs ml-auto text-sm border font-mono',
+            themeClasses(
+              'text-shade-100 bg-shade-0 border-neutral-400',
+              'text-shade-0 bg-shade-100 border-neutral-600',
+            ),
+          )
+        "
+        type="text"
+        placeholder="Resource Id"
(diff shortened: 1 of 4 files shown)


wendybujalski (dismissed):

app/web/src/newhotness/AttributePanel.vue
```diff
@@ -1,5 +1,48 @@
 <template>
   <div v-if="root">
+    <div
+      v-if="showImportArea"
+      class="mt-xs grid grid-cols-2 pl-xs gap-2xs relative"
+    >
+      <div class="flex flex-row items-center gap-2xs">
+        <TruncateWithTooltip>{{
+          importing ? "Importing Attributes" : "Import"
+        }}</TruncateWithTooltip>
```
wendybujalski, line 10: Generally we don't need a TruncateWithTooltip for text that isn't dynamic. Is there a reason why we're having one here?

app/web/src/newhotness/ComponentDetails.vue
```diff
@@ -89,11 +89,42 @@
...
+                size="sm"
+                tone="neutral"
+                label="Set attributes manually"
+                class="ml-2xs mr-xs font-normal"
+                @click.stop="
+                  () => {
+                    showResourceInput = false;
+                  }
+                "
+              />
```
wendybujalski, line 119: There's probably an elegant way to combine these two buttons into one with conditions for various properties.

wendybujalski (approved):

app/web/src/newhotness/ComponentDetails.vue
wendybujalski: Something like -

```
<VButton
  size="sm"
  tone="neutral"
  :label="showResourceInput ? 'Set attributes manually' : 'Import a Resource'"
  class="ml-2xs mr-xs font-normal"
  @click.stop="
    () => {
      showResourceInput = !showResourceInput;
    }
  "
/>
```

vbustamante: Yeah this makes sense. I'm still working on the area and will implement this on my next pr

app/web/src/newhotness/AttributePanel.vue
vbustamante: No real reason, just copied over from the attribute tree component. Does this not help responsiveness though? Like if someone makes the screen too narrow

PR #6389, opened 2025-06-11
feat(clover): add prop suggestions

sprutton1:
Take all primary identifiers that exist in a resource (meaning that spec created them) and match them in a very brute force way against as many props as possible. This should lead to suggestions that allow the ui to bubble the most likely candidate for our prop. Includes the asset func bits.

```typescript
    const VpcIdProp = new PropBuilder()
        .setName("VpcId")
        .setKind("string")
        .setHidden(false)
        .setWidget(new PropWidgetDefinitionBuilder()
            .setKind("text")
            .setCreateOnly()
            .build())
        .setValidationFormat(Joi.string().required())
        .setDocLink("https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ec2-subnet.html#cfn-ec2-subnet-vpcid")
        .setDocumentation("The ID of the VPC the subnet is in.\n If you update this property, you must also update the ``CidrBlock`` property.")
        .suggestSource({
            schema: "AWS::EC2::VPC",
            prop: "/resource_value/VpcId"
        })
        .build();
```
image

image


bin/clover/src/pipeline-steps/createSuggestionsAcrossAssets.ts
```diff
@@ -0,0 +1,106 @@
...
+      // stripping /root out
+      const propPath = "/" + prop.metadata.propPath.slice(1).join("/");
+      schemasToPrimaryIdents.set(specName, [
+        propPath,
+        nameVariants,
+      ]);
+    }, { skipTypeProps: true });
+  }
+
+  for (const spec of specs) {
```
jkeiser, line 78: This loop could use a comment saying what we're doing

bin/clover/src/spec/props.ts
```diff
@@ -610,6 +617,7 @@ export function createScalarProp(
     hidden: false,
     docLink: null,
     documentation: null,
+    ui_optionals: null,
```
jkeiser: This should be `uiOptionals` for parity with other ones, right?

PR #6398, opened 2025-06-12
Remove cross-MV dependency ordering when building MVs

jhelwig:
The intent of having strong references determine the build ordering of MVs was that MVs were going to have access to other MVs that had been built earlier within the same batch. This would have allowed them to know the checksum of the other MVs without having to build it for itself, which would duplicate the amount of work performed by edda.

Instead, we have been moving towards using weak references (only the kind & ID, with no checksum), which do not require either having access to MVs that may have been built within the same batch, or to re-built the MV every time it is going to be referenced from another MV.

Rather than limiting which MVs can be built at the same time in preparation for a direction we are moving away from, the only limit to the parallelism is now the `PARALLEL_BUILD_LIMIT`, and all MVs are assumed to be able to build without the output from any other MV being available (which is the current state of all MVs).

This also changes the only remaining `Reference` to be a `WeakReference`, removing the need to build every component's `IncomingConnections` MV an additional time to have the checksum in the ref.

### DIFF
--- a/lib/dal-materialized-views/src/incoming_connections_list.rs
+++ b/lib/dal-materialized-views/src/incoming_connections_list.rs
@@ -12,17 +12,13 @@ use telemetry::prelude::*;
 )]
 pub async fn assemble(ctx: DalContext) -> super::Result<IncomingConnectionsListMv> {
     let ctx = &ctx;
-    let component_ids = Component::list_ids(ctx).await?;
-    let mut component_connections = Vec::with_capacity(component_ids.len());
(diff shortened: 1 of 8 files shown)


jhelwig:
~This should be "ready", but I think we may have some questions around refs in general that might need to be answered before the `Reference -> WeakReference` part should ship.~ Should be good to go now.

jobelenus (approved):

britmyerss (approved):

PR #6399, opened 2025-06-12
feat(web): use a shared webworker to enable sqlite3 shared access pool

zacharyhamm:
Uses weblocks to ensure that only one tab's worker task has access to the sqlite3 database at a time. Proxies requests through a SharedWorker to the worker that holds the lock. As soon as a tab closes, all tabs will race for the lock, and the one that acquires it first will send a proxy of itself to the SharedWorker.

Requires all DBInterface methods that hit the database to be async (so they can wait for a tab remote to be ready)

### DIFF
--- a/app/web/src/store/realtime/heimdall.ts
+++ b/app/web/src/store/realtime/heimdall.ts
@@ -35,8 +35,22 @@ export const init = async (bearerToken: string, _queryClient: QueryClient) => {
     // eslint-disable-next-line no-console
     console.log("🌈 initializing bifrost...");
     const start = Date.now();
-    await db.setBearer(bearerToken);
-    await db.initBifrost();
+    await tabDb.setBearer(bearerToken);
+    const { port1, port2 } = new MessageChannel();
+
+    // This message fires when the lock has been acquired for this tab
+    port1.onmessage = () => {
+      tabDb.addListenerBustCache(Comlink.proxy(bustTanStackCache));
+      tabDb.addListenerInFlight(Comlink.proxy(inFlight));
+      tabDb.addListenerReturned(Comlink.proxy(returned));
+      tabDb.addListenerLobbyExit(Comlink.proxy(lobbyExit));
+
+      db.setRemote(Comlink.proxy(tabDb));
+    };
+
+    // We are deliberately not awaiting this promise, since it blocks forever on the tabs that do not get the lock
+    tabDb.initBifrost(Comlink.proxy(port2));
+
     const end = Date.now();
     token = bearerToken;
     queryClient = _queryClient;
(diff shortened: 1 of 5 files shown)


app/web/src/store/realtime/heimdall.ts
```diff
@@ -35,8 +35,22 @@ export const init = async (bearerToken: string, _queryClient: QueryClient) => {
...
+    port1.onmessage = () => {
+      tabDb.addListenerBustCache(Comlink.proxy(bustTanStackCache));
+      tabDb.addListenerInFlight(Comlink.proxy(inFlight));
+      tabDb.addListenerReturned(Comlink.proxy(returned));
+      tabDb.addListenerLobbyExit(Comlink.proxy(lobbyExit));
+
+      db.setRemote(Comlink.proxy(tabDb));
+    };
+
+    // We are deliberately not awaiting this promise, since it blocks forever on the tabs that do not get the lock
```
jobelenus, line 54: ❤️ this comment

app/web/src/workers/webworker.ts
```diff
@@ -2595,16 +2596,46 @@ const getMany = async (
...
+let returnedFn: RainbowFn;
 let lobbyExitFn: LobbyExitFn;
 
 let abortController: AbortController | undefined;
 
 const dbInterface: DBInterface = {
+  async broadcastMessage(_message) {
+    debug("only sharedworker should be sent broadcasts directly");
+  },
+  async receiveBroadcast(message: BroadcastMessage) {
```
jobelenus, line 2612: OK, I see the pattern...

1. The running tabDB web worker calls the respective heimdall fn, just like it does today
2. Only with a broadcast flag.
3. Heimdall tells the web worker to tell each tabDB to the operation to perform
4. Each running tabDB calls their respective main thread heimdall, without the broadcast flag

In general, the shared worker does very little coordination... most of the operations between the web worker and heimdall are the same...

app/web/src/workers/shared_webworker.ts
```diff
@@ -0,0 +1,330 @@
...
+    return j;
+  },
+
+  async handlePatchMessage(data: PatchBatch, span?: Span): Promise<void> {
+    await withRemote(
+      async (remote) => await remote.handlePatchMessage(data, span),
+    );
+  },
+
+  async handleHammer(msg: AtomMessage, span?: Span): Promise<void> {
```
jobelenus, line 290: im not sure the shared worker needs these three fns... since it doesnt even have a web socket where it would get the data that these functions operate on

zacharyhamm: ah makes sense. I just implemented everything. we should split up the interfaces

app/web/src/workers/webworker.ts
zacharyhamm: Right. Over the weekend I realized I should refactor this by making those callbacks all just listen to a broadcast channel, so they don't have to trigger their own broadcasting. Not necessary yet thoug

PR #6400, opened 2025-06-12
Fast Lists

jobelenus:
# Goal
- Limit the number of times data has to cross over the thread boundary
- When it does have to pass, send it via the most efficient means possible
    - currently a blob or string, not JSON, future: transferable object

## What is in the PR
- Move the outgoing count of computed out to the main thread via tan stack, and reference it in the tile component
- Component names is the biggest cause of References in MVs. Make a global lookup instead
- Bringing back the "do it all in JSON" for "get the first order references in a list"
- Don't cache bust lists on patches
- Instead push up individual list item changes
- List items have no references
- Only cache bust list on a hammer (when we can't patch)

### Timings
250 components loading in 1 second.

When I start _on the map_, it performs 2 big queries (note: this behavior is fixed on `main` but leaving this here for how "multiple things over the interface at once" results in slower performance):
- componentList of 250 in ~1.3 seconds
- IncomingConnections list of 250 in ~3.2 seconds

If I start on the grid, and navigate to the map, so it only has one query to perform:
- IncomingConnections list of 250 in ~1.5 seconds

### Testing Steps
1. Initial load, see the lists
2. Making a new change set, still see lists
3. Make a new component, hit back, see component in the list update without a LIST cache bust
4. Change the component name, hit back, see the list update without a cache bust
6. Erase a component, see the list update without a cache bust
7. Make a connection between components, see the incoming & outgoing numbers change on the list
8. Delete component that has a connection and watch numbers on the list change
9. Add a view, see the list change
10. Update a view name, see the list change
11. Delete a view, see the list change

Also—use multiple tabs!

### DIFF
--- a/app/web/src/newhotness/Workspace.vue
+++ b/app/web/src/newhotness/Workspace.vue
@@ -89,7 +89,7 @@ import {
   watch,
 } from "vue";
 import { Icon } from "@si/vue-lib/design-system";
-import { useQueryClient } from "@tanstack/vue-query";
+import { useQuery, useQueryClient } from "@tanstack/vue-query";
 import NavbarPanelLeft from "@/components/layout/navbar/NavbarPanelLeft.vue";
 import NavbarPanelRight from "@/components/layout/navbar/NavbarPanelRight.vue";
 import NavbarButton from "@/components/layout/navbar/NavbarButton.vue";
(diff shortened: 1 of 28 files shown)


app/web/src/newhotness/Workspace.vue
```diff
@@ -129,13 +136,85 @@ const realtimeStore = useRealtimeStore();
 const workspacePk = computed(() => props.workspacePk);
 const changeSetId = computed(() => props.changeSetId);
 
+const enabled = ref(false);
```
zacharyhamm: Our future selves will probably appreciate a more descriptive name for this (what is being enabled by it?)

app/web/src/workers/types/entity_kind_types.ts
```diff
@@ -26,13 +27,12 @@ export enum EntityKind {
   SchemaMembers = "SchemaMembers",
   ActionViewList = "ActionViewList",
   ActionPrototypeViewList = "ActionPrototypeViewList",
-  SecretList = "SecretList",
-  SecretDefinitionList = "SecretDefinitionList",
-  Secret = "Secret",
```
zacharyhamm, line 31: Curious why these went away

adamhjk:
We should absolutely ship this, once you're good with Zack's comments.

app/web/src/workers/types/entity_kind_types.ts
jobelenus: We just actually don't use the data in the front end in that way. We had removed it in the MVs earlier and just hadn't gotten rid of these defs

PR #6402, opened 2025-06-12
Fix: during mounting the ? returns false, which is actually undesirable because it fires the map query, enforce the default/undefined state better

jobelenus:
The `Map`'s incoming connection list query was firing when we didn't want it to

## How was it tested?
Refresh grid page, see no incoming connections list query

### DIFF
--- a/app/web/src/newhotness/Explore.vue
+++ b/app/web/src/newhotness/Explore.vue
@@ -299,7 +299,7 @@ const urlGridOrMap = computed(() => {
   if (keys.includes("map")) return "map";
   return "grid";
 });
-const showGrid = computed(() => groupRef.value?.isA);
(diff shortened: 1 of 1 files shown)


wendybujalski (approved):

app/web/src/newhotness/Explore.vue
```diff
@@ -299,7 +299,7 @@ const urlGridOrMap = computed(() => {
   if (keys.includes("map")) return "map";
   return "grid";
 });
-const showGrid = computed(() => groupRef.value?.isA);
```
wendybujalski, line 302: It can be so easy to overlook optional chaining as a source of unexpected data!

wendybujalski (approved):

wendybujalski (approved):

PR #6403, opened 2025-06-12
fix(web): Ensure we treat secrets as secret input in the new UX

stack72:
Screenshot 2025-06-12 at 22 52 10

### DIFF
--- a/app/web/src/newhotness/layout_components/ComponentSecretAttribute.vue
+++ b/app/web/src/newhotness/layout_components/ComponentSecretAttribute.vue
@@ -59,7 +59,7 @@
                           'border-destructive-500 z-100',
                       )
                     "
-                    type="text"
+                    :type="getFormFieldType(fieldname)"
                     :value="field.state.value"
                     tabindex="0"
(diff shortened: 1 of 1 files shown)


app/web/src/newhotness/layout_components/ComponentSecretAttribute.vue
```diff
@@ -296,6 +296,13 @@ const getPlaceholder = (fieldname: string) => {
   } else return "empty";
 };
 
+const getFormFieldType = (fieldname: string) => {
+  if (fieldname !== "Name" && fieldname !== "Description") {
+    return "password";
```
wendybujalski, line 301: Would probably be good to add a visible toggle to the fields.

wendybujalski (approved):

app/web/src/newhotness/layout_components/ComponentSecretAttribute.vue
stack72: I am going to follow on with it! I definitely want to add that

PR #6404, opened 2025-06-12
feat(web): Virtualize the prop connections select list

jkeiser:
This speeds up the prop connections dropdown list initialization and scrolling. Because there are so many possible connections (10-20x the number of components you have), the dropdown currently creates HTML elements for all of them, which overwhelms the browser's layout engine and takes forever. This uses virtualization to only create HTML elements for the items that will be displayed.

1. Makes the size of the list constant so that the virtualizer knows how big it is. There will basically always be more than 5 suggestions as long as there is at least one component.
2. If there are no possible connections (such as when there is only one component), we will not show possible connections at all. We added this because the previous code used `max-height` instead of a fixed `height`, which allowed it to be small when there were no suggestions. Since we modified it to a fixed height, we needed to handle this case a different way. This will do a little better, not showing any prop connection titles or sections at all when there is nothing to select.

## How was it tested?

- [X] Integration tests pass
- [X] Manual test: can scroll through all connections
- [X] Manual test: can select connections past the first page
- [X] Manual test: can still select values (not just connections)
- [X] Manual test: does not show connections when there is only one component

### DIFF
--- a/app/web/src/newhotness/layout_components/AttributeInput.vue
+++ b/app/web/src/newhotness/layout_components/AttributeInput.vue
@@ -338,7 +338,7 @@
           </div>
 
           <!-- select potential connection area -->
-          <template v-if="!isMap">
+          <template v-if="!isMap && filteredConnections.length > 0">
             <div
               v-if="!isSecret"
               :class="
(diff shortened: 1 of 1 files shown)


app/web/src/newhotness/layout_components/AttributeInput.vue
```diff
@@ -338,7 +338,7 @@
           </div>
 
           <!-- select potential connection area -->
-          <template v-if="!isMap">
+          <template v-if="!isMap && filteredConnections.length > 0">
```
wendybujalski, line 341: Do we have an empty state for if there are no connections?

app/web/src/newhotness/layout_components/AttributeInput.vue
```diff
@@ -350,82 +350,108 @@
...
+                      'px-xs py-2xs h-[30px]',
+                      isConnectionSelected(virtualItem.index) && [
+                        'input-selected-item',
+                        themeClasses('bg-action-200', 'bg-action-900'),
+                      ],
+                      themeClasses(
+                        'hover:border-action-500',
+                        'hover:border-action-300',
+                      ),
+                      false && themeClasses('bg-action-200', 'bg-action-900'),
```
wendybujalski: This line looks like dead code? Left in by mistake?

app/web/src/newhotness/layout_components/AttributeInput.vue
```diff
@@ -350,82 +350,108 @@
...
                     )
                   "
+                  :style="{
+                    position: 'absolute',
+                    top: 0,
+                    left: 0,
+                    width: '100%',
```
wendybujalski: These styles could be Tailwind classes. Best to avoid using style whenever possible.

app/web/src/newhotness/layout_components/AttributeInput.vue
```diff
@@ -1014,6 +1050,19 @@ watch(
...
+const filteredConnectionsList = useVirtualizer(computed(() => {
+  return {
+    count: filteredConnections.value.length,
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    getScrollElement: () => filteredConnectionsListRef.value!,
+    // getItemKey: (index: number) => filteredConnections.value[index]?.attributeValueId ?? "<unknown>",
+    estimateSize: () => 30,
+    // overscan: 3,
+  };
+}));
```
wendybujalski: Why comment out the explicit overscan amount? Also I'd prefer to see the computed stored in its own variable. But regardless, this does look like it should work!

app/web/src/newhotness/layout_components/AttributeInput.vue
```diff
@@ -748,6 +775,15 @@ const emit = defineEmits<{
...
 const defaultSelectedIndex = () => (props.isSecret ? 1 : 0);
+/**
+ * The index of the selected option or connection.
+ *
+ * This is an index into an imagined concatenated list containing both filteredOptions and filteredConnections.
+ *
+ * - If this is 0, nothing is selected.
+ * - If this is > 0 and <= filteredOptions.length, it references filteredOptions[selectedIndex - 1].
+ * - If this is > filteredOptions.length, it references filteredConnetions[selectedIndex - 1 - filteredOptions.length].
+ */
```
wendybujalski, line 798: A very helpful comment!

wendybujalski (changes requested):
Minor things to fix, then it'll be ready to go. Great work!

app/web/src/newhotness/layout_components/AttributeInput.vue
jkeiser: oh, you're right, it should be looking at whether there are any *potential* connections, not filtered!

app/web/src/newhotness/layout_components/AttributeInput.vue
jkeiser: yeah, it just got indented here. I'll remove it, it's clearly a "comment"

app/web/src/newhotness/layout_components/AttributeInput.vue
jkeiser: I'll go look them up, for sure :)

app/web/src/newhotness/layout_components/AttributeInput.vue
jkeiser: ohhh, yeah, I'll add overscan back in. I didn't make computed its own variable because it wasn't referenced elsewhere, but I'm happy to put it in its own variable if that's the direction we want to be consistent!

app/web/src/newhotness/layout_components/AttributeInput.vue
wendybujalski: I believe filteredConnections is the list of potential connections, as filtered by the fuzzy search. I mean what happens if the list is empty, either because there are no possible connections or the fuzzy search gives no results?

app/web/src/newhotness/layout_components/AttributeInput.vue
wendybujalski: Let me know if you have any trouble with that, happy to help. And those last two styles with the variables in them can't be Tailwind classes of course!

app/web/src/newhotness/layout_components/AttributeInput.vue
jkeiser: yeah, that's what I mean--we would not show the potential connection area at all. In that case I think the area should at least still show up, empty, so you at least know it's possible. I could imagine wanting to show "0 matches" or something as well.

wendybujalski: Not worth blocking on though, I can tidy up in another pass! Fix the linting and I'll hit approve.

jkeiser: We could definitely also set the size of the dialog based on the number of matching entries, though! It's just the max-height was set to 10rem, and the height of the entries is 30px, but rem is variable based on zoom and font size ... I'm not sure if I can tell CSS that height=min(10rem, (filteredConnections.length%5)*30px).

app/web/src/newhotness/layout_components/AttributeInput.vue
jkeiser: (Removed)

app/web/src/newhotness/layout_components/AttributeInput.vue
jkeiser: Modified all the styles I could to be tailwind classes. The only ones that didn't seem to work were `h-[${filteredConnectionsList.getTotalSize()}px]px` on the parent div, and `translate-y-[$virtualItem.start}px` on this one.
```ts
clsx(`absolute top-0 left-0 w-full h-[${virtualItem.size}px]`)
```

PR #6410, opened 2025-06-13
Stream patches to the front end from edda; try to reduce nats write load; priority queue for MV builds

jhelwig:
# Initial sketch of streaming patches to the front end from edda

The idea with streaming patches to the front end is that it will be able to start using the new MVs as soon as they are generated, before _all_ of the data is ready, and there will be a message with the MvIndex patch (message not created/sent yet) that will let the front end know that the stream of patches has finished, and which objects it should have by the end (fetching any it missed the patch for).

Right now, these streaming patches are sent in addition to the existing patch batches, but once we start using the streamed patches in the front end, we'll want to stop creating & sending the batched patches as it will be purely redundant from the streamed patches.

We will also want to figure out if the existing `IndexUpdate` message should be augmented to include both the from-checksum & patch for the MvIndex, or if we want an entirely new message to handle this.

# Try to reduce nats load by only inserting FrontEndObjects if they're not already in the KV store

This doesn't reduce the amount of information sent over the wrire to nats on an insert, but nats should be able to short-circuit and avoid doing any write in the case where the object has already been generated for any reason.

With the interface provided by the nats KV store, our options for only inserting if the key doesn't already exist looked to be:

* Send the (potentially) new thing over the wire and watching for a failure (`create` behaves this way).
* Try to get first (and receive the full value).
* Iterate over all of the keys to see if the key is already there.

Trying the insert using `create` and ignoring the "already exists" error kind seems to be the least-bad of the options.

# Use a priority queue to determine order of MV builds in edda

When streaming patches out to the front end, it would be helpful for the front end's reactivity to be able to send lists early so it can display skeletons for individual items it hasn't gotten the patch information for yet.

Right now, there are two "priorities":
  * Lists
  * Details (everything else)

By default an MV will use the "Details" priority. If the `build_priority` annotation is provided for the MV struct definition, then it will use whatever is specified in the annotation.

This also re-structures how MV build tasks are spawned so that we can get everything queued before letting the priority queue handle things from there, to ensure the higher-priority items are always built first. Rather than having to explicitly check if the Change's `entity_kind` is the same as the MV's `trigger_entity`, this check has been abstracted away into the `MaterializedViewInventoryItem` and is checked during the initial queue population. This means that whenever we go to spawn an MV build task, we already know that the change is relevant for that specific MV kind.

# Make streamed patches from edda a config setting instead of always on

Rather than always sending the patches for the front end both as individual patches when each is built, and as a single batch after all have been built, there is now a `SI_EDDA__STREAMING_PATCHES` setting to toggle between the two behaviors. The default behavior is to send the patches as a single batch, making the newer streaming behavior opt-in for development & testing.

### DIFF
--- a/lib/edda-server/src/change_set_processor_task/materialized_view.rs
+++ b/lib/edda-server/src/change_set_processor_task/materialized_view.rs
@@ -1,7 +1,7 @@
 use std::{
     collections::{
+        BinaryHeap,
         HashSet,
-        VecDeque,
     },
     time::Duration,
 };
@@ -28,10 +28,10 @@ use frigg::{
 };
 use si_events::{
     WorkspaceSnapshotAddress,
+    materialized_view::BuildPriority,
     workspace_snapshot::Change,
 };
 use si_frontend_mv_types::{
-    MaterializedView,
     action::{
         ActionPrototypeViewList as ActionPrototypeViewListMv,
         ActionViewList as ActionViewListMv,
@@ -49,7 +49,10 @@ use si_frontend_mv_types::{
         IncomingConnectionsList as IncomingConnectionsListMv,
     },
     index::MvIndex,
-    materialized_view::materialized_view_definitions_checksum,
+    materialized_view::{
+        MaterializedViewInventoryItem,
+        materialized_view_definitions_checksum,
+    },
     object::{
         FrontendObject,
         patch::{
@@ -79,7 +82,6 @@ use si_id::{
     WorkspacePk,
 };
 use si_layer_cache::LayerDbError;
-use strum::IntoEnumIterator;
 use telemetry::prelude::*;
 use telemetry_utils::metric;
 use thiserror::Error;
@@ -553,24 +555,17 @@ async fn build_mv_inner(
     let mut frontend_objects = Vec::new();
     let mut patches = Vec::new();
     let mut build_tasks = JoinSet::new();
-    let mut queued_mv_builds = VecDeque::new();
+    let mut queued_mv_builds = BinaryHeap::new();
 
-    // We'll spawn up to the first `parallel_build_limit` build tasks, and queue the rest to be spawned
-    // as other build tasks are completed.
+    // Queue everything so we can let the priority queue determine the order everything is built.
     for &change in changes {
-        for mv_kind in ReferenceKind::iter() {
-            if let Some(queued_build) = spawn_build_mv_task_for_change_and_mv_kind(
-                &mut build_tasks,
-                ctx,
-                frigg,
-                parallel_build_limit,
-                change,
-                mv_kind,
-                workspace_pk,
-            )
-            .await?
-            {
-                queued_mv_builds.push_back(queued_build);
+        for mv_inventory_item in ::inventory::iter::<MaterializedViewInventoryItem>() {
+            if mv_inventory_item.should_build_for_change(change) {
+                queued_mv_builds.push(QueuedBuildMvTask {
+                    change,
+                    mv_kind: mv_inventory_item.kind(),
+                    priority: mv_inventory_item.build_priority(),
+                });
             }
         }
     }
@@ -591,26 +586,23 @@ async fn build_mv_inner(
 
         // Spawn as many of the queued build tasks as we can, up to the concurrency limit.
         while !queued_mv_builds.is_empty() && build_tasks.len() < parallel_build_limit {
-            let Some(QueuedBuildMvTask { change, mv_kind }) = queued_mv_builds.pop_front() else {
+            let Some(QueuedBuildMvTask {
+                change, mv_kind, ..
+            }) = queued_mv_builds.pop()
+            else {
                 // This _really_ shouldn't ever return `None` as we just checked that
                 // `queued_mv_builds` is not empty.
                 break;
             };
-            // This _really_ shouldn't ever return `Some`, but better to be paranoid than to
-            // forget about pending work.
-            if let Some(queued_build) = spawn_build_mv_task_for_change_and_mv_kind(
+            spawn_build_mv_task_for_change_and_mv_kind(
                 &mut build_tasks,
                 ctx,
                 frigg,
-                parallel_build_limit,
                 change,
                 mv_kind,
                 workspace_pk,
             )
             .await?
-            {
-                queued_mv_builds.push_back(queued_build);
-            }
         }
 
         if let Some(join_result) = build_tasks.join_next().await {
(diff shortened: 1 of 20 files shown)


jhelwig:
An example of one of the streamed patches on the websocket:

```json
{"workspaceId":"01GWAJW3PQ0S1DVBN76XH9DBHV","changeSetId":"01JXN8KGNG2CK01ETFH33VBNJE","kind":"Component","id":"01JXN8MRMF6AZ50CTMXPKT741Y","fromChecksum":"616b523fb65078f0063ad308e6e7cf1b","toChecksum":"214531f044b7b66cc11976eb8fea6847","patch":[{"op":"replace","path":"/resourceDiff/current","value":"{\n  \"si\": {\n    \"name\": \"StackFleetAssociation\",\n    \"type\": \"component\",\n    \"color\": \"#FF9900\"\n  },\n  \"domain\": {\n    \"extra\": {\n      \"Region\": \"us-east-1\",\n      \"AwsResourceType\": \"AWS::AppStream::StackFleetAssociation\",\n      \"PropUsageMap\": \"{\\\"createOnly\\\":[],\\\"updatable\\\":[\\\"StackName\\\",\\\"FleetName\\\"],\\\"secrets\\\":[]}\"\n    }\n  }\n}"},{"op":"replace","path":"/resourceDiff/diff","value":"+{\n+  \"si\": {\n+    \"name\": \"StackFleetAssociation\",\n+    \"type\": \"component\",\n+    \"color\": \"#FF9900\"\n+  },\n+  \"domain\": {\n+    \"extra\": {\n+      \"Region\": \"us-east-1\",\n+      \"AwsResourceType\": \"AWS::AppStream::StackFleetAssociation\",\n+      \"PropUsageMap\": \"{\\\"createOnly\\\":[],\\\"updatable\\\":[\\\"StackName\\\",\\\"FleetName\\\"],\\\"secrets\\\":[]}\"\n+    }\n+  }\n+}"}],"messageKind":"StreamingPatch"}
```

Human readable-ish version:

```json
{
  "workspaceId": "01GWAJW3PQ0S1DVBN76XH9DBHV",
  "changeSetId": "01JXN8KGNG2CK01ETFH33VBNJE",
  "kind": "Component",
  "id": "01JXN8MRMF6AZ50CTMXPKT741Y",
  "fromChecksum": "616b523fb65078f0063ad308e6e7cf1b",
  "toChecksum": "214531f044b7b66cc11976eb8fea6847",
  "patch": [
    {
      "op": "replace",
      "path": "/resourceDiff/current",
      "value": "{\n  \"si\": {\n    \"name\": \"StackFleetAssociation\",\n    \"type\": \"component\",\n    \"color\": \"#FF9900\"\n  },\n  \"domain\": {\n    \"extra\": {\n      \"Region\": \"us-east-1\",\n      \"AwsResourceType\": \"AWS::AppStream::StackFleetAssociation\",\n      \"PropUsageMap\": \"{\\\"createOnly\\\":[],\\\"updatable\\\":[\\\"StackName\\\",\\\"FleetName\\\"],\\\"secrets\\\":[]}\"\n    }\n  }\n}"
    },
    {
      "op": "replace",
      "path": "/resourceDiff/diff",
      "value": "+{\n+  \"si\": {\n+    \"name\": \"StackFleetAssociation\",\n+    \"type\": \"component\",\n+    \"color\": \"#FF9900\"\n+  },\n+  \"domain\": {\n+    \"extra\": {\n+      \"Region\": \"us-east-1\",\n+      \"AwsResourceType\": \"AWS::AppStream::StackFleetAssociation\",\n+      \"PropUsageMap\": \"{\\\"createOnly\\\":[],\\\"updatable\\\":[\\\"StackName\\\",\\\"FleetName\\\"],\\\"secrets\\\":[]}\"\n+    }\n+  }\n+}"
    }
  ],
  "messageKind": "StreamingPatch"
}

```

jobelenus:
@jhelwig thank you for the example payload. Is that `id` field the "placeholder" (e.g. that value will be in the "end" message for the index update with the resulting index checksum?)

jhelwig:
The `id` in the sample payload is the ID of the atom/MV. At the time the streaming patch is sent, we don't yet know what the index checksum will be.

jhelwig:
The "placeholder"/"staging" index checksum should probably be the change set id since we mostly treat the checksum as an opaque string, and it will never overlap with a "real" checksum.

lib/edda-server/src/change_set_processor_task/materialized_view.rs
```diff
@@ -522,13 +532,30 @@ async fn build_mv_inner(
 
             match execution_result {
                 Ok((maybe_patch, maybe_frontend_object)) => {
+                    // We need to make sure the frontend object is inserted into the store first so that
+                    // a client can directly fetch it without racing against the object's insertion if the
+                    // client does not already have the base object to apply the streaming patch to.
```
fnichol, line 616: 👍🏽

fnichol (approved):
I like your approach to send both the incremental and the batch as we can migrate/experiment with both approaches.

Looks good!

lib/edda-server/src/change_set_processor_task/materialized_view.rs
```diff
@@ -754,356 +763,249 @@ where
...
+
+impl Ord for QueuedBuildMvTask {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        match self.priority.cmp(&other.priority) {
+            std::cmp::Ordering::Equal => {
+                // Within the same priority, items will be ordered by the triggering
+                // entity ID.
+                self.change.entity_id.cmp(&other.change.entity_id)
+            }
+            ord => ord,
```
jhelwig: The sorting here is actually backwards. This should have "higher priority" items as "larger".

lib/edda-server/src/change_set_processor_task/materialized_view.rs
```diff
@@ -542,31 +547,25 @@ async fn build_mv_inner(
...
+    edda_updates: &EddaUpdates,
     workspace_pk: si_id::WorkspacePk,
     change_set_id: ChangeSetId,
     changes: &[Change],
 ) -> Result<BuildMvInnerReturn, MaterializedViewError> {
     let mut frontend_objects = Vec::new();
     let mut patches = Vec::new();
     let mut build_tasks = JoinSet::new();
-    let mut queued_mv_builds = VecDeque::new();
+    let mut queued_mv_builds = BinaryHeap::new();
```
fnichol, line 558: Love the idea of the `BinaryHeap`

lib/edda-server/src/config.rs
```diff
@@ -223,6 +235,10 @@ fn default_parallel_build_limit() -> usize {
     DEFAULT_PARALLEL_BUILD_LIMIT
 }
 
+fn default_streaming_patches() -> bool {
+    false
```
fnichol, line 239: 👍🏽

lib/edda-server/src/updates.rs
```diff
@@ -65,6 +72,10 @@ impl EddaUpdates {
         fields()
     )]
     pub(crate) async fn publish_patch_batch(&self, patch_batch: PatchBatch) -> Result<()> {
+        if self.streaming_patches {
+            return Ok(());
+        }
```
fnichol, line 77: Well that was nice and simple

PR #6411, opened 2025-06-13
fix virtualization of component list

zacharyhamm:
WIP DONT MERGE.

@wendybujalski this is the fix for the virtualization list (although it means no grid)

### DIFF
--- a/app/web/src/newhotness/Explore.vue
+++ b/app/web/src/newhotness/Explore.vue
@@ -83,30 +83,46 @@
         </InstructiveVormInput>
       </div>
       <template v-if="showGrid">
-        <div
-          ref="scrollRef"
-          class="scrollable tilegrid grow"
-          @scroll="onScroll"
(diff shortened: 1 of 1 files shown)


app/web/src/newhotness/Explore.vue
```diff
@@ -83,30 +83,46 @@
...
+              v-for="(component, index) in componentVirtualItemsList"
+              :key="filteredComponents[component.index]!.id"
+              :style="{
+                position: 'absolute',
+                top: 0,
+                left: 0,
+                width: '100%',
+                height: `50px`,
+                transform: `translateY(${component.start}px)`,
+              }"
```
zacharyhamm, line 105: These styles are what matter, although also maybe the overflow-anchor: none

zacharyhamm: height is wrong here

PR #6414, opened 2025-06-13
Add ability to edit and delete views

nickgerace:
## Description

This change adds the ability to edit and delete views with protections against non-empty views. The existing "AddViewModal" was used as a frame of reference despite the new approval modal being closer to the desired designs.

Some complications in here included adding the "settings-edit" button, which behaves differently than "endLinkTo" in "DropdownMenuItem" as well as handling "canDeleteView" and query timing. Leveraging the "beforeOptions" and "afterOptions" for the button as well as allowing options to expand past the max width were part of getting here. This one was painful to get close, but we got there.

## View Switcher Example

Screenshot 2025-06-13 at 6 50 16 PM

## Edit View Modal Example with Deletion

Screenshot 2025-06-13 at 6 50 33 PM

## Edit View Modal Example without Deletion

Screenshot 2025-06-13 at 6 50 27 PM

### DIFF
--- a/app/web/src/newhotness/EditViewModal.vue
+++ b/app/web/src/newhotness/EditViewModal.vue
@@ -1,0 +1,126 @@
+<template>
+  <!-- NOTE: the Modal CSS for height in "max" doesn't work as we might expect -->
+  <Modal
+    ref="modalRef"
+    size="lg"
+    title="Edit View"
+    type="save"
+    saveLabel="Done"
+    @save="() => nameForm.handleSubmit()"
+  >
+    <label class="flex flex-row items-center relative">
+      <span>View Name</span>
+      <nameForm.Field name="name">
+        <template #default="{ field }">
+          <input
+            :class="
+              clsx(
+                'block w-72 ml-auto border',
+                themeClasses(
+                  'text-black bg-white border-neutral-600 disabled:bg-neutral-100',
+                  'text-white bg-black border-neutral-400 disabled:bg-neutral-900',
+                ),
+              )
+            "
+            :value="field.state.value"
+            type="text"
+            :disabled="wForm.bifrosting.value"
+            @input="
+              (e) => field.handleChange((e.target as HTMLInputElement).value)
+            "
+          />
+        </template>
+      </nameForm.Field>
+      <Icon
+        v-if="wForm.bifrosting.value"
+        class="absolute right-2xs"
+        name="loader"
+        size="sm"
+        tone="action"
+      />
+    </label>
+    <template #leftButton>
+      <div
+        v-tooltip="
+          canDeleteView
+            ? undefined
+            : 'Views containing one or more components cannot be deleted. To delete a view, first remove all components from it.'
+        "
+      >
+        <VButton
+          label="Delete View"
+          :disabled="!canDeleteView"
+          :tone="canDeleteView ? 'destructive' : 'neutral'"
+          @click="deleteView"
+        />
+      </div>
+    </template>
+  </Modal>
+</template>
+
+<script setup lang="ts">
+import { VButton, Modal, Icon, themeClasses } from "@si/vue-lib/design-system";
+import { computed, ref } from "vue";
+import clsx from "clsx";
+import { ViewId } from "@/api/sdf/dal/views";
+import { useApi, routes } from "./api_composables";
+import { useWatchedForm } from "./logic_composables/watched_form";
+
+const viewId = ref<ViewId>("");
+const canDeleteView = ref<boolean>(false);
+
+const modalRef = ref<InstanceType<typeof Modal>>();
+const deleteViewApi = useApi();
+const updateViewApi = useApi();
+
+const deleteView = async () => {
+  const call = deleteViewApi.endpoint<{ id: string }>(routes.DeleteView, {
+    viewId: viewId.value,
+  });
+  const { req } = await call.delete({});
+  if (deleteViewApi.ok(req)) {
+    close();
+    emit("deleted");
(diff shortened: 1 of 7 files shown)


app/web/src/newhotness/Explore.vue
```diff
@@ -31,14 +31,29 @@
...
+              @select="() => (selectedView = '')"
+            />
+          </template>
           <template #afterOptions>
             <DropdownMenuItem
+              class="border-t"
               label="Add a View"
               icon="plus"
-              @select="openViewModal"
+              disableCheckable
```
nickgerace, line 55: The bottom option cannot be checked and has a border above it. `disableCheckable` also ensures it isn't left-padded.

app/web/src/newhotness/Explore.vue
```diff
@@ -31,14 +31,29 @@
...
           @update:modelValue="(val) => (selectedView = val)"
         >
+          <template #beforeOptions>
+            <DropdownMenuItem
+              label="All Views"
+              value="''"
+              checkable
+              :checked="selectedView === ''"
+              @select="() => (selectedView = '')"
+            />
```
nickgerace, line 48: Originally included in the list, we do not want a "settingsEdit" option, so let's place it in the "beforeOptions". The key is that "All Views" is an empty string.

app/web/src/newhotness/Explore.vue
```diff
@@ -31,14 +31,29 @@
           class="rounded min-w-[128px]"
           :options="viewListOptions"
           :modelValue="selectedView"
+          minWidthToAnchor
```
nickgerace, line 34: This is subtle but allows the options to be longer

app/web/src/newhotness/EditViewModal.vue
```diff
@@ -0,0 +1,126 @@
...
+  },
+});
+
+const emit = defineEmits<{
+  (e: "deleted"): void;
+}>();
+
+const open = (openViewId: string, openCanDeleteView: boolean) => {
+  viewId.value = openViewId;
+  canDeleteView.value = openCanDeleteView;
```
nickgerace, line 117: I want this to be computed based on the provided `viewId` in the future, but due to performance and query timings, we need the caller to compute it at the moment

app/web/src/newhotness/Explore.vue
```diff
@@ -180,11 +195,18 @@
...
+    <!-- MODALS -->
     <AddComponentModal ref="addComponentModalRef" />
     <AddViewModal
       ref="addViewModalRef"
       :views="viewListQuery.data.value?.views"
     />
+    <!-- For the edit view modals, upon delete, change back to "All Views" -->
+    <EditViewModal
+      ref="editViewModalRef"
+      @deleted="() => (selectedView = '')"
```
nickgerace, line 208: Whenever we delete the view, go back to "All Views"

app/web/src/newhotness/Explore.vue
```diff
@@ -926,10 +949,25 @@ const openAddComponentModal = () => {
...
 
+const editViewModalRef = ref<InstanceType<typeof EditViewModal>>();
+
+const openEditViewModal = (viewId: string) => {
+  // FIXME(nick): right now, we need to switch the current view to edit it because we need the query data.
+  // Why? The current view may differ than the one wishing to be edited. I had a version of this where the
+  // modal had an inner query to account for this scenario (essentially two different "view component list"
+  // calls), but I scrapped it since the query would never load properly. Ideally, you should be able to
+  // delete a view or edit it without switching to it. For now, we will survive.
+  selectedView.value = viewId;
```
nickgerace, line 964: I did not want to do this, but spinning the wheels here isn't worth it at present. It would be nice to have a better way to have conditional query loading like "you asked for a different view than the one loaded in the Explore Vue component, okay I'll get its components", but doing that is a bit jank right now... for now, let's just switch to the view so that we get all of its components needed for the "canDeleteView" computation.

lib/vue-lib/src/design-system/menus/DropdownMenuItem.vue
```diff
@@ -8,7 +8,7 @@
       clsx(
         'flex gap-xs items-center group select-none',
         noInteract ? 'text-neutral-500' : 'cursor-pointer',
-        !endLinkTo && 'children:pointer-events-none',
+        !endLinkTo && !enableSettingsEdit && 'children:pointer-events-none',
```
nickgerace, line 11: New settings button `pointer-events-none` check was annoying to lock down, but we did it

lib/vue-lib/src/design-system/menus/DropdownMenuItem.vue
```diff
@@ -123,9 +129,27 @@
           <Icon v-else name="link" />
         </slot>
       </div>
+
       <template v-else-if="shortcut">
         {{ shortcut }}
       </template>
+
+      <div
+        v-else-if="enableSettingsEdit"
```
nickgerace, line 138: Very similar to `endLinkto`, but emits instead of redirects

lib/vue-lib/src/design-system/modals/Modal.vue
```diff
@@ -87,6 +87,7 @@
                   v-if="type === 'save' && !noWrapper"
                   class="py-3 flex flex-row justify-between gap-sm"
                 >
+                  <slot name="leftButton" />
```
nickgerace, line 90: Stuff the delete button here in an ergonomic manner

stack72:
@nickgerace works really well! Nice work :)

PR #6415, opened 2025-06-13
feat: Run mgmt funcs

vbustamante:

- You can now run the management funcs on the component details screen

#### Screenshots:

image

#### Out of Scope:

There seems to be an issue that the func run logs get wiped, so we don't show a success message. This will be tackled in a further PR.

## How was it tested?

- [X] Manual test: Can run an import func and see it create components


app/web/src/newhotness/ComponentDetails.vue
```diff
@@ -476,6 +486,67 @@ const hasImportFunc = computed(
 
 const showResourceInput = ref(false);
 
+// MGMT funcs
+const ctx = inject<Context>("CONTEXT");
+
+const funcRunQuery = useInfiniteQuery({
```
nickgerace, line 492: Huh I have never seen this before... neat.

app/web/src/newhotness/ComponentDetails.vue
```diff
@@ -476,6 +486,67 @@ const hasImportFunc = computed(
...
+    const req = await call.get(params);
+    if (api.ok(req)) {
+      return req.data;
+    }
+    return {
+      funcRuns: [],
+      nextCursor: null,
+    };
+  },
+  initialPageParam: undefined,
```
nickgerace, line 514: Makes sense

app/web/src/newhotness/ComponentDetails.vue
```diff
@@ -476,6 +486,67 @@ const hasImportFunc = computed(
...
+      params.append("cursor", pageParam);
+    }
+    const req = await call.get(params);
+    if (api.ok(req)) {
+      return req.data;
+    }
+    return {
+      funcRuns: [],
+      nextCursor: null,
+    };
```
nickgerace, line 512: Right... we'll always have something. Good idea

app/web/src/newhotness/ManagementFuncCard.vue
```diff
@@ -0,0 +1,109 @@
...
+const route = useRoute();
+
+const mgmtRunApi = useApi();
+const runMgmtFunc = async (funcId: string) => {
+  const call = mgmtRunApi.endpoint<{ success: boolean }>(
+    routes.RunMgmtPrototype,
+    {
+      prototypeId: funcId,
+      componentId: props.componentId,
+      viewId: explore?.viewId.value ?? "DEFAULT", // Should get the default view id
```
nickgerace, line 69: Feels like this should be an optional param in the backend, but I don't think it's worth blocking this PR to change

PR #6416, opened 2025-06-13
feat(web) - truncation and other UI fixes newhotness

wendybujalski:
Screenshot 2025-06-13 at 7 06 47 PM

This PR fixes a number of truncation bugs, and also adds the ability to show/hide secret fields.

### DIFF
--- a/app/web/src/newhotness/AttributeValueBox.vue
+++ b/app/web/src/newhotness/AttributeValueBox.vue
@@ -3,6 +3,7 @@
     :class="
       clsx(
         'flex flex-row items-center gap-xs border w-fit px-3xs rounded-sm',
+        'max-w-full',
         themeClasses(
           'text-neutral-600 border-neutral-400 bg-neutral-100',
           'text-neutral-400 border-neutral-600 bg-neutral-900',
--- a/app/web/src/newhotness/ComponentDetails.vue
+++ b/app/web/src/newhotness/ComponentDetails.vue
@@ -38,35 +38,36 @@
     <div
       :class="
         clsx(
-          'name items-center flex flex-row gap-xs p-xs',
+          'name flex flex-row items-center gap-xs p-xs',
           themeClasses('bg-neutral-200', 'bg-neutral-800'),
         )
       "
     >
       <VButton
         v-tooltip="'Back (Esc)'"
-        class="border-0 mr-2em"
+        class="border-0 mr-2em flex-none"
         icon="arrow--left"
         size="sm"
         tone="shade"
         variant="ghost"
         @click="back"
       />
-      <span>{{ component.schemaVariantName }}</span>
-      <span>/</span>
-      <span class="grow">
+      <div class="flex-none">{{ component.schemaVariantName }}</div>
+      <div class="flex-none">/</div>
+      <div class="flex-1 min-w-0">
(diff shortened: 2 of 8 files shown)


app/web/src/newhotness/AttributeValueBox.vue
```diff
@@ -3,6 +3,7 @@
     :class="
       clsx(
         'flex flex-row items-center gap-xs border w-fit px-3xs rounded-sm',
+        'max-w-full',
```
wendybujalski, line 6: Prevents the AttributeValueBox from growing beyond its parent element so that the truncation of the contents inside works.

app/web/src/newhotness/ComponentDetails.vue
```diff
@@ -38,35 +38,36 @@
...
               :loading="wForm.bifrosting.value"
               :loadingText="component.name"
-              class="border-0 font-normal"
+              class="border-0 font-normal max-w-full"
               iconRight="edit"
               loadingIcon="loader"
               size="sm"
               tone="shade"
               variant="ghost"
+              truncateText
```
wendybujalski, line 70: `VButton` components can now truncate their dynamic label if placed inside a constraining container.

app/web/src/newhotness/AttributeValueBox.vue
nickgerace: Yep! I had thought that's what I was reading. Great idea.

app/web/src/newhotness/layout_components/SecretInput.vue
```diff
@@ -0,0 +1,74 @@
...
+    "
+    />
+    <Icon
+      v-if="getFormFieldType(fieldname) === 'password'"
+      v-tooltip="secretShowing ? 'Hide Value' : 'Show Value'"
+      :name="secretShowing ? 'hide' : 'eye'"
+      size="xs"
+      class="mr-xs cursor-pointer"
+      @click="toggleSecretShowing"
+    />
```
wendybujalski, line 42: This should probably have a better hover state but I wanna come back later and find consistent styles for all of our "buttons that go inside of an input field" anyway.

lib/vue-lib/src/design-system/general/TruncateWithTooltip.vue
```diff
@@ -138,5 +138,5 @@ onBeforeUnmount(() => {
   clearInterval(lineClampRefresh.value);
 });
 
-defineExpose({ tooltipActive });
+defineExpose({ tooltipActive, tooltip });
```
wendybujalski, line 141: Sometimes we need to pass the tooltip up the DOM to another element for it to function properly in a particular spot, exposing it from the `TruncateWithTooltip` component allows us to do that.

lib/vue-lib/src/design-system/general/VButton.vue
```diff
@@ -3,6 +3,7 @@
     :is="htmlTagOrComponentType"
     v-bind="dynamicAttrs"
     ref="component"
+    v-tooltip="truncateRef?.tooltipActive ? truncateRef.tooltip : undefined"
```
wendybujalski, line 6: Here we are getting the tooltip data from the nested `TruncateWithTooltip` on the label and putting it on the top level element of the button.

PR #6422, opened 2025-06-16
Protect the default view from deletion

nickgerace:
## Description

This change protects the default view from deletion at the graph level as well as provides tooltip text for default views in the edit modal in the new UI.

In addition to that change, the local development docs have been edited to mention the SSL requirement when not using localhost. This was already recommended, but is now required with the recent webworker changes and its lock acquisition code.

### DIFF
--- a/app/web/src/newhotness/EditViewModal.vue
+++ b/app/web/src/newhotness/EditViewModal.vue
@@ -40,13 +40,7 @@
       />
     </label>
     <template #leftButton>
-      <div
-        v-tooltip="
-          canDeleteView
-            ? undefined
-            : 'Views containing one or more components cannot be deleted. To delete a view, first remove all components from it.'
-        "
-      >
+      <div v-tooltip="tooltipText">
         <VButton
           label="Delete View"
           :disabled="!canDeleteView"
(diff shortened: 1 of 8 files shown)


lib/dal/src/workspace_snapshot/split_snapshot.rs
```diff
@@ -1691,6 +1691,19 @@ impl ViewExt for SplitSnapshot {
...
+            view_id,
+            Incoming,
+            EdgeWeightKindDiscriminants::Use,
+        )? {
+            let edge_weight = edge_ref.weight();
+            if *edge_weight.kind() == EdgeWeightKind::new_use_default() {
+                return Err(WorkspaceSnapshotError::DefaultViewDeletionAttempt);
+            }
+        }
+
```
zacharyhamm, line 1706: Nice. Glad you covered this

app/web/src/newhotness/EditViewModal.vue
```diff
@@ -68,6 +62,17 @@ import { useWatchedForm } from "./logic_composables/watched_form";
 
 const viewId = ref<ViewId>("");
 const canDeleteView = ref<boolean>(false);
+const isDefaultView = ref<boolean>(false);
+
+const tooltipText = computed(() => {
+  if (isDefaultView.value) {
+    return "Cannot delete the default View.";
+  }
+  if (canDeleteView) {
```
zacharyhamm: canDeleteView is a computed ref, so you have to check `.value` here

nickgerace: Oops! That was an accident in removing the nested ternary. Wasn't picked up by the linter locally, but I changed it to the computed `tooltipText`.

zacharyhamm (approved):

PR #6423, opened 2025-06-16
fix(dal): refactor where we can build component sources

stack72:


lib/dal/src/component.rs
```diff
@@ -1337,6 +1342,57 @@ impl Component {
         Ok(input_socket_ids)
     }
 
+    pub async fn sources_for_id(
```
jkeiser: This should be the only version of this method on `Component`, so unless there is a `sources()` method already, let's name this `sources()` (trying to avoid littering the code with _by_id and _for_id where we can, and reduce the number of places we do it over time).

jkeiser (approved):

PR #6427, opened 2025-06-17
Make non-exempt mutation calls handle change sets

nickgerace:
## Description

This change makes all non-exempt mutation calls handle new change sets returned from SDF. An exempt mutation call is one that is marked in "CAN_MUTATE_ON_HEAD", which currently only corresponds to three calls related to actions. All exempt locations have been marked with a comment.

This change fixes the issue where mutating on HEAD did not properly move the user to the new change set as well as cements the developer pattern to either handle the new change set with post, put and delete methods, or to indicate that it doesn't need to be handled via a comment.

All calls were found with the following "ripgrep" command in the "newhotness" directory: ```rg 'call\.(put|post|delete)' -B 1```.

## Other Fixes

- The duplicate "new-hotness-view" route has been removed.
  - It is not being used and can cause a bug where the "secretId" would be missing when navigating to the route
- Multiple usages of the same "useApi" have been split up since they should only be used for one route each

### DIFF
--- a/app/web/src/newhotness/ActionCard.vue
+++ b/app/web/src/newhotness/ActionCard.vue
@@ -233,6 +233,8 @@ const hold = () => {
 const holdApi = useApi();
 const finishHold = async () => {
   const call = holdApi.endpoint(routes.ActionHold, { id: props.action.id });
+
+  // This route can mutate head, so we do not need to handle new change set semantics.
   await call.put({});
   confirmRef.value?.close();
 };
@@ -240,12 +242,16 @@ const finishHold = async () => {
 const retryApi = useApi();
 const retry = async () => {
   const call = retryApi.endpoint(routes.ActionRetry, { id: props.action.id });
+
+  // This route can mutate head, so we do not need to handle new change set semantics.
   await call.put({});
 };
 
 const removeApi = useApi();
 const remove = async () => {
   const call = removeApi.endpoint(routes.ActionCancel, { id: props.action.id });
+
+  // This route can mutate head, so we do not need to handle new change set semantics.
   await call.put({});
 };
 </script>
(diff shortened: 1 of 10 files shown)


app/web/src/newhotness/AddViewModal.vue
```diff
@@ -77,7 +79,7 @@ const nameForm = wForm.newForm({
       if (newChangeSetId) {
         api.navigateToNewChangeSet(
           {
-            name: "new-hotness-view",
+            name: "new-hotness",
```
nickgerace, line 82: Let's do what the comment above says

app/web/src/newhotness/ActionCard.vue
```diff
@@ -233,19 +233,25 @@ const hold = () => {
 const holdApi = useApi();
 const finishHold = async () => {
   const call = holdApi.endpoint(routes.ActionHold, { id: props.action.id });
+
+  // This route can mutate head, so we do not need to handle new change set semantics.
   await call.put({});
```
nickgerace, line 238: You'll see this in all locations where `put`, `post`, and `delete` calls are used for `CAN_MUTATE_ON_HEAD` routes. This should not be the default pattern for developers. These routes are the exception and we should notate it when working with them.

app/web/src/newhotness/AddViewModal.vue
```diff
@@ -25,7 +25,9 @@
             :value="field.state.value"
             type="text"
             :disabled="wForm.bifrosting.value"
-            @input="(e) => field.handleChange((e.target as HTMLInputElement).value)"
+            @input="
+              (e) => field.handleChange((e.target as HTMLInputElement).value)
+            "
```
nickgerace, line 30: I've re-run the linter a bunch... my editor does this but `fix-lint` doesn't fix it...

nickgerace:
Performed a clean rebase. No other changes.

jobelenus (approved):

PR #6430, opened 2025-06-17
fix(web): ensure we set lockAcquired = true on other tabs

zacharyhamm:
Ensures that if the database lock is acquired by one tab, every tab knows about it.

Tested by opening lots of tabs and seeing that they all render the change set eventually.


jobelenus (approved):
Very nice!

zacharyhamm:
@jobelenus had to run the linter

jobelenus (approved):

PR #6432, opened 2025-06-17
Update component name and attributes optimistically

jkeiser:
This makes it so when you edit a value in the Attributes Panel, the edited value shows immediately.

It does this for:

- Component Name label at the top
- Component Name special textbox
- All attributes

## How was it tested?

- [X] Manually check all three places
- [X] Check multiple tabs show updates to each other


app/web/src/newhotness/logic_composables/watched_form.ts
```diff
@@ -82,17 +82,23 @@ export const useWatchedForm = <Data>(label: string) => {
...
           changeSetId: ctx.changeSetId.value,
           form: label,
         });
+
+        // Submit the form and start waiting for the data to come back
         start = Date.now();
-        onSubmit(props);
+        await onSubmit(props);
         bifrosting.value = true;
         rainbow.add(ctx.changeSetId.value, label);
```
jobelenus, line 99: now that we're awaiting, we should move bifrosting and rainbow add _before_ the `onSubmit`

app/web/src/newhotness/logic_composables/watched_form.ts
```diff
@@ -82,17 +82,23 @@ export const useWatchedForm = <Data>(label: string) => {
...
         span.setAttributes({
           workspaceId: ctx.workspacePk.value,
           changeSetId: ctx.changeSetId.value,
           form: label,
         });
+
+        // Submit the form and start waiting for the data to come back
         start = Date.now();
-        onSubmit(props);
+        await onSubmit(props);
```
jobelenus: We need to catch errors here, set them on the form object, and reset the form/ensure it can be re-submitted

app/web/src/newhotness/ComponentDetails.vue
```diff
@@ -55,31 +55,30 @@
       <div class="flex-none">{{ component.schemaVariantName }}</div>
       <div class="flex-none">/</div>
       <div class="flex-1 min-w-0">
-        <EditInPlace ref="editInPlaceRef" @hidden="reset" @showing="focus">
-          <template #trigger>
-            <VButton
-              :label="component.name"
-              :loading="wForm.bifrosting.value"
```
jobelenus: We talked about how we don't want any spinner activity, so we can remove label and loading. And the `{{ component.name }}` needs to change, because once it flips back to the button, its not going to show the value the user typed in.

app/web/src/newhotness/logic_composables/watched_form.ts
jkeiser: Added the reset! However, the only thing I see on the form object is wForm.setErrorMap(), which seems to be about validation on individual fields; additionally, I'm not sure how they end up being shown on the form (especially if we're using things like valueForm.Field).

jobelenus: Yea, the tanstack field in the template has access to errors

app/web/src/newhotness/logic_composables/watched_form.ts
```diff
@@ -67,69 +74,70 @@ export const useWatchedForm = <Data>(label: string) => {
...
-      // so this will never get removed, this is just a UI fallback
-      setTimeout(() => {
-        rainbow.remove(ctx.changeSetId.value, label);
-      }, 750);
-    }
+    // Update form data as data changes
+    watch(
+      () => toValue(data),
+      (newData) => wForm.reset(newData),
+    );
```
jobelenus, line 140: we _may_ not _always_ want this behavior on every form... but we'll address that if it comes to it... see how people experience things, etc...

jkeiser: Yep! Though the way it's written now with MaybeRefOrGetter<Data>, if people don't want the behavior, they can just pass a literal instead of a computed value. Keeps the caller code clear; if you have a computed value or a ref, the implication is its value can change and consumers should react.

PR #6434, opened 2025-06-17
fix(web) - three newhotness fixes

wendybujalski:
- Fixed the empty state on the component grid.
- Fixed the navigate to selected component on click in the Map.
- Back button is now "Close" and sends you back to the Grid or Map, whichever you were on last.


app/web/src/newhotness/Explore.vue
```diff
@@ -1105,6 +1118,16 @@ const scrollCurrentTileIntoView = () => {
 const onMapDeselect = () => {
   searchRef.value?.blur();
 };
+
+const storeViewMode = () => {
+  if (!groupRef.value) return;
+
+  if (groupRef.value.isB) {
+    localStorage.setItem(VIEW_MODE_LOCAL_STORAGE_KEY, "grid");
```
jobelenus: I would put the `changeSetId` in the key of `setItem`, that way multiple tab usage doesnt create unexpected behavior

jobelenus (approved):

PR #6439, opened 2025-06-18
feat(web): Show new components instantly

jkeiser:
Right now, we wait for edda to finish indexing new components before we show their attributes. With this change, we display new components as soon as they are created.

This PR:

* Adds the attribute tree to the create_component response
* Caches the attribute tree MV from the response so that it can be used in the absence of a true updated index
* Uses the cached attribute tree MV when the index has not yet been updated
* Switches to use the real attribute tree MV when it becomes available

#### Out of Scope:

* This does not speed up the actual creation of the component (if anything, it slows it down, because it adds data to the response). This means that when you double click, there is still a long, perceptible pause (with no feedback) before the component is created and the new page opens.
* This will not add a "skeleton" component to the main screen, so if you hit escape because the component is taking too long to create, you won't see it show up in the screen.

## How was it tested?

- [x] Created a new component, verified the attribute tree shows right away
- [x] Temporarily disabled the active query (locally) to ensure the attribute tree looks correct and it wasn't just the MV being created quickly

## In Short

instant noodles

### DIFF
--- a/app/web/src/newhotness/AddComponentModal.vue
+++ b/app/web/src/newhotness/AddComponentModal.vue
@@ -231,6 +231,7 @@ import {
   UninstalledVariant,
   BifrostComponent,
   EddaComponent,
+  AttributeTree,
 } from "@/workers/types/entity_kind_types";
 import { bifrost, useMakeArgs, useMakeKey } from "@/store/realtime/heimdall";
 import FilterTile from "./layout_components/FilterTile.vue";
@@ -354,6 +355,7 @@ const onEnter = async () => {
   const call = api.endpoint<{
     componentId: string;
     materializedView: EddaComponent;
+    attributeTreeMaterializedView: AttributeTree;
     schemaVariantMaterializedView: SchemaVariant;
   }>(routes.CreateComponent, {
     viewId: explore.viewId.value,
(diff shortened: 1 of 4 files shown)


app/web/src/newhotness/AddComponentModal.vue
```diff
@@ -363,25 +365,40 @@ const onEnter = async () => {
...
+    if (newChangeSetId) {
+      componentQueryKey.forEach((v, idx) => {
+        if (v === route.params.changeSetId)
+          componentQueryKey[idx] = newChangeSetId;
+      });
+      attributeTreeQueryKey.forEach((v, idx) => {
+        if (v === route.params.changeSetId)
+          attributeTreeQueryKey[idx] = newChangeSetId;
+      });
+    }
```
nickgerace, line 387: Yes... I believe this is the right move.

app/web/src/newhotness/ComponentDetails.vue
```diff
@@ -305,29 +305,26 @@ const outgoing = computed(
...
-    return component;
-  },
+    )) ??
+    queryContext.client.getQueryData(
+      key(EntityKind.Component, componentId).value,
+    ),
 });
 
-const attributeTreeQuery = useQuery<AttributeTree | null>({
+const attributeTreeQuery = useQuery<AttributeTree | undefined>({
```
nickgerace, line 319: Better!

nickgerace (approved):

jobelenus (approved):

PR #6441, opened 2025-06-18
Add info about MV build timings to build all & build for changes info spans

jhelwig:
The build all MVs & build for changes info spans now include the number of MVs built, how long each one took on average, which kind was the slowest to build, and how long the slowest took to build. This information should help when investigating slower than expected MV building, and give us metrics to use to monitor edda's performance over time.

Example screenshot of Jaeger from local testing:

image

### DIFF
--- a/lib/edda-server/src/change_set_processor_task/materialized_view.rs
+++ b/lib/edda-server/src/change_set_processor_task/materialized_view.rs
@@ -1,6 +1,9 @@
-use std::collections::{
-    HashSet,
-    VecDeque,
+use std::{
+    collections::{
+        HashSet,
+        VecDeque,
+    },
+    time::Duration,
 };
 
 use dal::{
@@ -79,7 +82,10 @@ use strum::IntoEnumIterator;
 use telemetry::prelude::*;
 use telemetry_utils::metric;
 use thiserror::Error;
-use tokio::task::JoinSet;
+use tokio::{
+    task::JoinSet,
+    time::Instant,
+};
 
 use crate::updates::{
     EddaUpdates,
@@ -259,6 +265,10 @@ pub async fn reuse_or_rebuild_index_for_new_change_set(
         si.workspace.id = Empty,
         si.change_set.id = %ctx.change_set_id(),
         si.materialized_view.reason = reason_message,
+        si.edda.mv.count = Empty,
+        si.edda.mv.avg_build_elapsed_ms = Empty,
+        si.edda.mv.max_build_elapsed_ms = Empty,
+        si.edda.mv.slowest_kind = Empty,
     ),
 )]
 pub async fn build_all_mv_for_change_set(
@@ -278,7 +288,14 @@ pub async fn build_all_mv_for_change_set(
         ))
         .await?;
 
-    let (frontend_objects, patches) = build_mv_inner(
+    let (
+        frontend_objects,
+        patches,
+        build_count,
+        build_total_elapsed,
+        build_max_elapsed,
+        build_slowest_mv_kind,
+    ) = build_mv_inner(
         ctx,
         frigg,
         ctx.workspace_pk()?,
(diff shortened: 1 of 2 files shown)


britmyerss (dismissed):

lib/edda-server/src/change_set_processor_task/materialized_view.rs
```diff
@@ -368,8 +395,32 @@ pub async fn build_mv_for_changes_in_change_set(
...
+    span.record("si.edda.mv.count", build_count);
+    span.record(
+        "si.edda.mv.avg_build_elapsed_ms",
+        build_total_elapsed.as_millis() / build_count,
+    );
+    span.record(
+        "si.edda.mv.max_build_elapsed_ms",
+        build_max_elapsed.as_millis(),
+    );
+    span.record("si.edda.mv.slowest_kind", build_slowest_mv_kind);
```
fnichol, line 419: These `span.record()` calls would attempt to add values to this function's `materialized_view.build_mv_for_changes_in_change_set` Span, but these fields aren't set up in the `#[instrument()]` call, meaning that the values will be dropped.

Is it possible that you meant to add those 4 fields to this Span as well as to the `materialized_view.build_all_mv_for_change_set` Span?

jhelwig:
@britmyerss Force-pushed to fix a lint issue that came up from splitting apart this work from the configurable-build-parallelism work.

britmyerss (dismissed):

lib/edda-server/src/change_set_processor_task/materialized_view.rs
```diff
@@ -530,6 +589,12 @@ async fn build_mv_inner(
                     if let Some(frontend_object) = maybe_frontend_object {
                         frontend_objects.push(frontend_object);
                     }
+                    build_count += 1;
+                    if build_duration > build_max_elapsed {
+                        build_max_elapsed = build_duration;
+                        build_slowest_mv_kind = kind.to_string();
```
fnichol: I believe that if we added strum's [`IntoStaticStr`](https://docs.rs/strum/latest/strum/derive.IntoStaticStr.html) on `ReferenceKind`, then we'd have a `&'static str` representation of kind and it would save us the String allocation here.

fnichol (commented):
Nice work!

I had a question about some potential missing fields in a Span and a thought about saving some allocations and otherwise this looks good

lib/edda-server/src/change_set_processor_task/materialized_view.rs
jhelwig: I did mean to. Looks like those didn't get pulled over when I was splitting this out from other work I had going on concurrently. Will update.

lib/edda-server/src/change_set_processor_task/materialized_view.rs
jhelwig: Will update.

lib/edda-server/src/change_set_processor_task/materialized_view.rs
```diff
@@ -485,6 +542,11 @@ async fn build_mv_inner(
         }
     }
 
+    let mut build_total_elapsed = Duration::from_nanos(0);
+    let mut build_count: u128 = 0;
+    let mut build_max_elapsed = Duration::from_nanos(0);
+    let mut build_slowest_mv_kind: &str = "N/A";
```
fnichol, line 548: Oh right, you can do that now!

lib/si-frontend-mv-types-rs/src/reference.rs
```diff
@@ -31,6 +31,7 @@ pub use weak::WeakReference;
     strum::Display,
     strum::EnumIter,
     strum::EnumString,
+    strum::IntoStaticStr,
```
fnichol, line 34: I like this one, haven't used it yet, but might use it more often now

fnichol (approved):
Looks great!

PR #6445, opened 2025-06-18
feat(web): fetch all change sets on load, wait in the lobby

zacharyhamm:
Starts the user in the lobby, and runs "muspelheim", which runs niflheim against every open change set. Only then does it exit the lobby.

Lobby is controlled by the status of the niflheim queries, not the route.

Also sends the user to the lobby / muspelheims if the tab has been idle for 12 hours+.

### DIFF
--- a/app/web/src/newhotness/Workspace.vue
+++ b/app/web/src/newhotness/Workspace.vue
@@ -51,12 +51,9 @@
      min-h-0 prevents the main container from being *larger* than the max it can grow, no matter its contents -->
     <main class="grow min-h-0">
       <div v-if="tokenFail">Bad Token</div>
-      <div v-else-if="lobby" class="w-[50svh] mx-auto mt-[15svh]">
-        <h1 class="text-center text-2xl">Welcome!</h1>
-        <h2 class="text-center text-xl">Have a seat in our lobby</h2>
-        <h3 class="text-center text-lg">We are loading your workspace now</h3>
-        <Icon class="mx-auto" name="loader" size="full" />
-      </div>
+      <template v-else-if="lobby">
+        <Lobby />
+      </template>
       <template v-else-if="componentId">
         <ComponentDetails :componentId="componentId" />
       </template>
@@ -88,7 +85,6 @@ import {
   provide,
   watch,
 } from "vue";
-import { Icon } from "@si/vue-lib/design-system";
 import { useQuery, useQueryClient } from "@tanstack/vue-query";
 import NavbarPanelLeft from "@/components/layout/navbar/NavbarPanelLeft.vue";
 import NavbarPanelRight from "@/components/layout/navbar/NavbarPanelRight.vue";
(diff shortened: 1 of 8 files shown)


app/web/src/newhotness/Workspace.vue
```diff
@@ -336,17 +332,18 @@ watch(
...
-        name: "new-hotness",
-        params: {
-          workspacePk: props.workspacePk,
-          changeSetId: props.changeSetId,
-        },
-      });
+    if (!success) {
+      heimdall.muspelheimStatuses.value[props.changeSetId] = false;
     }
+    router.push({
```
jobelenus: We shouldn't need a router push here? Just make the lobby bool false so they see the page...

jobelenus (dismissed):
One nit in there. 

I do have a behavior question though; in the case that one of the index requests did 202... what is going to re-fire that after IndexUpdate tells us it's good to go?

app/web/src/newhotness/Workspace.vue
zacharyhamm: ahh. makes sense

There's another issue here: it's getting stuck forever when creating a new change set. I have to fix that.

zacharyhamm: The lobbyExit callback will mark the change set as ready to move it out of the lobby if it gets a 202.

app/web/src/workers/webworker.ts
```diff
@@ -625,15 +625,15 @@ const insertAtomMTM = (atom: Atom, indexChecksum: Checksum) => {
       sql: `insert into index_mtm_atoms
         (index_checksum, kind, args, checksum)
           VALUES
-        (?, ?, ?, ?)
+        (?, ?, ?, ?) on conflict do nothing
```
jobelenus: We can't do this because we're using ` ON CONFLICT REPLACE` on that very table.

jobelenus (changes requested):

app/web/src/workers/webworker.ts
jobelenus: We shouldn't actually hit any more mtm failures, bc everything is now sitting in queues so things shouldn't be racing

zacharyhamm:
No longer seeing the createMTM issues, which may have been solved by keiser's changes?

app/web/src/newhotness/Workspace.vue
```diff
@@ -306,51 +315,38 @@ onBeforeMount(async () => {
       if (connectionShouldBeEnabled.value && heimdall.initCompleted.value) {
         // NOTE(nick,wendy): this says "reconnect", but it must run once on startup.
         await heimdall.bifrostReconnect();
-      } else {
```
zacharyhamm, line 309: This was closing the websocket. Do we need to close it?

jobelenus: huh... well... if someone gets un-auth'd/turned off (i dont know how that would get pushed down to the app) it would close the websocket—which would be desirable

jobelenus (approved):

PR #6448, opened 2025-06-18
fix(dal): Remove subscriber APAs when removing components

jkeiser:
Right now, when you remove a component, any subscribing components will get into an invalid graph state where the subscription APA has no value edge. This causes downstream errors with some operations that check for this.

This PR fixes this by removing subscriber APAs pointing directly at an AV when that AV is removed. Subscribers should have the same value they would have if the subscription to was to an array element and the array element was removed.

### Factor Budget

A couple of small factors here:

- AttributeValue::remove_by_id -> AttributeValue::remove (it is accepted and often preferred for DAL methods to take IDs instead of self, so more and more _by_id will go away in the coming months). This was a pure rename, no changes to callers or to the function itself.
- subscription tests use idiomatic APIs -- this revealed some test bugs where we weren't really testing what we thought we were (when you have to actually type the word "subscriber" instead of component_id, it makes it clear exactly what you expected to happen).

## How was it tested?

- [X] Integration tests pass
- [X] New integration test for removing a subcribed component
- [x] Manual test: delete subscribed component, then try to duplicate the subscriber. See no errors.

subscription cancelled

### DIFF
--- a/lib/dal/src/attribute/value.rs
+++ b/lib/dal/src/attribute/value.rs
@@ -258,8 +258,6 @@ pub enum AttributeValueError {
     PropMoreThanOneChild(PropId),
     #[error("prop not found for attribute value: {0}")]
     PropNotFound(AttributeValueId),
-    #[error("trying to delete av that's not related to child of map or array: {0}")]
-    RemovingWhenNotChildOrMapOrArray(AttributeValueId),
     #[error("secret error: {0}")]
     Secret(#[from] Box<SecretError>),
     #[error("serde_json: {0}")]
@@ -1451,14 +1449,14 @@ impl AttributeValue {
                 warn!(
                     "Removing child AV {child_av_id} (no prop) parent object AV {parent_av_id} (parent prop {parent_prop_id})"
                 );
-                AttributeValue::remove_by_id(ctx, child_av_id).await?;
+                AttributeValue::remove(ctx, child_av_id).await?;
                 continue;
             };
             if let Some(duplicate_av_id) = existing_children.insert(child_prop_id, child_av_id) {
                 warn!(
                     "Removing duplicate child AV {duplicate_av_id} (prop {child_prop_id}) from parent object AV {parent_av_id} (parent prop {parent_prop_id})"
                 );
-                AttributeValue::remove_by_id(ctx, duplicate_av_id).await?;
+                AttributeValue::remove(ctx, duplicate_av_id).await?;
                 continue;
             }
         }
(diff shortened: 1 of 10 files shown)


lib/dal/src/attribute/value.rs
```diff
@@ -2536,36 +2536,67 @@ impl AttributeValue {
         Ok(pairs)
     }
 
-    pub async fn remove_by_id(ctx: &DalContext, id: AttributeValueId) -> AttributeValueResult<()> {
-        let parent_av_id = Self::parent_id(ctx, id)
-            .await?
-            .ok_or(AttributeValueError::RemovingWhenNotChildOrMapOrArray(id))?;
+    /// Remove an array or map element.
```
zacharyhamm: This will also remove object children... so the comment is not quite true

zacharyhamm (commented):
A small note. Otherwise it looks correct.

I think we'll also need a correction here. If the component is deleted in another change set, and the subscribers in this change set don't exist in that change set, they will continue to "dangle" when the deletion change set is applied

lib/dal/src/attribute/value.rs
zacharyhamm: A github comment on my review disappeared for some reason (eventual consistency? i guess!).

I wanted to say: why have this error at all. It only prevents deleting the root attribute value. I say remove the check. We should validate the input higher up than this call if we want to prevent the removal of non array or map elements (and this check does not prevent that!)

lib/dal/src/attribute/value.rs
```diff
@@ -2536,36 +2536,67 @@ impl AttributeValue {
...
+    pub async fn remove(ctx: &DalContext, id: AttributeValueId) -> AttributeValueResult<()> {
+        if !Self::parent_id(ctx, id).await?.is_some() {
```
zacharyhamm: Linter is going to ding you for not using `is_none` btw

jkeiser:
Agreed, we'll need a CT here. Another of those cases that almost feels like it should *only* be a CT--that maybe we just don't bother snipping dependencies like this until we reach the CT phase.

lib/dal/src/attribute/value.rs
jkeiser: Yeah; I didn't add the error, so I didn't want to remove it in case it was being depended upon, but I guess if someone is relying on an error for this, something's gone way wrong elsewhere anyway. I think I'm OK with removing the error right now ... I'll check for callers and make it go away before I ping for re-review!

PR #6449, opened 2025-06-18
Expose edda's parallel MV build limit as a configuration setting

jhelwig:
This maintains the previously hard-coded value as the default for the configuration setting, but it is now possible to change the limit on how many MVs edda will try to build at the same time without having to recompile edda.

By being able to adjust this limit without having to rebuild edda, it will be easier/faster to tune the value for the running hardware.

edda's parallel MV build limit is now configurable by setting the `SI_EDDA__EDDA_PARALLEL_BUILD_LIMIT` environment variable, as well as via edda's config file.


fnichol (approved):
Looks great!

PR #6465, opened 2025-06-23
fix(web): Ensure we default to the selected view

stack72:
When changing to a different change set, if the selected view doesn’t exist, then we default to `All Views`.


nickgerace (approved):
Brilliant

PR #6472, opened 2025-06-24
fix(web): Remove expand-logs button on FuncRunDetails

stack72:
This button didn’t do anything so I am removing it as it’s misleading

Screenshot 2025-06-24 at 11 17 14


app/web/src/newhotness/layout_components/GridItemWithLiveHeader.vue
```diff
@@ -11,7 +11,7 @@
     "
   >
     <div class="flex flex-row items-center justify-between px-sm py-2xs">
-      <h2 class="text-sm font-medium">{{ title }}</h2>
```
wendybujalski, line 14: Bigger text here is good too!

PR #6477, opened 2025-06-24
fix: close modal on view creation

aaron-dernley:

- Creating a view closes the modal when clicking the Create button
- Creating a view closes the modal when hitting Enter after entering a value
- Editing a view name pulls in the name of the view to the input

#### Video:

https://jam.dev/c/93a78fec-3a94-4af1-9b2a-133e1105dee1

## How was it tested?

- [X] Integration tests pass

### DIFF
--- a/app/web/src/newhotness/AddViewModal.vue
+++ b/app/web/src/newhotness/AddViewModal.vue
@@ -8,43 +8,45 @@
     saveLabel="Create"
     @save="() => nameForm.handleSubmit()"
   >
-    <label class="flex flex-row items-center relative">
-      <span>View Name</span>
-      <nameForm.Field name="name">
-        <template #default="{ field }">
-          <input
-            :class="
-              clsx(
-                'block w-72 ml-auto border',
-                themeClasses(
-                  'text-black bg-white border-neutral-600 disabled:bg-neutral-100',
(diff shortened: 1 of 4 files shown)


app/web/src/newhotness/AddViewModal.vue
```diff
@@ -8,43 +8,45 @@
...
-      </nameForm.Field>
-      <Icon
-        v-if="wForm.bifrosting.value"
-        class="absolute right-2xs"
-        name="loader"
-        size="sm"
-        tone="action"
-      />
-    </label>
+    <form @submit.prevent="nameForm.handleSubmit()">
```
aaron-dernley, line 11: I was unsure about this being the right way to handle this, however it does work..

wendybujalski: This should capture both pressing enter or clicking the submission button. Let me take this for a test run to check for edge case issues!

app/web/src/newhotness/AddViewModal.vue
wendybujalski, line 1: Took this PR for a spin, it looks like the AddViewModal also has other bugs which we should fix in this PR.

wendybujalski (changes requested):

wendybujalski (approved):

PR #6479, opened 2025-06-24
Being more defensive here around the potential for bad data

jobelenus:
Prevent exceptions. We were seeing a rare error where `d` was undefined so `d.id` threw an exception

### DIFF
--- a/app/web/src/store/realtime/heimdall.ts
+++ b/app/web/src/store/realtime/heimdall.ts
@@ -175,10 +175,11 @@ const updateCache = (
 ) => {
   // there is always more data attached, but we only care about accessing the ID
   // so thats all we need to type!
+  if (!removed && !data) return;
   const cachedData = queryClient.getQueryData(queryKey) as { id: string }[];
-  if (!cachedData) return;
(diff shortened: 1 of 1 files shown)


app/web/src/store/realtime/heimdall.ts
```diff
@@ -175,10 +175,11 @@ const updateCache = (
 ) => {
   // there is always more data attached, but we only care about accessing the ID
   // so thats all we need to type!
+  if (!removed && !data) return;
   const cachedData = queryClient.getQueryData(queryKey) as { id: string }[];
-  if (!cachedData) return;
+  if (!cachedData || cachedData.length === 0) return;
   // TODO removal
-  const idx = cachedData.findIndex((d) => d.id === id);
+  const idx = cachedData.findIndex((d) => d && d.id === id);
```
jkeiser, line 182: I don't think we should do this without at least adding an error log when we get a falsey value--defensive programming is good but we're probably missing something important here.

jkeiser (changes requested):

PR #6483, opened 2025-06-24
Fix: vue reactivity couldn't track an unassigned computed like this

jobelenus:
Thanks, vue


jkeiser (approved):

vbustamante (commented):
We should add a comment over here that says this is a load bearing variable

PR #6484, opened 2025-06-24
chore(dev): Include Luminork in initial build

jkeiser:
This includes Luminork in the initial "build everything" pass, so that we get everything compiled in parallel and started quicker when we first start Tilte.

When we don't do this, we often have to wait for Luminork to compile *anyway*, because only one of them can run the build at a time, and if Luminork gets the lock before sdf starts (for example), sdf has to wait for it to compile anyway!

## How was it tested?

- [X] Starting up tilt still works fine

### DIFF
--- a/dev/Tiltfile
+++ b/dev/Tiltfile
@@ -412,6 +412,7 @@ rust_build_targets = [
     "//bin/rebaser:rebaser",
     "//bin/sdf:sdf",
     "//bin/veritech:veritech",
+    "//bin/luminork:luminork",
 ]
 local_resource(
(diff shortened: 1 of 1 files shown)


stack72:
So we didn't do this because we wanted that luminork didn't slow down the startup of our system - if you think this is ok for us then thats ok for me :)

stack72 (approved):

jkeiser:
@stack72 we can revert if people are having a bunch of issues, but I fixed this because when I ran it, sdf wouldn't start up until Luminork compiled because of the lock :)

PR #6493, opened 2025-06-25
fix(web): Wait for bifrost to initialize before fetching the action view list from sqlite

britmyerss:

This is the one case where we're rendering data before bifrost is initialized and before cold start finishes, which was causing a whole host of strange issues. Adding this flag to the context, allows the Apply button to render with a default 0 action count, and the action count updates when the query can finally run.

## How was it tested?

Manually tested to ensure that loading an empty change set succeeds, loading a change set with actions shows 0 actions initially and the correct count after cold start completes, and flipping between change sets shows the accurate count as well. 

<div><br/>via <a href="https://giphy.com/dreamstage/">DREAMSTAGE</a> on <a href="https://giphy.com/gifs/dreamstage-live-show-saintmotel-fames-orchestra-6rdMX8uL9NaawLhX96">GIPHY</a></div>

### DIFF
--- a/app/web/src/newhotness/ApplyChangeSetButton.vue
+++ b/app/web/src/newhotness/ApplyChangeSetButton.vue
@@ -34,7 +34,7 @@
 </template>
 
 <script lang="ts" setup>
-import { computed, ref } from "vue";
+import { computed, inject, ref } from "vue";
 import * as _ from "lodash-es";
 import { VButton, PillCounter } from "@si/vue-lib/design-system";
 import { useQuery } from "@tanstack/vue-query";
@@ -47,9 +47,12 @@ import {
 } from "@/workers/types/entity_kind_types";
 import { bifrost, useMakeArgs, useMakeKey } from "@/store/realtime/heimdall";
 import ApprovalFlowModal from "./ApprovalFlowModal.vue";
+import { assertIsDefined, Context } from "./types";
 
 const changeSetsStore = useChangeSetsStore();
 const statusStore = useStatusStore();
+const ctx = inject<Context>("CONTEXT");
+assertIsDefined(ctx);
 
 const applyChangeSetReqStatus =
   changeSetsStore.getRequestStatus("APPLY_CHANGE_SET");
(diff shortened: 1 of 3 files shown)


app/web/src/newhotness/ApplyChangeSetButton.vue
```diff
@@ -34,7 +34,8 @@
 </template>
 
 <script lang="ts" setup>
-import { computed, ref } from "vue";
+import { Context } from "vm";
```
jobelenus: this is the wrong Context 🤣

app/web/src/newhotness/ApplyChangeSetButton.vue
```diff
@@ -47,9 +48,12 @@ import {
 } from "@/workers/types/entity_kind_types";
 import { bifrost, useMakeArgs, useMakeKey } from "@/store/realtime/heimdall";
 import ApprovalFlowModal from "./ApprovalFlowModal.vue";
+import { assertIsDefined } from "./types";
```
jobelenus: you want the `Context` from here

app/web/src/newhotness/Workspace.vue
```diff
@@ -229,6 +229,7 @@ const ctx = computed<Context>(() => {
     outgoingCounts,
     componentNames,
     schemaMembers,
+    queriesEnabled: computed(() => queriesEnabled.value),
```
jobelenus: this doesn't need to wrap in computed, since the whole context is computed

jobelenus: should just be `queriesEnabled`

app/web/src/newhotness/types.ts
```diff
@@ -13,6 +13,7 @@ export interface Context {
   outgoingCounts: ComputedRef<Record<ComponentId, number>>;
   componentNames: ComputedRef<Record<ComponentId, string>>;
   schemaMembers: ComputedRef<Record<SchemaId, SchemaMembers>>;
+  queriesEnabled: ComputedRef<boolean>;
```
jobelenus: and then this becomes `Ref<boolean>`

jobelenus (changes requested):

jobelenus (approved):
:shipit:

PR #6494, opened 2025-06-25
Fix: these fns should only be called after init has completed... 

jobelenus:

If you called these funcs before the system has initialized you're doing it wrong!

@zacharyhamm this is a QOL for a "fast failure" rather than getting hung in a promise that never watches because the watch died / lost scope. Basically, we'd rather yell upfront than have a weird "why didn't it update" behavior.

### DIFF
--- a/app/web/src/store/realtime/heimdall.ts
+++ b/app/web/src/store/realtime/heimdall.ts
@@ -298,7 +298,8 @@ export const bifrost = async <T>(args: {
   kind: Gettable;
   id: Id;
 }): Promise<Reactive<T> | null> => {
-  await waitForInitCompletion();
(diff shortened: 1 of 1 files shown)


app/web/src/store/realtime/heimdall.ts
```diff
@@ -421,7 +428,8 @@ export const getOutgoingConnections = async (args: {
   workspaceId: string;
   changeSetId: ChangeSetId;
 }) => {
-  await waitForInitCompletion();
+  if (!initCompleted.value)
+    throw new Error("You must wait for initialization")
```
jkeiser: While you are lint-fixing, may as well add a newline after this one like the others :)

zacharyhamm (approved):

aaron-dernley (approved):

PR #6495, opened 2025-06-25
Fix: dont use stores to find the list of open change sets

jobelenus:

Folks saw intermittent problems when loading the app. The problem was that the URL to fetch change sets had no `workspacePk` The stack trace went through heimdall and the cold start. That cold start uses the change set and workspace store. Those stores have "activation" steps that are async, which we cannot await. Even though the stores pull the workspace from the URL, other activities must have been getting in the way and slowing activation.

We don't need to use the store to get the change set data we're looking for. So lets just avoid the problem

## How was it tested?

Lobby still works!

### DIFF
--- a/app/web/src/store/realtime/heimdall.ts
+++ b/app/web/src/store/realtime/heimdall.ts
@@ -31,10 +31,12 @@ import {
   Prop,
   SchemaMembers,
 } from "@/workers/types/entity_kind_types";
-import { ChangeSetId } from "@/api/sdf/dal/change_set";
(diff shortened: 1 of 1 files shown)


zacharyhamm (approved):
I can't repro the issue, so can't be 100% sure this fixes. However it breaks nothing, and I think we should not use any of the old stores in the new UI if we can help it

fnichol (approved):
That's a nice step in the direction away from stores, works for me!

PR #6498, opened 2025-06-25
Add collapse, empty states and deep linking to group by

vbustamante:

- Makes component groups collapsible by the user
- Stores the groupby criteria on the URL so a user can link to specific UI states
- Shows empty groups with the new design

#### Screenshots:

image

### DIFF
--- a/app/web/src/newhotness/Explore.vue
+++ b/app/web/src/newhotness/Explore.vue
@@ -106,15 +106,18 @@
         </TabGroupToggle>
         <div>
           <DropdownMenuButton
-            :class="clsx('rounded')"
-            :options="groupByOptions"
+            v-if="showGrid"
+            class="rounded"
+            :options="groupByDropDownOptions"
             :modelValue="groupBySelection"
             placeholder="Group by"
             minWidthToAnchor
             checkable
             alwaysShowPlaceholder
             highlightWhenModelValue
-            @update:modelValue="(val) => (groupBySelection = val)"
+            @update:modelValue="
+              (val) => (groupBySelection = groupByFromString(val))
+            "
           >
             <template #beforeOptions>
               <DropdownMenuItem
@@ -122,7 +125,7 @@
                 value="''"
                 checkable
                 :checked="groupBySelection === ''"
-                @select="() => (groupBySelection = '')"
+                @select="() => (groupBySelection = GroupByCriteria.None)"
               />
             </template>
           </DropdownMenuButton>
@@ -486,8 +489,9 @@ const groupedComponents = computed(() => {
     }
   } else if (groupBySelection.value === "Qualification Status") {
     groups = {
-      Failed: [],
-      Passed: [],
+      "Failed qualifications": [],
+      Warnings: [],
+      "Passed qualifications": [],
     };
     for (const component of filteredComponents) {
       const title = getQualificationStatusTitle(component);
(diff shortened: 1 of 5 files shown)


app/web/src/newhotness/Explore.vue
```diff
@@ -486,8 +489,9 @@ const groupedComponents = computed(() => {
     }
   } else if (groupBySelection.value === "Qualification Status") {
     groups = {
-      Failed: [],
-      Passed: [],
+      "Failed qualifications": [],
+      Warnings: [],
+      "Passed qualifications": [],
```
nickgerace, line 494: Intentional ordering

app/web/src/newhotness/Explore.vue
```diff
@@ -927,24 +957,64 @@ const storeViewMode = () => {
...
 
+// Update the hash of the route (allowing for URL links) when the group by selection changes.
+watch([groupBySelection], () => {
+  const query: SelectionsInQueryString = {
+    ...router.currentRoute.value?.query,
+  };
+  delete query.map;
+  delete query.groupBy;
+
+  query.grid = "1";
```
nickgerace, line 991: Note: when you share the link you also need to be sent to the grid rather than the map. This handles that.

app/web/src/newhotness/Explore.vue
```diff
@@ -927,24 +957,64 @@ const storeViewMode = () => {
...
+
 const getQualificationStatusTitle = (component: ComponentInList) => {
   const status = getQualificationStatus(component);
   switch (status) {
     case "success":
-      return "Passed";
+      return "Passed qualifications";
     case "running":
-      return "Running";
+      return "Running qualifications";
```
nickgerace, line 1012: "Running" is not real and will die in a follow up PR. This is from before qualifications were on the prop tree... might be like... since late 2022 😅. For now, we'll keep it since we discovered this right as we were wrapping it. It is harmless.

app/web/src/newhotness/Explore.vue
```diff
@@ -106,23 +106,26 @@
         </TabGroupToggle>
         <div>
           <DropdownMenuButton
+            v-if="showGrid"
             :class="clsx('rounded')"
```
wendybujalski: Probably do not need to use clsx here!

app/web/src/newhotness/Explore.vue
```diff
@@ -927,24 +957,64 @@ const storeViewMode = () => {
...
+    query,
+  });
+});
+
 const getQualificationStatusTitle = (component: ComponentInList) => {
   const status = getQualificationStatus(component);
   switch (status) {
     case "success":
-      return "Passed";
+      return "Passed qualifications";
```
nickgerace, line 1010: Note: the section titles and icons are intentional and match what we have discussed.

app/web/src/newhotness/Workspace.vue
```diff
@@ -265,6 +265,7 @@ export type SelectionsInQueryString = Partial<{
   map: string;
   grid: string;
   c: string;
+  groupBy: "diff" | "qual" | "upgr";
```
nickgerace, line 268: These are the first four characters of each type. We may go down to one character or change this in the future, but we wanted something that could be backwards compatible to start.

app/web/src/newhotness/Explore.vue
```diff
@@ -927,24 +957,64 @@ const storeViewMode = () => {
...
+    query.groupBy = "qual";
+  } else if (groupBySelection.value === GroupByCriteria.Upgrade) {
+    query.groupBy = "upgr";
+  }
+
+  router.push({
+    query,
+  });
+});
+
```
wendybujalski: Such a clean and elegant watcher!

app/web/src/newhotness/explore_grid/ExploreGrid.vue
```diff
@@ -163,28 +161,49 @@ const gridRows = computed(() => {
     if (!components) continue;
 
     const count = components.length;
+    let collapsed = collapseTracker.value[groupName];
+
+    // Handle the very first time everything is loaded. We want empty sections to begin collapsed
+    // and non-empty sections to be expanded by default.
+    if (collapsed === undefined) {
+      collapsed = count === 0;
+    }
```
nickgerace, line 170: Here's the deal: this was originally a "toggle", but we needed to handle a ternary state. By default, empty sections needed to be collapsed and non-empty sections needed to be expanded. So, this is how we handle that.

app/web/src/newhotness/explore_grid/ExploreGridRow.vue
```diff
@@ -48,6 +50,24 @@
...
+        <Icon name="check-circle-outline" />
+      </div>
+      <span>
+        {{ emptyAreaData?.message ?? "Nothing to see here!" }}
+      </span>
+    </div>
+  </div>
+  <div v-else>
+    <!-- footer area -->
+  </div>
```
nickgerace, line 70: Subtle, but important to have a div here.

app/web/src/newhotness/Explore.vue
vbustamante: fixed! thanks for the reminder

jobelenus (approved):
Excellent router query work

PR #6504, opened 2025-06-25
feat(web): re-use atoms instead of throwing a hammer

zacharyhamm:
Combined with fetching HEAD first, this should improve load times for large workspaces by reusing atoms we have already fetched, and only fetching hammers for what does not exist in the local database at all.

Test this pr by creating a workspace with many change sets. Make some changes in some of the change sets (delete or modify components and connections). Ensure all change sets load. Then drop tables, close all but one tab. Then hard reload. Change set loading should be faster, since atoms will be shared between all change sets.

### DIFF
--- a/app/web/src/workers/webworker.ts
+++ b/app/web/src/workers/webworker.ts
@@ -710,20 +710,20 @@ const initIndexAndChangeSet = (meta: AtomMeta, span?: Span) => {
     throw new Error("Null value from SQL, impossible");
   }
 
-  if (
-    changeSetExists &&
-    meta.fromIndexChecksum &&
-    meta.fromIndexChecksum !== currentIndexChecksum
-  ) {
-    debug("🔥🔥 RAGNAROK", meta.fromIndexChecksum, currentIndexChecksum);
-    // throw new Ragnarok(
-    //   "From Checksum Doesn't Exist",
-    //   workspaceId,
-    //   changeSetId,
-    //   meta.fromIndexChecksum,
-    //   currentIndexChecksum,
-    // );
-  }
+  // if (
+  //   changeSetExists &&
+  //   meta.fromIndexChecksum &&
+  //   meta.fromIndexChecksum !== currentIndexChecksum
+  // ) {
+  //   debug("🔥🔥 RAGNAROK", meta.fromIndexChecksum, currentIndexChecksum);
+  //   // throw new Ragnarok(
+  //   //   "From Checksum Doesn't Exist",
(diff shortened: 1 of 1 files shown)


jobelenus (approved):

app/web/src/workers/webworker.ts
```diff
@@ -1526,23 +1545,52 @@ const niflheim = async (
...
+          insertAtomMTM(
+            {
+              kind,
+              id,
+              toChecksum: checksum,
+              workspaceId,
+              changeSetId,
+              // These are not used in the insert, but the type requires them. So the values here may not be correct
+              fromIndexChecksum: indexChecksum,
+              toIndexChecksum: checksum,
```
britmyerss: I think you'll want to use `indexChecksum` here for both values

britmyerss: in case we do start using them, we set to/from to the same indexChecksum if it doesn't represent a known patch from one to the other (which is true in this case)

jobelenus: IIRC for hammers the to and from index are the same, the `toIndexChecksum` should _definitely_ not be the MV checksum... the `fromIndexChecksum` can also be the same value as `toIndexChecksum`

app/web/src/workers/webworker.ts
```diff
@@ -1432,6 +1432,25 @@ const pruneAtomsForClosedChangeSet = async (
...
+    sql: `select * from atoms where atoms.kind = ? AND atoms.args = ? and atoms.checksum = ?;`,
+    bind: [kind, id, checksum],
+    returnValue: "resultRows",
+  });
+
+  for (const row of rows) {
+    result.push(row[0]);
+  }
+
+  return result.length > 0;
```
jhelwig: Why the manual row count, instead of doing `select count(*) ...` and have the query return it directly?

app/web/src/workers/webworker.ts
```diff
@@ -1432,6 +1434,85 @@ const pruneAtomsForClosedChangeSet = async (
...
+    updateChangeSetWithNewIndex({
+      toIndexChecksum: indexChecksum,
+      changeSetId,
+    });
+  }
+
+  bustCacheAndReferences(workspaceId, changeSetId, kind, id, false, true);
+};
+
+const decodedAtomCache: { [key: string]: AtomDocument } = {};
```
jobelenus, line 1473: we should empty this after we're done using it(?)

PR #6513, opened 2025-06-26
Only enqueue Update Actions during DVU if the Component does not already have any Actions

jhelwig:
Previously, we were only checking to see if the Component had an Update Action queued, to prevent queuing a second Update Action. This would still let us get into situations such as a Component that has a Destroy Action queued, would also end up with an Update Action queued by the DVU job.

We now prevent automatically queuing the Update Action if there are *any* Actions for the Component already, regardless of state or kind. This does not prevent manually queuing any Action kind, so a user would still be able to queue every Action available to a Component at the same time if they desired.

### DIFF
--- a/lib/dal/src/component.rs
+++ b/lib/dal/src/component.rs
@@ -2919,7 +2919,7 @@ impl Component {
     }
 
     /// If the attribute value is somewhere in 'root/domain', the component has a resource, and a single update function,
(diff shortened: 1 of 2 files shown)


stack72 (approved):
We talked this through as the right solution - if a user wants to enqueue multiple actions on the same component they it's an opt-in they will need to make

PR #6514, opened 2025-06-26
Fix: dont mutate when cachedData is undefined. Add first element to empty list.

jobelenus:

There was a TODO here. When HEAD is empty, and you make a new change set, and add a component, it doesn't show up in the list... b/c we didn't add it 😓. This would also crop up when making the first item of any of the big lists..

I was initially here to test & fix "i made a change set and applied it while HEAD was empty, but nothing was showing up after I applied!". **That did in fact work**, I assume Zack's MTM fix resolved that problem (its in the right area)


app/web/src/store/realtime/heimdall.ts
```diff
@@ -182,16 +182,17 @@ const updateCache = (
...
-      return cachedData;
+    if (!cachedData) {
+      return;
     }
 
     if (removed) {
       // Filter out the item if it is removed
       return cachedData.filter((d) => d?.id !== id);
     } else {
+      if (cachedData.length === 0) {
```
jkeiser: You should be able to remove this if statement entirely: if the list is empty, we'll go to the `else` case and it will do the same thing!

app/web/src/store/realtime/heimdall.ts
```diff
@@ -182,16 +182,17 @@ const updateCache = (
   if (!removed && !data) return;
 
   queryClient.setQueryData(queryKey, (cachedData: { id: string }[]) => {
-    // Don't mutate if we haven't got any data yet
-    // TODO it seems wrong not to add items to an empty list ...
-    if (!cachedData || cachedData.length === 0) {
-      return cachedData;
+    if (!cachedData) {
+      return;
```
jkeiser: This needs to return cachedData, right? That's at least how their examples worked--if you don't modify it you return it. (I guess the code you are writing would return undefined, but the intent is "leave it unchanged" rather than "set it to undefined.")

jkeiser (changes requested):

jkeiser:
Thanks for fixing this, I forgot I had written the comment by the time I put up the PR!

app/web/src/store/realtime/heimdall.ts
jobelenus: sure.. in this case `cachedData` _is_ undefined.. but i can be more explicit.

app/web/src/store/realtime/heimdall.ts
jobelenus: ah, right... since we removed the `length === 0` check above

jkeiser (approved):

PR #6515, opened 2025-06-26
feat(web): don't run coldStartComputed. Never decode the same atom twice

zacharyhamm:
Use the strategy used when re-using existing atoms to process *all* atoms from the index, instead of running the `coldStartComputed` promise. 

Chunk the atoms to make fewer sqlite3 queries.

Attempt to cache every document in a QuickLRU cache, to avoid JSON deserialization on the same document.


zacharyhamm:
Curious what we think about this change, specifically dropping coldStartComputed.

PR #6518, opened 2025-06-26
feat(web) - add a management connection via the ManagementPanel

wendybujalski:

This wraps up ENG-3083 so that the Management panel within ComponentDetails can be used to add a management connection in addition to displaying the existing management connections and functions.

The input field for selecting a component to manage is similar to but not the same as AttributeInput. Eventually those two components might have some of their parts merged into a general fuzzy search input component.

There are also a few small bug fixes and improvements wrapped into this PR as well, will point em out in the comments!

#### Screenshots:

Screenshot 2025-06-26 at 5 30 00 PM

#### Out of Scope:

The data loaded into the frontend about management connections is still not always accurate.


app/web/src/components/WorkspaceImportModal.vue
```diff
@@ -223,7 +223,7 @@
             <LoadingMessage>
               Importing your workspace!
               <template #moreContent>
-                <p class="italic font-sm">
+                <p class="italic text-sm">
```
wendybujalski, line 226: "font-sm" doesn't do anything! The correct Tailwind class is "text-sm"

app/web/src/newhotness/DeleteModal.vue
```diff
@@ -90,14 +91,14 @@ async function onConfirm() {
 }
 
 onMounted(() => {
-  window.addEventListener("keydown", onKeyDown);
+  keyEmitter.on("Enter", onKeyDown);
 });
 onBeforeUnmount(() => {
-  window.removeEventListener("keydown", onKeyDown);
+  keyEmitter.off("Enter", onKeyDown);
```
wendybujalski, line 97: We want to use the centralized emitters instead of creating additional addEventListeners

app/web/src/newhotness/logic_composables/emitters.ts
```diff
@@ -75,3 +75,36 @@ export const startWindowResizeEmitter = (window: Window) => {
...
+export const mouseEmitter: Emitter<MouseDetails> = mitt<MouseDetails>();
+
+let mouseEmitterStarted = false;
+
+export const startMouseEmitters = (window: Window) => {
+  if (mouseEmitterStarted) return;
+  mouseEmitterStarted = true;
+  window.addEventListener("click", onClick);
+  window.addEventListener("mousedown", onMouseDown);
+};
```
wendybujalski, line 110: New mouseEmitter for centrally handling click and mousedown events on the whole window.

app/web/src/newhotness/layout_components/SecretInput.vue
```diff
@@ -14,7 +14,7 @@
       ref="inputRef"
       :class="
         clsx(
-          'block h-lg w-full ml-auto font-sm font-mono',
+          'block h-lg w-full ml-auto text-sm font-mono',
```
wendybujalski, line 17: Once again, "font-sm" doesn't actually do anything 😅

app/web/src/newhotness/layout_components/ComponentSecretAttribute.vue
```diff
@@ -322,30 +323,33 @@ const openSecretForm = () => {
     }
   });
 };
+const closeSecretForm = () => {
+  secretFormOpen.value = false;
+  removeListeners();
```
wendybujalski, line 328: Fixed a bug where the listeners were not always being removed.

app/web/src/newhotness/ManagementConnectionInput.vue
wendybujalski, line 1: The guts of this component are definitely very similar to AttributeInput, but there are enough little differences to make merging them not worth it (for now).

app/web/src/newhotness/Explore.vue
```diff
@@ -860,7 +862,7 @@ const onResize = () => {
 
 // general click handler for the whole page
 // any click which doesn't do this behavior should have .stop on it!
-const onClick = (e: MouseEvent) => {
+const onClick = (e: MouseDetails["click"]) => {
```
jobelenus, line 865: Oooo

wendybujalski: Yeah, no reason to not apply our same "use one central event listener that emits events instead of a bunch of individual ones" structure for mouse events. Right now it's only "mousedown" and "click" but adding additional events to the same mouseEmitter when we need em should be easy :)

PR #6519, opened 2025-06-26
fix(web): Show func run status even under heavy load

jkeiser:
If there are a ton of functions running (which happens frequently), we never show func run status because we keep cancelling the current query each time we get a FuncRunLogUpdated event.

This PR:
* Don't send a trillion fetches to SDF--only once every 500ms
* If status updates during a fetch, refetch *after* the current request finishes instead of cancelling

#### Out of Scope:

* There is still a delay of 500ms when we first. We can't get rid of that until we fix the fact that FuncRunLogUpdated comes in *before* the func run status is available (we may need to create--or listen to--a different event).
* Does not fix the fact that the "5s ago" text in the func run status never updates over time

## How was it tested?

- [X] Make sure UI shows func run log status under load (3 component creates per second)
- [X] Confirmed we only send one request to SDF every 50ms under load
- [X] Confirmed that we still refetch if FuncRunLogUpdate comes in while we're fetching

### DIFF
--- a/app/web/src/newhotness/Workspace.vue
+++ b/app/web/src/newhotness/Workspace.vue
@@ -87,6 +87,7 @@ import {
   Ref,
   inject,
 } from "vue";
+import * as _ from "lodash-es";
 import { useQuery, useQueryClient } from "@tanstack/vue-query";
 import NavbarPanelLeft from "@/components/layout/navbar/NavbarPanelLeft.vue";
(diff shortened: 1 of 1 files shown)


jobelenus (approved):
Yep. This is great to ship. Brit has already made an attempt to get us more func run messages besides just "update"

aaron-dernley (dismissed):

PR #6520, opened 2025-06-26
fix(luminork): Ensure Edda knows about a new change set as per SDF behaviour

stack72:


fnichol (approved):
Looks *exactly* like what I was going to do, excellent work!

PR #6521, opened 2025-06-26
fix(luminork): Support "HEAD" in luminork URLs

jkeiser:
This supports the word "head" in the luminork URL instead of a change set id. e.g. `GET /v1/w/<id>/change-sets/head/components`

It changes `ChangeSetDalContext` extractor; any endpoint that uses that will . The word "head" is matched case-insensitive, so HEAD works too.

#### Out of Scope:

Some SDF endpoints now support head, because it is in a common extractor. Not all do, however, because they don't all use that extractor yet. This does not try to change SDF, but it seems just fine to allow at least some endpoints to support the new pattern.

## How was it tested?

- [X] Integration tests pass
- [X] Manual API test for head, HEAD, <change set id of head>, <other change set id>: get list of components, make sure it is from the right change set
- [X] Manual test: old UI can create and switch change sets, create components and view the workspace
- [X] Manual test: 400 if you use "blah" as change set id

### DIFF
--- a/lib/luminork-server/src/service/v1/workspaces.rs
+++ b/lib/luminork-server/src/service/v1/workspaces.rs
@@ -18,7 +18,7 @@ use super::common::ErrorIntoResponse;
 use crate::{
     AppState,
     extract::{
-        change_set::TargetChangeSetIdFromPath,
+        change_set::TargetChangeSetIdentFromPath,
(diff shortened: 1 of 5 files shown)


stack72 (approved):
This works incredibly well! Thank you :)

PR #6523, opened 2025-06-27
Add sort by functionality to new UI

nickgerace:
## Introduction

This change adds sort by functionality to the new UI. It uses the guts of the group by functionality to do so, but with some twists.

## Description

There are three sorts: latest to oldest, failed actions and running actions. By default, we will use latest to oldest, which sorts by each component's approximate creation time and uses their ULID-based IDs for sorting. That sort is performed before performing another sort so that we ensure relative ordering always makes sense.

Both the failed actions sort and the running actions sort leverage the pre-existing "actionViewList" in "Explore.vue". Originally, this change used an elaborate multi-point "component action state" that was summary of a component's actions. Not only was that complex and fragile, but it was nightmarish to read and arguably not essential. Now, there are two simple sets to cover "does this component have at least one failed action?" and "does this component have at least one running action?". These sets are arguably more correct anyway as they aren't opinionated summaries of the components and their actions.

The sort by feature was designed to be used with group by or in isolation. In doing so, it uses the deep linking system from the group by feature and expands deep linking strings for all selections to be full names rather than the first four characters of each selection. This should be easier for readability.

## Screenshot of VPC Template Created via API with Actively Sorting "Running Actions"

Screenshot 2025-06-27 at 1 16 38 AM

## Audits Trail Post Deletion (Just for fun!)

Screenshot 2025-06-27 at 1 26 47 AM

### DIFF
--- a/app/web/src/newhotness/Explore.vue
+++ b/app/web/src/newhotness/Explore.vue
@@ -104,9 +104,8 @@
             />
           </template>
         </TabGroupToggle>
-        <div>
+        <div v-if="showGrid" class="flex flex-row gap-xs">
           <DropdownMenuButton
-            v-if="showGrid"
             class="rounded"
             :options="groupByDropDownOptions"
             :modelValue="groupBySelection"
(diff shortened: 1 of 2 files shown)


app/web/src/newhotness/Explore.vue
```diff
@@ -451,6 +473,25 @@ const actionViewList = computed(
   () => actionViewListRaw.data.value?.actions ?? [],
 );
 
+const componentsWithFailedActions = computed(() => {
```
nickgerace, line 476: These can be combined into one to prevent two O(n) walks. Probably an interface with two sets.

Marking for the next PR. I don't think it blocks this one.

PR #6524, opened 2025-06-27
fix(web): Remove incorrect default of all passing for qualifications

stack72:
When we created a component in the new UI, we checked for failing or warning qualifications and if not found then we defaulted to `all passed`.

For new components, there will be no passing, no failing and no warning. So if we find that seet of groups then we default to `unknown` as the status is actually calculating the results.

This makes the flow that when you create a component that te qualifications show unknown until they work as expected

https://jam.dev/c/d382da8c-b166-4508-bbbc-449c69966b0c

### DIFF
--- a/app/web/src/newhotness/explore_grid/ExploreGridTile.vue
+++ b/app/web/src/newhotness/explore_grid/ExploreGridTile.vue
@@ -54,6 +54,25 @@
         <TextPill
           v-if="
             component.qualificationTotals.failed === 0 &&
+            component.qualificationTotals.warned === 0 &&
+            component.qualificationTotals.succeeded === 0
+          "
+          tighter
+          :class="
+            clsx(
+              'text-xs ml-auto',
+              themeClasses(
+                'border-neutral-500 bg-neutral-100 text-black',
+                'border-neutral-600 bg-neutral-900 text-white',
+              ),
+            )
+          "
+        >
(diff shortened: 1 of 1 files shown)


app/web/src/newhotness/explore_grid/ExploreGridTile.vue
```diff
@@ -85,7 +104,7 @@
           {{ component.qualificationTotals.failed }} Failed
         </TextPill>
         <TextPill
-          v-if="component.qualificationTotals.warned > 0"
+          v-else-if="component.qualificationTotals.warned > 0"
```
wendybujalski, line 107: I think technically this could just be a v-else assuming that the totals can never be negative.

app/web/src/newhotness/explore_grid/ExploreGridTile.vue
```diff
@@ -53,6 +53,25 @@
         <div>Qualifications</div>
         <TextPill
           v-if="
+            component.qualificationTotals.failed === 0 &&
+            component.qualificationTotals.warned === 0 &&
+            component.qualificationTotals.succeeded === 0
```
wendybujalski, line 58: Might make sense to also catch any of the totals being negative here, which _should_ never happen but if it does then "Unknown" would definitely be the right status to show!

PR #6528, opened 2025-06-27
Remove running qualifications from new UI

nickgerace:
## Description

This change removes running qualifications from the new UI. This is a very old concept that predates qualifications on the prop tree (late 2022!) and has been moved around over the years. DVU is the canonical source of prop tree functions in flight and qualification statuses are reflections of the result of DVU.

As a consequence of this work, the domain naming has been standardized across the newhotness Vue components.

An additional change here is that the two sets for action state for sort by functionality have been combined under one interface to only loop over the action list once.

Finally, there's now a comment clarifying why the "canUpgrade" variable was created in a previous PR for the grid.

### DIFF
--- a/app/web/src/newhotness/Explore.vue
+++ b/app/web/src/newhotness/Explore.vue
@@ -444,6 +444,8 @@ const upgrade = useUpgrade();
 const upgradeableComponents = computed(() => {
   const set: Set<ComponentId> = new Set();
   for (const component of filteredComponents) {
+    // This needs to be split out into a variable for reactivity. Keep this here or drown in
+    // sorrow and suffering. Relevant pull request: https://github.com/systeminit/si/pull/6483
     const canUpgrade = upgrade(
       component.schemaId,
       component.schemaVariantId,
@@ -473,23 +475,25 @@ const actionViewList = computed(
   () => actionViewListRaw.data.value?.actions ?? [],
 );
 
-const componentsWithFailedActions = computed(() => {
-  const componentIds: Set<ComponentId> = new Set();
+interface ComponentsHaveActionsWithState {
+  failed: Set<ComponentId>;
+  running: Set<ComponentId>;
+}
+
+const componentsHaveActionsWithState = computed(() => {
+  const results: ComponentsHaveActionsWithState = {
+    failed: new Set(),
+    running: new Set(),
+  };
   for (const action of actionViewList.value) {
-    if (action.componentId && action.state === ActionState.Failed) {
-      componentIds.add(action.componentId);
+    if (!action.componentId) continue;
+    if (action.state === ActionState.Failed) {
+      results.failed.add(action.componentId);
+    } else if (action.state === ActionState.Running) {
+      results.running.add(action.componentId);
     }
   }
-  return componentIds;
-});
-const componentsWithRunningActions = computed(() => {
-  const componentIds: Set<ComponentId> = new Set();
-  for (const action of actionViewList.value) {
-    if (action.componentId && action.state === ActionState.Running) {
-      componentIds.add(action.componentId);
-    }
-  }
-  return componentIds;
+  return results;
 });
 
 const kind = computed(() =>
@@ -531,7 +535,7 @@ const groupedComponents = computed(() => {
     const failed = [];
     const theRest = [];
     for (const component of components) {
-      if (componentsWithFailedActions.value.has(component.id)) {
+      if (componentsHaveActionsWithState.value.failed.has(component.id)) {
         failed.push(component);
       } else {
         theRest.push(component);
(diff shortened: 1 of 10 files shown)


app/web/src/newhotness/Explore.vue
```diff
@@ -1094,14 +1098,12 @@ watch([groupBySelection], () => {
...
-      return "Passed qualifications";
-    case "running":
-      return "Running qualifications";
+    case "failure":
+      return "Failed qualifications";
     case "warning":
       return "Warnings";
     default:
-      return "Failed qualifications";
+      return "Success qualifications";
```
nickgerace, line 1106: The pattern has been to default to success for our switch and "if-else" statements. This is just aligning to that and is merely cosmetic.

app/web/src/newhotness/Explore.vue
```diff
@@ -473,23 +475,25 @@ const actionViewList = computed(
...
-  return componentIds;
-});
-const componentsWithRunningActions = computed(() => {
-  const componentIds: Set<ComponentId> = new Set();
+interface ComponentsHaveActionsWithState {
+  failed: Set<ComponentId>;
+  running: Set<ComponentId>;
+}
+
+const componentsHaveActionsWithState = computed(() => {
```
nickgerace, line 483: Now we only do one O(n) walk rather than two.

app/web/src/newhotness/QualificationPanel.vue
```diff
@@ -82,34 +84,34 @@ const validations = computed(() => {
   return avsWithValidation;
 });
 
-// TODO(Wendy) - this is very annoying!
-// Why are statuses sometimes capitalized and sometimes not?!?!?!
-const fixStatus = (status: string): QualificationStatus => {
+const convertValidationStatusToQualificationStatus = (
+  status: ValidationOutputStatus,
+): QualificationStatus => {
   switch (status) {
```
nickgerace, line 90: This is annoying, but it is technically a different concept. We should look at these after we switch to the new UI because a lot of these status types are old.

app/web/src/newhotness/QualificationView.vue
```diff
@@ -5,12 +5,22 @@
     </div>
 
     <div class="flex flex-col gap-xs text-sm">
-      <StatusMessageBox :status="qualificationStatus" class="break-all">
+      <StatusMessageBox
+        :status="qualificationStatus ?? 'unknown'"
```
nickgerace, line 9: Technically, we cannot assume it is running if it never has ran potentially. I don't think this will ever happen unless the function execution layer is broken, but we should be prepared.

app/web/src/newhotness/QualificationView.vue
```diff
@@ -5,12 +5,22 @@
...
+        class="break-all"
+      >
         <template v-if="qualificationStatus === 'success'"> Passed! </template>
-        <template v-else-if="qualificationStatus === 'failure'">
+        <template
+          v-else-if="
+            qualificationStatus === 'failure' ||
+            qualificationStatus === 'warning'
```
nickgerace, line 16: Warning should behave similarly to failure here

app/web/src/newhotness/QualificationView.vue
```diff
@@ -5,12 +5,22 @@
...
+          "
+        >
           {{ qualification.message }}
         </template>
-        <template v-else> Qualification running, standby...</template>
+        <template v-else>
+          The qualification has not yet ran or is actively running. Standby...
+        </template>
```
nickgerace, line 23: Don't assume what's happening, but still be clear to the user

app/web/src/newhotness/QualificationView.vue
```diff
@@ -52,22 +62,22 @@ import * as _ from "lodash-es";
...
 }>();
 
 const showDetails = ref(false);
 
 const qualification = toRef(props, "qualification");
 
 const qualificationStatus = computed(() => {
-  if (_.isNil(props.qualification.result)) return "running";
-  return props.qualification.result;
+  if (_.isNil(props.qualification.status)) return undefined;
```
nickgerace, line 79: We cannot assume this means running nor will this ever really happen unless something is broken. `undefined` (or `null`) is more accurate.

app/web/src/newhotness/QualificationPanel.vue
```diff
@@ -18,13 +18,15 @@ import {
...
 import { findAvsAtPropPath } from "./util";
 
-export interface QualItem {
+export type QualificationStatus = "success" | "failure" | "warning";
+
+export interface Qualification {
   name?: string;
   message?: string;
-  result?: QualificationStatus;
+  status?: QualificationStatus;
```
nickgerace, line 29: Aligning with other language.

wendybujalski (approved):

PR #6531, opened 2025-06-27
chore(dal): Specify management functions by name in tests

jkeiser:
This is a small test refactor PR to continue the pattern of using names for things in tests, which is working very nicely for readability:

* Lets `component::execute_management_func()` specify the function by name (you can still use ID)
* Makes you specify management func name in `variant::create_management_func()` so it can be referred to
* Adds `id()` test funcs like `component::id(ctx, "ComponentName")`, just to make it less of an eyesore in the few tests that use it
* Adds a couple of #[ignored] tests I am using to fix an actual bug in a later PR :)

The `ComponentKey` (etc.) stuff is where I've been prototyping ways to specify stuff with nice names, and often later finds its way into actual endpoint JSON specifications, so it's worth getting right :)

## How was it tested?

- [X] Integration tests pass


stack72:
I've tested this locally and we get a successful set of components created from it!

PR #6539, opened 2025-06-30
Add ability to pin a component

nickgerace:
## Introduction

This change adds the ability to pin a component. Pinning a component results in two groups being shown underneath it: its incoming and outgoing connecting components. The pinned component is placed at the top of the grid and can be focused.

## More Information

There are two ways to interact with a pinned component via the keyboard. You can press "ENTER" to navigate to it or you can press "P" to unpin it. Conversely, pinning a component can be done by pressing "P" or using the component context menu. Note: the component context menu is disabled for pinned components and "group by" is disabled when pinning.

As a consequence of the primary change of this commit, component qualification status badging and iconography has been moved into its own Vue component. Not only does it make it reusable for the grid tile and the pinned component card, but it also allows these details to be available in the delete modal.

## Other Changes

- Add unfocusing when using group by, sort by and pinning due to invalid indexing
- Remove "Standby..." from the qualification helper text
- Remove leftover "Running qualifications" switch check from grid row
- Fix qualification status missing "unknown" for qualification view

## Known Missing Feature

We did not add the pin icon on the upper right of the grid tiles themselves.

## Screenshot: Pinned Component

Screenshot 2025-06-30 at 6 44 26 PM

## Screenshot: Delete Modal with Qualification Status

Screenshot 2025-06-30 at 6 44 40 PM

## Screenshot: Pinning in Keyboard Controls

Screenshot 2025-06-30 at 7 38 04 PM

### DIFF
--- a/app/web/src/newhotness/Explore.vue
+++ b/app/web/src/newhotness/Explore.vue
@@ -114,6 +114,7 @@
             checkable
             alwaysShowPlaceholder
             highlightWhenModelValue
+            :disabled="pinnedComponentId !== undefined"
             @update:modelValue="
               (val) => (groupBySelection = groupByFromString(val))
             "
@@ -165,6 +166,7 @@
           :components="groupedComponents"
           :focusedComponentIdx="focusedComponentIdx"
           @childClicked="componentClicked"
+          @unpin="() => (pinnedComponentId = undefined)"
           @scrollend="fixContextMenuAfterScroll"
           @scroll="onScroll"
         />
@@ -270,6 +272,7 @@
       onGrid
       enableKeyboardControls
       @edit="navigateToFocusedComponent"
+      @pin="(c) => (pinnedComponentId = c)"
     />
   </section>
 </template>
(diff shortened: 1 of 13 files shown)


app/web/src/newhotness/Explore.vue
```diff
@@ -522,6 +561,82 @@ const componentList = computed(() => componentListRaw.data.value ?? []);
...
+  // Make sure we have the entire component and not just the ID. Why? The URL could contain a query
+  // for a pinned component that no longer exists.
+  enabled: () => pinnedComponent.value !== undefined,
+  queryKey: key(EntityKind.OutgoingConnections),
+  queryFn: async () => {
+    if (!pinnedComponentId.value) return [];
+    const byComponents = await getOutgoingConnections(
+      args(EntityKind.OutgoingConnections),
+    );
+    const mine = byComponents.get(pinnedComponentId.value);
```
jobelenus, line 585: This part *has* to change... follow this pattern: https://github.com/systeminit/si/blob/main/app/web/src/newhotness/ConnectionsPanel.vue#L77

The problem here is that the GLOBAL `queryKey: key(EntityKind.OutgoingConnections),` now gets associated with ONLY the outgoing for the specific pinned component id _in the entire app_

PR #6541, opened 2025-06-30
fix(web): Don't allow renaming HEAD

stack72:


app/web/src/components/layout/navbar/ChangeSetPanel.vue
```diff
@@ -147,10 +147,17 @@ const selectedChangeSetName = computed(
...
-  ..._.map(openChangeSets.value, (cs) => ({ value: cs.id, label: cs.name })),
+  ..._.map(openChangeSets.value, (cs) => ({
+    value: cs.id,
+    label: cs.name,
+  })),
   // { value: "NEW", label: "+ Create new change set" },
 ]);
 
+const calculateShowSecondaryAction = (option: { label: string }) => {
+  return option.label.toUpperCase() !== "HEAD";
```
vbustamante, line 158: We could also check the value instead of the label, since that's the ID and we can figure out the HEAD ID, but this works!

PR #6549, opened 2025-07-01
Fix: possible connections invalidations

jobelenus:
- First, we can't add "extras" to `queryKey`—because the web worker has no way of knowing how to bust it
- Because we can't add extras, we cannot account for the last 2 arguments of "what do i want a suggestion for???"...
    - send the full possible list over the boundary
    - do the "what order ought things to be in" based on the prop, within the component itself.
- Third, we missed the web worker cache bust on the `PossibleConnections`, it was set to bust on `ChangeSetId`, needed to be `WorkspaceId`.

I screwed up what goes on what side of the boundary in my first WIP on this one... which led to folks adding "extras" to force it to work—which means busting fails.

### Testing steps
1. On HEAD, I have Cred, Region, and EC2::Instance w/ resource all linked together
2. Create a change set
3. Make another EC2::Instance
4. Name it `Deleted`
5. Navigate to the first EC2::Instance
6. Open `AttributeInput`, see `Deleted` in the list of options
7. Back to `Explore`
8. Right Click & Erase the `Deleted` EC2::Instance
9. Go back to the first EC2::Instance
10. Open `AttributeInput`, no more `Deleted`!!

To ensure that the "suggested props" are ordered correctly... I compare the order of the list on any input vs the `Region` input which always puts the region first.

### DIFF
--- a/app/web/src/newhotness/layout_components/AttributeInput.vue
+++ b/app/web/src/newhotness/layout_components/AttributeInput.vue
@@ -547,6 +547,7 @@ import {
 } from "@/store/realtime/heimdall";
 import CodeViewer from "@/components/CodeViewer.vue";
 import { PropKind } from "@/api/sdf/dal/prop";
+import { CategorizedPossibleConnections } from "@/workers/types/dbinterface";
 import {
   attributeEmitter,
   MouseDetails,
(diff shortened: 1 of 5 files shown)


britmyerss (approved):
ooo great find!

PR #6554, opened 2025-07-01
feat(web): Ensuring we store the viewId in the queryString

stack72:
We will not store the viewId if it’s default or all views - but we will if it’s a custom view and we will load the correct view from that queryString on deep linking

### DIFF
--- a/app/web/src/newhotness/Explore.vue
+++ b/app/web/src/newhotness/Explore.vue
@@ -10,13 +10,13 @@
         <DropdownMenuButton
           class="rounded min-w-[128px]"
           :options="viewListOptions"
-          :modelValue="selectedView"
+          :modelValue="selectedViewId"
           minWidthToAnchor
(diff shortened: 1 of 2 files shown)


app/web/src/newhotness/Explore.vue
```diff
@@ -424,14 +423,35 @@ const viewListOptions = computed(() => {
...
+  const query: SelectionsInQueryString = {
+    ...router.currentRoute.value?.query,
+  };
+  delete query.map;
+  delete query.viewId;
+  query.grid = "1";
+  if (
+    selectedViewId.value !== "" &&
+    selectedViewId.value !== defaultView.value?.id
+  ) {
```
nickgerace, line 447: Nice. Now I know why it was showing an empty query all the time earlier.

nickgerace (approved):
Great work!

PR #6556, opened 2025-07-01
Feat/add cypress remote envs

johnrwatson:
Adds the ability to run cypress tests against a remote node in AWS orchestrated via SI, pretty cool!

**This is the workflow:**
- User selects ec2-node as environment in the github workflow (Cypress E2E Tests)
- SI is invoked to create a change set and launch a machine via an apply that will run the SI Stack of main
- The remote machine is SSH tunnelled to present localhost:3020 (bedrock) and localhost:8080 (webapp)
- Bedrock seeds the `CI Remote Execution Environment` workspace via a restore point `W=01JYPR32SD5RKR3AMG298J7263-CS=01JZ3W5XX6QHQZ6PYSBHK4SB3K (39 components)` that has some failed actions and components on HEAD
- Services serve
- CI detects web app and continues to execution the suite of cypress tests against the remote node

**Yep, it's the disclaimer (or bonus information!) section**
- This whole code fork isn't used anywhere, it's currently only manually selected
- Branch-selection isn't supported yet, but the framework is there to support it
- It can only run with a concurrency of one, as we have no way of merging multiple change sets at once (ideally each github pipeline would be able to merge it's own Change Set, but this is fraught with difficulty at the minute)
- Automatic cleanup will come sharpish, this can cost a lot if we aren't careful
- I wish I used tailscale but I got stuck in a `tailscale serve` groove and couldn't get over some "workspace url needs to be static for each test" problems
- SSH key needed is in the Shared Engineering Vault, it's literally only used for these machines, for debugging we can connect directly from local to the execution machines via the same portforwarding or SSH methods
- It's a ~12 minute compile from scratch at the minute with the machine type selected

<div></div>

### DIFF
--- a/.ci/provision.sh
+++ b/.ci/provision.sh
@@ -1,0 +1,56 @@
+#!/bin/bash
+set -euxo pipefail
+
+# 0. Remove broken EC2 repo if it exists
+if grep -q "\[ec2\]" /etc/pacman.conf; then
+  echo "[INFO] Removing broken 'ec2' repo from /etc/pacman.conf"
+  sed -i '/^\[ec2\]/,/^$/d' /etc/pacman.conf
+fi
+
+# 1. Update base packages & generate keychain
+pacman -Sy --noconfirm
+pacman-key --init
+pacman-key --populate archlinux
+
+# 2. Install tools:
+pacman -S --noconfirm base-devel git wget postgresql docker
+
+# 2(a) install yay for aur repo usage for us to grab ssm-agent
+#git clone https://aur.archlinux.org/yay.git
+#makepkg -si --noconfirm -D ./yay/
+#yay -Sy amazon-ssm-agent --noconfirm
+
+# (b) Enable and start SSM Agent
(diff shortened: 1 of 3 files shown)


.ci/provision.sh
```diff
@@ -0,0 +1,56 @@
...
+pacman-key --init
+pacman-key --populate archlinux
+
+# 2. Install tools:
+pacman -S --noconfirm base-devel git wget postgresql docker
+
+# 2(a) install yay for aur repo usage for us to grab ssm-agent
+#git clone https://aur.archlinux.org/yay.git
+#makepkg -si --noconfirm -D ./yay/
+#yay -Sy amazon-ssm-agent --noconfirm
```
johnrwatson, line 21: This is commented out as I was thinking about using SSM agent for the portforwarding, but decided to leave it here in case I change my mind. It'd certainly be better than a static SSH key

stack72 (approved):

PR #6558, opened 2025-07-01
Fix component list and connections in Explore

nickgerace:
## Introduction

Welcome. This is "one of those" changes. Buckle up.

## Description

Here's the deal: I originally set out to just fix connections for pinning. As I clawed at every little bug fix, the rabbit hole got deeper. I fixed the connections when pinning! Wait, now the indexes are messed up. I fixed the indexes sometimes. Wait, it still happens. Okay those look better, but... oh DEAR GOD THE COMPONENT CONTEXT MENU IS FLYING OFF THE SCREEN.

There turned out to be something evil lurking within...the component list, filtered component list, and grouped component list all did not account for pinning. Sure, they accounted for views, but not pinning. For indices, tabbing, component context menu and general ease of use, you need your component list to be absolute. You cannot just drop components from that list when grouping.

I combed all of Explore and now my brain is fried. It is happy, but fried. Each section is now titled in the file and there is a clear split between determining the absolute component list and performing filtering, sorting and grouping of that list. More stuff in the logic composables directory in the future? Likely yes.

Now, the original guts of this change are still here, despite all the drama. That change fixes connections for pinning and makes connections reusable for multiple contexts. Now, the connections panel and pinned connections use the same underlying logic. Alongside that work, we now bust the output count alongside the outgoing connections to ensure both are working in tandem with one another.

All of these changes result in grid that is much more trustworthy and reliable when working with pinning and connections.

### DIFF
--- a/app/web/src/newhotness/Explore.vue
+++ b/app/web/src/newhotness/Explore.vue
@@ -163,7 +163,7 @@
         <ExploreGrid
           v-else
           ref="exploreGridRef"
-          :components="groupedComponents"
+          :components="sortedAndGroupedComponents"
           :focusedComponentIdx="focusedComponentIdx"
           @childClicked="componentClicked"
           @unpin="() => (pinnedComponentId = undefined)"
@@ -299,22 +299,19 @@ import {
   Icon,
 } from "@si/vue-lib/design-system";
 import clsx from "clsx";
-import { useQuery, useQueryClient } from "@tanstack/vue-query";
+import { useQuery } from "@tanstack/vue-query";
 import { Fzf } from "fzf";
 import {
   bifrost,
   bifrostList,
   useMakeArgs,
   useMakeKey,
-  getOutgoingConnections,
 } from "@/store/realtime/heimdall";
 import {
   BifrostActionViewList,
   EntityKind,
   ComponentInList,
   View,
-  Connection,
-  IncomingConnections,
 } from "@/workers/types/entity_kind_types";
 import RealtimeStatusPageState from "@/components/RealtimeStatusPageState.vue";
 import { ComponentId } from "@/api/sdf/dal/component";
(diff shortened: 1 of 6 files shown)


stack72 (approved):
I was able to run this locally and it has fixed the woes I was seeing yesterday with the outgoing connections getting reset

app/web/src/newhotness/explore_grid/ExploreGrid.vue
```diff
@@ -296,6 +296,8 @@ const virtualizerOptions = computed(() => ({
   // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
   getScrollElement: () => scrollRef.value!,
   estimateSize: (i: number) => rowHeights.value[i] ?? 0,
+  // The item key is essential for reactivity and was found by @vbustamante and his valiant
+  // efforts. Without this, the virtualizer will not re-compute, even with new components.
```
jobelenus, line 300: I'd like to better understand the issue this solves. Good work!

app/web/src/newhotness/Explore.vue
```diff
@@ -852,20 +837,22 @@ watch([focusedComponentIdx], () => {
   }
 });
 
-const focusedGridComponent = computed(() =>
+const focusedGridComponentRef = computed(() =>
   focusedComponentIdx.value > -1
-    ? exploreGridRef.value?.getGridComponentByIndex(focusedComponentIdx.value)
+    ? exploreGridRef.value?.getGridComponentRefByIndex(
```
jobelenus, line 842: Nick you suggested last night that this is what was missing for your "flying context menu" but... diff suggests it was here the whole time??

jobelenus (approved):

app/web/src/newhotness/explore_grid/ExploreGrid.vue
nickgerace: This is actually @vbustamante's work from a week ago. I think it the rows are the same length, but everything is re-sorted that's passed into the props, you can sometimes end up in a situation where the rows aren't re-sorted too.

app/web/src/newhotness/Explore.vue
nickgerace: This is just a rename

nickgerace: It was the component list not accounting for pinning that was the issue. All the indices were frequently messed up or could become inconsistent with enough iterations of using pinning.

PR #6561, opened 2025-07-02
fix(web): queries should not be enabled before init completion, or while the lobby is visible

zacharyhamm:
This should resolve some issues where the ApplyChangeSetButton is making queries too soon.


jobelenus (approved):

zacharyhamm:
@jobelenus or @wendybujalski the linter, of course, had to be made happy.

jobelenus (dismissed):

PR #6564, opened 2025-07-02
feat(web): replace delayed loader with an animated skeleton

zacharyhamm:
Shimmering "skeletons!".

Currently only mimics the component grid.

### DIFF
--- a/app/web/src/newhotness/skeletons/DelayedSkeleton.vue
+++ b/app/web/src/newhotness/skeletons/DelayedSkeleton.vue
@@ -1,0 +1,25 @@
+<template>
+  <div v-if="show">
+    <ExploreGridSkeleton />
+  </div>
+</template>
+
+<script setup lang="ts">
+import { onMounted, ref } from "vue";
+import ExploreGridSkeleton from "./ExploreGridSkeleton.vue";
+
+/**
+ * The idea of this component is that, we only want to show a big loader
+ * when folks are actually waiting a significant amount of time
+ *
+ * We don't want to "flash" a loader at them for a few frames bc the tanstack
+ * query took ~90ms to return...
+ */
+const DELAY = 300; // 300ms
+const show = ref(false);
+onMounted(() => {
+  setTimeout(() => {
+    show.value = true;
+  }, DELAY);
+});
+</script>
--- a/app/web/src/newhotness/skeletons/ExploreGridSkeleton.vue
+++ b/app/web/src/newhotness/skeletons/ExploreGridSkeleton.vue
@@ -1,0 +1,279 @@
+<template>
+  <section
+    :class="
+      clsx('grid h-full explore', themeClasses('bg-white', 'bg-neutral-950'))
+    "
+  >
+    <div
+      :class="
+        clsx(
+          'main pt-xs flex flex-col gap-xs items-stretch [&>div]:mx-[12px]',
+          themeClasses('bg-white', 'bg-neutral-950'),
+        )
+      "
+    >
+      <!-- view dropdown / search bar -->
+      <div class="flex-none flex flex-row items-center gap-xs">
+        <!-- View dropdown skeleton -->
+        <div
+          :class="
+            clsx(
+              'rounded min-w-[128px] h-8 skeleton-shimmer',
+              themeClasses(
+                'bg-neutral-200 border border-neutral-400',
+                'bg-neutral-700 border border-neutral-600',
+              ),
+            )
+          "
+        ></div>
+
+        <!-- search bar -->
+        <div
+          :class="
+            clsx(
+              'rounded grow h-8 skeleton-shimmer flex items-center px-2 gap-2',
+              themeClasses(
+                'bg-neutral-200 border border-neutral-400',
+                'bg-neutral-700 border border-neutral-600',
+              ),
+            )
+          "
+        ></div>
+      </div>
+
+      <div class="flex-none flex flex-row items-center gap-xs justify-between">
+        <!-- grid/map toggle -->
+        <div class="flex flex-row gap-1">
+          <div
+            :class="
+              clsx(
+                'px-3 py-1 rounded h-8 skeleton-shimmer min-w-[128px]',
+                themeClasses('bg-neutral-200', 'bg-neutral-700'),
+              )
+            "
+          ></div>
+        </div>
+
+        <div class="flex flex-row gap-xs">
+          <!-- group by dropdown skeleton -->
+          <div
+            :class="
+              clsx(
+                'rounded min-w-[100px] h-8 skeleton-shimmer',
+                themeClasses(
+                  'bg-neutral-200 border border-neutral-400',
+                  'bg-neutral-700 border border-neutral-600',
+                ),
+              )
+            "
+          ></div>
+
+          <!-- sort by dropdown skeleton -->
+          <div
+            :class="
+              clsx(
+                'rounded min-w-[100px] h-8 skeleton-shimmer',
+                themeClasses(
+                  'bg-neutral-200 border border-neutral-400',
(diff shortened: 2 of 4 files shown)


app/web/src/newhotness/explore_grid/ExploreGridTile.vue
```diff
@@ -23,9 +23,9 @@
     >
       <Icon :name="getAssetIcon(component.schemaCategory)" size="lg" />
       <h2>
-        <TruncateWithTooltip class="pb-xs">{{
-          component.name
-        }}</TruncateWithTooltip>
+        <TruncateWithTooltip class="pb-xs"
+          >name: {{ component.name }}</TruncateWithTooltip
```
wendybujalski: Why the addition of "name:"? We already don't have much room to show the name in this spot.

app/web/src/newhotness/skeletons/ExploreGridSkeleton.vue
```diff
@@ -0,0 +1,289 @@
+<template>
+  <section
+    :class="
+      clsx('grid h-full explore', themeClasses('bg-white', 'bg-neutral-950'))
```
wendybujalski, line 4: `bg-neutral-950` is not a valid Tailwind class. You can use `bg-neutral-900` or `bg-black` instead.

wendybujalski: Make sure you correct the other spots you are using `bg-neutral-950` as well!

app/web/src/newhotness/explore_grid/ExploreGridTile.vue
jobelenus: is this right? having `name: ` in the text?

jobelenus (approved):

app/web/src/newhotness/skeletons/DelayedSkeleton.vue
```diff
@@ -0,0 +1,25 @@
+<template>
+  <div>
```
vbustamante: Did you intend to use show here like this? otherwise we don't need this div
```suggestion
  <div v-if="show">
```

app/web/src/newhotness/explore_grid/ExploreGridTile.vue
zacharyhamm: whoops!

zacharyhamm: That was me figuring out which tile was being rendered. And no, i didn't mean to leave it

app/web/src/newhotness/skeletons/ExploreGridSkeleton.vue
wendybujalski, line 1: A lot of using numeric Tailwind classes (such as `py-2`) instead of abstract ones (equivalent would be `py-xs`). Please try to prioritize abstract size classes over numeric ones when possible.

app/web/src/newhotness/explore_grid/ExploreGridTile.vue
zacharyhamm: ok @jobelenus removed

app/web/src/newhotness/skeletons/DelayedSkeleton.vue
zacharyhamm: It's so the skeletons only show after the timeout. Or am I confused?

app/web/src/newhotness/skeletons/ExploreGridSkeleton.vue
```diff
@@ -0,0 +1,289 @@
...
+  windowHeight.value = window.innerHeight;
+};
+
+onMounted(() => {
+  window.addEventListener("resize", updateWindowSize);
+});
+
+onUnmounted(() => {
+  window.removeEventListener("resize", updateWindowSize);
+});
```
wendybujalski: In the new UI we want to avoid adding/removing event listeners directly inside of individual Vue components as much as possible. Instead, we have unified event listeners in `emitters.ts` that give you the same functionality without adding a new listener to each individual Vue component.

app/web/src/newhotness/skeletons/DelayedSkeleton.vue
jobelenus: i think he means the commit as-is doesnt have the `v-if="show"`

vbustamante: @zacharyhamm Yeah but the way it's written on the PR the show variable does not seem to be used

app/web/src/newhotness/skeletons/ExploreGridSkeleton.vue
```diff
@@ -0,0 +1,289 @@
...
+import clsx from "clsx";
+import { themeClasses } from "@si/vue-lib/design-system";
+import GridTileSkeleton from "./GridTileSkeleton.vue";
+
+const MIN_GRID_TILE_WIDTH = 250;
+const GRID_TILE_GAP = 16;
+const GRID_TILE_HEIGHT = 233;
+
+const windowWidth = ref(window.innerWidth);
+const windowHeight = ref(window.innerHeight);
```
wendybujalski: See `emitters.ts` for the better way to source this data.

app/web/src/newhotness/skeletons/DelayedSkeleton.vue
zacharyhamm: We set it to true after the timeout? Just belw

vbustamante: You could also do it like this, it's what i'd suggest
```
<template>
  <ExploreGridSkeleton v-if="show" />
</template>


```

zacharyhamm: OH i don't know how i dropped that

app/web/src/newhotness/skeletons/ExploreGridSkeleton.vue
```diff
@@ -0,0 +1,289 @@
...
+}
+
+@keyframes shimmer {
+  0% {
+    left: -100%;
+  }
+  100% {
+    left: 100%;
+  }
+}
```
wendybujalski, line 262: Very cool!

app/web/src/newhotness/skeletons/DelayedSkeleton.vue
zacharyhamm: Fixed

app/web/src/newhotness/skeletons/GridTileSkeleton.vue
```diff
@@ -0,0 +1,98 @@
...
+}
+
+@keyframes shimmer {
+  0% {
+    left: -100%;
+  }
+  100% {
+    left: 100%;
+  }
+}
```
wendybujalski: Probably better to not have this be scoped so that we don't have the same code duplicated twice.

app/web/src/newhotness/skeletons/DelayedSkeleton.vue
```diff
@@ -0,0 +1,25 @@
+<template>
+  <div v-if="show">
+    <ExploreGridSkeleton />
+  </div>
```
wendybujalski, line 4: This `v-if` could probably just be directly on ExploreGridSkeleton?

zacharyhamm: This is going to have different skeletons based on where you are at some point (component view skel, and map skel). So was leaving that structure in place for the future

wendybujalski: Ah okay awesome that makes sense! 👍

PR #6565, opened 2025-07-02
Fix: map works when looking at a view

jobelenus:
Note: when looking at the default view, everything was working as expected. Needed a view that didn't have *everything*.

The set of `matchingIds` needed to include all the components passed in (which are the components in a view) when there isn't a search string. The logic of "ignore whether component is in matchingIds if we dont have a search string" was incorrect

image


stack72 (approved):
Thank you so much!!

stack72 (approved):

stack72 (approved):

PR #6566, opened 2025-07-02
Fix: when performing a no-op patch, we need the atom document for postProcessing

jobelenus:
This was the cause of the "MVs don't update" most frequently seen with Action List. `Apply Changes` expresses a lot of the "no-op" behavior.. and it was the no-op on an AttributeTree that was causing the exception

When performing a no-op patch, we need the atom document for postProcessing. Without the document we ran into a `throw` that stopped a batch of patches from processing.

This _ought_ to prevent the `Atom is not removed, but no data for post processing` case... and changing that to not `throw` to a console error so it doesn't block other atoms

### DIFF
--- a/app/web/src/workers/webworker.ts
+++ b/app/web/src/workers/webworker.ts
@@ -949,6 +949,8 @@ const applyPatch = async (atom: Required<Atom>, indexChecksum: Checksum) => {
         atom.toChecksum,
         upToDateAtomIndexes,
       );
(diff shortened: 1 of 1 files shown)


zacharyhamm (dismissed):

britmyerss (dismissed):

stack72 (approved):
just reapproving because you linted

PR #6567, opened 2025-07-02
feat(web) - a few more UI improvements

wendybujalski:
- Create only props now get disabled for editing after they have a resource.
- Fixed a bug with the add item form for maps/arrays.
- Attributes panel and secrets form now supports textarea and code editor inputs


nickgerace (approved):
https://www.youtube.com/shorts/ro0Wsq4w1As

PR #6571, opened 2025-07-03
Add system status in the new UI

nickgerace:
## Introduction

This change adds system status to the new UI.

## Description

The bucket currently has two kinds of items in it right now: components in flight (DVU roots) and "rainbow" items (MVs in flight). More kinds of items are planned to be added in the future, like management functions. As a result of these changes, the rainbow spinner has been removed.

The underlying mechanism opens the door to varying levels of granularity of status and detail. Right now, the bucket is generic to all kinds of status and the level of detail is low. However, the core logic should be supple enough for us to go nuts.

This change also disables all modals related to the old UI when using change set apply except for the one that indicates to users that they've been moved to HEAD. Applying a change set should feel much faster and the apply flow no longer relies on the old change sets store. It now uses the new "useApi" composable.

One more thing too... the message saying "Because you are currently on Head"... I changed "Head" to "HEAD" for consistency. It drove me nuts!

Finally, there is a known issue with this change: the new status icon isn't perfectly centered to the change set buttons to the left of it.

## Screenshots

Here is what happens when there is at least one unexpired item in the bucket. It is animated too!

Screenshot 2025-07-03 at 12 29 43 AM

Here is what happens when changes are synced. It will hold for a few seconds and then fade away automatically.

Screenshot 2025-07-03 at 12 30 12 AM


app/web/src/components/layout/navbar/NavbarPanelLeft.vue
```diff
@@ -39,6 +39,8 @@
 
     <ChangeSetPanel ref="changeSetPanelRef" />
 
+    <StatusPanel v-if="useNewUI" />
```
nickgerace, line 42: Put as much of this in `newhotness` as possible... well, and this is cleaner separation.

app/web/src/newhotness/AddComponentModal.vue
```diff
@@ -35,7 +35,7 @@
             >
               <Icon name="info-circle" />
               <div class="grow">
-                Because you are currently on Head, when you create a component a
+                Because you are currently on HEAD, when you create a component a
```
nickgerace, line 38: I call this out in the description, but this bothered me a ton. We call it "HEAD" everywhere else, so I wanted to make it consistent.

app/web/src/newhotness/ApplyChangeSetButton.vue
```diff
@@ -66,16 +65,9 @@ const openApprovalFlowModal = () => {
 const disabled = computed(
   () =>
     changeSetsStore.selectedChangeSet?.status !== ChangeSetStatus.Open ||
-    changeSetsStore.headSelected ||
-    statusStoreUpdating.value,
```
nickgerace, line 70: This should be the new status actually... Hmm... I didn't run into issues, but now that I'm re-reading this, it might be best to re-instate until we have all the new bucket items in place just to make sure we don't apply pre-maturely.

Thoughts @jobelenus @jhelwig?

app/web/src/newhotness/StatusPanel.vue
```diff
@@ -0,0 +1,84 @@
...
+watch(
+  () => componentsInFlight.value,
+  (newComponentsInFlight) => {
+    if (newComponentsInFlight.length > 0 && !bucket.componentsInFlight) {
+      bucket.componentsInFlight = Date.now();
+    } else if (newComponentsInFlight.length < 1 && bucket.componentsInFlight) {
+      delete bucket.componentsInFlight;
+    }
+  },
+);
```
nickgerace: This is our first bucket item: components or flights. I'd love to use a set here with just timers, but knowing _**what**_ is being tracked will be important soon.

app/web/src/newhotness/StatusPanel.vue
```diff
@@ -0,0 +1,84 @@
...
+    if (now - value > timeoutMs) {
+      delete bucket[key];
+    }
+  }
+
+  // TODO(nick): if we saw at least one failure, we should not say everything is synced.
+  // if (deletedAtLeastOne) {
+  // status.value = undefined;
+  // }
+});
```
nickgerace, line 89: On every trigger, check if anything timed out. We may want to maybe show a third state like an "x" with hover text "some changes may not have synced" if at least one item timed out... however, that's not actionable for a user either, so... eh....

Thoughts @stack72 @VictoriaMolgado?

app/web/src/newhotness/StatusPanel.vue
```diff
@@ -0,0 +1,84 @@
...
+watch(
+  () => ({ ...bucket }),
+  (newBucket) => {
+    if (Object.keys(newBucket).length < 1) {
+      status.value = "synced";
+    } else {
+      status.value = "syncing";
+    }
+  },
+);
```
nickgerace, line 100: Change the status based on the bucket. The bucket being a `reactive` is important to detect key deletion and shape.

app/web/src/newhotness/StatusPanel.vue
```diff
@@ -0,0 +1,84 @@
...
+
+onMounted(() => {
+  const interval = setInterval(() => {
+    trigger.value = !trigger.value;
+  }, tickMs);
+
+  onUnmounted(() => {
+    clearInterval(interval);
+  });
+});
```
nickgerace, line 110: Dummy trigger system on a timer

app/web/src/newhotness/StatusPanelIcon.vue
```diff
@@ -0,0 +1,59 @@
+<template>
+  <Transition
+    enterActiveClass="transition-opacity duration-300"
```
nickgerace, line 3: We transition in slightly faster than we do out. This makes it less jarring on the way in while making sure it is obvious when it is fading on the way out.

app/web/src/newhotness/StatusPanelIcon.vue
```diff
@@ -0,0 +1,59 @@
...
+    enterFromClass="opacity-0"
+    leaveActiveClass="transition-opacity duration-500"
+    leaveToClass="opacity-0"
+    appear
+  >
+    <!-- We may want to make this "invisible" to keep the space in the future. -->
```
nickgerace, line 9: Yeah, so... right now it just disappears. I think that's fine, but if we get a lot of element popping in the future, we may just want to reserve the space.

app/web/src/newhotness/StatusPanelIcon.vue
```diff
@@ -0,0 +1,59 @@
...
+  (newStatus) => {
+    if (!newStatus) {
+      show.value = false;
+    } else if (newStatus === "syncing") {
+      show.value = true;
+    } else if (newStatus === "synced") {
+      show.value = true;
+      setTimeout(() => {
+        show.value = false;
+        emit("faded");
```
nickgerace, line 49: We have to emit that it faded so that the parent can reset it to `undefined`. This component is intentionally not any smarter than it has to be.

lib/vue-lib/src/design-system/icons/Icon.vue
```diff
@@ -115,6 +118,17 @@ const AUTO_SPIN_ICONS_COUNTER_CLOCKWISE = ["refresh-active"];
...
+    transform: scaleY(-1);
+
+    > :deep(svg) {
+      // Note that we spin counter-clockwise because we've flipped it. That's also why we don't
+      // have the flip and the spin separated: it would be confusing to the caller to request
+      // counter-clockwise and flip when they truly want clockwise.
+      @apply animate-spin-cc;
+      animation-duration: 5s;
+    }
+  }
```
nickgerace, line 131: Why are these grouped and not just two options? If you flip, you must also invert spin direction... and we want it slower for this one too!

@VictoriaMolgado this uses your `5s` duration, by the way.

app/web/src/workers/types/entity_kind_types.ts
```diff
@@ -447,3 +448,8 @@ export interface Secret {
   // If the secret can be used on this workspace
   isUsable: boolean;
 }
+
+export interface DependentValueComponentList {
```
nickgerace, line 452: This uses the MV that @jhelwig added in #6551

app/web/src/newhotness/StatusPanel.vue
stack72: I'd ignore that for now - we can just transition the spinner when it times out

app/web/src/newhotness/ApplyChangeSetButton.vue
britmyerss: I think it's ok - we're protected by the route, and need to fix up the response to not be a 500 anyways as even with the MV there's a chance you hit the button between receiving updates

nickgerace: Okay great. I also just added the ability to block it anyway for DVU roots. So... crisis(?) averted.

jobelenus: Sorry, I'm not at all sure what this comment is wondering/worried about?

app/web/src/newhotness/ApprovalFlowModal.vue
```diff
@@ -182,10 +176,10 @@ async function applyButtonHandler() {
   //   }
   // }
 
-  // TODO(nick): we should make sure this isn't possible...
-  if (!ctx || !authStore.user) return;
+  // This should be impossible.
+  if (!ctx) return;
 
-  const resp = await changeSetsStore.APPLY_CHANGE_SET(authStore.user.name);
+  const resp = await changeSetsStore.APPLY_CHANGE_SET_NEW_HOTNESS();
```
jobelenus: why are we doing this in a store???

app/web/src/store/change_sets.store.ts
```diff
@@ -260,9 +263,12 @@ export function useChangeSetsStore() {
               window.localStorage.removeItem(headkey);
             },
             onSuccess: (_response) => {
-              // this.changeSetsById[response.changeSet.pk] = response.changeSet;
-              const statusStore = useStatusStore();
-              statusStore.resetWhenChangingChangeset();
+              // NOTE(nick): do not use the status store in the new UI.
+              const route = router.currentRoute.value;
```
jobelenus: this is all simpler if we just get out of the store

app/web/src/newhotness/StatusPanel.vue
```diff
@@ -0,0 +1,85 @@
...
+    }
+  }
+
+  // TODO(nick): if we saw at least one failure, we should not say everything is synced.
+  // if (deletedAtLeastOne) {
+  // status.value = undefined;
+  // }
+});
+
+watch(
```
jobelenus, line 91: this needs to also use the "rainbowCounter" data... right now, as implemented we've got 2 spinners next to each other... we should consolidate into 1.

jobelenus (changes requested):

app/web/src/newhotness/ApplyChangeSetButton.vue
nickgerace: @jobelenus in the old UI, we disabled the button if either a rebase or a DVU was in flight. At the time of this comment, I had neither protection in place because I removed the `statusStoreUpdating` check and did not replace it with anything else. However, I amended the commit and now this checks if `status.value === 'syncing'`, which uses the new status work introduced by this PR to disable the button.

app/web/src/newhotness/StatusPanel.vue
nickgerace: Okay! Will do.

app/web/src/newhotness/ApplyChangeSetButton.vue
jobelenus: Right, we still know how to disable the button.

app/web/src/store/change_sets.store.ts
nickgerace: Will do

app/web/src/newhotness/ApprovalFlowModal.vue
nickgerace: I was more concerned with getting it functional first for this PR. That being said, it's a good question because this is further digging into a frankenstein setup. I'll look into splicing it out.

nickgerace:
Putting this back into drafts for a moment to address @jobelenus' comments on using the store.

lib/vue-lib/src/design-system/icons/Icon.vue
```diff
@@ -22,6 +22,8 @@
         computedRotate && `--rotate-${computedRotate}`,
         AUTO_SPIN_ICONS_CLOCKWISE.includes(name) && '--spin',
         AUTO_SPIN_ICONS_COUNTER_CLOCKWISE.includes(name) && '--spin-cc',
+        AUTO_SPIN_SLOW_WITH_HORIZONTAL_FLIP_CLOCKWISE.includes(name) &&
```
wendybujalski, line 25: we love a verbose and clear name

wendybujalski (approved):

jobelenus (approved):

app/web/src/newhotness/StatusPanel.vue
nickgerace: We now have two bucket items since Rainbow moved into this

app/web/src/newhotness/StatusPanel.vue
nickgerace: Resolving with that in mind!

PR #6574, opened 2025-07-03
fix(web): get base change set id correctly

zacharyhamm:
Legacy workspaces will never have a change set that has no base change set. Just grab the base change set off the workspace metadata.

### DIFF
--- a/app/web/src/store/realtime/heimdall.ts
+++ b/app/web/src/store/realtime/heimdall.ts
@@ -31,7 +31,7 @@ import {
   EntityKind,
   SchemaMembers,
 } from "@/workers/types/entity_kind_types";
-import { ChangeSetId, ChangeSetStatus } from "@/api/sdf/dal/change_set";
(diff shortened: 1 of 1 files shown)


app/web/src/store/realtime/heimdall.ts
```diff
@@ -483,30 +483,24 @@ const MUSPELHEIM_CONCURRENCY = 1;
...
+const fetchOpenChangeSets = async (
+  workspaceId: string,
+): Promise<WorkspaceMetadata> => {
   const resp = await sdf<WorkspaceMetadata>({
     method: "GET",
     url: `v2/workspaces/${workspaceId}/change-sets`,
   });
-  return resp.data.changeSets.filter((cs) =>
-    [
-      ChangeSetStatus.Open,
```
zacharyhamm, line 493: Filtering here is unnecessary. The backend gives us only "active" change sets

PR #6575, opened 2025-07-03
feat: Implement the lobby with messages

vbustamante:
Changes lobby to overlay whole screen, fade in and out and show messages terminal style

Screenshot 2025-07-03 at 16 06 40

### DIFF
--- a/app/web/src/newhotness/LobbyTerminalOutputLine.vue
+++ b/app/web/src/newhotness/LobbyTerminalOutputLine.vue
@@ -1,0 +1,80 @@
+<template>
+  <div class="flex flex-row items-center">
+    <Icon
+      name="logo-si"
+      size="sm"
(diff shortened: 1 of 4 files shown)


jobelenus (approved):

app/web/src/newhotness/LobbyTerminalOutputLine.vue
```diff
@@ -0,0 +1,80 @@
+<template>
+  <div class="flex items-center">
```
wendybujalski: Try to avoid using `flex` without an explicit direction (in this case `flex-row`), as it can make the code confusing later on down the line if more classes are added.

wendybujalski (approved):
Super cool, excellent work!

wendybujalski (approved):

PR #6576, opened 2025-07-03
Jf/mv component go faster

fnichol:


lib/dal-materialized-views/src/component.rs
```diff
@@ -27,32 +29,43 @@ pub async fn assemble_in_list(
     component_id: ComponentId,
 ) -> crate::Result<ComponentInListMv> {
     let ctx = &ctx;
-    let schema_variant_id = Component::schema_variant_id(ctx, component_id).await?;
-    let schema_variant = SchemaVariant::get_by_id(ctx, schema_variant_id).await?;
-    let schema = SchemaVariant::schema_for_schema_variant_id(ctx, schema_variant_id).await?;
+    // TODO(fnichol): remove `.instrument()` calls
```
fnichol, line 32: Note that I've left these commented out `.instrument()` calls in here on purpose (and the other `TODO` items) as I'll likely be un-commenting these in the next phase of improvements--this was how we determined where our function time was being allocated and which function calls to improve first. Cool, no?

PR #6587, opened 2025-07-04
fix: Fail on self rebases everywhere

vbustamante:

Block merging a changeset if it is the same as the base changeset on ctx. This is meant to prevent occurrences where rogue API requests try to apply HEAD, which if no other changesets exist in the workspace leaves the system unusable 

#### Screenshots:

image

## How was it tested?

- [X] Manual test: Created and deleted a security group via actions
- [X] Manual test: tried to merge via direct api call


stack72 (approved):
Nice addition!

PR #6590, opened 2025-07-04
fix(web): Remove PropTree from AddComponentModal

stack72:
This doesn’t quite work as we expected so we can revist this at a later date

### DIFF
--- a/app/web/src/newhotness/AddComponentModal.vue
+++ b/app/web/src/newhotness/AddComponentModal.vue
@@ -175,23 +175,12 @@
             </div>
           </div>
         </div>
-        <!-- Right side - documentation and attributes -->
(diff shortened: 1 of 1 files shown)


app/web/src/newhotness/AddComponentModal.vue
```diff
@@ -224,8 +213,8 @@ import EditingPill from "@/components/EditingPill.vue";
 import {
   BifrostSchemaVariantCategories,
   CategoryVariant,
-  Prop,
-  PropTree,
+  // Prop,
+  // PropTree,
```
vbustamante:
```suggestion
```
We should just delete these

vbustamante (approved):

PR #6594, opened 2025-07-04
feat: Move validations to show as icons on the prop editor

vbustamante:
- Removes individual validations from qualifications list
- Shows validations status icons besides the prop editor
- On fail, hovering over the validation icons shows the error

Screenshot 2025-07-04 at 18 56 45


stack72 (approved):
This is such a great step forward!

PR #6605, opened 2025-07-07
feat(web): When pinning a component, have an unconnected section

stack72:
This allows a user to be able to step in / pin another component easily rather than needing to unpin to find another component

### DIFF
--- a/app/web/src/newhotness/Explore.vue
+++ b/app/web/src/newhotness/Explore.vue
@@ -660,17 +660,9 @@ const componentList = computed(() => {
   if (!pinnedComponent.value || !pinnedComponentConnections.value)
     return componentListUnchecked.value;
 
-  // If we are dealing with pinning, we are only concerned with relevant components.
-  const relevantComponentsWhenPinning = [pinnedComponent.value];
-  for (const component of componentListUnchecked.value) {
-    if (
-      pinnedComponentConnectionSets.value.incoming.has(component.id) ||
-      pinnedComponentConnectionSets.value.outgoing.has(component.id)
-    ) {
-      relevantComponentsWhenPinning.push(component);
-    }
-  }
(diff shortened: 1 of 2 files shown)


app/web/src/newhotness/explore_grid/ExploreGrid.vue
```diff
@@ -199,9 +199,10 @@ const gridRows = computed(() => {
     let collapsed = collapseTracker.value[groupName];
 
     // Handle the very first time everything is loaded. We want empty sections to begin collapsed
-    // and non-empty sections to be expanded by default.
+    // and non-empty sections to be expanded by default. The "Unconnected" section should always
+    // start collapsed.
     if (collapsed === undefined) {
-      collapsed = count === 0;
+      collapsed = count === 0 || groupName === "Unconnected";
```
nickgerace, line 205: Love this detail

app/web/src/newhotness/Explore.vue
```diff
@@ -660,17 +660,9 @@ const componentList = computed(() => {
...
-      pinnedComponentConnectionSets.value.incoming.has(component.id) ||
-      pinnedComponentConnectionSets.value.outgoing.has(component.id)
-    ) {
-      relevantComponentsWhenPinning.push(component);
-    }
-  }
-  return relevantComponentsWhenPinning;
+  // When pinning, include all components so they can be properly grouped
+  // into connected and unconnected sections
+  return componentListUnchecked.value;
```
nickgerace, line 665: We may be able to simplify all of this in the future, but I agree with keeping the list and unchecked list separate for now. The second that we have a product need for a list of components that is smaller than the full list, we'll want this separation

nickgerace:
Accidentally moved to queue. Up to you!

PR #6608, opened 2025-07-07
fix(web): Only show context menu on right click

stack72:


app/web/src/newhotness/Map.vue
```diff
@@ -449,7 +451,7 @@ const connections = useQuery<IncomingConnections[]>({
           const c = d.find((c) => c.id === fillDefault.value);
           if (c && c.id) {
             // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-            selectComponent(componentsById.value[c.id]!);
+            selectedComponent.value = componentsById.value[c.id]!;
```
wendybujalski, line 454: Don't we need to check for the context menu here as well?

wendybujalski: Ah wait I guess it can't be open at this point, nevermind.

PR #6610, opened 2025-07-07
Don't use func run id inside the actions list MV

jobelenus:

The func run id isn't on the graph, and won't cause the MV to update as it changes. Instead, look up the func run id from the action, and then look up the logs for that func run id

### Testing
Load action details in the URL and see that it still displays data and is hitting the new endpoint


app/web/src/newhotness/LatestFuncRunDetails.vue
```diff
@@ -12,46 +12,37 @@
...
 
-const key = useMakeKey();
-const args = useMakeArgs();
-const actionViewListRaw = useQuery<BifrostActionViewList | null>({
-  queryKey: key(EntityKind.ActionViewList),
-  queryFn: async () =>
-    await bifrost<BifrostActionViewList>(args(EntityKind.ActionViewList)),
+const actionFuncRunQuery = useQuery<string>({
+  queryKey: ["action_func_run_id", props.actionId],
+  staleTime: 100,
```
jobelenus, line 31: note that we add a stale time here, because there is nothing that can invalidate this data, so each subsequent request to the details page ought to re-fire the query. (stale time is defaulted to `infinity`/never)

stack72 (approved):

PR #6611, opened 2025-07-07
feat(web): Add colors to connected edges on the Map view

stack72:


app/web/src/newhotness/Map.vue
```diff
@@ -115,6 +115,59 @@
...
+          <path
+            d="M 0,0 L 8,3.5 L 0,7"
+            fill="none"
+            :stroke="themeClasses('#d1d5db', '#4b5563')"
+            stroke-width="1.5"
+            stroke-linejoin="miter"
+            opacity="0.3"
+          />
+        </marker>
+      </defs>
```
wendybujalski, line 170: This is so good! But we will definitely want to replace those hex codes with variables eventually.

app/web/src/newhotness/Map.vue
```diff
@@ -647,6 +700,51 @@ watch(selectedComponent, () => {
...
+        edgeElement.setAttribute(
+          "marker-end",
+          isConnectedToSelected
+            ? "url(#arrowhead-highlighted)"
+            : "url(#arrowhead-greyed)",
+        );
+      } else {
+        // Reset to default when no component is selected - theme aware
+        const isDark = document.body.classList.contains("dark");
+        const defaultColor = isDark ? "#9ca3af" : "#6b7280"; // neutral-400 : neutral-500
```
wendybujalski, line 738: Same here! Appreciate the call out in the comment tho.

PR #6613, opened 2025-07-07
fix(web): Allow insertion of array items under array items

jkeiser:
Presently, when you try to insert an array item underneath an array item, it fails with a 400 error.

The source of this bug was that we were using the prop path (munged to remove the initial "root") instead of the attribute path; because of this, we send the parent array item prop name instead of the actual array index.

This makes the frontend always use AttributePath instead of munging PropPath when editing / inserting / removing attributes.

It also fixes a bug where you can't insert map items when the map item is an array/object/map.

### Factor Budget

* Made AttributeTree.AttributeValue.path non-optional (which it already is)
* Stopped passing AttributeValueId around in the attribute panel; use path everywhere (which we mostly already were)
* Require "/" at the beginning of AttributePath. This gives nice autocomplete, but more importantly, makes it so you can't pass a prop path to something expecting an AttributePath.

#### Screenshots:

image

## How was it tested?

- [X] Integration tests pass
- [X] Manual test: inserting and deleting deep children of array and map items works
- [x] Manual test: updating deep children of array items works
- [x] Manual test: insert and delete of top-level array and map items works

### DIFF
--- a/app/web/src/newhotness/layout_components/AttributeInput.vue
+++ b/app/web/src/newhotness/layout_components/AttributeInput.vue
@@ -594,6 +594,7 @@ import CodeViewer from "@/components/CodeViewer.vue";
 import { PropKind } from "@/api/sdf/dal/prop";
 import { CategorizedPossibleConnections } from "@/workers/types/dbinterface";
 import StatusIndicatorIcon from "@/components/StatusIndicatorIcon.vue";
+import { AttributePath, ComponentId } from "@/api/sdf/dal/component";
 import {
   attributeEmitter,
   MouseDetails,
@@ -608,8 +609,7 @@ type UIPotentialConnection = PossibleConnection & {
 };
 
 const props = defineProps<{
-  attributeValueId: string;
-  path: string;
+  path: AttributePath;
   value: string;
   kind?: PropertyEditorPropWidgetKind | string;
   prop?: Prop;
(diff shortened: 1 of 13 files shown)


app/web/src/newhotness/layout_components/AttributeInput.vue
```diff
@@ -634,16 +634,9 @@ const selectKeyString = "Tab";
...
-  // fix the MV! for arrays path": "root\u000bdomain\u000btags\u000btag"
-  let path = props.path;
-  path = path.replaceAll("\u000b", "/");
-  return path;
-});
-
 type AttrData = { value: string };
 const wForm = useWatchedForm<AttrData>(
-  `component.av.prop.${props.attributeValueId}`,
+  `component.av.prop.${props.path}`,
```
jobelenus: i think that ID is better here, b/c its unique per component, correct? E.g. if someone was editing the same path on two components it would overwrite one another... alternatively, stick the component ID in there with the path

lib/si-frontend-mv-types-rs/src/component/attribute_tree.rs
```diff
@@ -69,7 +69,7 @@ pub struct ExternalSource {
 pub struct AttributeValue {
     pub id: AttributeValueId,
     pub key: Option<String>,
-    pub path: Option<String>,
+    pub path: String,
```
jobelenus, line 72: so glad this is not an Option

app/web/src/newhotness/layout_components/AttributeInput.vue
jkeiser: oh, I see! yeah, I'll fix

PR #6618, opened 2025-07-08
fix(web): Don't show caution-tape or passing validatons

stack72:
We should only show failing validations to avoid noise. We also need to ensure that we show a readonly input that isn’t caution tape

### DIFF
--- a/app/web/src/newhotness/layout_components/AttributeInput.vue
+++ b/app/web/src/newhotness/layout_components/AttributeInput.vue
@@ -17,10 +17,10 @@
           <TruncateWithTooltip>{{ displayName }}</TruncateWithTooltip>
           <div class="flex flex-row items-center ml-auto gap-2xs">
             <StatusIndicatorIcon
-              v-if="validation"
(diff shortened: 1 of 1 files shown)


app/web/src/newhotness/layout_components/AttributeInput.vue
```diff
@@ -17,10 +17,10 @@
           <TruncateWithTooltip>{{ displayName }}</TruncateWithTooltip>
           <div class="flex flex-row items-center ml-auto gap-2xs">
             <StatusIndicatorIcon
-              v-if="validation"
+              v-if="validation && validation.status !== 'Success'"
               v-tooltip="validation.message"
               type="qualification"
-              :status="validation.status === 'Success' ? 'success' : 'failure'"
```
vbustamante, line 23: This will make the icon neutral, I believe. the `validations.status` values are uppercase, while the components expects lowercase values. 
The ideal thing would be setting `status="failure"` over here.

PR #6622, opened 2025-07-08
wip

zacharyhamm:
## How was it tested?

- [X] Integration tests pass


lib/dal/src/attribute/value/dependent_value_graph.rs
```diff
@@ -173,71 +256,32 @@ impl DependentValueGraph {
...
-                    self.value_depends_on(subscriber_av_id, current_attribute_value.id());
-                }
+                let new_work_queue_values = self
+                    .process_subscription_dependencies(
+                        ctx,
+                        current_attribute_value.id(),
+                        &workspace_snapshot,
+                    )
+                    .await?;
+                work_queue.extend(new_work_queue_values);
```
jhelwig, line 295: Should `work_queue.reserve(new_work_queue_values.len());` before the `extend` since `extend` could do multiple allocations. It'll be really nice once [`extend_reserve`](https://doc.rust-lang.org/std/iter/trait.Extend.html#method.extend_reserve) is stabilized.

lib/dal/tests/integration_test/audit_logging.rs
```diff
@@ -89,7 +89,7 @@ async fn round_trip(ctx: &mut DalContext, audit_database_context: AuditDatabaseC
 
     // Check that everything looks as we expect.
     {
-        let expected_total = 5;
```
zacharyhamm, line 92: @nickgerace For some reason, this change added 2 more audit log entries in this test. I don't really know why? Any red flags here?

lib/dal/src/attribute/value/dependent_value_graph.rs
```diff
@@ -158,6 +165,82 @@ impl DependentValueGraph {
...
+                                }
+
+                                // For an inter-component subscription, the subscriber should depend on the other
+                                // component's *root* attribute value, so that the entire other component's data is
+                                // processed by the dvu and available to the subscriber before we try to run the
+                                // subscriber's function.
+                                self.value_depends_on(
+                                    subscriber_av_id,
+                                    component_root_attribute_value_id,
+                                );
```
jhelwig: Why is depending on the resolved subscribed to AV not sufficient? The subscription shouldn't be using any data that isn't the resolved subscribed to AV or one of its descendants.

zacharyhamm: You're not wrong. The real fix here was ensuring we walked to the parent for all values added to the graph during dependency discovery. This fixed only some of the problems before I realized that, so the comment is probably out of date. Let me run the subscription tests with the direct connection only.

zacharyhamm: On testing again, the new test fails if we only make the subscribe_av_id depend on the resolved_av_id. I believe this is because of the way the subscriptions resolve values when they're executing, but I'd have to investigate it further.

jhelwig: Yeah, having to depend on the root feels like papering over a bug elsewhere that should be fixed instead.

zacharyhamm: Using the "controlling avs" for each seems to work

jhelwig: That sounds like the right thing to do, since that'll be the function that generates the value (if it's not from the subscribed-to AV's prototype). 👍🏻

jhelwig (approved):

lib/dal/tests/integration_test/audit_logging.rs
nickgerace: @zacharyhamm I wouldn't worry at the moment, but I'll look tomorrow. Worst case it is overly chatty for a bit.

PR #6624, opened 2025-07-08
Completed observability pass on the web worker

jobelenus:
This contains roughly all the data that is outputted via DEBUG. Turning DEBUG off for non-local!

We get a span for:
1. Each WS message received and added to a queue
2. Each element of the queue processed
3. Each patch applied.

We'll definitely have to keep our eyes on how these numbers affect the total budget. I can see us, after gaining more confidence, dropping the first one there...

### Testing
Ensure that patches for new and modified components work, and that cold start works.


nickgerace (dismissed):
Lovely!

britmyerss (approved):

PR #6626, opened 2025-07-08
Observe the lobby

jobelenus:
Add a span that ends when a lobby ends

### Testing
In `main.ts` turn on the HoneycombWebSDK `debug: true` and you can see one of the very last spans that gets emitted is the `lobby` event


app/web/src/newhotness/Workspace.vue
```diff
@@ -147,6 +151,22 @@ const lobby = computed(() => {
   return false;
 });
 
+watch(
+  lobby,
+  () => {
+    if (!span.value && lobby.value) {
+      span.value = tracer.startSpan("lobby");
+      span.value.setAttributes({ ...props, user: authStore.user?.pk });
```
britmyerss: let's also add si.workspace.id and si.changeset.id here as attributes!

britmyerss (approved):

PR #6627, opened 2025-07-08
Add apply change set API to the status indicator

nickgerace:
## Description

This change adds the "apply change set" API call to the status indicator. This ensures that active calls to apply the change set block additional calls.


app/web/src/newhotness/logic_composables/change_set.ts
```diff
@@ -20,3 +21,17 @@ export const useCurrentChangeSet = () => {
...
+
+  const loading = computed(() => api.inFlight.value);
+
+  const performApply = async () => {
+    const call = api.endpoint(routes.ApplyChangeSet);
+    const { req } = await call.post({});
+    return { success: api.ok(req) };
+  };
+
+  return { loading, performApply };
```
nickgerace, line 38: Return both the computed for it loading and the function to make the apply call.

app/web/src/newhotness/logic_composables/change_set.ts
```diff
@@ -20,3 +23,34 @@ export const useCurrentChangeSet = () => {
...
+};
+
+export const useDisableApplyChangeSet = () => {
+  const ctx = inject<Context>("CONTEXT");
+  assertIsDefined(ctx);
+
+  const currentChangeSet = useCurrentChangeSet();
+  const status = useStatus();
```
nickgerace, line 46: I'm a little sussed out by using composables in composables, but these are both very basic ones. I think for this one off, it is fine.

jobelenus (dismissed):

nickgerace:
Taking some offline feedback from @jobelenus to combine the composables into one. New commit in a bit...

PR #6635, opened 2025-07-09
Remove async_recursion from dal

jhelwig:
While convenient, `async_recursion` introduces a fair amount of overhead, and can end up blowing through the stack size limit much faster than expected. Especially when recursing over user-defined data, as is done in `AttributeValue::view`.
 
In addition to making `AttributeValue::view` iterative, instead of recursive, this moves a lot of "graph level" code out of  the domain objects and down into the graph level traits allowing for more direct usage of the underlying graph, more re-use of code between the graph level trait methods, and minimizing the construction of full domain-objects only to pull specific pieces of information from them and to gain access to the helper methods that require the full domain-object instead of just an ID.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2189,7 +2189,6 @@ dependencies = [
 name = "dal"
 version = "0.1.0"
 dependencies = [
- "async-recursion",
  "async-trait",
  "audit-database",
  "audit-logs-stream",
--- a/lib/dal-test/src/helpers.rs
+++ b/lib/dal-test/src/helpers.rs
@@ -288,9 +288,9 @@ pub async fn get_component_input_socket_value(
         .collect_vec()
         .pop()
         .ok_or(eyre!("no input socket match found"))?;
-    let input_socket_av =
-        AttributeValue::get_by_id(ctx, component_input_socket.attribute_value_id).await?;
-    Ok(input_socket_av.view(ctx).await?)
+    AttributeValue::view(ctx, component_input_socket.attribute_value_id)
+        .await
+        .map_err(Into::into)
 }
 /// Gets the [`Value`] for a specific [`Component`]'s [`InputSocket`] by the [`InputSocket`] name
 pub async fn get_component_input_socket_attribute_value(
@@ -333,9 +333,9 @@ pub async fn get_component_output_socket_value(
         .collect_vec()
         .pop()
         .ok_or(eyre!("no input socket match found"))?;
-    let output_socket_av =
-        AttributeValue::get_by_id(ctx, component_output_socket.attribute_value_id).await?;
-    Ok(output_socket_av.view(ctx).await?)
+    AttributeValue::view(ctx, component_output_socket.attribute_value_id)
+        .await
+        .map_err(Into::into)
 }
 
 /// Update the [`Value`] for a specific [`AttributeValue`] for the given [`Component`](ComponentId) by the [`PropPath`]
@@ -381,10 +381,9 @@ pub async fn get_attribute_value_for_component_opt(
         .ok_or(eyre!("unexpected, no attribute values found for prop"))?;
     assert!(attribute_value_ids.is_empty());
 
-    let attribute_value = AttributeValue::get_by_id(ctx, attribute_value_id).await?;
-
-    let value = attribute_value.view(ctx).await?;
-    Ok(value)
+    AttributeValue::view(ctx, attribute_value_id)
+        .await
+        .map_err(Into::into)
 }
 
 /// Encrypts a message with a given [`KeyPairPk`](KeyPair).
@@ -420,11 +419,9 @@ pub async fn fetch_resource_last_synced_value(
         return Err(eyre!("unexpected: more than one attribute value found"));
     }
 
-    let last_synced_value = AttributeValue::get_by_id(ctx, attribute_value_id)
-        .await?
-        .view(ctx)
-        .await?;
-    Ok(last_synced_value)
+    AttributeValue::view(ctx, attribute_value_id)
+        .await
+        .map_err(Into::into)
 }
 
 /// Extracts the value and validation from a raw property edtior value.
--- a/lib/dal/src/attribute/prototype.rs
+++ b/lib/dal/src/attribute/prototype.rs
@@ -82,6 +82,7 @@ use crate::{
             content_node_weight,
             traits::SiNodeWeight,
         },
+        traits::attribute_prototype::AttributePrototypeExt as _,
     },
 };
 
@@ -107,6 +108,8 @@ pub enum AttributePrototypeError {
     LayerDb(#[from] LayerDbError),
     #[error("attribute prototype {0} is missing a function edge")]
     MissingFunction(AttributePrototypeId),
+    #[error("Attribute prototype {0} has multiple functions")]
+    MultipleFunctionsFound(AttributePrototypeId),
     #[error("no arguments to identity func at {0}")]
     NoArgumentsToIdentityFunction(AttributePrototypeId),
     #[error("No attribute values for: {0}")]
@@ -220,19 +223,9 @@ impl AttributePrototype {
         ctx: &DalContext,
         prototype_id: AttributePrototypeId,
     ) -> AttributePrototypeResult<FuncId> {
-        let workspace_snapshot = ctx.workspace_snapshot()?;
-        for node_index in workspace_snapshot
-            .outgoing_targets_for_edge_weight_kind(prototype_id, EdgeWeightKindDiscriminants::Use)
-            .await?
-        {
-            let node_weight = workspace_snapshot.get_node_weight(node_index).await?;
-            let node_weight_id = node_weight.id();
-            if NodeWeightDiscriminants::Func == node_weight.into() {
-                return Ok(node_weight_id.into());
-            }
-        }
-
-        Err(AttributePrototypeError::MissingFunction(prototype_id))
+        ctx.workspace_snapshot()?
+            .attribute_prototype_func_id(prototype_id)
+            .await
     }
 
     pub async fn func(
@@ -733,27 +726,9 @@ impl AttributePrototype {
         ctx: &DalContext,
         ap_id: AttributePrototypeId,
     ) -> AttributePrototypeResult<Vec<AttributePrototypeArgumentId>> {
-        let mut apas = vec![];
-
-        let snap = ctx.workspace_snapshot()?;
-
-        for idx in snap
-            .outgoing_targets_for_edge_weight_kind(
-                ap_id,
-                EdgeWeightKindDiscriminants::PrototypeArgument,
-            )
-            .await?
-        {
-            apas.push(snap.get_node_weight(idx).await?.id().into())
-        }
-        // Whenever we're dealing with the arguments, we always want them to have stable
-        // ordering. For function execution, we want the arguments to have the same relative
-        // order from run to run. For use when generating checksums and other lists we also
-        // want them to have consistent order to avoid unnecessary churn in the data sent to
-        // & used by the front end.
-        apas.sort();
-
-        Ok(apas)
+        ctx.workspace_snapshot()?
+            .attribute_prototype_arguments(ap_id)
+            .await
     }
 
     // Returns the argument to this identity (or unset) function call.
--- a/lib/dal/src/attribute/value.rs
+++ b/lib/dal/src/attribute/value.rs
@@ -7,9 +7,7 @@ use std::{
     sync::Arc,
 };
 
-use async_recursion::async_recursion;
 pub use dependent_value_graph::DependentValueGraph;
-use indexmap::IndexMap;
 pub use is_for::ValueIsFor;
 use petgraph::prelude::*;
 use serde::{
@@ -28,6 +26,7 @@ use si_pkg::{
     AttributeValuePath,
     KeyOrIndex,
 };
+use si_split_graph::SplitGraphError;
 use subscription::ValueSubscription;
 use telemetry::prelude::*;
 use thiserror::Error;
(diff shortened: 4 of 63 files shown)


lib/dal-test/src/helpers.rs
```diff
@@ -288,9 +288,9 @@ pub async fn get_component_input_socket_value(
         .collect_vec()
         .pop()
         .ok_or(eyre!("no input socket match found"))?;
-    let input_socket_av =
```
jkeiser, line 291: <3 moving to `id` methods so people don't have to pay the assembly cost!

lib/dal-test/src/helpers.rs
```diff
@@ -420,11 +419,9 @@ pub async fn fetch_resource_last_synced_value(
...
     }
 
-    let last_synced_value = AttributeValue::get_by_id(ctx, attribute_value_id)
-        .await?
-        .view(ctx)
-        .await?;
-    Ok(last_synced_value)
+    AttributeValue::view(ctx, attribute_value_id)
+        .await
+        .map_err(Into::into)
```
jkeiser, line 424: Is the `.map_err(Into::into)` here to avoid `Ok(<expr>?)`? The latter can end up looking better (and being fewer lines) if it works.

jhelwig: That is why the `map_err`. I think it looks & reads better than `Ok(<expr>?)`.

Cargo.lock
```diff
@@ -2189,7 +2189,6 @@ dependencies = [
 name = "dal"
 version = "0.1.0"
 dependencies = [
- "async-recursion",
```
fnichol, line 2192: 💥

lib/dal/src/attribute/prototype.rs
```diff
@@ -220,19 +223,9 @@ impl AttributePrototype {
...
-            let node_weight_id = node_weight.id();
-            if NodeWeightDiscriminants::Func == node_weight.into() {
-                return Ok(node_weight_id.into());
-            }
-        }
-
-        Err(AttributePrototypeError::MissingFunction(prototype_id))
+        ctx.workspace_snapshot()?
+            .attribute_prototype_func_id(prototype_id)
+            .await
```
fnichol, line 228: Nice!

lib/dal/src/attribute/value.rs
```diff
@@ -900,17 +898,15 @@ impl AttributeValue {
...
+            let av_id = OutputSocket::component_attribute_value_id(
                 ctx,
-                OutputSocket::component_attribute_value_id(
-                    ctx,
-                    inferred_connection.output_socket_id,
-                    inferred_connection.source_component_id,
-                )
-                .await?,
+                inferred_connection.output_socket_id,
+                inferred_connection.source_component_id,
```
fnichol, line 904: Better too!

lib/dal/src/attribute/value.rs
```diff
@@ -1355,81 +1351,21 @@ impl AttributeValue {
             .ok_or(AttributeValueError::NoChildWithName(id, name.to_string()))
     }
 
-    #[async_recursion]
```
fnichol, line 1358: 💥

lib/dal/src/attribute/value.rs
```diff
@@ -1355,81 +1351,21 @@ impl AttributeValue {
...
+            ValueIsFor::Prop(_) => {
+                ctx.workspace_snapshot()?
+                    .attribute_value_view(ctx, attribute_value_id)
+                    .await
+            }
+            ValueIsFor::OutputSocket(_) | ValueIsFor::InputSocket(_) => {
+                let attribute_value = AttributeValue::get_by_id(ctx, attribute_value_id).await?;
+                Ok(attribute_value.value(ctx).await?)
+            }
+        }
```
fnichol, line 1368: That's a pretty glorious refactor

lib/dal/src/component.rs
```diff
@@ -514,8 +515,10 @@ impl Component {
         for value_id in Component::attribute_values_for_prop_id(ctx, id, root_prop_id).await? {
             let value_component_id = AttributeValue::component_id(ctx, value_id).await?;
             if value_component_id == id {
-                let root_value = AttributeValue::get_by_id(ctx, value_id).await?;
-                return Ok(root_value.view(ctx).await?);
+                return Ok(ctx
+                    .workspace_snapshot()?
+                    .attribute_value_view(ctx, value_id)
+                    .await?);
```
fnichol, line 521: 👌🏽

lib/dal/src/component/diff.rs
```diff
@@ -148,8 +148,7 @@ impl Component {
         // Now we need serialized, textual versions of the component on this change set and on head
         // in order to calculate a textural diff
         //
-        // NOTE(fnichol): this ported implementation is still way too expensive and relies on
-        // `AttributeValue.view()` which uses `async_recursion`. This needs to change...
+        // NOTE(fnichol): this ported implementation is still way too expensive
```
fnichol, line 151: hahah, I still agree with this comment change!

lib/dal/src/workspace_snapshot/traits/func.rs
```diff
@@ -0,0 +1,33 @@
...
+    },
+};
+
+#[async_trait]
+pub trait FuncExt {
+    /// A non-dynamic Func is an Intrinsic func that returns a fixed value, set by a StaticArgumentValue in the graph
+    /// opposingly, a dynamic Func is a func that returns a non statically predictable value, possibly user defined.
+    ///
+    /// It's important to note that not all Intrinsic funcs are non-dynamic. Identity, for instance, is dynamic.
+    async fn func_is_dynamic(&self, func_id: FuncId) -> FuncResult<bool>;
```
fnichol, line 18: Yeah, I like that

lib/dal/src/workspace_snapshot/graph/traits/attribute_value.rs
```diff
@@ -1,13 +1,122 @@
...
+    pub attribute_value_id: AttributeValueId,
+    pub prop_id: PropId,
+    pub prop_kind: PropKind,
+}
+
+/// The tree of [`AttributeValueId`] starting from the selected root [`AttributeValueId`].
+#[derive(Debug)]
+pub struct AttributeValueTree {
+    root_av_entry: AttributeValueTreeEntry,
+    children: HashMap<AttributeValueId, Vec<AttributeValueTreeEntry>>,
```
jkeiser, line 57: Curious why have this here instead of a `Vec<>` in `AttributeValueTreeEntry`?

lib/dal/src/workspace_snapshot/traits/func.rs
```diff
@@ -0,0 +1,33 @@
...
+        let func = self
+            .get_node_weight(func_id)
+            .await?
+            .get_func_node_weight()?;
+        match IntrinsicFunc::maybe_from_str(func.name()) {
+            None => Ok(true),
+            Some(intrinsic_func) => Ok(intrinsic_func.is_dynamic()),
+        }
+    }
+}
```
fnichol, line 33: Very cool, and I suspect much cheaper/faster to compute this way

lib/dal/src/qualification.rs
```diff
@@ -214,20 +215,20 @@ impl Ord for QualificationView {
 impl QualificationView {
     pub async fn new(
         ctx: &DalContext,
-        attribute_value: AttributeValue,
```
fnichol, line 217: aye 😢

lib/dal/src/workspace_snapshot/graph/v4/attribute_value.rs
```diff
@@ -1,21 +1,167 @@
...
+                    attribute_value_id: child_id,
+                    prop_id: child_prop_id,
+                    prop_kind: child_prop_kind,
+                });
+            }
+            tree.add_children(attribute_value_id, child_tree_entries);
+        }
+
+        Ok(tree)
+    }
```
fnichol, line 74: Yes yes yes!

lib/dal/src/workspace_snapshot/graph/traits/attribute_value.rs
jhelwig: The tree started out as just the IDs when initially working on this, and `AttributeValueTreeEntry` was added mid-way through when I wanted to also keep the prop ID & kind with each entry. If it were to change in that way, then the entries would no longer be a Copy type, there wouldn't be a need for both tree & tree entry, and getting (and maintaining) the total number of AVs involved would be a lot more complex & time-consuming than `self.children.len()`.

lib/dal/src/workspace_snapshot/traits/attribute_value.rs
```diff
@@ -1,31 +1,175 @@
...
+        let av_tree = self
+            .working_copy()
+            .await
+            .attribute_value_tree(attribute_value_id)?;
+        let mut sub_views = HashMap::new();
+        let mut forward_queue = VecDeque::from_iter([av_tree.root()]);
+        let mut work_stack = Vec::with_capacity(av_tree.count());
+
+        // Build the stack of work, ensuring that all children appear after their parent in the list.
+        while let Some(av_tree_entry) = forward_queue.pop_front() {
```
fnichol, line 55: :boom:

fnichol (approved):
Wow, that *was* quite a refactor and it spread to connected code which makes sense. Nicely done!!

lib/dal/src/workspace_snapshot/graph/v4/attribute_value.rs
```diff
@@ -1,21 +1,167 @@
...
+            .get_node_weight_by_id(root_prop_id)?
+            .get_prop_node_weight()?
+            .kind();
+        let mut tree =
+            AttributeValueTree::new(root_attribute_value_id, root_prop_id, root_prop_kind);
+        let mut work_queue = VecDeque::from([root_attribute_value_id]);
+
+        while let Some(attribute_value_id) = work_queue.pop_front() {
+            let children = self.child_attribute_values_in_order(attribute_value_id)?;
+            work_queue.reserve(children.len());
```
jkeiser: reserve() isn't needed here: when you extend a Vec or VecDequeue with a Vec or slice, Rust ends up calling [append_elements()](https://doc.rust-lang.org/src/alloc/vec/mod.rs.html#2590), which reserves any needed capacity. It should be smart enough to do that with *any* `ExactSizeIterator`, but I didn't find that code so I'm not sure.

lib/dal/src/workspace_snapshot/graph/v4/attribute_value.rs
```diff
@@ -1,21 +1,167 @@
...
+
+        Ok(tree)
+    }
+
+    fn child_attribute_values(
+        &self,
+        attribute_value_id: AttributeValueId,
+    ) -> AttributeValueResult<Vec<AttributeValueId>> {
+        let node_index = self.get_node_index_by_id(attribute_value_id)?;
+        let child_idxs: Vec<_> = self
```
jkeiser: Since you're in `sync` code here, you can avoid the first `Vec` allocation entirely (just don't `.collect()`). You can probably do `try_collect()` as well, which automatically does `with_capacity()` if `.targets()` produces an `ExactSizeIterator`:

```rust
let child_idxs = self.targets(node_index, EdgeWeightKindDiscriminants::Contain);
let child_ids = child_idxs.map(|child_idx| self.node_index_to_id(child_idx).map(Into::into)
  .ok_or_else(|| WorkspaceSnapshotGraphError::NodeWithIndexNotFound(child_idx))
);
Ok(child_ids.try_collect()?);
```

lib/dal/src/workspace_snapshot/split_snapshot/prop.rs
```diff
@@ -0,0 +1,63 @@
...
+
+        match self
+            .attribute_prototype_arguments(prototype_id)
+            .await?
+            .first()
+        {
+            Some(&apa_id) => self
+                .attribute_prototype_argument_static_value(ctx, apa_id)
+                .await
+                .map_err(Into::into),
```
jkeiser, line 46: `?` with surrounding `Ok()` will likely end up more compact here (at the least, the formatter is willing to turn that into a single line if it's small enough).

lib/dal/src/workspace_snapshot/traits/attribute_value.rs
```diff
@@ -1,31 +1,175 @@
...
+                            .await?
+                            .map(Into::into)
+                            .ok_or_else(|| {
+                                WorkspaceSnapshotError::MissingContentFromStore(
+                                    av_tree_entry.attribute_value_id.into(),
+                                )
+                            })?,
+                        PropKind::Array => {
+                            let children = av_tree.children_of(av_tree_entry.attribute_value_id);
+                            let mut entries = Vec::with_capacity(children.len());
```
jkeiser, line 107: Is this what we want here, given it will overallocate if the `if let Some()` is in the loop? (It might be what we want, especially if the map is likely to be close to the same size, because it means we don't have to do multiple allocations while growing. I'm not 100% clear what sub_views is for.)

lib/dal/src/workspace_snapshot/graph/v4/attribute_value.rs
jhelwig: When I was tracing through things before it looked like `Vec::append` reserved, while `Vec::extend` did not, but I should have made a left turn at Albuquerque as I was following the code paths.

lib/dal/src/workspace_snapshot/traits/attribute_value.rs
jhelwig: It would almost certainly overallocate if it was allowed to grow with its normal algorithm since it doesn't always grow by one, and would end up doing multiple allocations in the process.

lib/dal/src/workspace_snapshot/graph/v4/attribute_value.rs
jhelwig: `try_collect` hasn't stabilized yet: https://github.com/rust-lang/rust/issues/94047 https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.try_collect

PR #6637, opened 2025-07-09
feat: only run bedrock into PG once, multiple times is not cool

johnrwatson:
Adds tracer bullet test for the new UI, leveraging bedrock to seed the workspace.

Some work still needed to figure out how to avoid `cy.wait` but the fundamentals are there.

This test will not be executed via the crons due to where it lives in the folder structure. I'm going to write a new cron that executes this specifically on main.

### DIFF
--- a/app/web/cypress/e2e/get-tiles.cy.ts
+++ b/app/web/cypress/e2e/get-tiles.cy.ts
@@ -1,0 +1,101 @@
+/// @ts-check
+///<reference path="../global.d.ts"/>
+
+const AUTH0_USERNAME = Cypress.env('VITE_AUTH0_USERNAME') || import.meta.env.VITE_AUTH0_USERNAME;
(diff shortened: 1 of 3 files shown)


app/web/cypress/e2e/get-tiles.cy.ts
```diff
@@ -0,0 +1,101 @@
...
+    cy.sendPosthogEvent(Cypress.currentTest.titlePath.join("/"), "test_uuid", UUID);
+    
+    cy.wait(30000)
+
+    cy.visit({
+        url: SI_WORKSPACE_URL + '/n/' + SI_WORKSPACE_ID + '/auto',
+        failOnStatusCode: false
+    });
+
+    cy.wait(30000);
```
johnrwatson, line 35: Oh man, I hate this so much but I couldn't quite figure out how to get a reliable test unless I did something like this (page load complete / an element existing, etc all seemed to flake on me)

stack72 (approved):

PR #6639, opened 2025-07-09
feat: Search for components by attribute

jkeiser:
This adds `vpcId:vpc-1234` syntax to the search bar, allowing you to refine a search based on arbitrary attributes in a component.

* **Attribute matching:** `/domain/Foo/Bar:val` matches components where /domain/Foo/Bar = val
* **Partial attribute matching:** `Bar:val` and `Foo/Bar:val` will all match components where /domain/Foo/Bar = val.
  * Partial matches will match partial *paths*, not partial attribute names: `Bar:val` will not match /domain/BarMaid or /domain/OpenBar.
* **Exact matches:** Bar:val will not match /domain/Bar = "evaluation"
* **Progressive matches:** Bar:val will match /domain/Bar = "valuation" so that it progressively refines the set of components as you continue typing each letter the word "Bar:valuation"
  * Empty value does not match attributes with no value: `SecondaryIp:` will match every instance with a SecondaryIp defined, but will not match instances where SecondaryIp: has no value (was never filled in). Yes, I know SecondaryIp is not a thing, I had to make up something that sounded optional.
* **Wildcards matches**: `InstanceType:*.large` will match any large instance type.

#### Screenshots:

image

#### Out of Scope:

* Can't search for boolean and int values. This sucks!
* Default value searches. Not sure if we want x:y to match unset values if the default value of x is y; we should consider.

## How was it tested?

- [X] Integration tests pass
- [X] Manual test: fuzzy search, schema:AWS::EC2::VPC, schema::VPC, InstanceType:, /domain/InstanceType:

### DIFF
--- a/app/web/src/newhotness/Explore.vue
+++ b/app/web/src/newhotness/Explore.vue
@@ -1,6 +1,6 @@
 <template>
   <DelayedSkeleton
-    v-if="componentListRaw.isLoading.value"
+    v-if="componentListQuery.isLoading.value"
     :skeleton="urlGridOrMap"
   />
   <section v-else :class="clsx('grid h-full', showGrid ? 'explore' : 'map')">
@@ -157,7 +157,9 @@
       </div>
       <template v-if="showGrid">
         <div
-          v-if="componentList.length === 0 && componentListRaw.isSuccess.value"
+          v-if="
+            componentList.length === 0 && componentListQuery.isSuccess.value
+          "
           class="flex-1 overflow-hidden flex flex-row items-center justify-center"
         >
           <EmptyState icon="component" text="No components in view" />
@@ -262,7 +264,7 @@
           class="flex-none rounded p-xs"
           variant="key2"
         >
-          Total: {{ componentListRaw.data.value?.length ?? 0 }}
+          Total: {{ componentList.length }}
         </TextPill>
         <TextPill
           v-if="resourceCount > 0"
@@ -313,6 +315,7 @@
 
 <script lang="ts" setup>
 import * as _ from "lodash-es";
+import { computedAsync } from "@vueuse/core";
 import {
   computed,
   inject,
@@ -339,6 +342,7 @@ import { Fzf } from "fzf";
 import {
   bifrost,
   bifrostList,
+  bifrostQueryAttributes,
   useMakeArgs,
   useMakeKey,
 } from "@/store/realtime/heimdall";
@@ -438,14 +442,13 @@ watch(showGrid, () => {
 
 // ================================================================================================
 // SETUP THE FILTERED COMPONENTS REACTIVE AND UPGRADEABLES
-const filteredComponents = reactive<ComponentInList[]>([]);
 const upgrade = useUpgrade();
 const upgradeableComponentIds = computed(() => {
   const set: Set<ComponentId> = new Set();
 
   // TODO(nick): try to swap this with the component list to see if we recompute this less
   // frequently. This is not a problem today, but could be tomorrow.
-  for (const component of filteredComponents) {
+  for (const component of filteredComponents.value) {
     // This needs to be split out into a variable for reactivity. Keep this here or drown in
     // sorrow and suffering. Relevant pull request: https://github.com/systeminit/si/pull/6483
     const canUpgrade = upgrade(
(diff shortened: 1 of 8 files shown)


app/web/src/newhotness/Explore.vue
```diff
@@ -802,49 +795,117 @@ const sortedAndGroupedComponents = computed(() => {
...
+        break;
+      default:
+        break;
     }
+  }
+
+  // Regular fuzzy search across all fields
+  const fzf = new Fzf(components, {
+    casing: "case-insensitive",
+    selector: (c) => `${c.name} ${c.schemaName} ${c.id}`,
```
jobelenus: are we sure thats all we want the fuzzy search on? this removes several things...

adamhjk: yeah, given the existence of the refinment logic. By constraining it this way, the fuzzy search is more likely to find an exact match

app/web/src/newhotness/Explore.vue
```diff
@@ -802,49 +795,117 @@ const sortedAndGroupedComponents = computed(() => {
...
+    return undefined;
+  }
 
-    // Check if the search term contains "schema:" queries
-    const schemaMatches = searchTerm.match(/schema:([^\s]+)/gi);
-    if (schemaMatches) {
-      const schemaNames = schemaMatches.map((match) =>
-        match.substring(7).trim().toLowerCase(),
-      );
+  const result = await bifrostQueryAttributes({
```
jobelenus, line 847: We need to make a change here:
1. Either, wrap this in useQuery with a key that the web worker can bust when AV changes come across.
2. Add something inside this computed that will trigger a re-query when AV changes happen.

The first is easier

app/web/src/newhotness/Explore.vue
```diff
@@ -802,49 +795,117 @@ const sortedAndGroupedComponents = computed(() => {
...
+ *
+ *     {
+ *       attrSearchTerms: [ { key: "vpcId", value: "vpc-123" } ],
+ *       knownSearchTerms: [ { key: "schema", value: "AWS::EC2::Subnet" } ],
+ *       fuzzyTerms: [ "MyComponent", "AWS::EC2::Subnet" ]
+ *     }
+ */
+const debouncedSearchTerms = computed(() => {
+  const attrSearchTerms = new Array<{ key: string; value: string }>();
+  const knownSearchTerms = new Array<{ key: "schema"; value: string }>();
```
jobelenus, line 815: i feel like we'll have a better chance of making the fuzzy search stricter/less pervasive than special casing a list of all potential exact matches but 🤷

jobelenus (changes requested):

app/web/src/newhotness/Explore.vue
jkeiser: Agreed; we need to handle live update! It's noted, and we're planning to address that in a followup (this will only affect the vpcId:xxx terms and not the fuzzy ones, note, so no regression here).

app/web/src/workers/webworker.ts
```diff
@@ -2560,6 +2561,53 @@ from
...
+  terms: { key: string; value: string }[],
+) => {
+  const sql = `
+    SELECT atoms.args AS component_id
+       FROM changesets
+       JOIN indexes ON changesets.index_checksum = indexes.checksum
+       JOIN index_mtm_atoms ON indexes.checksum = index_mtm_atoms.index_checksum
+       JOIN atoms ON atoms.kind = index_mtm_atoms.kind AND atoms.args = index_mtm_atoms.args AND atoms.checksum = index_mtm_atoms.checksum
+       JOIN json_each(jsonb_extract(CAST(atoms.data as text), '$.attributeValues')) AS attr
+      WHERE changesets.change_set_id = ?
```
britmyerss, line 2576: 💥

app/web/src/newhotness/Explore.vue
jkeiser: Adding category back in; for Clover it's redundant, but it's reasonable for people to make schemas where category = AWS::EC2 and name = VPC.

app/web/src/newhotness/Explore.vue
jkeiser: Discussed in zoom to make sure we had the right scope--we added category in to handle this.

jobelenus (dismissed):

jobelenus (approved):
The plan is to use tanStack queries around the attribute search. And we'll use a key ala
```
queryKey: [workspaceId, changeSetId, "AttributeQuery", "<search terms | ulid | `explore`>"]
```
With whatever specificity we need on the front end. And then in `postProcess` when we invalidate, we specifically do:
```
queryClient.invalidateQueries({ filter: [workspaceId, changeSetId, "AttributeQuery"]})
```
and that goes and finds all the other _more specific_ keys

PR #6640, opened 2025-07-09
fix(web): support multiple workspaces

zacharyhamm:
The bifrost websocket is workspace specific, so we need to connect a new socket for every workspace that the user loads.

The sdf client injects the token when it is created, so we also need an sdf client for every workspace that the user loads.

This pr ensures that we initialize a new socket and sdf client for every workspace that the user loads, and that we re-initialize all the necessary sockets and sdf clients when a new tab grabs the lock for the database.

## How was it tested?

- [X] Manual test: switch workspaces in a single tab. Ensure bifrost events flow.
- [X] Open multiple workspaces in multiple tabs. Test updates by editing component properties.
- [x] Close the tab that holds the lock and ensure websockets are connected in the tab that now has the lock.

### DIFF
--- a/app/web/src/workers/shared_webworker.ts
+++ b/app/web/src/workers/shared_webworker.ts
@@ -32,10 +32,10 @@ function debug(...args: any | any[]) {
   if (_DEBUG) console.debug(args);
 }
 
-let bearerToken: string;
 let currentRemote: Comlink.Remote<TabDBInterface> | undefined;
 let currentRemoteId: string | undefined;
 const remotes: { [key: string]: Comlink.Remote<TabDBInterface> } = {};
+const bearerTokens: { [key: string]: string } = {};
 
 const hasRemoteChannel = new MessageChannel();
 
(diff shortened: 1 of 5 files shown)


app/web/src/workers/shared_webworker.ts
```diff
@@ -90,24 +90,23 @@ const dbInterface: SharedDBInterface = {
...
     if (!currentRemote) {
       throw new Error(`remote {$remoteId} not registered`);
     }
     currentRemoteId = remoteId;
 
-    // Ensure we reconnect the websocket if we already had a remote
-    // (otherwise we let heimdall decide when to connect the websocket)
-    if (wasConnected) {
-      currentRemote.bifrostReconnect();
+    for (const workspaceId in bearerTokens) {
```
jobelenus: btw, `Object.entries` gets you both keys and values, so you dont need the `if(objectThatIsDefinitelyIntheMap)`

app/web/src/workers/webworker.ts
```diff
@@ -1235,6 +1235,13 @@ const mjolnirBulk = async (
     inFlightFn(changeSetId, `${o.kind}.${o.id}`);
   });
 
+  const sdf = sdfClients[workspaceId];
+  if (!sdf) {
+    debug("MJOLNIR BULK FAILED: no SDF client for workspace", workspaceId);
```
jobelenus: we should `error` this as well as instrumenting it in the span so we can alert on it

app/web/src/workers/webworker.ts
```diff
@@ -1376,6 +1383,12 @@ const mjolnirJob = async (
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   let req: undefined | AxiosResponse<IndexObjectMeta, any>;
 
+  const sdf = sdfClients[workspaceId];
+  if (!sdf) {
+    debug("no sdf client found for", workspaceId);
```
jobelenus: we should `error` this as well as instrumenting it in the span so we can alert on it

app/web/src/workers/webworker.ts
```diff
@@ -1650,6 +1663,12 @@ const niflheim = async (
   changeSetId: ChangeSetId,
 ): Promise<boolean> => {
   return await tracer.startActiveSpan("niflheim", async (span: Span) => {
+    const sdf = sdfClients[workspaceId];
+    if (!sdf) {
+      debug("No SDF client found for workspaceId", workspaceId);
```
jobelenus: we should `error` this as well as instrumenting it in the span so we can alert on it

jobelenus (approved):
Whenever we need fewer web sockets, we know where to turn to, and splat them all into the shared worker... it means shuttling more stuff around but..

jobelenus (approved):

PR #6654, opened 2025-07-10
feat(web): Adding a keyboard shortcut to toggle minimap

stack72:


app/web/src/newhotness/Explore.vue
```diff
@@ -1323,6 +1325,14 @@ const onR = (e: KeyDetails["r"]) => {
...
   }
 };
+const onM = (e: KeyDetails["m"]) => {
+  e.preventDefault();
+  if (showGrid.value) {
+    // Do nothing in grid mode
+    return;
+  }
+  mapRef.value?.onM(e);
+};
```
wendybujalski, line 1335: Thank you for putting this inside of Explore and then passing the event to the Map!

wendybujalski (approved):
image

app/web/src/newhotness/Explore.vue
stack72: I am just following your pattern :)

PR #6655, opened 2025-07-10
feat(web) - better add component modal newhotness

wendybujalski:

Many small improvements to the add component modal, including -

- Improved fuzzy search!
- It starts with "All" selected and the component categories collapsed
- Reduced the amount of filters to just the useful ones, will add more later
- Better controls, including tab and shift + tab, shift + up or down to jump between categories, and left or right to switch filters
- Cleaner automatic scrolling
- Improved documentation panel with better styles and formatting
- Empty states
- It doesn't randomly change size anymore

#### Screenshots:

Screenshot 2025-07-10 at 7 24 27 PM

#### Out of Scope:

We will definitely want to virtualize the component list eventually, the small amount of lag sometimes experienced in the panel is 100% due to the large amount of DOM elements and will only get worse with more components.

The improved fuzzy search should be ported to replace fuzzy searching elsewhere in the newhotness UI.

Currently we have the SI logo as the icon for a few categories (All, Templates) and will probably want to come up with specific icons for each of those.

We will definitely be adjusting and adding to the filters list. Might be good to have one called "Other" that has everything not in any other category (ignoring All of course!)

## How was it tested?

Oodles of manual testing!

### DIFF
--- a/app/web/src/newhotness/AddComponentModal.vue
+++ b/app/web/src/newhotness/AddComponentModal.vue
@@ -1,11 +1,18 @@
 <template>
   <!-- NOTE: the Modal CSS for height in "max" doesn't work as we might expect -->
-  <Modal ref="modalRef" noWrapper hideExitButton size="max" @click="onClick">
+  <Modal
+    ref="modalRef"
+    noWrapper
+    hideExitButton
+    size="max"
+    class="w-[max(800px,66vw)]"
+    @click="onClick"
+  >
     <div class="h-[45svh]">
       <div
         :class="
           clsx(
-            'grid createcomponent gap-xs [&>*]:border max-h-full',
+            'grid createcomponent gap-xs [&>*]:border h-full',
             selectedAsset && 'grid grid-cols-2',
             themeClasses(
               '[&>*]:bg-shade-0 border-neutral-400 [&_*]:border-neutral-400',
@@ -27,7 +34,7 @@
             <SiSearch
               ref="searchRef"
               v-model="fuzzySearchString"
-              placeholder="Start typing to find components"
+              placeholder="Start typing to filter components"
               :borderBottom="false"
               class="flex-none border"
               @blur="searchRef?.focusSearch()"
@@ -35,6 +42,9 @@
               @keydown.enter.prevent="onEnter"
               @keydown.up.prevent="onUp"
               @keydown.down.prevent="onDown"
+              @keydown.left.prevent="onLeft"
+              @keydown.right.prevent="onRight"
+              @keydown.tab.prevent="onTab"
             >
               <template #right>
                 <div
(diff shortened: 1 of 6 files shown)


stack72 (approved):
so happy about these changes :)

app/web/src/newhotness/AddComponentModal.vue
```diff
@@ -1,11 +1,18 @@
 <template>
   <!-- NOTE: the Modal CSS for height in "max" doesn't work as we might expect -->
-  <Modal ref="modalRef" noWrapper hideExitButton size="max" @click="onClick">
+  <Modal
+    ref="modalRef"
+    noWrapper
+    hideExitButton
+    size="max"
+    class="w-[max(800px,66vw)]"
```
wendybujalski, line 8: Currently we just have the modal be a fixed size, 800 pixels wide or 66% of the window width, whichever is bigger. This will be an issue later if we want the UI to work on viewports under 800 pixels in width.

app/web/src/newhotness/MarkdownRender.vue
```diff
@@ -8,7 +8,39 @@
...
+      if (props.removeMargins) {
+        el.classList.add(tw`m-0`);
+      }
+
+      Array.from(el.children).forEach((child) => walkChildren(child));
+    };
+
+    walkChildren(root);
+  }
+});
```
wendybujalski, line 45: The only way I could figure out to directly add styles to the contents of our Markdown renderer.

app/web/src/newhotness/layout_components/FilterTile.vue
```diff
@@ -6,8 +6,8 @@
         'flex flex-row gap-xs items-center',
         'p-2xs max-w-24 h-10 text-sm border cursor-pointer',
         themeClasses(
-          'border-neutral-400 hover:border-action-500',
-          'border-neutral-600 hover:border-action-300',
+          'border-neutral-400 hover:border-action-500 active:bg-action-500',
+          'border-neutral-600 hover:border-action-300 active:bg-action-300',
```
wendybujalski, line 10: Active styles help to make interactions feel more coherent, we could use em in other places across the UI.

app/web/src/newhotness/AddComponentModal.vue
```diff
@@ -491,11 +629,12 @@ const filteredAssetsFlat = computed(() => {
 });
 
 const showResults = computed(
-  () => !!(debouncedSearchString.value !== "" || selectedFilter.value),
+  () => true, // !!(debouncedSearchString.value !== "" || selectedFilter.value),
```
wendybujalski, line 632: It was easier to just make this computed be true instead of ripping out the "showResults" system. We can remove it later once we're sure we like the results always showing.

app/web/src/newhotness/logic_composables/fzf.ts
wendybujalski, line 1: A logic composable for our ideal fuzzy search! We will definitely want to expand on this and use it all over the new UI.

PR #6657, opened 2025-07-11
feat(sdf,web): use gzip compression when sending attribute update calls

fnichol:
This change adds a middleware to SDF's PUT `component/<id>/attributes` route which requires the client to send the JSON body request compressed using Gzip compression.

On the `app/web` side, the SDF API client now supports calls that must be compressed, of which the "attribute update" call is the first and only one.

The purpose of compressing this request body is in an effort to avoid a Web Application Firewall (i.e. a WAF) from erroneously detecting suspicious or malicious client content being sent to the server backend. In this case some of the attribute value's contents are themselves JSON objects and other elaborate data structures that have been detected and rejected by AWS' WAF product.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -205,7 +205,7 @@ checksum = "3109e49b1e4909e9db6515a30c633684d68cdeaa252f215214cb4fa1a5bfee2c"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.103",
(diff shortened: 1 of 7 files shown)


stack72 (approved):
We verified that this works as expected :)

fnichol:
> We verified that this works as expected :)

YES!!

PR #6660, opened 2025-07-11
Feat/add cron for main check

johnrwatson:
Adds a new pipeline that runs a "validation on merge" run that runs the new UX tests, including the tracer bullet test for the tile view.

Modifies existing e2e pipeline to accept a new "path" variable to allow us to test different groups of tests at once.


.github/workflows/deploy-service-restart.yml
```diff
@@ -84,6 +84,7 @@ jobs:
     uses: ./.github/workflows/e2e-validation.yml
     with:
       environment: ${{ inputs.environment }}
+      test_dir: cypress/e2e
```
stack72, line 87: why did all the test_dirs have to change?

johnrwatson: They haven't changed - I just needed a method of specifying an alternative folder

stack72 (approved):

PR #6661, opened 2025-07-11
feat(web): gracefully handle the deployment of a new shared webworker

zacharyhamm:
If a new shared worker is detected, hotswap in our comlink to that worker and update our remotes and bearer tokens.

Note that if the tab-local worker changes in breaking ways, or if the interface contract between the tab and the shared worker changes, a reload will still be required for other tabs to get the latest code. But that should be handled by our CachedAppNotification component already.

To test this, load the app in a few tabs (for a real test, load more than one workspace locally, with at least 3 tabs). Then go into `heimdall.ts` and modify the `sharedWebWorkerName`. Then reload in just *one* tab. Ensure that all tabs work correctly and that updates flow across the websocket(s).

Then open `chrome://inspect/#workers` or
`about:debugging#/runtime/this-firefox`, There should only be one shared webworker. If you see two, reload the inspect page (it is not always immediately reactive).

Note that after this is deployed, every tab will have to be closed to support the new shared worker code, but only one final time. Afterwards, we should gracefully detect new shared workers without requiring all tabs to close.

### DIFF
--- a/app/web/src/store/realtime/heimdall.ts
+++ b/app/web/src/store/realtime/heimdall.ts
@@ -39,9 +39,14 @@ import { sdfApiInstance as sdf } from "@/store/apis.web";
 import { WorkspaceMetadata } from "@/api/sdf/dal/workspace";
 import { useChangeSetsStore } from "../change_sets.store";
 import { useWorkspacesStore } from "../workspaces.store";
+import {
+  cachedAppEmitter,
+  SHOW_CACHED_APP_NOTIFICATION_EVENT,
+} from "./cached_app_emitter";
 
-// We want an id right away, not later. But ulid fails if run in this context (something about crypto randomValues).
-// we do not need crypto-secure ulids. We just want every tab to have a different one. Which this will get us.
+// We want an id right away, not later. But ulid fails if run in this context
+// (something about crypto randomValues).  we do not need crypto-secure ulids.
+// We just want every tab to have a different one. Which this will get us.
 const ulid = monotonicFactory(() => Math.random());
 
 const ranInit = ref<boolean>(false);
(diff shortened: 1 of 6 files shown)


app/web/src/store/realtime/heimdall.ts
```diff
@@ -50,8 +51,71 @@ const tabDbId = ulid();
...
+const tabDb: Comlink.Remote<TabDBInterface> = Comlink.wrap(tabWorker);
+
+const onSharedWorkerBootBroadcastChannel = new BroadcastChannel(
+  SHARED_BROADCAST_CHANNEL_NAME,
+);
+
+onSharedWorkerBootBroadcastChannel.onmessage = async (msg) => {
+  const name = msg.data as string;
+  if (name !== sharedWebWorkerName) {
+    // This will ensure that the new shared worker is the one we use to
```
britmyerss, line 99: should we hook into the 'something has changed please reload' message to handle this?

zacharyhamm: Yeah we can definitely hook that up here as well. I thought that would appear automatically though?

britmyerss: ah I suppose so! So new sharedwebworker would only ship with a new build entirely, which means if you're on the app during a deploy, we tell you to reload and it would reload it all... yeah?

app/web/src/workers/shared_webworker.ts
```diff
@@ -80,11 +83,25 @@ const dbInterface: SharedDBInterface = {
...
+        }
+      }, SHUTDOWN_DELAY_MS);
+    }
   },
-  registerRemote(id: string, remote: Comlink.Remote<TabDBInterface>) {
+  async registerRemote(id: string, remote: Comlink.Remote<TabDBInterface>) {
     if (!remotes[id]) {
       debug("register remote in shared", id);
       remotes[id] = remote;
+      if (!currentRemote && (await remote.hasDbLock())) {
```
jobelenus, line 102: how come we are asking if it has the lock here?

app/web/src/store/realtime/heimdall.ts
```diff
@@ -61,39 +125,41 @@ export const init = async (
...
+  // If both tabs are refreshed at the same time, this can falsely indicate that
+  // a tab has the lock, but that tab has actually been refreshed just *after*
+  // this call, so *we* now have the lock.  adding 1.5 second timeout here
+  // ensures that there is enough time for the lock to be resolved in a multitab
+  // scenario before we begin cold start. (This only matters if 2+ tabs are
+  // refreshed at more or less the same time, in the normal scenario we will
+  // indicate lock acquisition via the broadcast channel)
   setTimeout(async () => {
-    if (await db.hasRemote()) {
+    if ((await db.hasRemote()) && !(await tabDb.hasDbLock())) {
```
jobelenus, line 166: similarly here, if we `hasRemote` what does it matter that "its not my tabdb that has it"?

app/web/src/workers/shared_webworker.ts
zacharyhamm: When we reboot the shared webworker, we need to (1) re-register all the current tabs, and (2) set the current remote again. So here we check to see if the remote that was just registered has the lock. If it does, make it the current remote.

app/web/src/store/realtime/heimdall.ts
```diff
@@ -50,8 +51,71 @@ const tabDbId = ulid();
...
+    // webworker code is detected. But, note that if the interface changes, this
+    // tab will still have to be reloaded for that communication to work.
+
+    // eslint-disable-next-line no-console
+    console.log("🌈 new shared worker detected, reconnecting");
+    const currentBearers = await db.getBearers();
+    db.unregisterRemote(tabDbId);
+    sharedWorker = spawnSharedWorker(name);
+    sharedWebWorkerName = name;
+    db = Comlink.wrap(sharedWorker.port);
```
jobelenus, line 110: is it correct that without unregistering the remote on the current shared web worker, when we re-spawn the new web worker here _the old one actually won't go away_? (e.g. do we have to clear out all the references from the old shared web worker in order for it to "shut down" b/c I don't see an explicit "shut down" here?)

zacharyhamm: We're shutting down in the shared webworker. On unregister. If the shared webworker has no remotes for 5 seconds, it shuts down

jobelenus: 💡 ahhh missed that, excellent.

app/web/src/store/realtime/heimdall.ts
zacharyhamm: The guard here was mostly to ensure that we set lockAcquired to true in one place only for each scenario:

1. Being the tab that has the lock.
2. Getting the broadcast because the lock has just been acquired in another tab.
3. Detecting the lock after this timeout.


I needed to see which of the three was happening for each tab.

app/web/src/store/realtime/heimdall.ts
zacharyhamm: So yes, the unregister has to happen for the old worker to shut down. If that's what you're asking

app/web/src/store/realtime/heimdall.ts
```diff
@@ -50,8 +55,76 @@ const tabDbId = ulid();
...
+    db.registerRemote(tabDbId, Comlink.proxy(tabDb));
+    if (await tabDb.hasDbLock()) {
+      await db.setRemote(tabDbId);
+    }
+    db.addBearers(currentBearers);
+    showCachedAppNotification();
```
britmyerss, line 116: 🎉

PR #6663, opened 2025-07-11
feat: Allow search with multiple values for a same attribute at once

vbustamante:

Adds the ability to filter by more than one value when searching, by using a | or a comma, like so:
```
EC2::Instance InstanceType:*.large,m8g.medium
 ```

#### Screenshots:
image

## How was it tested?

### CODE REVIEW

File: app/web/src/newhotness/Explore.vue [hunk 1]
```diff
@@ -704,7 +704,7 @@ const sortedAndGroupedComponents = computed(() => {
   // First, always sort by latest to oldest. This relies on the fact ULIDs are time-based.
   // NOTE: We also do this to get a new array, so that later sort() calls do not mutate the
   // filteredComponents array.
-  let components = _.sortBy(filteredComponents.value, (c) => c.id);
+  let components = _.reverse(_.sortBy(filteredComponents.value, (c) => c.id));
```
jkeiser, line 707: Worth a comment on why we're doing reverse (because sortBy doesn't let you sort descending)

app/web/src/newhotness/logic_composables/search.ts
```diff
@@ -145,10 +145,30 @@ class SearchParser {
     // If there is a ":", parse it as an attribute (take the stuff after the :, which might
     // have other colons in it).
     if (this.consume(":")) {
+      const values: string[] = [];
+
+      // Always get first string after :, and if we hit an OR separator, get the next string
+      do {
+        const attrString = this.consumeUntil([
```
jkeiser, line 152: oh no! I wonder if you rename this `val` or `attrVal` if it will be one line again :)

jkeiser (approved):

PR #6664, opened 2025-07-11
feat: clear down old machines, stablise tiles test

johnrwatson:
Tidy up nodes that are spun up and various fixes to the execution of the test variant. Make sure we only try and clean up once, and for the variants that run ec2-node type.

```
🔍 Creating a change set for deletion...
✅ Change set created: 01JZXTQYYPAWTF1DQJHX5WESRH
🔍 Searching for EC2 instance components...
⚠️ Found 1 EC2 components. Deleting...
🗑️ Deleting component: 01JZXSPY4DKEB6YYVE323AXPCH
Waiting for DVU
🚀 Applying change set to execute deletions...
✅ All EC2 components scheduled for deletion.
```

Chore: Extend the wait in the get-tiles test as the lobby is loading slower now with the same test conditions.

**Testing:**
Pass with tools / normal - https://github.com/systeminit/si/actions/runs/16231089952

Pass with ec2-node / new -
https://github.com/systeminit/si/actions/runs/16230918213

### DIFF
--- a/app/web/cypress/ux/tile-view/get-tiles.cy.ts
+++ b/app/web/cypress/ux/tile-view/get-tiles.cy.ts
@@ -32,7 +32,7 @@ describe('web', () => {
         failOnStatusCode: false
     });
 
-    cy.wait(30000);
+    cy.wait(60000);
 
     let attempt = 0;
     let maxAttempts = 10;
--- a/.ci/delete-stacks.py
+++ b/.ci/delete-stacks.py
@@ -1,0 +1,81 @@
+import os
+import requests
+import time
+
+API_TOKEN = os.environ.get('SI_API_TOKEN')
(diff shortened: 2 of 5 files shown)


app/web/cypress/ux/tile-view/get-tiles.cy.ts
```diff
@@ -32,7 +32,7 @@ describe('web', () => {
         failOnStatusCode: false
     });
 
-    cy.wait(30000);
+    cy.wait(60000);
```
johnrwatson, line 35: This test started to flake today, it was getting stuck at the lobby.

I'd love to refactor this to do some kind of watcher, it'd be far more accurate.

stack72 (approved):

PR #6672, opened 2025-07-14
feat(web): compute possible connections per component

zacharyhamm:
We calculated possible connections across every component, with an algo that grew in time linearly with the number of attribute values in the entire change set. But we only need to calculate them per-component!

This improves cold start loads of prod saas in tools (in the no hammer case) from 100+ seconds to around 60 for me.

Please test this before approving. In particular, test the connections panel and ensure that continues to work as expected. Are the right connections there? And do they get removed if a component is removed?


britmyerss (approved):
Tested locally - looks good!!

PR #6675, opened 2025-07-14
feat: Search features: literal booleans fix schema: and support exact values for attrs

vbustamante:

- Support exact attribute matches with quotes: `region:"us-east-1"`
  - Alternatives, wildcards and mixed exact+starts-with matches supported: `InstanceType:*.larger|"m8g.large"`
- Fix schema name searches with `schema:name`
  - Add support for `schema:"AWS::EC2::Subnet"` (exact match), `schema:AWS*Instance` (wildcard) and alternatives, in keeping with attr: syntax
  - attributes named "schema" will still match
- Support searching booleans with `value:true` and `value:false`
- Force exact matches for numeric values: `InstanceCount:1` does not match "10"

## How was it tested?

- [X] Manual tests for each search type, with wildcards and alternatives and exact matches (as well as mixes thereof)


app/web/src/newhotness/Explore.vue
```diff
@@ -924,16 +924,30 @@ const filteredComponents = computedAsync(async () => {
...
           value,
+          op: "startsWith" as const,
         }));
+        const exactTerms = term.exact.map((value) => ({
+          key: term.key,
+          value,
+          op: "exact" as const,
+        }));
+
+        if (exactTerms.length === 0 && startTerms.length === 0) {
```
jkeiser: Could use a comment here!

app/web/src/newhotness/logic_composables/search.ts
```diff
@@ -145,30 +145,42 @@ class SearchParser {
...
-          "&",
-          "!",
-          '"',
-          "|",
-          ",",
-        ]);
-        // Drop empty strings that will be generated by trailing commas or |s
-        if (values.length === 0 || attrString !== "") {
-          values.push(attrString);
+        if (this.consume('"')) {
```
jkeiser, line 154: This branch, and the else branch, are big enough they could probably use comments on the if and on the else.

app/web/src/workers/webworker.ts
```diff
@@ -2682,38 +2683,72 @@ from
 const queryAttributes = (
   _workspaceId: WorkspacePk,
   changeSetId: ChangeSetId,
-  terms: { key: string; value: string }[],
+  terms: QueryAttributesTerm[],
 ) => {
+  // Generate the SQL statements and the respective binds
```
jkeiser, line 2681: I like that these are together now!

app/web/src/workers/webworker.ts
```diff
@@ -2682,38 +2683,72 @@ from
...
+  const sqlTerms = terms.flatMap((term) => {
+    const key = term.key.startsWith("/") ? term.key : `%/${term.key}`;
+    // Extract the correct SQL like statement for our values, using * as the wildcard and respecting exact vs startsWith
+    let value = term.value;
+    if (!term.value.match(/^\d+$/)) {
```
jkeiser, line 2687: This, the boolean, and the schema case are weird enough to deserve comments, as well (at the very least so people don't just remove them thinking they are stray code!)

jkeiser (dismissed):

stack72 (approved):

PR #6679, opened 2025-07-15
fix(auth-api): Upgrade version of posthog-node

stack72:
We believe that this was causing issues sending events to posthog

Screenshot 2025-07-15 at 22 05 44


stack72:
The new container build worked so we are good here - https://buildkite.com/system-initiative/si-merge-queue/builds/6388/steps/canvas?sid=01980f7d-fd11-459e-b02f-fbb96754e7d6

PR #6681, opened 2025-07-15
feat(web) - adjustments to the AttributePanel newhotness

wendybujalski:
Changes to the AttributePanel in slices C & D for ENG-3120. Also some other minor UI fixes.


app/web/src/newhotness/AttributePanel.vue
```diff
@@ -199,6 +199,7 @@
       />
     </div>
   </div>
+  <EmptyState v-else text="No attributes to display" icon="code-circle" />
```
wendybujalski, line 202: Empty states should be inside of the related vue component unless they are different for different usages of that component.

PR #6684, opened 2025-07-15
feat(web): on cold start, do database operations in bulk

zacharyhamm:
Insert atoms, mtms, and do cleanup in single SQL queries to improve frontend performance.

Removes `coldStartComputed`. Handling of atoms is now entirely in `postProcess`.
    
Use QuickLRU for the decodedAtomCache to prevent runaway atom cache sizes.

Also included: use queriesEnabled from the context in more places to ensure we do not query before the database is ready

## Testing this PR

Load a big workspace. Ensure everything is as expected :).

Confirm management edges and possible connections are correct.

### DIFF
--- a/app/web/src/store/realtime/heimdall.ts
+++ b/app/web/src/store/realtime/heimdall.ts
@@ -184,8 +184,6 @@ const bustTanStackCache: BustCacheFn = (
   noBroadcast?: boolean,
 ) => {
   const queryKey = [workspaceId, changeSetId, kind, id];
-  // eslint-disable-next-line no-console
-  console.log("💥 bust tanstack cache for", queryKey);
   queryClient.invalidateQueries({ queryKey });
   if (!noBroadcast) {
     db.broadcastMessage({
--- a/app/web/src/workers/webworker.ts
+++ b/app/web/src/workers/webworker.ts
@@ -1,5 +1,6 @@
 import * as Comlink from "comlink";
 import { applyPatch as applyOperations } from "fast-json-patch";
+import QuickLRU from "quick-lru";
 import sqlite3InitModule, {
   Database,
   ExecBaseOptions,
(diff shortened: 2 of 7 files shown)


app/web/src/workers/webworker.ts
```diff
@@ -404,6 +405,28 @@ const newIndex = (meta: AtomMeta, fromIndexChecksum: string | undefined) => {
   }
 };
 
+const bulkRemoveAtoms = async (
```
jobelenus, line 408: don't need async here?

app/web/src/workers/webworker.ts
```diff
@@ -682,32 +705,88 @@ const handleHammer = async (msg: AtomMessage) => {
...
-    if (exists.length > 0) {
-      return false; // no-op
-    }
-
     db.exec({
       sql: `insert into index_mtm_atoms
         (index_checksum, kind, args, checksum)
           VALUES
         (?, ?, ?, ?)
+        on conflict (index_checksum, kind, args) do nothing
```
jobelenus: we _can't_ do this, the table explicitly has a `ON CONFLICT REPLACE`

jobelenus (changes requested):
I really like that this folds everything into postProcess and we don't need to have a separate coldStartComputed. Nice work.

app/web/src/workers/webworker.ts
zacharyhamm: but it works!

app/web/src/store/realtime/heimdall.ts
```diff
@@ -184,8 +184,6 @@ const bustTanStackCache: BustCacheFn = (
   noBroadcast?: boolean,
 ) => {
   const queryKey = [workspaceId, changeSetId, kind, id];
-  // eslint-disable-next-line no-console
```
zacharyhamm, line 187: This was wayyy too noisy and i don't think it tells us much anymroe

PR #6686, opened 2025-07-16
Add resource values panel

nickgerace:
## Description

This change adds the "ResourceValuesPanel" for looking at a read-only tree of values corresponding resource of a given component. There are three states for this panel: non-existence (the component will not have a resource), empty state (the component does not have a resource) and populated (the component has a resource). The empty state design is copied from "ResourcePanel".

At a nuts and bolts level, the new panel is a stripped down version of the "AttributePanel". In fact, so much has been re-used that some logic has been moved into a new composable library and likely more logic can be moved there in the near future.

## Screenshot: "Will Not Have a Resource"

Screenshot 2025-07-15 at 8 40 23 PM

## Screenshot: "Does Not Have a Resource"

Note: the empty state text has changed since this screenshot was taken!

Screenshot 2025-07-15 at 8 40 38 PM

## Screenshot: "Has a Resource"

Screenshot 2025-07-15 at 8 39 34 PM

### DIFF
--- a/app/web/src/newhotness/ComponentDetails.vue
+++ b/app/web/src/newhotness/ComponentDetails.vue
@@ -112,6 +112,21 @@
             :importFuncRun="latestFuncRuns[importFunc?.id ?? '']"
           />
         </CollapsingFlexItem>
+        <CollapsingFlexItem
+          v-if="hasResourceValueProps"
+          ref="resourceRef"
+          :expandable="false"
+        >
+          <template #header>
+            <div class="flex place-content-between w-full">
+              <span>Resource Values</span>
+            </div>
+          </template>
+          <ResourceValuesPanel
+            :component="component"
+            :attributeTree="attributeTree"
+          />
+        </CollapsingFlexItem>
         <CollapsingFlexItem ref="actionRef" :expandable="false">
           <template #header>Actions</template>
           <ActionsPanel
@@ -234,6 +249,7 @@ import { Context, assertIsDefined } from "@/newhotness/types";
 import { FuncRun } from "@/newhotness/api_composables/func_run";
 import { useRealtimeStore } from "@/store/realtime/realtime.store";
 import AttributePanel from "./AttributePanel.vue";
+import ResourceValuesPanel from "./ResourceValuesPanel.vue";
 import { attributeEmitter, keyEmitter } from "./logic_composables/emitters";
 import CollapsingFlexItem from "./layout_components/CollapsingFlexItem.vue";
 import DelayedLoader from "./layout_components/DelayedLoader.vue";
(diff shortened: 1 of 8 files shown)


app/web/src/newhotness/ComponentDetails.vue
```diff
@@ -305,6 +328,25 @@ const attributeTreeQuery = useQuery<AttributeTree | undefined>({
 });
 const attributeTree = computed(() => attributeTreeQuery.data.value);
 
+// Determines if the component is able to have a resource. For example, for "AWS::EC2::KeyPair",
+// this would return 'true', but fore "AWS Region", this would return false.
+const hasResourceValueProps = computed(() => {
```
nickgerace, line 313: Something about this computed feels messy to me. I think it's because only the panel itself will introspect the tree... we do need some way to know this information though and it's not on the component itself.

app/web/src/newhotness/ResourceValuesPanel.vue
```diff
@@ -0,0 +1,215 @@
...
+
+const q = ref("");
+
+const props = defineProps<{
+  component: BifrostComponent;
+  attributeTree?: AttributeTree;
+}>();
+
+// TODO(nick): move the root computation to a shared location since this is a copy from "AttributePanel".
+const root = computed<AttrTree>(() => {
```
nickgerace, line 85: I would like for both panels to share this in the future. I think you could do it with a `useRoot` composable that returns a computed for given parameters... in this case, just one parameter: the `attributeTree` from the `props`.

app/web/src/newhotness/ResourceValuesPanel.vue
```diff
@@ -0,0 +1,215 @@
...
+const resourceValue = computed(() =>
+  root.value?.children.find((c) => c.prop?.name === "resource_value"),
+);
+
+const filtered = reactive<{ tree: AttrTree | object }>({
+  tree: {},
+});
+
+// TODO(nick): move the root computation to a shared location since this is a copy from "AttributePanel".
+watch(
```
nickgerace, line 117: I think this watch could also move to a composable or maybe even another component. I'm a little more iffy on the ergonomics of how this would work, but I think it's possible.

app/web/src/newhotness/ResourceValuesPanel.vue
```diff
@@ -0,0 +1,215 @@
...
+  });
+  focusSearch();
+});
+
+const focusSearch = () => {
+  searchRef.value?.focusSearch();
+};
+
+onBeforeUnmount(() => {
+  keyEmitter.off("Tab");
```
nickgerace, line 216: I question the efficacy of tabbing when everything is always read-only. Thoughts @wendybujalski?

app/web/src/newhotness/ComponentDetails.vue
nickgerace: DO I SMELL ANOTHER COMPOSABLE???

app/web/src/newhotness/ComponentDetails.vue
```diff
@@ -118,6 +118,28 @@
...
+            v-if="attributeTree && component.hasResource"
+            :component="component"
+            :attributeTree="attributeTree"
+          />
+          <!-- TODO(nick): find a non-wasteful way to share this empty state with "ResourcePanel" -->
+          <EmptyState
+            v-else
+            icon="check-hex"
+            text="This component has no resource"
+          />
```
wendybujalski: It would be cleaner if this `EmptyState` was rolled up inside `ResourceValuesPanel`

app/web/src/newhotness/ComponentDetails.vue
wendybujalski: A messy, complicated conditional is exactly the kind of thing to put into a computed value! Unless we can find a simpler way to figure this boolean out, I think the way you've done it here is best.

wendybujalski (approved):

app/web/src/newhotness/ResourceValuesPanel.vue
wendybujalski: Nah we still want users to be able to navigate as usual when a field is read only.

PR #6687, opened 2025-07-16
feat(web): Show the component resource Id when we have it

stack72:
We know a component has a resource and we should have a resourceId, so 
we can calculate that in the MV and sendit across to use it

Also ensured that the GridTile showed items correctly and that when we 
have a required field in the AttributePanel that we show it as a 
redAsterix

### DIFF
--- a/app/web/src/newhotness/explore_grid/ExploreGridTile.vue
+++ b/app/web/src/newhotness/explore_grid/ExploreGridTile.vue
@@ -55,16 +55,31 @@
         <!-- We need the "grow" here so that the qualification status expands fully. -->
         <ComponentQualificationStatus class="grow" :component="component" />
       </li>
-      <li>
-        <template v-if="component.hasDiff">
+
+      <!-- Rows 2-3: Dynamic content based on priority -->
+      <li v-for="(rowContent, index) in dynamicRows" :key="index">
+        <template v-if="rowContent === 'resource'">
+          <StatusIndicatorIcon
+            v-if="component.resourceId"
+            v-tooltip="'Resource'"
+            type="resource"
+            size="sm"
+            status="exists"
+          />
+          <StatusIndicatorIcon
+            v-else
+            type="resource"
+            size="sm"
+            status="exists"
+          />
+          <div v-if="component.resourceId" class="text-xs opacity-75">
+            {{ component.resourceId }}
+          </div>
+          <div v-else>Resource</div>
+        </template>
+        <template v-else-if="rowContent === 'diff'">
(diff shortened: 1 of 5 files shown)


app/web/src/newhotness/explore_grid/ExploreGridTile.vue
```diff
@@ -156,6 +171,21 @@ const canBeUpgraded = computed(() =>
...
+const dynamicRows = computed(() => {
+  const rows: (string | null)[] = [];
+
+  // Priority order: resource first, then diff
+  if (props.component.hasResource) rows.push("resource");
+  if (props.component.hasDiff) rows.push("diff");
+
+  // Ensure we always have exactly 2 rows (for rows 2-3)
+  while (rows.length < 2) {
+    rows.push(null);
```
stack72, line 183: This keeps us at a consistent layout style for grid tiles!

app/web/src/newhotness/explore_grid/ExploreGridTile.vue
```diff
@@ -55,18 +55,33 @@
...
+            type="resource"
+            size="sm"
+            status="exists"
+          />
+          <div v-if="component.resourceId" class="text-xs opacity-75">
+            {{ component.resourceId }}
+          </div>
+          <div v-else>Resource</div>
+        </template>
+        <template v-else-if="rowContent === 'diff'">
```
wendybujalski, line 80: This is very cool! I assume that you made sure that the `GRID_TILE_HEIGHT` defined on line 205 is still accurate?

app/web/src/newhotness/layout_components/AttributeInput.vue
```diff
@@ -15,10 +15,25 @@
         <!-- Attribute name -->
         <div class="flex flex-row items-center gap-2xs pl-xs">
           <TruncateWithTooltip>{{ displayName }}</TruncateWithTooltip>
+          <span
+            v-if="showRequiredAsterisk"
+            v-tooltip="`${displayName} is a required value`"
+            :class="
+              clsx(
+                'shrink-0 text-lg font-bold',
+                themeClasses('text-destructive-600', 'text-destructive-300'),
```
wendybujalski, line 24: I am surprised that these colors are not our more common 500/400 combo for this type of thing.

wendybujalski (approved):

app/web/src/newhotness/layout_components/AttributeInput.vue
stack72: This should provide better contrast ratios for both themes while keeping the asterisk appropriately subtle.

PR #6689, opened 2025-07-16
feat: New empty explore page

vbustamante:

- Updates design for empty components grid
- Adds some new arguments for empty state component (iconSize, iconBg and slot for secondary content) 

#### Screenshots:

image


app/web/src/newhotness/EmptyState.vue
```diff
@@ -3,33 +3,42 @@
...
+} from "@si/vue-lib/design-system";
 import clsx from "clsx";
 import { PropType } from "vue";
 
 defineProps({
   text: { type: String, required: true },
   secondaryText: { type: String },
   icon: { type: String as PropType<IconNames>, required: true },
+  iconSize: { type: String as PropType<IconSizes> },
+  iconBg: { type: Boolean },
```
wendybujalski: Perhaps the default for `iconBg` should be true? Or maybe it should be `noIconStyles` instead?

app/web/src/newhotness/EmptyState.vue
```diff
@@ -3,33 +3,42 @@
     <div
       :class="
         clsx(
-          'mt-sm p-sm rounded-full',
-          themeClasses('bg-neutral-200', 'bg-neutral-700'),
+          iconBg && 'mt-sm p-sm rounded-full',
+          iconBg && themeClasses('bg-neutral-200', 'bg-neutral-700'),
```
wendybujalski: You can put all of these styles under one conditional like this -

`iconBg && ['mt-sm p-sm rounded-full', themeClasses('bg-neutral-200', 'bg-neutral-700')]`

app/web/src/newhotness/Explore.vue
```diff
@@ -160,9 +160,58 @@
...
+              <a href="https://discord.gg/system-init" target="_blank">
+                Join the community
+              </a>
+              <a
+                :href="`https://auth.systeminit.com/workspace/${ctx?.workspacePk.value}`"
+                target="_blank"
+                >Invite a team mate</a
+              >
+            </div>
+          </div>
```
wendybujalski: It looks like the styles in this section are for dark mode only?

app/web/src/newhotness/EmptyState.vue
wendybujalski: Basically we should avoid having to set this prop to true so many times in so many places when the default behavior is clearly to show the icon background.

app/web/src/newhotness/Explore.vue
vbustamante: @wendybujalski now there's light mode stuff

wendybujalski (approved):

stack72:
Ncie work @vbustamante

PR #6702, opened 2025-07-17
fix(web): Truncate with tooltiip the resource Id if it's too long

stack72:


app/web/src/newhotness/explore_grid/ExploreGridTile.vue
```diff
@@ -75,9 +75,12 @@
             size="sm"
             status="exists"
           />
-          <div v-if="component.resourceId" class="text-xs opacity-75">
+          <TruncateWithTooltip
+            v-if="component.resourceId"
+            class="text-xs pb-xs opacity-75"
```
wendybujalski: This should be `py-xs` instead of `pb-xs` - having the padding only on the bottom causes the text to be pushed upward out of alignment with everything else on the row.

wendybujalski (approved):

PR #6703, opened 2025-07-17
Do not coerce value for subscriptions

nickgerace:
## Descriptrion

This change ensures that we do not coerce a value when dealing with subscriptions. This bug was found when trying to subscribe from (to?) a prop with kind "integer", but the value would be the prop path, which is a "string". This would result in "NaN" being the coerced value and then "null" sent to the attributes API. Now, we don't even try to coerce if we are dealing with a subscription.

Screenshot 2025-07-17 at 3 16 21 PM


stack72 (approved):

aaron-dernley:
Legend

jobelenus (approved):

PR #6705, opened 2025-07-17
Jobelenus/replace navbar

jobelenus:
# Goal

- Decouple the navbar in new hotness from the old system
- Remove `useChangeSetsStore` from newhotness.
- Secondary, remove `useWorkspacesStore`, but this is less impactful.

I _wasn't_ entirely successful in removing the change set store from everywhere... that thing found itself into so many places, and I had to stop somewhere. So I am choosing here 😅 

1. Wrap loading change sets into the `change_set` composable so they can be used both by the top-level `Workspace.vue` as well as the `ChangeSetsPanel`.
2. This is a bit of a catch-22 move, because I need the `Context` in order to `useApi`, but in `Workspace` we're _creating_ it. So I made a small work-around, both for `watch`ing data that comes back from loading the change sets data—as well as passing the `ctx` into `useApi`
3. I also moved around some of Nick's pieces for the composable he was using to apply change sets.
4. `NavbarPanelLeft` now loads the workspace query. Which means `useApi` also knows how to hit the auth API now.
5. We're making a few override decisions in creating the URL for the API with some one-offs for these change set and workspace endpoints since they don't fit nicely into the v2 structure.
6. Copy navbar elements and remove the workspace & change set store from them replacing with the values from `Context`
7. And we've now got some more expanded data inside the `Context` object.
8. In order to remove change set store from `heimdall` i needed to remove the deprecated bits from the old UI that used `makeKey` and `makeArgs` that didn't use the `Context`. Which led to deleting some of the `mead-hall` components that were "off" but referencing `heimdall`.

Since I couldn't entirely eliminate every loading of the change set store, I have copied and modified (mostly simplified) the WSEvent subscriptions in `Workspace` but left them commented out. We can't uncomment these subscription until the change set store is fully excised otherwise we'll get duplicate toasts and operations.

#### Out of Scope:
There are a few change set store pieces remaining (one in `CodeEditor`?!). We'll get there.

I left a fair bit of the `workspacesStore` in the nav bar which dealt with import / export / integrations, etc. That store is lower priority.

## How was it tested?

- Creating a change set
- Abandoning a change set & _make sure you only see one toast_
- **All subsequent tests when re-using older components or have any potential of bringing back in the change set store should involve abandoning a change set. If you see TWO toasts, you brought it back in**


app/web/src/newhotness/api_composables/index.ts
```diff
@@ -142,15 +158,33 @@ export class APICall<Response> {
   }
 
   url(): string {
+    if (
```
jobelenus, line 161: Special case these paths since they don't fit into the v2 pattern

app/web/src/newhotness/api_composables/index.ts
```diff
@@ -142,15 +158,33 @@ export class APICall<Response> {
...
+      return this.path;
+    }
+    if ([_routes.ChangeSets].includes(this.path)) {
+      return `v2/workspaces/${this.workspaceId}/change-sets`;
+    }
     const API_PREFIX = `v2/workspaces/${this.workspaceId}/change-sets/${this.changeSetId}`;
     return `${API_PREFIX}${this.path}`;
   }
 
+  api(): AxiosInstance {
```
jobelenus, line 177: Identify which backend to point at

app/web/src/newhotness/api_composables/index.ts
```diff
@@ -365,11 +407,30 @@ export const useApi = () => {
...
 
+  type ApiRequestStatus = {
+    isRequested: boolean;
+    isPending: boolean;
+    isFirstLoad: boolean;
+    isError: boolean;
+    isSuccess: boolean;
+  };
+
+  const requestStatuses = computed<ApiRequestStatus>(() => {
```
jobelenus, line 418: Returning a structure that matches what the Icon statuses expect for easy parity!

app/web/src/newhotness/layout_components/AttributeInput.vue
```diff
@@ -586,7 +586,7 @@
       <CodeEditorModal
         ref="codeEditorModalRef"
         :title="`Set Value For ${displayName}`"
-        :codeEditorId="`${displayName}-${changeSetId}-${prop?.id}`"
```
jobelenus, line 589: code editor actually pulls in the change set store to prepend the changeSetId to the key, so this isn't needed

app/web/src/newhotness/logic_composables/change_set.ts
```diff
@@ -42,14 +79,11 @@ const useDisableApplyChangeSetInner = () => {
...
 
-  const currentChangeSet = useCurrentChangeSet();
   const status = useStatus();
 
-  const changeSet = computed(() => currentChangeSet.value);
-
   return computed(
     () =>
-      changeSet.value?.status !== ChangeSetStatus.Open ||
+      ctx.changeSet.value?.status !== ChangeSetStatus.Open ||
```
jobelenus, line 86: Now, the `Context` has the `changeSet` the user is on, so we don't need to re-use the other composable

app/web/src/newhotness/nav/WorkspaceSettingsMenuItems.vue
wendybujalski, line 1: ooo i like this!

app/web/src/newhotness/nav/NavbarPanelRight.vue
```diff
@@ -0,0 +1,223 @@
...
+            "
+          />
+        </template>
+      </NavbarButton>
+    </template>
+
+    <template v-if="!collapse">
+      <NavbarButton
+        icon="grid"
+        tooltipText="Use New UI"
```
wendybujalski: Shouldn't this be "Use Old UI?"

wendybujalski (dismissed):

PR #6707, opened 2025-07-17
fix: Remove old values when migrating subscriptions

jkeiser:
Migrating array socket connections creates duplicate values right now.

### The Migration Problem

Specifically, when you have a item connected to an array via socket (and si:normalizeToArray), and try to migrate:

1. Initial state: `Foo: /domain/InstanceIds` is connected to `Bar: /resource_value/InstanceId` through a socket.
   a. `/domain/InstanceIds` has no component-specific prototype. It gets its value from the "default" (the prop's prototype).
   b. The prop's prototype is `si:normalizeToArray`, and the socket connection is its argument.
   c. It has 1 child attribute value, created by DVU, representing the value at the other end of the socket connection.
2. Migration: first, we remove the socket connection.
   a. We remove the socket connection argument from the prop's prototype.
   b. This does *not* remove the actual child value from the array. It still has 1 element.
   c. This would happen if DVU ran, but we don't give it time to run here.
3. Migration: next, we create a prop subscription `TaskDef: /domain/ContainerDefinitions/-` <- ContainerDefinition
   a. We override the connection prototype by creating a manual array, setting component-specific prototype `si:setArray([])` so we can add a child.
   b. We do *not* currently remove children from the array. This means it still has 1 element.
   c. We add a new child array element to `/domain/ContainerDefinitions` and put the subscription on it. Now it has 2 elements.

In the end, we end up with 2 elements: the migrated prop subscription, and the "manual" array element that came from the no-longer-existent socket connection. If either 2b or 3b had happened, we would be OK here. We solve this by making 3b happen.

## The Fix

This PR fixes the issue by modifying `vivify_json_pointer()` (used in PUT /attributes, luminork, and migrations) to properly override the pointer. In particular:
* When `/Foo` is driven by a dynamic "default" function from its prop (e.g. si:normalizeToArray, si:identity, socket connections)
* And you want to set `/Foos/-` (or for maps, `/Foos/Bar`)
* The vivify() method currently *overrides* the Foo "default" prototype from the schema using `set_value()`, using `AttributeValue::update([] or {})`.
* This removes any existing array or map values, that may have been set by the schema prototype. Before this, we used `set_value()`, which did not remove children.

## How was it tested?

- [X] Integration tests pass
- [X] New test: migrating a child value of a connection removes the data from that connection (for map/object/array).
- [X] Manual test: migrate TaskDefinition -> ContainerDefinition connection, see only one value

### DIFF
--- a/lib/dal/src/attribute/value.rs
+++ b/lib/dal/src/attribute/value.rs
@@ -114,6 +114,7 @@ use crate::{
             NodeWeight,
             NodeWeightDiscriminants,
             NodeWeightError,
+            PropNodeWeight,
         },
         serde_value_to_string_type,
         traits::attribute_value::AttributeValueExt,
@@ -156,8 +157,8 @@ pub enum AttributeValueError {
     CannotCreateSocketValueWithoutComponentId,
     #[error("cannot explicitly set the value of {0} because it is for an input or output socket")]
     CannotExplicitlySetSocketValues(AttributeValueId),
-    #[error("cannot set child of value {0} because it has a dynamic func {1}")]
-    CannotSetChildOfDynamicValue(AttributeValueId, FuncId),
+    #[error("cannot set child of value {0} because it has a dynamic prototype")]
+    CannotSetChildOfDynamicValue(AttributeValueId),
     #[error("change set error: {0}")]
     ChangeSet(#[from] ChangeSetError),
     #[error(
@@ -858,11 +859,10 @@ impl AttributeValue {
         ctx: &DalContext,
         input_attribute_value_id: AttributeValueId,
     ) -> AttributeValueResult<Vec<Value>> {
-        let maybe_input_socket_id =
-            match AttributeValue::is_for(ctx, input_attribute_value_id).await? {
-                ValueIsFor::InputSocket(input_socket_id) => Some(input_socket_id),
-                _ => None,
-            };
+        let maybe_input_socket_id = match Self::is_for(ctx, input_attribute_value_id).await? {
+            ValueIsFor::InputSocket(input_socket_id) => Some(input_socket_id),
+            _ => None,
+        };
 
         let Some(input_socket_id) = maybe_input_socket_id else {
             return Ok(vec![]);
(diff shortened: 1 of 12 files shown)


lib/dal/src/attribute/value.rs
```diff
@@ -2193,9 +2171,9 @@ impl AttributeValue {
     pub async fn prop(
         ctx: &DalContext,
         attribute_value_id: AttributeValueId,
-    ) -> AttributeValueResult<Prop> {
+    ) -> AttributeValueResult<PropNodeWeight> {
```
jhelwig, line 2174: Don't know that this needs to change right now, but changing this to return a `PropNodeWeight` feels "leaky"; like it's exposing information that shouldn't make it out this far. Especially in a `pub` function. If it weren't `pub` I don't think it would be as much of a concern.

jkeiser: I think it's important that none of the callers use the CAS parts of Prop, though, and I'm sort of loathe to revert it so they are grabbing it ... would it be worth changing all the callers to do `Prop::node_weight(ctx, AttributeValueId::prop_id(ctx, av_id).await?).await?`?

Maybe `AttributeValue::prop_node()` or something.

jkeiser: Many of the users probably use just `kind`, too--but there are many who use both `name` and `kind`, so a `prop_kind()` method on AV probably doesn't give the necessary ergonomics benefit.

jhelwig: I'd prefer to not have _any_ public interface on `AttributeValue` (or any of the other domain objects) that returns a `NodeWeight`. If that were not `pub`, using it as a helper internally for the `AttributeValue` methods would make sense, and be fine.

Looking at the usage of `AttributeValue::prop` outside of `AttributeValue` it looks like they're entirely concerned with either the kind, or the name (and I don't see any places that are concerned with both at the same time). If there were `AttributeValue::prop_kind` and `AttributeValue::prop_name` methods, I think those would be reasonable conveniences (that probably should internally delegate to `Prop::kind` & `Prop::name` methods that actually deal with the `PropNodeWeight` itself).

jhelwig: > Many of the users probably use just `kind`, too--but there are many who use both `name` and `kind`, so a `prop_kind()` method on AV probably doesn't give the necessary ergonomics benefit.

I don't actually see any internal to `AttributeValue` that are interested in both the `name` and `kind`. I do see one place that is interested in both the `id` and the `kind`, though.

jkeiser: Sounds good! I think the one that wants `id` and `kind` can *reasonably* do two separate calls anyway--it's only people who want both name and kind that I'd worry about. I'll take a stab at this, thanks!

jkeiser: I'll make this change in a followup PR, so we don't overburden this one!

PR #6709, opened 2025-07-18
feat(web,edda): Add RunTemplate management func kind and expose it at the top of the attributes panel

britmyerss:
Adds another `ManagementFuncKind` to the system for our templates.  Eventually we'll want something better than name matching, but this is serving our needs for now. Also, when a component has a template function, we will expose it at the top of the attributes panel as we do for `Import` functions. 

#### Screenshots:
image

#### Out of Scope:
Better interaction between this and the management panel. I temporarily disabled the ability to run this kind of management function from the Management Panel, but I didn't remove it entirely so we can continue to have a way to view the details of the run. 

## How was it tested?
Installed the new VPC template, confirmed that the management func kind is correct and clicking the button runs the func!

<div><br/>via <a href="https://giphy.com/theoffice/">The Office</a> on <a href="https://giphy.com/gifs/theoffice-gQm37vKLwkUVjOvVmp">GIPHY</a></div>

### DIFF
--- a/app/web/src/newhotness/ComponentDetails.vue
+++ b/app/web/src/newhotness/ComponentDetails.vue
@@ -105,6 +105,20 @@
                   "
                 />
               </template>
+              <template v-else-if="runTemplateFunc">
+                <VButton
+                  size="sm"
+                  label="Run Template"
+                  loadingText="Running"
+                  tone="neutral"
+                  class="ml-2xs mr-xs font-normal"
+                  :loading="
+                    dispatchedFunc ||
+                    latestFuncRuns[runTemplateFunc?.id]?.state === 'Running'
+                  "
+                  @click.stop="runMgmtFunc(runTemplateFunc?.id)"
+                />
+              </template>
             </div>
           </template>
           <AttributePanel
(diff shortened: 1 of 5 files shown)


app/web/src/newhotness/ComponentDetails.vue
```diff
@@ -391,9 +406,46 @@ export type NameFormData = {
...
+  dispatchedFunc.value = true;
+  setTimeout(() => {
+    dispatchedFunc.value = false;
+  }, 2000);
+
+  // NOTE(nick): need to make sure this makes sense after the timeout.
+  if (mgmtRunApi.ok(req) && newChangeSetId) {
+    mgmtRunApi.navigateToNewChangeSet(
+      {
+        name: "new-hotness",
```
jobelenus: shouldn't this be `new-hotness-component` with a `componentId`?

britmyerss: oo yes - thank you. fixing!

app/web/src/newhotness/ManagementFuncCard.vue
```diff
@@ -98,10 +99,11 @@ const runMgmtFunc = async (funcId: string) => {
   if (mgmtRunApi.ok(req) && newChangeSetId) {
     mgmtRunApi.navigateToNewChangeSet(
       {
-        name: "new-hotness",
+        name: "new-hotness-component",
```
britmyerss, line 102: @jobelenus I copied from here so updating this one too so we redirect appropriately.

PR #6710, opened 2025-07-18
feat(web, sdf): Ability to manually re-enqueue a qualification to run

britmyerss:
Adds an SDF route to enqueue an attribute value in DVU if it is set by a dynamic function. 

NOTE: This is permitted to be run on head! 

#### Screenshots:
image

https://jam.dev/c/b69d826c-e6ba-4542-8a7f-73fa8616530c 

#### Out of Scope:
I didn't wire up the FuncRunDetails page to let you re-enqueue because we'll have to restructure that a bit to be ensure we reload when the new func runs. 

## How was it tested?

<div><br/>via <a href="https://giphy.com/pbs/">PBS</a> on <a href="https://giphy.com/gifs/pbs-season-3-drama-hotel-portofino-DL3z7KqrZjq8nZX2Jp">GIPHY</a></div>


app/web/src/store/change_sets.store.ts
```diff
@@ -585,12 +585,14 @@ export function useChangeSetsStore() {
...
                     // if you're on a single-item page in the UI while just bring them to explore
+                    // but only if you're not already on head!
                     if (
                       ![
                         "new-hotness",
                         "new-hotness-workspace-auto",
                         "new-hotness-head",
-                      ].includes(route.name as string)
+                      ].includes(route.name as string) &&
+                      this.selectedChangeSetId !== this.headChangeSetId
```
jobelenus, line 595: why are we adding this?

app/web/src/newhotness/api_composables/index.ts
```diff
@@ -28,6 +28,7 @@ export enum routes {
   DeleteComponents = "DeleteComponents",
   DeleteView = "DeleteView",
   DuplicateComponents = "DuplicateComponents",
+  EnqueueAttributeValue = "EnqueueAttributeValue",
```
jobelenus, line 34: if this can run against HEAD it also needs to be placed inside the `CAN_MUTATE_ON_HEAD` array

jobelenus (changes requested):

jobelenus:
@britmyerss so long as we invalidate this query inside `<LatestFuncRunDetails>` (where the URL is the action ID)
```
const actionFuncRunQuery = useQuery<string>({
  queryKey: ["action_func_run_id", props.actionId],
```
When you hit "retry", you're all set & reactive 😉

app/web/src/newhotness/api_composables/index.ts
britmyerss: ah thank you! I swear I had added this but must have gotten lost in my cleanup.

app/web/src/store/change_sets.store.ts
britmyerss: This fixes a bug I found when testing this on head where if you mutate head (or head applies) while you're not on the grid view, you were being sent back to grid every time.

lib/sdf-server/src/service/v2/component/attributes.rs
```diff
@@ -85,3 +94,40 @@ async fn update_attributes(
...
+        ChangeSetId,
+        ComponentId,
+    )>,
+    Json(values): Json<Vec<AttributeValueIdent>>,
+) -> Result<()> {
+    let mut did_enqueue = false;
+    for value in values {
+        if let Some(attribute_value_id) = value.resolve(ctx, component_id).await? {
+            if AttributeValue::is_set_by_dependent_function(ctx, attribute_value_id).await? {
+                ctx.add_dependent_values_and_enqueue(vec![attribute_value_id])
```
zacharyhamm, line 112: That should do it

jobelenus (approved):

PR #6718, opened 2025-07-18
fix: Makes sure the components list always reflects what's on the search bar

vbustamante:
invalidate the search query immediately when mounting explore page so we don't render cached component lists that don't match the search terms


jobelenus (approved):

jkeiser (approved):
Looks good :) I wish I could think of a simple way to actually reuse the results of the last search we did as long as things haven't changed!

PR #6720, opened 2025-07-18
fix(web): force leader election, restart queries if leader changed mid query

zacharyhamm:
This pr makes three changes that should increase the reliablity of the
multitab experience:

  1. If lock acquisition has not been detected after 5 seconds, attempt
     to force leader election.
  2. If the leader changes in the middle of a request, that request will
     likely hang. So detect this change using Promise.race and retry the
     call in the case that leader election fails.
  3. Additionally, we have given `sqlite` it's correct type (it's possibly undefined). If we run into an undefined sqlite during leader re-election, we will also retry (with a sleep).

Also upgrades sqlite-wasm to 3.50.3-build1, which was required to force
leader election by closing the database and pausing the VFS.

Test this: 

1. Open the same workspace in a lot of tabs, on the new UI.
2. Have the console open on them all. Go into tilt and restart the web service. 
3. Once the tabs are all loaded and in the lobby, figure out which tab has the lock (it will be the one with lots of "busts processed" and "`? sql`" messages). Hit refresh on that tab, preferably in between a NIFLHEIM but before a DONE message on another tab.
4. Ensure all workspaces still load.

You'll want a workspace with a good number of components and change sets, so that it loads slow enough to exercise the restart.

### DIFF
--- a/app/web/src/workers/webworker.ts
+++ b/app/web/src/workers/webworker.ts
@@ -7,6 +7,7 @@ import sqlite3InitModule, {
   ExecReturnResultRowsOptions,
   ExecRowModeArrayOptions,
   FlexibleString,
+  SAHPoolUtil,
   Sqlite3Static,
   SqlValue,
 } from "@sqlite.org/sqlite-wasm";
@@ -46,6 +47,7 @@ import {
   Checksum,
   Common,
   ComponentInfo,
+  FORCE_LEADER_ELECTION,
   Gettable,
   Id,
   IndexObjectMeta,
@@ -61,6 +63,7 @@ import {
   Ragnarok,
   RainbowFn,
   TabDBInterface,
+  DB_NOT_INIT_ERR,
   UpdateFn,
 } from "./types/dbinterface";
 import {
@@ -131,7 +134,8 @@ function debug(...args: any | any[]) {
 /**
  *  INITIALIZATION FNS
  */
-let sqlite: Database;
+let sqlite: Database | undefined;
+let poolUtil: SAHPoolUtil | undefined;
 const sdfClients: { [key: string]: AxiosInstance } = {};
 
 const getDbName = (testing: boolean) => {
@@ -150,7 +154,11 @@ const start = async (sqlite3: Sqlite3Static, testing: boolean) => {
   const dbname = getDbName(testing);
 
   if ("opfs" in sqlite3) {
-    const poolUtil = await sqlite3.installOpfsSAHPoolVfs({});
+    if (!poolUtil) {
+      poolUtil = await sqlite3.installOpfsSAHPoolVfs({});
+    } else if (poolUtil.isPaused()) {
+      await poolUtil.unpauseVfs();
+    }
     sqlite = new poolUtil.OpfsSAHPoolDb(`/${dbname}`);
     debug(
       `OPFS is available, created persisted database in SAH Pool VFS at ${sqlite.filename}`,
@@ -184,7 +192,7 @@ const dropTables = () => {
   DROP TABLE IF EXISTS changesets;
   DROP TABLE IF EXISTS weak_references;
   `;
-  sqlite.exec({ sql });
+  sqlite?.exec({ sql });
 };
 
 // INTEGER is 8 bytes, not large enough to store ULIDs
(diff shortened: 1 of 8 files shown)


app/web/src/workers/webworker.ts
```diff
@@ -150,7 +153,11 @@ const start = async (sqlite3: Sqlite3Static, testing: boolean) => {
   const dbname = getDbName(testing);
 
   if ("opfs" in sqlite3) {
-    const poolUtil = await sqlite3.installOpfsSAHPoolVfs({});
+    if (!poolUtil) {
+      poolUtil = await sqlite3.installOpfsSAHPoolVfs({});
+    } else if (poolUtil.isPaused()) {
```
jobelenus, line 159: @zacharyhamm i'd love to better understand this bit, why / when does this promise pause?

zacharyhamm: in order to force leader election, we have to close the database and pause the vfs. but then the tabs will race to be leader again, and the tab that paused could be the one that wins. in which case, it needs to unpause.

app/web/src/workers/webworker.ts
```diff
@@ -1928,38 +1935,42 @@ const niflheim = async (
...
-        bustCacheAndReferences(
-          workspaceId,
-          changeSetId,
-          atom.kind,
-          atom.id,
-          false,
-          true,
-        );
+        // Placing this in a promise to yield control back to the event loop
+        await processAtom(atom);
```
jobelenus, line 2034: I assume this is to let the `Promise.race` fail earlier, rather than have to wait? 🤔 Not sure I'm thinking about that correctly just yet

app/web/src/workers/webworker.ts
```diff
@@ -3261,12 +3275,16 @@ const dbInterface: TabDBInterface = {
...
+          forceLeaderElectionBroadcastChannel.onmessage = () => {
+            abortController?.abort(FORCE_LEADER_ELECTION);
+          };
           abortController?.signal.addEventListener("abort", () => {
             hasTheLock = false;
+            db.close();
+            poolUtil?.pauseVfs();
             this.bifrostClose();
-            reject(abortController?.signal.reason);
+            resolve(abortController?.signal.reason);
```
jobelenus, line 3388: Whats the thinking behind resolving, rather than rejecting here? Just curious

app/web/src/workers/webworker.ts
zacharyhamm: This is to ensure messages to the webworker that has the lock can be processed in a timely way even while in the middle of a cold start. It's only tangentially related, mostly I found in testing that the leader election message would wait until a niflheim was finished before the message landed

app/web/src/workers/webworker.ts
jobelenus: Hrm, how come we've been able to do leader election without this pausing (until now of course)?

app/web/src/workers/webworker.ts
jobelenus: How does the `await` help them be done "timely" if they can just be done synchronously?

app/web/src/workers/webworker.ts
zacharyhamm: Wrapping this in `try catch` never caught the error when I tested it, I think because of the comlink? But with resolve it worked.

app/web/src/workers/webworker.ts
zacharyhamm: The history of this one is that I was trying out a strategy that would try and round trip from the current tab to the tab with the lock through the shared worker, with a `ping`. But the pings would timeout if the tab with the lock was running a cold start, since it would never reply until the cold start was over (since the event loop was stuck processing the atoms). This made the pings respond. Although I removed that code in the end, it seems ideal to prevent the event loop from locking up, and this is simpler than adding a `globalThis.scheduler?.yield()` polyfill.

app/web/src/workers/webworker.ts
zacharyhamm: In the other cases, leader election happened when a tab was closed, which kills the sahpool and closes the database. But in this case we're trying to force leader election with all tabs still open

jobelenus (approved):

PR #6721, opened 2025-07-18
Add new validation failure UI

nickgerace:
## Description

This change adds a new UI element for failed validations. When a validation is failing that is not a "requirement" one, we now expand the area around the input box and indicate within the panel why the field is failing. This to make the usability flow for both keyboard and mouse users better. The tooltip has been removed and we will show the full message..

This change also ensures that the asterisk is present when the possible connections box is present.

## Screenshot: New Validation UI

Standard property:

Screenshot 2025-07-21 at 10 13 49 AM

Array item property:

Screenshot 2025-07-21 at 10 14 36 AM

## Screenshot: Asterisk in the Possible Connections UI

Screenshot 2025-07-18 at 5 42 29 PM

### DIFF
--- a/app/web/src/newhotness/layout_components/AttributeInput.vue
+++ b/app/web/src/newhotness/layout_components/AttributeInput.vue
@@ -9,37 +9,27 @@
             'grid grid-cols-2 items-center gap-2xs relative text-sm font-normal',
             inputOpen && 'hidden',
             isSecret && 'mb-[-1px]',
+            validationStatus === 'failing' && [
+              'pr-xs',
+              themeClasses('bg-destructive-200', 'bg-[#5A4949]'), // FIXME(nick): replace hex color with palette color... this is 702C2C with 25% opacity on a 525252 and was made using alpha blending to avoid opacity usage
+            ],
           )
         "
       >
         <!-- Attribute name -->
-        <div class="flex flex-row items-center gap-2xs pl-xs">
-          <TruncateWithTooltip>{{ displayName }}</TruncateWithTooltip>
-          <span
-            v-if="showRequiredAsterisk"
-            v-tooltip="`${displayName} is a required value`"
-            :class="
-              clsx(
-                'shrink-0 text-lg font-bold',
-                themeClasses('text-destructive-600', 'text-destructive-300'),
-              )
-            "
-          >
-            *
-          </span>
+        <div
+          :class="
+            clsx(
+              'flex flex-row items-center gap-2xs pl-xs',
+              validationStatus === 'failing' && 'mt-xs',
+            )
+          "
+        >
+          <AttributeInputRequiredProperty
+            :text="displayName"
+            :showAsterisk="validationStatus === 'missingRequiredValue'"
+          />
           <div class="flex flex-row items-center ml-auto gap-2xs">
-            <StatusIndicatorIcon
-              v-if="
-                validation &&
-                validation.status !== 'Success' &&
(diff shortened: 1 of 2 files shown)


adamhjk:
Why would we still elipses the failure? Put it the line below and let that thing breathe!

app/web/src/newhotness/layout_components/AttributeInput.vue
```diff
@@ -9,37 +9,49 @@
             'grid grid-cols-2 items-center gap-2xs relative text-sm font-normal',
             inputOpen && 'hidden',
             isSecret && 'mb-[-1px]',
+            validationStatus === 'failing' && 'pr-xs', // additional right padding for failed validations
+            validationStatus === 'failing' &&
+              themeClasses(
+                'bg-destructive-200',
+                'bg-[#702C2C] bg-opacity-25', // FIXME(nick): use new palette and avoid opacity
```
wendybujalski: ah yeah, no opacity allowed, we should be able to achieve the same color without it

app/web/src/newhotness/layout_components/AttributeInput.vue
```diff
@@ -9,37 +9,49 @@
             'grid grid-cols-2 items-center gap-2xs relative text-sm font-normal',
             inputOpen && 'hidden',
             isSecret && 'mb-[-1px]',
+            validationStatus === 'failing' && 'pr-xs', // additional right padding for failed validations
```
wendybujalski: You could combine these two sections with the same conditional like so -

`validationStatus === 'failing' && [ 'pr-xs', themeClasses('bg-destructive-200', 'bg-[#702C2C] bg-opacity-25') ]`

wendybujalski (changes requested):
Looks good, but we should test this rebased on ENG-3128 before merging it.

app/web/src/newhotness/layout_components/AttributeInput.vue
nickgerace: Ahhh I was wondering how to do that. Thank you!

app/web/src/newhotness/layout_components/AttributeInput.vue
nickgerace: Yeah I kept trying to use destructive levels and none of them looked quite the same... maybe we can choose one together

wendybujalski: Yeah I mean it might be best to just pick a destructive level that's close and then when we do the frontend Tailwind revamp we're gonna change what all the abstract colors are anyway to match the new design language of the new UI. 🤔

app/web/src/newhotness/layout_components/AttributeInput.vue
wendybujalski: clsx can parse arrays! :)

app/web/src/newhotness/layout_components/AttributeInput.vue
nickgerace: Fair point. All of them clashed somewhat with the text and border, but I'll pick one and mark it with a TODO.

nickgerace:
Taking this to drafts. @adamhjk is right and there's some things to do for @wendybujalski's comments.

nickgerace:
My eyes deceived me and I found a mistake: there is an interior border that should not be there. Fixing...

stack72:
@nickgerace this looks great!!

PR #6725, opened 2025-07-19
Refresh actions can run in change sets

britmyerss:
When importing resources, it is quite painful to have to apply a change set in order to propagate resource_values to use in downstream components. Similarly, if you know a resource has changed, it is unnecessary to have to force a new change set to get the updated values. 

Users get two key benefits from these changes: 
1. A user can import a full suite of resources, and create connections as they go, without having to do the apply/open new change set dance.
2. A user can trigger a refresh from their change set, and see the results flow back to the change set (as opposed to switching to head, hitting the button, then going back to your change set)

## SDF Changes
A new route was added to SDF that handles refresh actions **only** with the following logic: 
1. If refresh was triggered on head, enqueue the action
2. If there's already a failed/on hold refresh action for that component, re-enqueue it 
3. If refresh was triggered in an open change set (NOT head), check if the component exists on head. If it does, we enqueue the refresh action on head (and upon completion the updates will be replayed back to the change set). If it does NOT exist on head, we hide the button entirely (NOTE: we are not actually preventing this from dispatching via SDF. If the same checks hold (component doesn't exist on head, has a resource), SDF will enqueue it, mark it as dispatched, and dispatch it directly from the change set! 

## Management Func Job Changes
In the management function job, if we're running an Import function, and we're NOT on head, AND the component we're running import for does not exist on head, we will now dispatch that refresh action job from the change set so that we can set the resource and propagate values as needed.

## Frontend Changes
Added a `Refresh` Button to the Resource Tab in ComponentDetails which calls the new SDF route: 
image
(yes, I know, all this for one button). 

As there was lots of duplication for handling the various places we show actions, I created a new composable `useComponentActions` which pulls all of this logic into one place, and is consumed where needed. 
 
In order to hide the refresh button if a component doesn't exist on head, I've made 2 new composables that allow us to make args/keys for queries for head from other change sets. We'll see if this is a good idea or not. 

## To Do Still
This should probably be hooked up to be able to multi-select components and enqueue them all for refresh. Did not do that yet. 

Also did not touch the map at all, we might want to consider something there?

## How was it tested?
Manual testing via the UI - NOTE: there seems to be a reactivity issue with our front end Context, punting this problem to look into separately
Adding dal integration tests to cover the various scenarios - NOTE: Some of these *might* be flaky, as it's a lot of "commit this change and wait exactly long enough to check that all the side effects have completed".  Something to keep an eye on. 

<div><br/>via <a href="https://giphy.com/anchormanmovie/">Anchorman Movie</a> on <a href="https://giphy.com/gifs/anchormanmovie-will-ferrell-anchorman-ron-burgundy-huHHuZ1JS9vyg">GIPHY</a></div>

### DIFF
--- a/app/web/src/newhotness/ActionWidget.vue
+++ b/app/web/src/newhotness/ActionWidget.vue
@@ -20,12 +20,7 @@
       {{ actionPrototypeView.displayName || actionPrototypeView.name }}
     </div>
 
-    <Icon
-      v-if="removeApi.bifrosting.value || addApi.bifrosting.value"
-      name="loader"
-      class="ml-auto"
-      size="sm"
-    />
+    <Icon v-if="actionBifrosting" name="loader" class="ml-auto" size="sm" />
   </div>
 </template>
 
@@ -33,14 +28,13 @@
 import * as _ from "lodash-es";
 import clsx from "clsx";
 import { Icon, themeClasses, Toggle } from "@si/vue-lib/design-system";
-import { useRoute } from "vue-router";
 import StatusIndicatorIcon from "@/components/StatusIndicatorIcon.vue";
 import { ActionId } from "@/api/sdf/dal/action";
 import {
   ActionPrototypeView,
   BifrostComponent,
 } from "@/workers/types/entity_kind_types";
-import { routes, useApi } from "./api_composables";
+import { useComponentActions } from "./logic_composables/component_actions";
 
 const props = defineProps<{
   component: BifrostComponent;
@@ -48,41 +42,12 @@ const props = defineProps<{
   actionId?: ActionId;
 }>();
 
-const route = useRoute();
-const removeApi = useApi();
-const addApi = useApi();
-const clickHandler = async () => {
-  if (props.actionId) {
-    const call = removeApi.endpoint(routes.ActionCancel, {
-      id: props.actionId,
-    });
-    removeApi.setWatchFn(() => props.actionId);
+const { toggleActionHandler } = useComponentActions(() => props.component);
 
-    // This route can mutate head, so we do not need to handle new change set semantics.
-    await call.put({});
-  } else {
-    const call = addApi.endpoint(routes.ActionAdd);
-    addApi.setWatchFn(() => props.actionId);
-    const { req, newChangeSetId } = await call.post<{
-      componentId: string;
-      prototypeId: string;
-    }>({
-      componentId: props.component.id,
-      prototypeId: props.actionPrototypeView.id,
-    });
-    if (newChangeSetId && addApi.ok(req)) {
-      addApi.navigateToNewChangeSet(
-        {
-          name: "new-hotness-component",
-          params: {
-            workspacePk: route.params.workspacePk,
-            changeSetId: newChangeSetId,
-            componentId: props.component.id,
-          },
-        },
-        newChangeSetId,
-      );
-    }
-  }
-};
+const { handleToggle, bifrosting: actionBifrosting } = toggleActionHandler(
+  props.actionPrototypeView,
+  () => props.actionId,
+);
+
+const clickHandler = handleToggle;
 </script>
--- a/app/web/src/newhotness/ActionsPanel.vue
+++ b/app/web/src/newhotness/ActionsPanel.vue
@@ -22,83 +22,23 @@
       v-for="actionPrototypeView in actionPrototypeViews"
       :key="actionPrototypeView.id"
       :actionPrototypeView="actionPrototypeView"
-      :actionId="actionByPrototype[actionPrototypeView.id]"
+      :actionId="actionByPrototype[actionPrototypeView.id]?.id"
       :component="props.component"
     />
   </div>
 </template>
 
 <script lang="ts" setup>
-import { useQuery } from "@tanstack/vue-query";
-import { computed, inject } from "vue";
-import { bifrost, useMakeArgs, useMakeKey } from "@/store/realtime/heimdall";
-import {
-  ActionPrototypeViewList,
-  BifrostActionViewList,
-  BifrostComponent,
-  EntityKind,
-} from "@/workers/types/entity_kind_types";
-import { ActionId, ActionPrototypeId } from "@/api/sdf/dal/action";
+import { BifrostComponent } from "@/workers/types/entity_kind_types";
 import ActionWidget from "./ActionWidget.vue";
 import EmptyState from "./EmptyState.vue";
-import { assertIsDefined, Context } from "./types";
+import { useComponentActions } from "./logic_composables/component_actions";
 
 const props = defineProps<{
   component: BifrostComponent;
 }>();
 
-// The code below is the same as in AssetActionsDetails in the mead hall
-
-// This is the core materialized view for this component. We need it to know what action prototypes
-// are available for the given component.
-const makeKey = useMakeKey();
-const makeArgs = useMakeArgs();
-
-const ctx = inject<Context>("CONTEXT");
-assertIsDefined(ctx);
-
-const queryKeyForActionPrototypeViews = makeKey(
-  EntityKind.ActionPrototypeViewList,
-  props.component.schemaVariant.id,
+const { actionPrototypeViews, actionByPrototype } = useComponentActions(
+  () => props.component,
 );
-const actionPrototypeViewsRaw = useQuery<ActionPrototypeViewList | null>({
-  enabled: ctx.queriesEnabled,
-  queryKey: queryKeyForActionPrototypeViews,
-  queryFn: async () =>
-    await bifrost<ActionPrototypeViewList>(
-      makeArgs(
-        EntityKind.ActionPrototypeViewList,
-        props.component.schemaVariant.id,
-      ),
-    ),
-});
-const actionPrototypeViews = computed(
-  () => actionPrototypeViewsRaw.data.value?.actionPrototypes ?? [],
-);
-
-// Use the materialized view for actions to know what actions exist for a given prototype and the
-// selected component.
-const queryKeyForActionViewList = makeKey(EntityKind.ActionViewList);
-const actionViewList = useQuery<BifrostActionViewList | null>({
-  queryKey: queryKeyForActionViewList,
-  queryFn: async () =>
-    await bifrost<BifrostActionViewList>(makeArgs(EntityKind.ActionViewList)),
-});
-const actionByPrototype = computed(() => {
-  if (!actionViewList.data.value?.actions) return {};
-  if (actionViewList.data.value.actions.length < 1) return {};
-
-  const result: Record<ActionPrototypeId, ActionId> = {};
-  for (const action of actionViewList.data.value.actions) {
-    if (action.componentId === props.component.id) {
-      // NOTE(nick): this assumes that there can be one action for a given prototype and component.
-      // As of the time of writing, this is true, but multiple actions per prototype and component
-      // aren't disallowed from the underlying graph's perspective. Theorhetically, you could
-      // enqueue two refreshes back-to-back. What then? I don't think we'll expose an interface to
-      // do that for awhile, so this should be sufficient.
-      result[action.prototypeId] = action.id;
-    }
-  }
-  return result;
-});
 </script>
--- a/app/web/src/newhotness/ComponentContextMenu.vue
+++ b/app/web/src/newhotness/ComponentContextMenu.vue
@@ -33,13 +33,10 @@ import { RouteLocationRaw, useRoute } from "vue-router";
 import { bifrost, useMakeArgs, useMakeKey } from "@/store/realtime/heimdall";
 import { ComponentId } from "@/api/sdf/dal/component";
 import {
-  BifrostActionViewList,
-  ActionPrototypeViewList,
   ComponentInList,
   EntityKind,
   SchemaVariant,
 } from "@/workers/types/entity_kind_types";
-import { ActionId, ActionPrototypeId } from "@/api/sdf/dal/action";
 import CreateTemplateModal from "@/newhotness/CreateTemplateModal.vue";
 import EraseModal from "./EraseModal.vue";
 import DeleteModal, { DeleteMode } from "./DeleteModal.vue";
@@ -47,6 +44,7 @@ import { useApi, routes } from "./api_composables";
 import { assertIsDefined, ExploreContext } from "./types";
 import { useComponentDeletion } from "./composables/useComponentDeletion";
 import { useComponentUpgrade } from "./composables/useComponentUpgrade";
+import { useComponentActions } from "./logic_composables/component_actions";
 
 const props = defineProps<{
   onGrid?: boolean;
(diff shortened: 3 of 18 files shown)


app/web/src/newhotness/ActionWidget.vue
```diff
@@ -20,69 +20,34 @@
...
-const route = useRoute();
-const removeApi = useApi();
-const addApi = useApi();
-const clickHandler = async () => {
-  if (props.actionId) {
-    const call = removeApi.endpoint(routes.ActionCancel, {
-      id: props.actionId,
-    });
-    removeApi.setWatchFn(() => props.actionId);
+const { toggleActionHandler } = useComponentActions(() => props.component);
```
britmyerss, line 45: using new composable!

app/web/src/newhotness/ActionsPanel.vue
```diff
@@ -22,83 +22,23 @@
...
-const makeKey = useMakeKey();
-const makeArgs = useMakeArgs();
-
-const ctx = inject<Context>("CONTEXT");
-assertIsDefined(ctx);
-
-const queryKeyForActionPrototypeViews = makeKey(
-  EntityKind.ActionPrototypeViewList,
-  props.component.schemaVariant.id,
+const { actionPrototypeViews, actionByPrototype } = useComponentActions(
```
britmyerss, line 41: using new composable!

app/web/src/newhotness/ComponentContextMenu.vue
```diff
@@ -88,44 +86,14 @@ const atLeastOneNormalComponent = computed(() =>
 const singleComponent = computed(() =>
   components.value.length === 1 ? components.value[0] : undefined,
 );
+
+// Use the composable for action functionality
+const { actionPrototypeViews, actionByPrototype, toggleActionHandler } =
```
britmyerss, line 91: using new composable!

app/web/src/newhotness/ComponentDetails.vue
```diff
@@ -318,6 +318,20 @@
         <CollapsingFlexItem expandable>
           <template #header><span class="text-sm">Resource</span></template>
           <template #headerIcons>
+            <!-- <div class="ml-auto my-sm"> -->
+            <VButton
```
britmyerss, line 372: _all this for a button?_

app/web/src/newhotness/api_composables/index.ts
```diff
@@ -66,6 +67,7 @@ const CAN_MUTATE_ON_HEAD: readonly routes[] = [
   routes.CreateChangeSet,
   routes.AbandonChangeSet,
   routes.EnqueueAttributeValue,
+  routes.RefreshAction,
```
britmyerss, line 70: Refresh Action can run on head!

app/web/src/newhotness/logic_composables/component_actions.ts
```diff
@@ -0,0 +1,265 @@
...
+        const call = removeApi.endpoint(routes.ActionCancel, {
+          id: action,
+        });
+        removeApi.setWatchFn(() => toValue(actionId));
+        await call.put({});
+      } else {
+        // Adding new action - check if it's a refresh action
+        const onHead = ctx.onHead.value;
+        const isRefresh = actionPrototypeView.kind === ActionKind.Refresh;
+        if (onHead && isRefresh) {
```
britmyerss, line 203: note: a little special logic here. Since the `AddAction` route needs `ForceNewChangeSet` (so users can't enqueue create/update/manual actions directly on head without a change set) we need to detect if the user is toggling a Refresh Action on head, and if so, use the new route to enqueue it (which does not use `ForceNewChangeSet`

app/web/src/newhotness/logic_composables/component_actions.ts
```diff
@@ -0,0 +1,265 @@
...
+  const ctx = inject<Context>("CONTEXT");
+  assertIsDefined(ctx);
+  const route = useRoute();
+
+  const component = computed(() => toValue(componentRef));
+  const schemaVariantId = computed(() => {
+    const comp = component.value;
+    if (!comp) return "";
+
+    // Handle both string (ComponentInList) and WeakReference (BifrostComponent) types
```
britmyerss, line 47: ContextMenu uses `ComponentInList` and ComponentDetails uses `Component` so we need to support both here

app/web/src/newhotness/logic_composables/component_actions.ts
```diff
@@ -0,0 +1,265 @@
...
+                },
+              },
+              newChangeSetId,
+            );
+          }
+        }
+      }
+    };
+
+    const bifrosting = computed(() => activeApi.value.bifrosting ?? false);
```
britmyerss, line 241: this is so the bifrosting status (the spinner when you toggle an action and are waiting for patches to come back) is reactive as without the composable

lib/dal-materialized-views/src/mgmt_prototype_view_list.rs
```diff
@@ -53,10 +45,19 @@ pub async fn assemble(
...
-            kind,
+            kind: convert_kind(kind),
         });
     }
 
     views.sort_by_key(|v| v.id);
     Ok(views)
 }
+
+fn convert_kind(kind: dalMgmtFuncKind) -> ManagementFuncKind {
```
britmyerss, line 56: moved the name matching check to the dal, but we've got no place to impl From for these structs

lib/sdf-server/src/service/v2/action.rs
```diff
@@ -144,23 +158,20 @@ pub struct AddActionRequest {
...
     Json(AddActionRequest {
         component_id,
         prototype_id,
     }): Json<AddActionRequest>,
-) -> ActionResult<()> {
-    let ctx = builder
-        .build(access_builder.build(change_set_id.into()))
-        .await?;
-    let prototype = ActionPrototype::get_by_id(&ctx, prototype_id).await?;
+) -> ActionResult<ForceChangeSetResponse<()>> {
```
britmyerss, line 167: this should always have forced a new change set (at least until we rip this out everywhere)

lib/dal/src/job/definition/management_func.rs
```diff
@@ -263,7 +280,32 @@ impl ManagementFuncJob {
...
 
         Self::success_state(ctx, execution_state_id).await?;
-
+        // if the management prototype is for a Import func and we're not on head, let's dispatch it!
+        let kind = ManagementPrototype::kind_by_id(ctx, self.prototype_id).await?;
+        if kind == ManagementFuncKind::Import
+            && ctx.get_workspace_default_change_set_id().await? != ctx.change_set_id()
+            && !Component::exists_on_head_by_ids(ctx, &[self.component_id])
+                .await?
+                .contains(&self.component_id)
```
jhelwig: Because you're only interested in a single component here, you could use `!Component::exists_on_head_by_ids(ctx, &[self.component_id]).await.is_empty()`.

app/web/src/newhotness/logic_composables/component_actions.ts
```diff
@@ -0,0 +1,265 @@
...
+        // NOTE(nick): this assumes that there can be one action for a given prototype and component.
+        // As of the time of writing, this is true, but multiple actions per prototype and component
+        // aren't disallowed from the underlying graph's perspective. Theorhetically, you could
+        // enqueue two refreshes back-to-back. What then? I don't think we'll expose an interface to
+        // do that for awhile, so this should be sufficient.
+        result[action.prototypeId] = action;
+      }
+    }
+    return result;
+  });
```
nickgerace, line 116: Ah I thought I recognized this.

app/web/src/newhotness/logic_composables/component_actions.ts
```diff
@@ -0,0 +1,265 @@
...
+
+  const actionIsRunning = (actionId?: ActionId | null) => {
+    if (!actionId) return false;
+    const state = actionViewList.data.value?.actions.find(
+      (action) => action.id === actionId,
+    )?.state;
+    if (!state) return false;
+    return [ActionState.Dispatched, ActionState.Running].includes(state);
```
nickgerace, line 124: I like this a lot

lib/dal/src/job/definition/management_func.rs
```diff
@@ -263,7 +280,32 @@ impl ManagementFuncJob {
...
+        {
+            let actions = Action::find_for_kind_and_component_id(
+                ctx,
+                self.component_id,
+                crate::action::prototype::ActionKind::Refresh,
+            )
+            .await?;
+            if actions.len() == 1 {
+                if let Some(action_id) = actions.first() {
```
jhelwig: If you bring in `itertools` (it's already a dependency of `dal`), you can use `.exactly_one` and condense this down into a single check.

```rust
use itertools::Itertools as _;

...

if let Some(action_id) = actions.exactly_one().ok() {
    ...
}
```

app/web/src/newhotness/ComponentDetails.vue
nickgerace: that tends to happen a lot in this UI :)

lib/dal-materialized-views/src/mgmt_prototype_view_list.rs
nickgerace: I feel the pain here. These conversions become annoying.

lib/dal-test/src/test_exclusive_schemas/legos/small.rs
```diff
@@ -840,22 +840,45 @@ pub(crate) async fn migrate_test_exclusive_schema_small_even_lego(
 
     // Build Create Action Func
     let create_action_code = "async function main() {
-                return { payload: { \"poop\": true }, status: \"ok\" };
+                return { payload: { \"poop\": \"created\", \"refresh_count\": 0 }, status: \"ok\" };
```
nickgerace, line 843: Something very weird about this new sentence

app/web/src/newhotness/logic_composables/component_actions.ts
```diff
@@ -0,0 +1,265 @@
...
+  const toggleActionHandler = (
+    actionPrototypeView: ActionPrototypeView,
+    actionId?: MaybeRefOrGetter<ActionId | undefined>,
+  ) => {
+    const removeApi = useApi();
+    const addApi = useApi();
+    const refreshApi = useApi();
+    // Track which API is currently active as we could call 3 different APIs and need to propagate
+    // the bifrosting status reactively for the correct one
+    const activeApi = ref(addApi);
```
nickgerace, line 185: This is clever and makes sense to me. Something feels clunky, but I can't quite put my finger on it. It's certainly not how you used it. In fact, this is probably the most ideal way to have dealt with multiple bifrosting values. I'll think on this for the future, but looks great to me.

lib/dal/src/management/prototype.rs
```diff
@@ -409,6 +420,27 @@ impl ManagementPrototype {
         self.description.as_deref()
     }
 
+    pub async fn kind_by_id(
+        ctx: &DalContext,
+        id: ManagementPrototypeId,
+    ) -> ManagementPrototypeResult<ManagementFuncKind> {
+        let func_id = Self::func_id(ctx, id).await?;
+        let func = Func::get_by_id(ctx, func_id).await?;
+        // TODO: Make Management Func Kinds a real thing
```
nickgerace, line 429: I think this will eventually bite us, especially when multi-cloud comes into play. Good call out

lib/dal/src/action.rs
```diff
@@ -759,6 +776,142 @@ impl Action {
...
+                .await?
+                .is_none()
+            {
+                Self::enqueue_refresh(ctx, component_id, false).await
+            }
+            // last case, component doesn't exist on head, but has a resource (which can only be true if we ran an import or other mgmt func)
+            // so we can enqueue and dispatch right away
+            else {
+                Self::enqueue_refresh(ctx, component_id, true).await
+            }
```
nickgerace, line 814: I get the desire to refresh HEAD from a change set, but I'd expect it to only refresh for my change set. I feel like this might be confusing. Thoughts?

lib/dal/src/action.rs
```diff
@@ -759,6 +776,142 @@ impl Action {
...
+                func_id,
+                func_display_name: func.display_name,
+                func_name: func.name.clone(),
+            },
+            func.name,
+        )
+        .await?;
+        Ok(action.id())
+    }
 }
```
nickgerace, line 902: Good idea to bundle audit log in too

lib/dal/src/action.rs
```diff
@@ -759,6 +776,142 @@ impl Action {
...
+            ActionKind::Refresh,
+        )
+        .await?;
+        if refresh_actions.len() > 1 {
+            return Err(ActionError::UnexpectedNumberOfActionKinds(
+                ActionKind::Refresh,
+                schema_variant_id,
+            ));
+        }
+        if let Some(&prototype_id) = refresh_actions.first() {
```
jhelwig: You could use `exactly_one` from `itertools` here, too, and map the error to be an `ActionError`.

lib/dal/tests/integration_test/action.rs
nickgerace, line 1: These are great!!!

nickgerace (approved):

lib/dal/src/action.rs
```diff
@@ -759,6 +776,142 @@ impl Action {
...
+            // we'll make this a 'retry' as this whole function exists to interpret what the user wanted
+            // which just means if the existing refresh action is Failed/On Hold, queue it up!
+            if !maybe_duplicate_action.is_empty() && !should_dispatch {
+                if maybe_duplicate_action.len() > 1 {
+                    return Err(ActionError::UnexpectedNumberOfActionsEnqueuedForComponent(
+                        ActionKind::Refresh,
+                        component_id,
+                    ));
+                }
+                if let Some(action_id) = maybe_duplicate_action.first() {
```
jhelwig: [`Itertools::at_most_one`](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.at_most_one) could probably tidy this up a bit.

lib/dal/src/action.rs
```diff
@@ -759,6 +776,142 @@ impl Action {
...
+                    let action = Action::get_by_id(ctx, *action_id).await?;
+                    match action.state() {
+                        ActionState::Failed | ActionState::OnHold => {
+                            Action::set_state(ctx, *action_id, ActionState::Queued).await?
+                        }
+                        ActionState::Dispatched | ActionState::Queued | ActionState::Running => {
```
jhelwig, line 862: Do we want `ActionState::Queued` to be an error here? Given the behavior with `Failed` and `OnHold`, I'd expect the behavior to be that it goes "you want it queued, and it already is, so everything's as you want it" (it'd be idempotent).

lib/dal/src/action.rs
```diff
@@ -759,6 +776,142 @@ impl Action {
...
+                            return Err(ActionError::ActionAlreadyEnqueued(prototype_id));
+                        }
+                    }
+                }
+            }
+            // if we're dispatching and we've got a refresh action enqueued, check where that action was enqueued from
+            // if it's enqueued in this same change set, dispatch it.
```
jhelwig: `If the originating change set is this change set, dispatch it` might be clearer, since it's definitely enqueued in this change set as it's the one were looking at to get the list of queued actions.

lib/dal/tests/integration_test/management.rs
```diff
@@ -125,6 +128,107 @@ async fn exec_mgmt_func_and_operate(
...
+    .await?;
+    ChangeSetTestHelpers::commit_and_update_snapshot_to_visibility(ctx).await?;
+
+    let actions = Action::all_ids(ctx).await?;
+    // there should be 1 import action now
+    assert!(actions.len() == 1);
+    for action_id in actions {
+        let action = Action::get_by_id(ctx, action_id).await?;
+        dbg!(action);
+    }
```
jhelwig: Intentionally left in?

lib/dal/tests/integration_test/management.rs
```diff
@@ -125,6 +128,106 @@ async fn exec_mgmt_func_and_operate(
...
+    )
+    .await?;
+    ChangeSetTestHelpers::commit_and_update_snapshot_to_visibility(ctx).await?;
+
+    let actions = Action::all_ids(ctx).await?;
+    // there should be 1 import action now
+    assert!(actions.len() == 1);
+    for action_id in actions {
+        let action = Action::get_by_id(ctx, action_id).await?;
+    }
```
jhelwig: Doesn't appear to be doing anything now that the `dbg!()` is gone?

jhelwig (dismissed):

lib/dal/src/action.rs
britmyerss: the guiding principal here is we don't want a change set to have a _newer_ representation than head. If the resource is different, all open change sets should know that.

stack72 (approved):

PR #6728, opened 2025-07-21
feat(sdf): use a JoinSet to parallelize multi_mjolnir requests

zacharyhamm:
Bulk mjolnir requests are kind of slow. For the "production saas" workspace in tools, the HEAD fetch can take 15 seconds. Since they are just fetching from a KV store, the boundary here should be how fast we can get the data out of NATS. To try and speed it up, threw all the requests in a join set. Locally, this cuts a 1.5second requests for ~2000 objects down to 500ms. Using a concurrency limit of 200 simultaneous requests.

### DIFF
--- a/lib/sdf-server/src/service/v2/index/get_front_end_object.rs
+++ b/lib/sdf-server/src/service/v2/index/get_front_end_object.rs
@@ -1,3 +1,5 @@
+use std::sync::Arc;
+
 use axum::{
     Json,
     extract::{
@@ -11,6 +13,7 @@ use dal::{
 };
 use sdf_core::index::{
     FrontEndObjectMeta,
+    IndexError,
     front_end_object_meta,
 };
 use serde::{
@@ -19,6 +22,10 @@ use serde::{
 };
 use si_frontend_types::FrontEndObjectRequest;
 use telemetry::prelude::*;
+use tokio::{
+    sync::Semaphore,
+    task::JoinSet,
+};
 
 use super::{
     AccessBuilder,
@@ -58,6 +65,8 @@ pub struct MultipleFrontEndObjectResponse {
     failed: Vec<FrontEndObjectRequest>,
 }
 
+const BULK_CONCURRENCY_LIMIT: usize = 200;
+
 pub async fn get_multiple_front_end_objects(
     HandlerContext(builder): HandlerContext,
     AccessBuilder(access_builder): AccessBuilder,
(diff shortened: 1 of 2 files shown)


lib/sdf-server/src/service/v2/index/get_front_end_object.rs
```diff
@@ -71,10 +74,25 @@ pub async fn get_multiple_front_end_objects(
 
     let mut successful = Vec::new();
     let mut failed = Vec::new();
+
+    let mut join_set = JoinSet::new();
+    let frigg_arc = Arc::new(frigg);
```
zacharyhamm: I was unsure if the frigg client should be cloned directly

jhelwig: Cloning directly is just fine. That's what we do when building the MVs in parallel: https://github.com/systeminit/si/blob/65807882fc7e4c20ae8f653ce9aba77bbf77764d/lib/edda-server/src/change_set_processor_task/materialized_view.rs#L521

jhelwig:
Should we limit the number processed in parallel here (I suspect yes, but I don't know what the limit should be)? Similar to what we do when building the MVs themselves. https://github.com/systeminit/si/blob/65807882fc7e4c20ae8f653ce9aba77bbf77764d/lib/edda-server/src/change_set_processor_task/materialized_view.rs#L590

zacharyhamm:
> Should we limit the number processed in parallel here (I suspect yes, but I don't know what the limit should be)? Similar to what we do when building the MVs themselves.
> 
> https://github.com/systeminit/si/blob/65807882fc7e4c20ae8f653ce9aba77bbf77764d/lib/edda-server/src/change_set_processor_task/materialized_view.rs#L590

That's probably a good idea, but I'm not sure what the limit should be either

jhelwig:
> > Should we limit the number processed in parallel here (I suspect yes, but I don't know what the limit should be)? Similar to what we do when building the MVs themselves.
> > https://github.com/systeminit/si/blob/65807882fc7e4c20ae8f653ce9aba77bbf77764d/lib/edda-server/src/change_set_processor_task/materialized_view.rs#L590
> 
> That's probably a good idea, but I'm not sure what the limit should be either

Maybe start with 100, and have it configurable? Even if we decide that's lower than we'd like, it should still significantly speed up this endpoint for large requests.

zacharyhamm:
@fnichol @jhelwig updated this with a sempahore for concurrency control

app/web/src/workers/mjolnir_queue.ts
```diff
@@ -4,7 +4,7 @@ import { ChangeSetId } from "@/api/sdf/dal/change_set";
 import { BustCacheFn, Id } from "./types/dbinterface";
 import { EntityKind } from "./types/entity_kind_types";
 
-const _DEBUG = import.meta.env.VITE_SI_ENV === "local";
+const _DEBUG = false;
```
jhelwig: Intentional change?

zacharyhamm: nope! should be good now though

jhelwig (approved):

PR #6730, opened 2025-07-21
feat: Retain the Explore page's filter and grouping options everywhere

vbustamante:

We had no easy way to retrieve the filter, grouping and sorting  values when coming back to the explore page from inner pages. This PR caches those values so we can use a redirect from anywhere, using the query key `retainSessionState` to re use that cached value. This way any page can go to the last seen explore page without needing to know any details of the implementation.

The cache is only retained per tab, and goes away when it's closed so we don't have any clashes with other sessions 

## How was it tested?

- [X] Manual test: Navigate back from component details
- [X] Manual test: Navigate back from func details
- [X] Manual regression test: Follow link, see it does NOT use the cached values

### DIFF
--- a/app/web/src/newhotness/Explore.vue
+++ b/app/web/src/newhotness/Explore.vue
@@ -448,7 +448,7 @@ import ExploreSearchBarSkeleton from "@/newhotness/skeletons/ExploreSearchBarSke
 import ExploreGridSkeleton from "@/newhotness/skeletons/ExploreGridSkeleton.vue";
 import ExploreRightColumnSkeleton from "@/newhotness/skeletons/ExploreRightColumnSkeleton.vue";
 import Map from "./Map.vue";
-import { collapsingGridStyles, preserveExploreState } from "./util";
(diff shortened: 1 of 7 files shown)


app/web/src/newhotness/Explore.vue
```diff
@@ -487,6 +487,34 @@ const args = useMakeArgs();
 const VIEW_MODE_LOCAL_STORAGE_KEY = "newhotness-view-mode";
 const viewModeStorageKey = () =>
   `${VIEW_MODE_LOCAL_STORAGE_KEY}: ${ctx.changeSetId}`;
+const storeViewMode = () => {
+  if (!groupRef.value) return;
+
+  const key = viewModeStorageKey();
+
+  if (groupRef.value.isB) {
+    localStorage.setItem(key, "grid");
```
jobelenus, line 496: Do we still want to be using local storage across tabs? Given this very tab centric approach?

jobelenus (approved):

PR #6734, opened 2025-07-22
Overhaul management function UX and special case funcs

nickgerace:
## Introduction

This change is a general overhaul to the UX and UI for management funcs with special case behavior for "import" and "run template" funcs. This design went through a bunch of iterations, but this one is a good start towards handling the balance of knowing the state of a management func (and not caring about whether the function or operations failed) and
being able to debug it with some granularity.

[image](https://myheroacademia.fandom.com/wiki/Kai_Chisaki)

## Description

Essentially, this change ensures that the management func tab and general status tracking for management funcs includes both the func run and the operator. Originally, this change used the status "bucket" from the StatusPanel, but given the requirement to view live WsEvents as well as the need for both FuncRunDetails and ManagementFuncCard to know the status of the operator upon initial load, using the job state database table resulted in a cleaner approach.

The status for management funcs is now centralized and is available for both the management func card and the component details page. It uses "funcRunStatus" because that function populates the func run status icons on the Explore page and the individual func run pages. The management func card now disallows "run template" funcs are they are special-cased in addition to "import" funcs.

From these changes, we also now allow error hinting via banners. There is one user of this subsystem now: the management funcs. The hint and raw error message are provided by checking the func run state and the job state as well as using the new "message" column for the job state table. It indicates the reason behind the current state. This is only used for failing states at the moment and is type guarded.

## Additional Changes:

- Add ErrorBanner for use in both ComponentDetails (grid compliant with a new "fit-content" usage... thank you @jobelenus) and on the func run page
- Add a new "si_id" library within "newhotness" to account for IDs without a home that are present in "lib/si-id"
- Add destructive color for newhotness designs within the Tailwind colors to prevent hex color sprawl (this affects the existing usages of the color in validations land)

## Screenshot: Component Details with failing Run Template

Screenshot 2025-07-24 at 4 15 52 PM

## Screenshot: Func Run Details with failing Run Template

Screenshot 2025-07-24 at 4 16 00 PM


app/web/src/newhotness/api_composables/management_func_job_state.ts
```diff
@@ -0,0 +1,28 @@
...
+import { Timestamp } from "@/api/sdf/dal/timestamp";
+import { FuncRunId } from "./func_run";
+import { UserPk, WorkspacePk } from "./si_id";
+
+export type ManagementFuncJobStateId = string;
+
+export interface ManagementFuncJobState {
+  id: ManagementFuncJobStateId;
+  workspaceId: WorkspacePk;
+  change_set_id: ChangeSetId;
```
nickgerace: Need to fix casing

app/web/src/newhotness/logic_composables/management.ts
```diff
@@ -0,0 +1,44 @@
+import { computed, ComputedRef, inject, ref } from "vue";
+import { useQuery } from "@tanstack/vue-query";
+import { FuncKind, FuncRun } from "../api_composables/func_run";
+import { ManagementFuncJobState } from "../api_composables/management_func_job_state";
+import { useApi, routes } from "../api_composables";
+import { assertIsDefined, Context } from "../types";
+
+export const useManagementFuncJobState = (
```
nickgerace, line 8: Need to double check that this can be re-used by multiple cards and can have shared state when dealing with the same func run

app/web/src/newhotness/logic_composables/management.ts
```diff
@@ -0,0 +1,44 @@
...
+      const call = api.endpoint<ManagementFuncJobState>(
+        routes.MgmtFuncGetJobState,
+        {
+          funcRunId: funcRun.value.id,
+        },
+      );
+      const req = await call.get();
+      if (api.ok(req)) {
+        pollInterval.value = ["executing", "operating", "pending"].includes(req.data.state)
+          ? 5000
```
nickgerace, line 37: Check this timer

PR #6736, opened 2025-07-22
fix: Don't run more than once search at a time

jkeiser:
With really fast typers, complex queries and large workspaces, searches can take a long time if you type really really fast. This is because we currently re-run the search query on every character you type; so they can get piled up one on top of the other.

This PR fixes the issue by limiting searches to one at a time (no matter how long they take):
* When searchString, componentList or any AttributeTree MV is generated, component search will re-run.
* If there is already a component search running (long running query), the search will be queued up and will re-run later.

To support this:
* Moved search implementation from Explore.vue to search.ts composable `useComponentSearch()`
* The ComponentSearch frontend-generated MV is now "QueryAttributes". This MV does nothing besides provide a reactive `queryAttributes` function for you, which will re-run anytime an AttributeTree MV is updated.
* `computedAsyncDebounce()` is similar to `computed()` except it lets you specify an async function, and handles the "debounce if it's already running" logic for you.

## How was it tested?

- [X] Integration tests pass
- [X] Manual test: searches come back fast and don't pile up when you type fast in a larger workspace
- [X] Manual test: searches update when component list updates
- [X] Manual test: searches update when users update attribute trees


app/web/src/newhotness/Explore.vue
```diff
@@ -296,7 +296,7 @@
         v-else
         ref="mapRef"
         :active="!showGrid"
-        :components="filteredComponents"
+        :components="filteredComponents ?? []"
```
jobelenus: can we put these coerce `?? []` into the return of the composable so everything that uses the composable doesn't need to do this operation?

app/web/src/store/realtime/heimdall.ts
```diff
@@ -43,6 +43,7 @@ import {
   cachedAppEmitter,
   SHOW_CACHED_APP_NOTIFICATION_EVENT,
 } from "./cached_app_emitter";
+import { WorkspacePk } from "../workspaces.store";
```
jobelenus: lets move this type out of the store, because we dont want to be importing stores.

jobelenus (changes requested):
Small nit, but looks great!

app/web/src/newhotness/Explore.vue
jkeiser: Yep, if we change componentList itself so it's always defined, filteredComponents will be as well! I'll look at it.

jkeiser: (It already has it defined so the input list harmonizes with the output list)

jkeiser: ... and it looks like I had removed that harmony, which was the source of the issue. Fixing.

jobelenus (approved):

PR #6738, opened 2025-07-22
Ensure new dynamic AttributeValues are in the DVU roots when migrating SchemaVariants

jhelwig:
While migrating a `Component` from one `SchemaVariant` to another, if the old version of the `Component` had an `AttributeValue` that the new version did not have an equivalent for (but was still appropriate for the new `Prop` tree structure), and the old `AttributeValue` used a dynamic function through its (or its `Prop`'s) `AttributePrototype`, we would continue early in the match arm after creating the new `AttributeValue` and preserving the `AttributePrototype` information. This meant that the `AttributeValue` never ended up in the list of DVU roots that was being built as that happened in the skipped over portion of the match arm.

We now ensure that the newly created `AttributeValue` with the dynamic function makes it into the DVU roots to be re-evaluated as part of the migrated `Component`.

### DIFF
--- a/lib/dal/src/component.rs
+++ b/lib/dal/src/component.rs
@@ -1100,6 +1100,10 @@ impl Component {
                         )
                         .await?;
 
+                        // We need to make sure this AV makes its way into the DVU roots, even though
+                        // we would normally add it at the end of the match arm, because we're
+                        // skipping the rest of the match arm by continuing here.
+                        dvu_roots.push(DependentValueRoot::Unfinished(inserted_value.id().into()));
                         continue;
                     }
 
--- a/lib/dal/tests/integration_test/schema/variant/authoring/regenerate.rs
+++ b/lib/dal/tests/integration_test/schema/variant/authoring/regenerate.rs
@@ -1,4 +1,5 @@
 use dal::{
+    AttributeValue,
     Component,
     ComponentType,
     DalContext,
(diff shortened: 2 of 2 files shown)


jhelwig:
Without the change to `lib/dal/src/component.rs`:
```
running 1 test
test integration_test::schema::variant::authoring::regenerate::dynamic_functions_on_new_attribute_values_are_rerun ... FAILED

failures:

---- integration_test::schema::variant::authoring::regenerate::dynamic_functions_on_new_attribute_values_are_rerun stdout ----
Message:  assertion `left == right` failed
  left: Some(Array [String("Source Component")])
 right: Some(Array [])
Location: lib/dal/tests/integration_test/schema/variant/authoring/regenerate.rs:755

failures:
    integration_test::schema::variant::authoring::regenerate::dynamic_functions_on_new_attribute_values_are_rerun

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 321 filtered out; finished in 17.00s
```

With the change to `lib/dal/src/component.rs`:
```
running 1 test
 (took 17 seconds) test integration_test::schema::variant::authoring::regenerate::dynamic_functions_on_new_attribute_values_are_rerun ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 321 filtered out; finished in 17.29s
```

zacharyhamm (approved):
Test looks legit. I read through the function being modified and this looks good to me

zacharyhamm (dismissed):

PR #6745, opened 2025-07-22
feat: updated template generation

adamhjk:
This PR implements updated template generation. It introduces the 'si-generate-template' crate, which uses the askama templating library to generate typescript function code. It uses the 'attributes' syntax  for setting all properties, uses variable declarations in the management function for easier maintenance, and sets sources correctly.

If you use AWS components in the template, it will automatically wire in the credential and the region from the template. 

It uses a 'Name Prefix' field to automatically prefix the name of the generated components. If there are components with that prefix already attached via a management edge, the function will fail (preventing multiple runs with the same prefix.) No attempt is made to make the template idempotent.

It updates the generated templates to use variables and the attribute syntax, plus some sugar for AWS components.

## How was it tested?

* Add a VPC and two subnets
* Add some prop to prop connections between them
* Generate a new template from the old UI
* Add a name prefix and run the template function


lib/sdf-server/src/service/v2/management/generate_template.rs
```diff
@@ -71,42 +80,107 @@ pub async fn generate_template(
     PosthogClient(posthog_client): PosthogClient,
     OriginalUri(original_uri): OriginalUri,
     Host(host_name): Host,
-    Path((_workspace_pk, change_set_id, view_id)): Path<(WorkspacePk, ChangeSetId, ViewId)>,
```
vbustamante, line 74: Since we're editing this endpoint, it's a good chance  to upgrade the extractors to use `ChangeSetDalContext`, like thus : https://github.com/systeminit/si/blob/63fdd1d46f2de7008a0ce24af190c2543ce7a3a7/lib/sdf-server/src/service/v2/view/paste_component.rs#L53-L59

This way you can get rid of most of these arguments (and also ignore the changeset in path, since that gets deal with by the extractor)

lib/sdf-server/src/service/v2/management/generate_template.rs
```diff
@@ -71,42 +80,107 @@ pub async fn generate_template(
...
+                }
+            }
+        }
+    }
+
+    let asset_schema_template = AssetSchema::new(has_aws);
+    let schema_variant_code = asset_schema_template.render()?;
+    let mut mgmt_func_template = RunTemplateMgmtFunc::new(template_components);
+    mgmt_func_template.sort_components_by_dependencies();
+    let mgmt_func_code = mgmt_func_template.render()?;
```
britmyerss, line 173: 💥 love this

britmyerss (approved):

PR #6747, opened 2025-07-22
fix(web): TanStack Query keys need to be reactive

britmyerss:
Fixes https://linear.app/system-initiative/issue/BUG-902/not-updating-attribute-tree-when-changing-between-components-in

I also did a pass at which query keys aren't reactive and updated ManagementPanel.vue as well.

## How was it tested?

- [X] Manual test: switching components doesn't show the wrong data!

<div><br/>via <a href="https://giphy.com/gifs/thank-you-thanks-thumbs-up-BYoRqTmcgzHcL9TCy1">GIPHY</a></div>

### DIFF
--- a/app/web/src/newhotness/ManagementPanel.vue
+++ b/app/web/src/newhotness/ManagementPanel.vue
@@ -73,10 +73,11 @@ const mgmtFuncs = computed(
 const key = useMakeKey();
 const args = useMakeArgs();
 
-const componentId = computed(() => props.component?.id);
+const componentId = computed(() => props.component?.id ?? "");
 
 const mgmtConnectionsQuery = useQuery<ManagementConnections | null>({
-  queryKey: key(EntityKind.ManagementConnections, componentId.value),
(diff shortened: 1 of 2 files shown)


app/web/src/newhotness/ManagementPanel.vue
```diff
@@ -73,10 +73,11 @@ const mgmtFuncs = computed(
 const key = useMakeKey();
 const args = useMakeArgs();
 
-const componentId = computed(() => props.component?.id);
+const componentId = computed(() => props.component?.id ?? "");
 
 const mgmtConnectionsQuery = useQuery<ManagementConnections | null>({
-  queryKey: key(EntityKind.ManagementConnections, componentId.value),
+  enabled: () => componentId.value !== "",
```
jobelenus, line 79: do you want `computed(() => componentId.value !== "")` ???

britmyerss: because it's wrapped in () => this works!

PR #6749, opened 2025-07-22
Enable admins to trigger migrations and snapshot validations

britmyerss:
This creates 2 new routes nested under /admin for the dry run and the actual migration.  I'm using the same internals as the existing routes to limit the blast for now. During cleanup we can decide if the old routes need to stay around or not. 

Also updated the admin store to use these new routes

## How was it tested?

- [X] Manual test: I can click the button still, and it does the dry run/migration!

<div><br/>via <a href="https://giphy.com/challenger/">Challenger</a> on <a href="https://giphy.com/gifs/challenger-challenger23-government-fbi-S99uqwwvMY6Yw">GIPHY</a></div>

### DIFF
--- a/lib/sdf-server/src/service/v2/admin.rs
+++ b/lib/sdf-server/src/service/v2/admin.rs
@@ -33,6 +33,7 @@ use dal::{
     Workspace,
     WorkspacePk,
     WorkspaceSnapshotAddress,
+    WsEventError,
     cached_module::CachedModuleError,
     func::runner::FuncRunnerError,
     slow_rt::SlowRuntimeError,
@@ -67,11 +68,13 @@ mod get_snapshot;
 mod innit;
 mod kill_execution;
 mod list_change_sets;
+mod migrate_connections;
 mod search_workspaces;
 mod set_concurrency_limit;
 mod set_snapshot;
 mod update_module_cache;
 mod upload_cas_data;
+mod validate_snapshot;
 
 // 1GB
 const MAX_UPLOAD_BYTES: usize = 1024 * 1024 * 1024;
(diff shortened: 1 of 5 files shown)


lib/sdf-server/src/service/v2/admin.rs
```diff
@@ -221,6 +222,14 @@ pub fn v2_routes(state: AppState) -> Router<AppState> {
             "/workspaces/:workspace_id/change_sets/:change_set_id/upload_cas_data",
             post(upload_cas_data::upload_cas_data),
         )
+        .route(
```
jkeiser, line 243: Curious why we didn't remove the old routes? The admin interface is the only thing that uses them, I think we'd probably be safe.

lib/sdf-server/src/service/v2/admin/migrate_connections.rs
```diff
@@ -0,0 +1,48 @@
...
+            AdminAPIResult,
+            AdminUserContext,
+        },
+        change_set,
+    },
+    track_no_ctx,
+};
+
+pub async fn migrate_connections(
+    AdminUserContext(mut ctx): AdminUserContext,
```
jkeiser, line 39: Sad we don't have an AdminChangeSetContext, but we don't have many of these endpoints yet anyway.

jkeiser (changes requested):
I think we're better off removing / moving the old routes rather than duplicating--there's nothing depending on them so we may as well not make work for our future selves :)

lib/sdf-server/src/service/v2/admin.rs
britmyerss: ok!

lib/sdf-server/src/service/v2/admin.rs
```diff
@@ -109,6 +126,8 @@ pub enum AdminAPIError {
     WorkspaceSnapshotAddressNotFound(ChangeSetId),
     #[error("workspace snapshot {0} for change set {1} could not be found in durable storage")]
     WorkspaceSnapshotNotFound(WorkspaceSnapshotAddress, ChangeSetId),
+    #[error("ws event error: {0}")]
+    WsEvent(#[from] WsEventError),
```
jkeiser, line 130: nit: where we can, we should qualify the error namespaces inline here--no need to specify them multiple times, and it makes them more copy/pasteable.

jkeiser (approved):

stack72 (approved):

PR #6757, opened 2025-07-23
Bulk Editing

jobelenus:

- Select multiple components
- select `B` for bulk edit
- see the common AV paths for the components that you can edit
- don't show values in the fields

**Note:** don't do this on HEAD right now 😄 (see TODO)

Under the hood, we build an AVTree based on _one_ of the selected components. We actually use _those_ AVs, IDs, values, etc, and then remove the values & external sources. And once you save, we run the `PUT /attributes` for each component ID at that path and the selected value/sub.

There are changes to `<ComponentAttribute>`, its a *proper* `layout_component` now, it no longer makes API calls, it emits up to its parent just like `save`, et al. That way the `<AttributePanel>` and `<AttributePanelBulk>` can re-use the form & logic components as they are, while making one vs many API calls.

#### Screenshots:
image
image

#### Out of Scope / TODO in the next turns

- right now it's only `/domain`, need to get `/secrets` in there too
- users need to see the values of the different components
- when hitting bulk, we need to make a new change set (solves a few technical problems)
    - in order to support this, we need to add the selected component IDs into the URL and a "bulk=1"
- need to handle prop-prop subscriptions better & with more clarity 
- remove `Save` button, and save on blur


app/web/src/newhotness/AttributePanel.vue
```diff
@@ -334,37 +303,84 @@ const save = async (
...
-      // effectively an unset operation.
-      coercedVal = null;
-    } else if (propKind === PropKind.Boolean) {
-      coercedVal = value.toLowerCase() === "true" || value === "1";
-    } else if (propKind === PropKind.Integer) {
-      coercedVal = Math.trunc(Number(value));
-    } else if (propKind === PropKind.Float) {
-      coercedVal = Number(value);
-    }
+  const payload = makeSavePayload(path, value, propKind, connectingComponentId);
```
nickgerace, line 306: Thank you for making this a function!

app/web/src/newhotness/Explore.vue
```diff
@@ -547,6 +553,8 @@ const showGrid = computed(() => urlGridOrMap.value === "grid");
 const gridMapSwitcherValue = computed(
   () => groupRef.value && groupRef.value.isA,
 );
+// TODO — if youre on HEAD and you start bulk editing, create a change set right away
+const bulkEditing = ref(false);
```
nickgerace, line 557: Copy that

app/web/src/newhotness/explore_grid/AttributePanelBulk.vue
```diff
@@ -0,0 +1,462 @@
...
+        size="sm"
+        @click="() => emit('close')"
+      />
+      <VButton
+        label="Save Bulk Edit"
+        tone="action"
+        size="sm"
+        :disabled="Object.values(valsToSave).length === 0"
+        :loading="saving > 0"
+        :loadingText="`Saving ${saving} Components`"
```
nickgerace, line 16: nit: should we have `Saving 1 Component` non-plural? You can ignore this nit. It's silly.

app/web/src/newhotness/explore_grid/AttributePanelBulk.vue
```diff
@@ -0,0 +1,462 @@
...
+        @click="() => emit('close')"
+      />
+      <h1 class="text-md font-bold my-xs">Bulk Editing</h1>
+    </div>
+    <ul class="flex flex-row gap-xs">
+      <li v-for="component in selectedComponents" :key="component.id">
+        <ComponentCard :component="component" />
+      </li>
+    </ul>
+    <DelayedLoader v-if="treesPending" :size="'full'" />
```
nickgerace, line 45: Good idea

app/web/src/newhotness/layout_components/ComponentAttribute.vue
```diff
@@ -245,11 +246,11 @@ import {
 import AttributeChildLayout from "./AttributeChildLayout.vue";
 import AttributeInput from "./AttributeInput.vue";
 import { AttrTree } from "../logic_composables/attribute_tree";
-import { useApi, routes, componentTypes } from "../api_composables";
+import { useApi, UseApi } from "../api_composables";
 import { useWatchedForm } from "../logic_composables/watched_form";
 
 const props = defineProps<{
-  component: BifrostComponent;
+  component: BifrostComponent | ComponentInList;
```
nickgerace, line 253: Ah I see the problem: when bulk editing, you don't have the full object

lib/dal-materialized-views/src/schema_variant/prop_tree.rs
```diff
@@ -76,6 +76,21 @@ pub async fn assemble_prop(
...
+            dal::PropKind::Array => PropKind::Array,
+            dal::PropKind::Boolean => PropKind::Boolean,
+            dal::PropKind::Integer => PropKind::Integer,
+            dal::PropKind::Json => PropKind::Json,
+            dal::PropKind::Map => PropKind::Map,
+            dal::PropKind::Object => PropKind::Object,
+            dal::PropKind::String => PropKind::String,
+            dal::PropKind::Float => PropKind::Float,
+        })
+    }
```
nickgerace, line 92: nit if you want to avoid mutability (feel free to ignore)
```suggestion
    let children = Prop::direct_child_props_ordered(ctx, prop_id).await?;
    let child_kind = if children.is_empty() {
        None
    } else {
        Some(match children[0].kind {
            dal::PropKind::Array => PropKind::Array,
            dal::PropKind::Boolean => PropKind::Boolean,
            dal::PropKind::Integer => PropKind::Integer,
            dal::PropKind::Json => PropKind::Json,
            dal::PropKind::Map => PropKind::Map,
            dal::PropKind::Object => PropKind::Object,
            dal::PropKind::String => PropKind::String,
            dal::PropKind::Float => PropKind::Float,
        })
    }l
```

app/web/src/newhotness/explore_grid/AttributePanelBulk.vue
jobelenus: can't open bulk edit with just 1 ;)

PR #6767, opened 2025-07-24
fix(dal): When duplicating components, only retain internal management connections, recreate internal subscriptions, and retain external subscriptions

britmyerss:
- Modified handling of management edges to only create if they're between duplicated components
- Recreate subscriptions between duplicated components

## How was it tested?

- [X] Wrote new integration tests to cover off on the permutations of behavior

<div><br/>via <a href="https://giphy.com/alcrego/">A. L. Crego</a> on <a href="https://giphy.com/gifs/alcrego-loop-eternal-TlK63Eu0ViBklB4v5RK">GIPHY</a></div>

### DIFF
--- a/lib/dal/src/component.rs
+++ b/lib/dal/src/component.rs
@@ -98,6 +98,7 @@ use crate::{
     },
     attribute::{
         attributes::{
+            self,
             AttributeValueIdent,
             Source,
         },
@@ -214,6 +215,8 @@ pub enum ComponentError {
     AttributePrototype(#[from] Box<AttributePrototypeError>),
     #[error("attribute prototype argument error: {0}")]
     AttributePrototypeArgument(#[from] Box<AttributePrototypeArgumentError>),
+    #[error("attributes error: {0}")]
+    Attributes(#[from] Box<attributes::Error>),
     #[error("attribute value error: {0}")]
     AttributeValue(#[from] Box<AttributeValueError>),
     #[error(
@@ -501,6 +504,11 @@ impl From<WorkspaceError> for ComponentError {
 
 impl From<WsEventError> for ComponentError {
     fn from(value: WsEventError) -> Self {
+        Box::new(value).into()
+    }
+}
+impl From<attributes::Error> for ComponentError {
+    fn from(value: attributes::Error) -> Self {
         Box::new(value).into()
     }
 }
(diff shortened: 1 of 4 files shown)


lib/dal/src/component.rs
```diff
@@ -3581,10 +3594,45 @@ impl Component {
...
+        }
+        let maybe_pasted = |id: ComponentId| to_pasted_id.get(&id).copied().unwrap_or(id);
+        // Copy correct connections (prop to prop and management)
+        for (&component_id, &pasted_component_id) in &to_pasted_id {
             // Copy manager connections
             for manager_id in Component::managers_by_id(ctx, component_id).await? {
-                Component::manage_component(ctx, manager_id, pasted_component.id).await?;
+                // If we were managed by a component that was also pasted, we should be managed by
+                // the pasted version--otherwise we're unmanaged!
+                if pasted_component_ids.contains(&maybe_pasted(manager_id)) {
```
jkeiser: You could use `if let Some(&pasted_manager_id) = pasted_component_ids.get(&manager_id) {` here, and then use it on the next line--that would make the intent clear with a few less steps in the way :)

jkeiser: And then you probably won't need `maybe_pasted` at all, either :)

lib/dal/src/component.rs
```diff
@@ -3581,10 +3594,45 @@ impl Component {
...
+                    AttributePrototypeArgument::prototype_id(ctx, subscriber_apa_id).await?;
+                let Some(subscriber_av_id) =
+                    AttributePrototype::attribute_value_id(ctx, subscriber_ap_id).await?
+                else {
+                    continue;
+                };
+
+                let subscriber_id = AttributeValue::component_id(ctx, subscriber_av_id).await?;
+
+                if pasted_component_ids.contains(&subscriber_id) {
```
jkeiser: I think you need to do `if let Some(pasted_subscriber_id) = pasted_component_ids.get(&subscriber_id)` here, and use that on line 3628 ... otherwise, you're subscribing the pasted destination component to the non-pasted source!

britmyerss: that's precisely what I want to do :) 

For example, select 50 components that are connected to a region + credential, but did not select the region/credential, we will recreate the subscriptions to the region credential

PR #6768, opened 2025-07-24
fix: Maintain attribute order in templates and PUT /attributes

jkeiser:
This makes the template generator output attributes in their original order (schema order + array order) rather than sorting them. On the other side, it honors that order in PUT /attributes. Before, templates would order `/domain/Array/10` before `/domain/Array/2`, which caused the PUT to fail.

image

As a side effect, it also means we can support what the docs say the API does, and do multiple appends:

```PUT /attributes
{
  "/domain/Foo/-": 1,
  "/domain/Foo/-": 2,
  "/domain/Foo/-": 3,
}
```

This PR fixes the problem by serializing and deserializing the list of attribute updates using the `tuple_vec_crate` to preserve ordering and duplicates.

## Out of Scope

It might also be desirable to be *able* to handle attributes no matter what order they are specified in the template.

## How was it tested?

- [X] Integration tests pass
- [X] New integration test testing duplicates and ordering
- [X] Manual test: generate template with lots of array items, make sure they show up in the proper order
- [X] Manual test: generate template with subscriptions, make sure they still show up

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2288,6 +2288,7 @@ dependencies = [
  "ringmap",
  "serde",
  "serde-aux",
+ "serde-tuple-vec-map",
  "serde_json",
  "serde_with",
  "shuttle-server",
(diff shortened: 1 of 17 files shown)


adamhjk (dismissed):
lgtm

PR #6790, opened 2025-07-28
Bring back ability to add/remove components from a view

nickgerace:
## Introduction

This change brings back the ability to add components to a view and to remove components from a view.

## Description

There are a lot of changes to the webworker and surrounding areas, but the primary change involves cache busting "ViewComponentList". This is because "ViewComponentList" and other listables are not supposed to be cache-busted. However, this is the first time where we need to do it until we come up with something better because, without it, making changes to components in views will not be reflected in the UI. I tested with this ~300 components locally, and while the initial load wasn't great, subsequent use of the components worked well.

Having fixed the reactivity issue, the issue then became ensuring that the dropdown data was sufficient. Not only did we need to make sure that components could be added and removed from views (rather than just tanking 403 errors), but we also needed to make sure that components would not be orphaned. This results in two new queries that are also special cased in the cache busting world: "componentsInViews" and "componentsInOnlyOneView". These are included in a composable that is used by the context menu to determine what to show. Fortunately, performance here is solid with around ~300 components.

## History

Originally, this used the "current" view for everything, which made the initial implementation functional, but very limited and unusable for adding components to anything other than the default view. That's because you could have only selected  what you saw, so if it was not in the view yet, you couldn't select it to add to that view. If you were in the "All Views" mode, you could add it to the default view, but that was the only way that adding "worked". Removing from a view worked well because you could remove what you could see, but that was still a limited experience compared to what this change now includes.

Both of these worked with keyboard shortcuts because we knew the view that you wanted to add/remove from. Those have been removed as a result of needing a breakout menu.

## Additional Changes

This change includes shortcut modal changes, including the addition of the "right click" option. There would not have been any shortcut modal changes, but since this change originally included shortcuts for the add/remove functionality, they were already in place.

### DIFF
--- a/app/web/src/newhotness/ComponentContextMenu.vue
+++ b/app/web/src/newhotness/ComponentContextMenu.vue
@@ -169,22 +169,37 @@ const showViewInRemoveFromViewMenuOptions = (
   // If there's nothing in the view, there's nothing to remove from it.
   if (componentIdsInView.size < 1) return false;
 
-  // For the selected components, if any of them are in the view, we can remove from it... provided
-  // that the view isn't the only one that the component exists in.
+  // For the selected components, only show the option if all of them are in the view and that view
+  // isn't the final view for any of the components.
   for (const componentId of componentIds.value) {
     const soleViewIdForCurrentComponent =
       componentsAndViews.componentsInOnlyOneView.value[componentId];
     if (
-      componentIdsInView.has(componentId) &&
-      soleViewIdForCurrentComponent !== viewId
-    ) {
-      return true;
-    }
+      !componentIdsInView.has(componentId) ||
+      soleViewIdForCurrentComponent === viewId
+    )
+      return false;
   }
 
-  // If none of the selected components are in the view, we cannot remove any of them from it.
-  return false;
+  // If we don't hit the exit clause, then we are good to include this option.
+  return true;
 };
+
+const removeFromViewTooltip = computed(() => {
+  if (availableViewListOptions.value.removeFromView.length > 0)
+    return undefined;
+  const unprocessedOptions = props.viewListOptions ?? [];
+  for (const unprocessedOption of unprocessedOptions) {
+    const viewId = unprocessedOption.value;
+    for (const componentId of componentIds.value) {
+      const soleViewIdForCurrentComponent =
+        componentsAndViews.componentsInOnlyOneView.value[componentId];
+      if (soleViewIdForCurrentComponent === viewId)
+        return "Cannot remove components from their final view. A given component must exist in at least one view.";
+    }
+  }
+  return undefined;
+});
 
 // ================================================================================================
 // BEGIN CREATING THE MENU OPTIONS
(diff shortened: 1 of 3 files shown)


app/web/src/newhotness/ComponentContextMenu.vue
```diff
@@ -111,8 +117,77 @@ const schemaVariantQuery = useQuery<SchemaVariant | null>({
...
+  if (componentIdsInView.size < 1) return false;
+
+  // For the selected components, if any of them are in the view, we can remove from it... provided
+  // that the view isn't the only one that the component exists in.
+  for (const componentId of componentIds.value) {
+    const soleViewIdForCurrentComponent =
+      componentsAndViews.componentsInOnlyOneView.value[componentId];
+    if (
+      componentIdsInView.has(componentId) &&
+      soleViewIdForCurrentComponent !== viewId
```
nickgerace: Don't orphan

app/web/src/newhotness/ComponentContextMenu.vue
```diff
@@ -250,6 +311,38 @@ const rightClickMenuItems = computed(() => {
...
+          componentsRemoveFromView(option.value, componentIds.value),
+      });
+    }
+    items.push({
+      icon: "minus",
+      label: "Remove from View",
+      submenuItems,
+      submenuVariant: "contextmenu",
+    });
+  }
```
nickgerace: All of that... for this...

app/web/src/workers/webworker.ts
```diff
@@ -2484,6 +2534,98 @@ const getComponentDetails = (
   return details;
 };
 
+const getComponentsInViews = (
```
nickgerace, line 2536: Is it more expensive to do two queries, or to do one query with two data structures in the return? This feels like the better pattern given that we would need to track components that are in only one view in TypeScript, which could involve a massive set of N components that are seen in at least two views. Then again, hogging up database query resources may not be the right move.

app/web/src/workers/webworker.ts
```diff
@@ -2246,6 +2279,23 @@ const postProcess = (
     atomUpdatedFn(workspaceId, changeSetId, kind, id, doc, listIds, removed);
   }
 
+  // FIXME(nick): do not bust lists and find a way to support add/remove component(s) from a view
+  // without it.
+  if (kind === EntityKind.ViewComponentList && bust) {
```
jobelenus: move this to `bustCacheAndReferences`

app/web/src/workers/webworker.ts
```diff
@@ -1288,7 +1318,10 @@ const applyPatch = async (
 
     if (
       COMPUTED_KINDS.includes(atom.kind) ||
-      LISTABLE_ITEMS.includes(atom.kind)
+      LISTABLE_ITEMS.includes(atom.kind) ||
+      // FIXME(nick): do not bust lists and find a way to support add/remove component(s) from a view
+      // without it.
+      atom.kind === EntityKind.ViewComponentList
```
jobelenus: Remove this, `ViewComponentList` goes through `bustCacheAndReferences`

app/web/src/workers/webworker.ts
zacharyhamm: What we want to avoid is N queries where N grows linearly. In this case, the difference between two queries and one query is probably not critical

zacharyhamm (commented):
No objections from a browser database standpoint

PR #6793, opened 2025-07-28
BUG-906 Fix: cannot return a fn from a tanstack queryfn. Invalidate the query on search term change.

jobelenus:

In order to de-bounce search results, we still need to call `invalidateQueries` when the search terms change. We can't return a fn from a tanstack `queryFn` because tanstack attempts to put that result into a store/cache. Once it attempts to use that value, search stops working.

## How was it tested?

- refresh the Explore page
- see that searching working
- click on a component to go to the details page
- click the X to go back to the Explore page
- searching stopped working. With these fixes, it works again.

### DIFF
--- a/app/web/src/newhotness/logic_composables/search.ts
+++ b/app/web/src/newhotness/logic_composables/search.ts
@@ -1,5 +1,13 @@
-import { useQuery } from "@tanstack/vue-query";
-import { MaybeRefOrGetter, Ref, toValue } from "vue";
+import { useQuery, useQueryClient } from "@tanstack/vue-query";
+import {
+  computed,
+  MaybeRefOrGetter,
+  reactive,
+  Ref,
+  toRaw,
+  toValue,
+  unref,
+} from "vue";
 import { Fzf } from "fzf";
 import { bifrostQueryAttributes, useMakeKey } from "@/store/realtime/heimdall";
 import { ComponentInList, EntityKind } from "@/workers/types/entity_kind_types";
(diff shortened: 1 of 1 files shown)


app/web/src/newhotness/logic_composables/search.ts
```diff
@@ -164,23 +191,26 @@ export function useComponentSearch(
...
+  terms: Ref<QueryAttributesTerm[]> | ComputedRef<QueryAttributesTerm[]>,
+) {
   const ctx = useContext();
   const key = useMakeKey();
   // Use tanstack `useQuery` so we can bust the cache of QueryAttributes whenever an
   // AttributeTree is updated. We don't actuall;y do the query in here though, we just return
   // a function that *can* do the query!
-  const attributeTreesUpdatedQuery = useQuery({
+  return useQuery({
     queryKey: key(EntityKind.QueryAttributes),
```
jkeiser: If we're doing this it's probably fine to include the terms as part of the key--they can be reactive AFAIK!

jobelenus: Done, pulled it all in to the same `fn` to simplify

app/web/src/newhotness/logic_composables/search.ts
```diff
@@ -33,8 +41,36 @@ export function useComponentSearch(
...
+    staleTime: 0,
+    queryFn: async () =>
+      await bifrostQueryAttributes(
+        ctx.workspacePk.value,
+        ctx.changeSetId.value,
+        terms.value,
+      ),
+  });
+
+  const componentIds = computed<Set<string>>(() => {
```
jkeiser, line 71: note: <Set<string>> isn't strictly needed here, it's implied by the new Set(). Could just be:

```
  const componentIds = computed(() => new Set(attrQuery.data.value ?? []));
```

PR #6797, opened 2025-07-28
fix: Fix queries breaking when tanstack is loaded from cache

jkeiser:
Sometimes searching for attributes doesn't work. This is because tanstack can't really cache functions, so the function is undefined when it's redhydrated from cache!

This PR fixes the problem by putting an arbitrary value into the cache (new Date()) when AttributeTree MVs get busted, but not putting the function into the cache.

Fixes BUG-906.

#### Out of Scope:

@jobelenus's fix [here](https://github.com/systeminit/si/pull/6793) goes further and tries to cache the results as well, which might be really useful and is worth considering!

## How was it tested?

- [X] Integration tests pass
- [X] Manual test: Do searches, click component, click "X," see search results return
- [X] Manual test: Do searches: refresh, forward and back, see search results return
- [X] Manual test: Complex queries with multiple attr: searches still work


app/web/src/newhotness/logic_composables/search.ts
```diff
@@ -170,17 +170,20 @@ export function useQueryAttributes() {
...
-        terms,
-      ),
+    queryFn: () => new Date(),
   });
-  return (terms: MaybeRefOrGetter<QueryAttributesTerm[]>) =>
-    attributeTreesUpdatedQuery.data?.value?.(toValue(terms));
+  return (terms: MaybeRefOrGetter<QueryAttributesTerm[]>) => {
+    // Just mentioning this will cause us to recompute when any attribute trees update.
+    // eslint-disable-next-line @typescript-eslint/no-unused-expressions
+    attributeTreesUpdatedAt.data.value;
```
jobelenus: this is not reactive (its not inside a computed)... i don't believe this fix will update the search results when the underlying MVs change

jobelenus (approved):

jobelenus:
We can merge this for now. Since it stops caching a fn that truly break things.

@jkeiser The proper fix is to unwind the recursive parse & search calls, and create many `QueryOptions` to use with`useQueries`, and operate on that reactive data, along with the fuzzy search data. That way when `queryKey` get busted, everything re-runs as expected.

PR #6803, opened 2025-07-29
Fix BUG-910: dont prevent saving input on mount.

jobelenus:

### DIFF
--- a/app/web/src/newhotness/AttributePanel.vue
+++ b/app/web/src/newhotness/AttributePanel.vue
@@ -538,7 +538,7 @@ watch(
       saveResourceId();
     },
     500,
-    { leading: true },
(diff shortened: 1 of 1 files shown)


app/web/src/newhotness/AttributePanel.vue
```diff
@@ -601,7 +601,6 @@ const importing = computed(
 );
 
 onMounted(() => {
-  resettingResourceId.value = true;
```
vbustamante, line 604: Was this causing an issue or just unnecessary?

jobelenus: Yes, that was the issue. That prevents on paste from being able to save

PR #6804, opened 2025-07-29
fix(web): ensure we serve the correct workspace token

stack72:
This token isn’t a user token, it’s a token generated for the workspace 
that can be used as part of assume role policies


jobelenus (approved):

jobelenus:
We'll get the workspace store out of here eventually

PR #6805, opened 2025-07-29
feature: si-mcp-server

adamhjk:
Adds an MCP server! Currently supports:

* Managing change sets
* Managing components
* Troubleshooting functions 
* Managing actions
* Reviewing schmea for AWS components only

Doesn't work yet for credentials, IAM policy, and a few other things. Has some buck2 integration, and wants to ship a docker container, but needs a bit of work to actually figure that out.

It allows you to run an MCP server with something like claude code and do fun things.

## How was it tested?

All by hand. We'll have to roll our own test suite at some point!

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -7,7 +7,8 @@
     "./lib/jsr-systeminit/remove-empty",
     "./lib/jsr-systeminit/ai-agent",
     "./lib/jsr-systeminit/cf-db",
-    "./bin/si-luminork-api-tests"
+    "./bin/si-luminork-api-tests",
+    "./bin/si-mcp-server"
   ],
   "imports": {
     "@cliffy/command": "jsr:@cliffy/command@^1.0.0-rc.7",
--- a/BUCK
+++ b/BUCK
@@ -130,6 +130,7 @@ deno_workspace(
         "//lib/jsr-systeminit/cf-db:deno.json",
         "//lib/jsr-systeminit/ai-agent:deno.json",
         "//bin/si-luminork-api-tests:deno.json",
+        "//bin/si-mcp-server:deno.json",
     ],
     visibility = ["PUBLIC"],
 )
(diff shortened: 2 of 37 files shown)


adamhjk:
@stack72, this should ready for a review. it has a buckfile but we're not doing much with it yet :) 

I'll need y'alls help to actually push the docker image, as I think you can only do it via the bot now?

lib/luminork-server/src/service/v1/actions/get_actions.rs
```diff
@@ -40,10 +40,13 @@ pub async fn get_actions(
 ) -> ActionsResult<Json<GetActionsV1Response>> {
     tracker.track(ctx, "api_get_actions", json!({}));
 
+    println!("list topologically");
```
stack72: @adamhjk do we need these?

deno.json
```diff
@@ -7,7 +7,8 @@
     "./lib/jsr-systeminit/remove-empty",
     "./lib/jsr-systeminit/ai-agent",
     "./lib/jsr-systeminit/cf-db",
-    "./bin/si-luminork-api-tests"
+    "./bin/si-luminork-api-tests",
```
stack72, line 10: I am pretty sure this will break the luminork api tests but lets test this out and see before we merge it

adamhjk: It was already there, we're just adding another one line below it?

stack72: yeah my mistake - i tested and it was totally fine :D

stack72 (dismissed):

lib/luminork-server/src/service/v1/actions/get_actions.rs
adamhjk: no! good catch!

adamhjk: fixed

stack72 (approved):

PR #6806, opened 2025-07-29
Add a configurable eviction loop to pgpool, with defaults smaller than our AWS configs

vbustamante:
Our database client code, pgpool, was expecting database connections to live indefinitely and only doing some light checking on connection health when reusing pool connections. This is suspected to be causing service issues where it would keep bad connections on the pool and try to use them when asked, since other services like RDS proxy are much more disconnect prone than us.

This PR adds an async task to our db library that will preemptively evict connection based on time since creation and idle time, hopefully removing connections from our pool before they become unusable, but without adding a penalty to each recycling attempt (which we can also already configure too, but aren't)

The configs for this task allow it to be disabled and for the loop time and the max times to be configured, with defaults that should work on production, but the loop itself disabled so we can roll it out and test it without full redeploys

### DIFF
--- a/lib/si-data-pg/src/lib.rs
+++ b/lib/si-data-pg/src/lib.rs
@@ -8,12 +8,14 @@
 #![allow(clippy::missing_errors_doc)]
 
 use std::{
+    cell::RefCell,
     cmp,
     fmt::{
         self,
         Debug,
     },
     net::ToSocketAddrs,
+    rc::Rc,
     sync::Arc,
     time::Duration,
 };
@@ -174,6 +176,10 @@ pub struct PgPoolConfig {
     pub pool_timeout_wait_secs: Option<u64>,
     pub pool_timeout_create_secs: Option<u64>,
     pub pool_timeout_recycle_secs: Option<u64>,
+    pub pool_total_connection_lifetime_secs: u64,
+    pub pool_idle_connection_lifetime_secs: u64,
+    /// If set to `None`, the eviction task won't be started
+    pub pool_lifetime_check_interval_secs: Option<u64>,
     pub recycling_method: Option<RecyclingMethodConfig>,
 }
 
@@ -193,6 +199,9 @@ impl Default for PgPoolConfig {
             pool_timeout_wait_secs: None,
             pool_timeout_create_secs: None,
             pool_timeout_recycle_secs: None,
+            pool_total_connection_lifetime_secs: 20 * 3600, // 20 hours (RDS Proxy has a 24-hour max)
+            pool_idle_connection_lifetime_secs: 6 * 3600,   // 6 hours
+            pool_lifetime_check_interval_secs: None,
             recycling_method: None,
         }
     }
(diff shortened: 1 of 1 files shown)


lib/si-data-pg/src/lib.rs
```diff
@@ -174,6 +175,10 @@ pub struct PgPoolConfig {
     pub pool_timeout_wait_secs: Option<u64>,
     pub pool_timeout_create_secs: Option<u64>,
     pub pool_timeout_recycle_secs: Option<u64>,
+    pub pool_total_connection_lifetime_secs: u64,
+    pub pool_idle_connection_lifetime_secs: u64,
+    /// If set to `None`, the eviction task won't be started
+    pub pool_lifetime_check_interval_secs: Option<u64>,
```
vbustamante, line 182: The expectation here is that the AWS parameter store will inject this values without any other code changes, but I may be wrong

lib/si-data-pg/src/lib.rs
```diff
@@ -2888,3 +2909,30 @@ impl PgOwnedTransaction {
...
+            let connection_count = Rc::new(RefCell::new(0));
+            let evicted_connections = Rc::new(RefCell::new(0));
+            pg_pool.retain(|_, metrics| {
+                *connection_count.borrow_mut() += 1;
+                if metrics.last_used() > max_idle_time {
+                    debug!("Evicting connection due to idle time");
+                    *evicted_connections.borrow_mut() += 1;
+                    false
+                } else if metrics.age() > max_age {
+                    debug!("Evicting connection due to total age");
```
sprutton1: should we `*evicted_connections.borrow_mut() += 1;` after this?

lib/si-data-pg/src/lib.rs
sprutton1: LayerdLoad just treats ParameterStore as another source, like a config file or env var, so we only need to set values there if we want them to be different than the defaults

lib/si-data-pg/src/lib.rs
vbustamante: most definitely

vbustamante: fixed

lib/si-data-pg/src/lib.rs
```diff
@@ -193,6 +199,9 @@ impl Default for PgPoolConfig {
             pool_timeout_wait_secs: None,
             pool_timeout_create_secs: None,
             pool_timeout_recycle_secs: None,
+            pool_total_connection_lifetime_secs: 20 * 3600, // 20 hours (RDS Proxy has a 24-hour max)
+            pool_idle_connection_lifetime_secs: 6 * 3600,   // 6 hours
+            pool_lifetime_check_interval_secs: None,
```
fnichol, line 204: Yeah, those all seem reasonable, nice that you noted the reason for the total lifetime is because of RDS Proxy's defaults.

lib/si-data-pg/src/lib.rs
```diff
@@ -2888,3 +2915,40 @@ impl PgOwnedTransaction {
 async fn test_connection_infallible_and_warm_up_pool_task(check_pool: PgPool) {
     let _result = check_pool.test_connection().await;
 }
+
+fn evict_old_and_idle_pgpool_connections(
+    pg_pool: Pool,
+    check_interval: Duration,
+    max_age: Duration,
+    max_idle_time: Duration,
+) -> impl Future<Output = ()> {
```
fnichol: Yeah, this'll work, although you may also be able to make this function signature `async fn evict_old_and_idle_pgpool_connections(...) {}`, meaning you can drop the `async move {}` surrounding scope inside the function. My guess would be this might have been an editor refactoring helper? I might be missing something though!

lib/si-data-pg/src/lib.rs
```diff
@@ -2888,3 +2915,40 @@ impl PgOwnedTransaction {
...
+    let min_check_interval = Duration::from_secs(30);
```
fnichol, line 2927: I'm a little curious for the minimum value--is that for a little safety?

fnichol (approved):
Yeah, nice work, nice and tidy! Only a couple of questions, but nothing blocking a merge. Excited to see this working in prod

lib/si-data-pg/src/lib.rs
vbustamante: Yeah, I wanted to make sure we definitely wouldn't configure the system to run this too much, since it blocks the pool. Just a bit of safety. I'm adding a warn! to make this clearer

lib/si-data-pg/src/lib.rs
vbustamante: I didn't know you could just pass a promise to tokio spawn, so I copied the block! Thanks for the callout

lib/si-data-pg/src/lib.rs
vbustamante: also added a comment to explain

lib/si-data-pg/src/lib.rs
fnichol: Pretty wicked, huh?

PR #6812, opened 2025-07-30
Selected components in url & save on blur

jobelenus:
## Changes

1. Selected components go in the URL
2. Whether you're bulk editing goes in the URL
3. Bulk edit no longer has a save button, its all "on blur/enter" now.

### Testing Steps
1. select components, refresh page, see that they remain selected
2. enter bulk editing, refresh page, see that it all remains
3. select / change a value to see that it still saves

### DIFF
--- a/app/web/src/newhotness/Explore.vue
+++ b/app/web/src/newhotness/Explore.vue
@@ -118,9 +118,7 @@
               alwaysShowPlaceholder
               highlightWhenModelValue
               :disabled="pinnedComponentId !== undefined"
-              @update:modelValue="
-                (val) => (groupBySelection = groupByFromString(val))
-              "
+              @update:modelValue="updateGroupBy"
             >
               <template #beforeOptions>
                 <DropdownMenuItem
@@ -128,7 +126,7 @@
                   value="''"
                   checkable
                   :checked="groupBySelection === ''"
-                  @select="() => (groupBySelection = GroupByCriteria.None)"
+                  @select="clearGroupBy"
                 />
               </template>
             </DropdownMenuButton>
@@ -559,10 +557,19 @@ const gridMapSwitcherValue = computed(
 // TODO — if youre on HEAD and you start bulk editing, create a change set right away
 const bulkEditing = ref(false);
 
+watch(bulkEditing, () => {
+  const query: SelectionsInQueryString = {
+    ...router.currentRoute.value?.query,
+  };
+  if (bulkEditing.value) query.b = "1";
+  else delete query.b;
+  storeFilterAndGroup(query);
+  router.push({ query });
+});
+
 watch(gridMapSwitcherValue, (newShowGrid) => {
   // If this is nil, groupRef is unmounted, and we don't care about the change.
   if (_.isNil(newShowGrid)) return;
-  clearSelection();
   const query: SelectionsInQueryString = {
     ...router.currentRoute.value?.query,
   };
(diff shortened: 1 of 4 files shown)


app/web/src/newhotness/explore_grid/AttributePanelBulk.vue
nickgerace, line 1: The simplification work here is masterclass.

app/web/src/newhotness/Explore.vue
```diff
@@ -559,10 +559,20 @@ const gridMapSwitcherValue = computed(
...
+  else delete query.b;
+  storeFilterAndGroup(query);
+  router.push({ query });
+});
+
 watch(gridMapSwitcherValue, (newShowGrid) => {
   // If this is nil, groupRef is unmounted, and we don't care about the change.
   if (_.isNil(newShowGrid)) return;
-  clearSelection();
+  // clearSelection();
```
nickgerace: Should we just remove this and the other one?

app/web/src/newhotness/Explore.vue
```diff
@@ -1608,7 +1647,7 @@ onBeforeUnmount(() => {
   removeEmitters();
   mouseEmitter.off("click", onClick);
 });
-watch([router.currentRoute], setSelectionsFromQuery);
+// watch([router.currentRoute], setSelectionsFromQuery);
```
nickgerace: This too

app/web/src/newhotness/Explore.vue
```diff
@@ -1597,6 +1624,18 @@ const setSelectionsFromQuery = () => {
...
+    const indexes = new Set(query.s.split("|").map((idx) => parseInt(idx)));
+    selectedComponentIndexes.clear();
+    indexes.forEach((idx) => {
+      selectedComponentIndexes.add(idx);
+    });
+  }
+
+  if (query.b && query.b === "1") {
+    bulkEditing.value = true;
+  }
```
nickgerace: Nice. This should work well.

app/web/src/newhotness/Explore.vue
```diff
@@ -1041,7 +1051,7 @@ watch(filteredComponents, () => {
   if (mapRef.value && typeof mapRef.value.deselect === "function") {
     mapRef.value.deselect();
   }
-  clearSelection();
+  // clearSelection();
```
vbustamante: Filtered components may change when the filter changes or when the components list itself changes. For the first case this is unnecessary, but in the second, not clearing the selection may result in the selected component changing (if the list changes in a way the the selected index points to a different component). I'm not sure whats the best behaviour here but keeping it is the safest in a way

app/web/src/newhotness/Explore.vue
```diff
@@ -559,10 +559,20 @@ const gridMapSwitcherValue = computed(
...
+  if (bulkEditing.value) query.b = "1";
+  else delete query.b;
+  storeFilterAndGroup(query);
+  router.push({ query });
+});
+
 watch(gridMapSwitcherValue, (newShowGrid) => {
   // If this is nil, groupRef is unmounted, and we don't care about the change.
   if (_.isNil(newShowGrid)) return;
-  clearSelection();
```
vbustamante, line 565: yeah this one feels like it can go. If the mode switcher changes the user is definitely not focusing on a component already

app/web/src/newhotness/Explore.vue
vbustamante: Removing this breaks retainSessionState, since when it loads up it just pushes values back to the url. In general, this enforces that we react to route changes that don't unmount this component, so we can let the URL control the page

app/web/src/newhotness/Explore.vue
```diff
@@ -1608,6 +1646,9 @@ onBeforeUnmount(() => {
   removeEmitters();
   mouseEmitter.off("click", onClick);
 });
+
+// without this watch the `retainSessionState` functionality doesn't fire
+// perhaps that can live in `beforeEnter` for the route?
 watch([router.currentRoute], setSelectionsFromQuery);
```
nickgerace, line 1652: Strange...

nickgerace (approved):

PR #6819, opened 2025-07-30
feat(dal): default subscriptions for scalar values

zacharyhamm:
Adds a new category node, `DefaultSubscriptionSources`, and a new edge kind: `DefaultSubscriptionSource` which points from an attribute value to the new category.

This allows us to calculate default subscriptions based on the `suggestSources` and `suggestAsSourceFor` maps in `ui_optionals` for a schema. If a suggested connection matches one of the default subscription sources unambiguously, we will be able to create that subscription when the component is created.

This PR implements the initial logic for discovering the default subscriptions and adds the new category node. Default subscriptions will not yet be created on component create (the calculation of suggested sources is expensive and needs to be done ahead of time and cached).

Only scalars supported in this PR. A follow up is coming for arrays, objects, and maps.

Includes a refactor to remove the never used not once source id argument to get_category_node

### DIFF
--- a/lib/dal/src/attribute/value/default_subscription.rs
+++ b/lib/dal/src/attribute/value/default_subscription.rs
@@ -1,0 +1,242 @@
+use std::collections::{
+    BTreeMap,
+    BTreeSet,
+    VecDeque,
+};
+
+use si_id::{
+    AttributeValueId,
+    ComponentId,
+    PropId,
+    ulid::Ulid,
+};
+
+use super::{
+    AttributeValue,
+    AttributeValueResult,
+    subscription::ValueSubscription,
+};
+use crate::{
+    Component,
+    DalContext,
+    EdgeWeight,
+    EdgeWeightKind,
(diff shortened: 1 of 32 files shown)


lib/dal/src/attribute/value/default_subscription.rs
```diff
@@ -0,0 +1,244 @@
...
+    pub async fn set_as_default_subscription_source(
+        ctx: &DalContext,
+        id: AttributeValueId,
+    ) -> AttributeValueResult<()> {
+        let snapshot = ctx.workspace_snapshot()?;
+
+        let category_id = get_or_create_default_subscription_category(ctx).await?;
+
+        snapshot
+            .add_edge(
```
vbustamante: have we decided to stop using the `implement_add_edge_to!` macros? It felt like they made understanding nodes easier at a glance

zacharyhamm: Good call

PR #6838, opened 2025-07-31
fix(template-gen): add region prop suggestion

aaron-dernley:

- Adds a source suggestion of domain/region to the Region prop in the aws template generator code


stack72:
Works a treat!

Screenshot 2025-07-31 at 21 59 58

PR #6844, opened 2025-08-01
fix(web): Optimise the performance of the AddComponentModal

stack72:
Recreating the fuzzy search instance each time was proving to be 
expensive. So we memoize it so we can reuse it

### DIFF
--- a/app/web/src/newhotness/AddComponentModal.vue
+++ b/app/web/src/newhotness/AddComponentModal.vue
@@ -571,6 +571,24 @@ const buildKey = (sv: CategoryVariant) => {
   return key;
 };
 
+// Memoized fuzzy search instance to avoid recreating on every search
+// Recreating the search instance on each debounced request was expensive
+const fzfInstance = computed(() => {
+  const filteredResults: UICategory[] = [];
+
+  // Get the same filtering logic for consistency
+  if (selectedFilter.value) {
+    filteredResults.push(
+      ...getCategoriesAndCountForFilterString(selectedFilter.value).categories,
+    );
+  } else {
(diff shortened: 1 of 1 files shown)


app/web/src/newhotness/AddComponentModal.vue
```diff
@@ -649,15 +660,12 @@ const updateDebouncedSearch = debounce(
...
+// Watch for changes to fuzzySearchString and selectedFilter
 const justCleared = ref(false);
 
 watch([fuzzySearchString, selectedFilter], ([newFuzzySearchString]) => {
   updateDebouncedSearch(newFuzzySearchString);
-
-  nextTick(() => {
-    fixCollapse();
-  });
+  fixCollapse();
```
jobelenus, line 668: are you sure its safe to remove it? have you talked to the Chesterton whose fence you are removing?

jobelenus:
If only git gave a better diff that would have been a lot easier to understand.
**Before:** new fzf every time the search term changes (often and in quick succession).
**After:** new fzf every time the filter changes (once)

jobelenus (approved):

app/web/src/newhotness/AddComponentModal.vue
stack72: 1. fixCollapse() doesn't depend on Vue's DOM updates - it just calls toggleIsOpen() on existing TreeNode components
  2. Eliminates unnecessary async delay - no need to wait for the next tick when we can act immediately
  3. Simpler and more direct - fewer moving parts means better performance and easier debugging
  4. The tree nodes are already rendered - we're just changing their state, not waiting for new DOM elements

PR #6848, opened 2025-08-01
docs: update vpc how to guide for new ui

aaron-dernley:
## Changes:
- New UI Screenshots and Terminology
- Temporarily removed other How-to Guides (apart from the API)


adamhjk:
@aaron-dernley what if we guided the user straight to the template instead?

aaron-dernley:
> @aaron-dernley what if we guided the user straight to the template instead?

@adamhjk I was thinking about that at one point. It's a good idea, we could even off the back of this How-to guide show a user how to create a template from this initial walk through in a How to create a template guide?

PR #6851, opened 2025-08-01
feat(web): Disable the ability for a use to toggle back and forward to the old UX

stack72:
There’s no turning back when we launch our new UX. This means when the 
flag goes on, it stays on :D A user can’t toggle between the old and new
UX - they are in or they are out!

### DIFF
--- a/app/web/src/components/layout/navbar/NavbarPanelRight.vue
+++ b/app/web/src/components/layout/navbar/NavbarPanelRight.vue
@@ -5,81 +5,7 @@
     <Collaborators />
     <Notifications />
 
-    <template v-if="featureFlagsStore.SQLITE_TOOLS">
-      <NavbarButton icon="odin" size="sm">
-        <template #dropdownContent>
-          <DropdownMenuItem
-            v-if="
-              changeSetsStore.selectedWorkspacePk &&
-              changeSetsStore.selectedChangeSetId
-            "
-            icon="niflheim"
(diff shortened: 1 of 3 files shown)


app/web/src/components/layout/navbar/NavbarPanelRight.vue
```diff
@@ -5,81 +5,7 @@
     <Collaborators />
     <Notifications />
 
-    <template v-if="featureFlagsStore.SQLITE_TOOLS">
```
stack72, line 8: this is removing the sqlite tools from the old NavBar

app/web/src/components/layout/navbar/NavbarPanelRight.vue
```diff
@@ -96,146 +22,24 @@
     </template>
 
     <ProfileButton :showTopLevelMenuItems="collapse" />
-
-    <ApplyChangeSetButton v-if="useNewUI" />
```
stack72, line 100: the change set button is not in the bar on the old UI

app/web/src/newhotness/Workspace.vue
```diff
@@ -30,7 +30,7 @@
       >
         <NavbarButton
           tooltipText="Compose"
-          icon="diagram"
+          icon="grid"
```
stack72, line 33: changes to the new Grid icon for the new UX nav bar :D

PR #6856, opened 2025-08-02
Indicate whether Components (and attribute values) have socket connections so we can warn users migration wasn't cleanly possible

britmyerss:

This change implements a method `has_socket_connections()` on `WorkspaceSnapshotSelector` which determines whether a component has at least one input socket with a connection, using lower level abstractions to reduce the performance hit of this work.

We include this data in the `ComponentInListMV` and, when true, for each `AttributeValue` we'll attempt to find if the attribute value is bein set by a socket. 

I've also removed `canBeSetBySocket` from the MV as we do not use this data in newhotness

While we have a process for migration from a socket world to a prop subscription world, some of our older Builtins and various assets authored are not able to be cleanly migrated automatically. This leaves us in a bit of a mixed state - the Component's data is correct in the new experience, but how it came to be is no longer functional.  As most workspaces ARE able to migrate MOST socket connections cleanly, we're adding this to the MV so we can indicate in the new experience when there's a component that is incompatible for modification in the new experience. 

## How was it tested?

- [X] Integration tests - added a new integration test that ensures these values are correct, including accounting for a simple migration case

<div><br/>via <a href="https://giphy.com/hyperrpg/">Hyper RPG</a> on <a href="https://giphy.com/gifs/hyperrpg-yes-agree-agreed-1g0NvvLSFiBjZEqYny">GIPHY</a></div>

### DIFF
--- a/lib/dal-materialized-views/src/component/attribute_tree.rs
+++ b/lib/dal-materialized-views/src/component/attribute_tree.rs
@@ -1,5 +1,6 @@
 use std::collections::{
     HashMap,
+    HashSet,
     VecDeque,
 };
 
@@ -8,11 +9,15 @@ use dal::{
     AttributeValue,
     Component,
     DalContext,
+    InputSocketId,
     Prop,
     Secret,
     component::ControllingFuncData,
     validation::ValidationOutputNode,
-    workspace_snapshot::traits::func::FuncExt as _,
+    workspace_snapshot::traits::{
+        component::ComponentExt,
+        func::FuncExt as _,
+    },
 };
 use si_frontend_mv_types::component::attribute_tree::{
     self,
@@ -41,6 +46,19 @@ pub async fn assemble(ctx: DalContext, component_id: ComponentId) -> crate::Resu
         Component::attribute_value_for_prop(ctx, component_id, &["root", "secrets"]).await?;
     let secret_ids_by_key = Secret::list_ids_by_key(ctx).await?;
 
+    let component_has_socket_connection = ctx
+        .workspace_snapshot()?
+        .has_socket_connections(component_id)
+        .await?;
+    let sockets_on_component: HashSet<InputSocketId> = if component_has_socket_connection {
+        Component::incoming_connections_for_id(ctx, component_id)
+            .await?
+            .into_iter()
+            .map(|c| c.to_input_socket_id)
+            .collect()
+    } else {
+        HashSet::new()
+    };
     let mut attribute_values = HashMap::new();
     let mut props = HashMap::new();
     let mut tree_info = HashMap::new();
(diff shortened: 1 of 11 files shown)


lib/dal-materialized-views/src/component/attribute_tree.rs
```diff
@@ -41,6 +46,19 @@ pub async fn assemble(ctx: DalContext, component_id: ComponentId) -> crate::Resu
...
+        .has_socket_connections(component_id)
+        .await?;
+    let sockets_on_component: HashSet<InputSocketId> = if component_has_socket_connection {
+        Component::incoming_connections_for_id(ctx, component_id)
+            .await?
+            .into_iter()
+            .map(|c| c.to_input_socket_id)
+            .collect()
+    } else {
+        HashSet::with_capacity(0)
```
jkeiser: HashSet::new() will do the same :)

lib/dal-materialized-views/src/component/attribute_tree.rs
```diff
@@ -145,6 +161,17 @@ pub async fn assemble(ctx: DalContext, component_id: ComponentId) -> crate::Resu
             av_id,
             is_dynamic_func,
         };
+        // could probably also convert list_input_socket_sources_for_id into a dfs graph walk but
+        // we shouldn't need to maintain this and it's at least guarded by whether it's worth the work or not
+        let has_socket_connection = if component_has_socket_connection {
+            let sockets_for_av =
+                AttributeValue::list_input_socket_sources_for_id(ctx, av_id).await?;
```
jkeiser, line 168: Input socket sources should only be for attribute functions, right? Like, from source=InputSocket -> dest=Prop on the same component. But even if this is right, I think this will false-flag components--it's not checking whether the connections are from this component, just whether they are from the same socket.

jkeiser: Oh, now I get what you are doing. I think it'll work, but I think it's the long way around. I don't know if we care though. If you want it more direct, it looks like:

- Grab AttributePrototype for this AV (only really need it if it's on the schema, but this should be cheap anyway)
- If the AV's prototype has any APAs pointed at InputSocket:
  - Grab the prototype for the InputSocket
  - Check if any APAs on the prototype have .argument_targets.destination_component_id = <this component>
  - If so, this AV has socket connections. If not, it doesn't.

britmyerss: Yeah I punted this for now, it's what we were doing before to ship this data. Think it's worth changing?

jkeiser (approved):

lib/dal-materialized-views/src/component/attribute_tree.rs
jkeiser: Not unless it's causing problems. What you wrote should work!

lib/dal/src/workspace_snapshot/graph/v4/component.rs
```diff
@@ -168,6 +170,64 @@ impl WorkspaceSnapshotGraphV4 {
             _ => Ok(petgraph::visit::Control::Continue),
         }
     }
+    fn has_socket_connections_dfs_event(
+        event: DfsEvent<NodeIndex>,
+        has_connections: &mut bool,
+        component_id: ComponentId,
+        graph: &Self,
+    ) -> WorkspaceSnapshotGraphResult<Control<()>> {
+        match event {
```
jkeiser, line 179: This whole thing will be *much much* faster if you just look at every node directly in a single `for node_idx in graph.nodes()`. Again, not sure it matters unless perf is bad, but we should keep an eye out.

PR #6867, opened 2025-08-04
Re-read the workspace tokens after you write them

jobelenus:

Brand new workspace's were failing to read their tokens correctly

## How was it tested?

1. create a new local dev workspace
2. hit `go to workspace`
3. get stuck on a blank UI — now this works 😄

### DIFF
--- a/app/web/src/newhotness/Workspace.vue
+++ b/app/web/src/newhotness/Workspace.vue
@@ -83,6 +83,10 @@
     >
       <Lobby v-if="!tokenFail && lobby" />
     </Transition>
+
+    <main v-if="lobby && tokenFail" class="grow min-h-0">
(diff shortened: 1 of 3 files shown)


app/web/src/newhotness/Workspace.vue
```diff
@@ -83,6 +83,10 @@
     >
       <Lobby v-if="!tokenFail && lobby" />
     </Transition>
+
+    <main v-if="lobby && tokenFail" class="grow min-h-0">
+      <div>Bad Token</div>
```
jobelenus, line 88: if this were to happen in the future somehow, users would see "Bad Token" now.

zacharyhamm (approved):

PR #6816, opened 2025-07-30
Resurrect the workspace approvals modal

nickgerace:

PR #6633, opened 2025-07-08
Port approvals spike

nickgerace:

PR #6868, opened 2025-08-04
Resurrect workspace approvals

nickgerace:
## Introduction

This change brings back workspace-level approvals and the approvals subsystem with an eye for the future. The inset modal has been converted into a box within the flow modal. The notifications system, including approving from another change set, has been brought back too. Grab a chair. This ended up being a big one.

## Strategy

The conversion and naming of "InsetApprovalModal" were both intentional. It is no longer inset, but retains the name temporarily for reviewers and to cleanly convey intent about what it does. There will be a quick follow-up change to fix any inconsistent naming, potential anti-patterns, etc. This name will have to change.

The style of the buttons and inner content of that modal are retained as well. You may notice that the apply button is the old one. This is entirely intentional. In the last attempt (PR https://github.com/systeminit/si/pull/6816), part of what made it a total pain to debug was that the outer "apply" button in "ApprovalFlowModal.vue" was also taking on the loading, disabling, labeling, clicking, and styling functionality of the inner "InsertApprovalModal". We have specifically aimed to not touch the style since only two workspaces are using this future.

The strategy was essentially to cleanly port that inset modal into a box within the flow modal, but the key was to ensure that clicking "withdraw request" and "request approval" would flip between the two modes. The modal dynamically sizes with a transition when you flip between modes... and I think it looks cool!

In the long term, we should consider integrating the two Vue components further (particularly the apply/request button), but for now, this should be functional, easy to debug, and a clean separation of the approvals feature from the main flow.

## Only One Approval Kind

This change does not explicitly block other kinds of approvals from existing in the backend and on the graph, but since the UI does not expose the ability to customize approvals, it will show all existing requirements as workspace-level changes. Since there are only two workspaces using approvals and they are workspace-level approvals, we are covered.

## Feature Flagging

Due to the nest of pinia stores in the old UI, the feature was flagged pretty aggressively. In this change, the feature is only flagged in "ApprovalFlowModal.vue". Why? Well, everything else is driven by a prop being passed down through components, like "Notifications.vue". That prop has all change sets with a status of "NeedsApproval". You cannot request approval for a change set outside of the flow modal, so all affected components will not be running queries or even loaded into the UI without the feature enabled.

## Testing

This was tested using a workspace with an owner and a collaborator in two different browsers. Both components with and without actions were tested. The flag was enabled/disabled via code by changing the computed value in "ApprovalFlowModal.vue".

## Other Things We Benefit From

- Added ability to work on another change set from another change set using the "newCtx" pattern (will require a helper eventually)
- Added ability to provide a "WORKSPACE" context to anyone who needs it in the new UI (this was all @jobelenus who put the now-squashed commit in my branch)
- Added ability to navigate to an existing change set with a helper
- Added ability to list and track users in the workspace

## Near Future Work

- Rename and refactor anything related to "inset" language
- Fix the change set rename modal anti-pattern and replace it with the "newCtx" pattern (seen in PR https://github.com/systeminit/si/pull/6816)
- Move all types from stores to either "@/api/sdf/dal" or any of the new directories (e.g. ActionProposedView, approversForChangeSet)

## Screenshots

Waiting to be approved...

Screenshot 2025-08-05 at 1 10 02 PM

...approved!

Screenshot 2025-08-05 at 1 10 12 PM

## Context

This change is a successor to PR https://github.com/systeminit/si/pull/6816 and both were inspired by PR https://github.com/systeminit/si/pull/6633. This required multiple attempts due to the nature of converting an inset approval into a regular component in the new UI, dealing with entirely non-MV content and dealing with workspace-level objects (e.g. users and the "approvals enabled" flag). This attempt was successful due to re-learning about the domain and finding all the dragons, but also taking a commit-by-commit, Vue component-by-component approach to a compiling "newhotness" UI with additional functionality on an iterative basis. Big commit squash party!

### DIFF
--- a/app/web/src/newhotness/InsetApprovalModal.vue
+++ b/app/web/src/newhotness/InsetApprovalModal.vue
@@ -1,0 +1,452 @@
+<template>
+  <!-- 'xl:max-w-[50vw] xl:min-w-[680px] max-h-full', -->
+  <div
+    v-if="mode !== 'error'"
+    :class="
+      clsx(
+        'flex flex-col gap-sm p-sm',
+        'rounded shadow-2xl overflow-hidden',
+        themeClasses('bg-shade-0 border', 'bg-neutral-900'),
+      )
+    "
+  >
+    <!-- HEADER -->
+    <div class="flex flex-row flex-none gap-md mb-sm items-center">
+      <div class="flex flex-col max-w-[66%] min-w-[50%]">
+        <TruncateWithTooltip class="font-bold italic pb-2xs">
+          {{ changeSetName }}
+        </TruncateWithTooltip>
+        <TruncateWithTooltip class="font-bold pb-2xs">{{
+          modalData.title
+        }}</TruncateWithTooltip>
+        <div v-if="modalData.date" class="text-sm italic">
+          <Timestamp :date="modalData.date" showTimeIfToday size="extended" />
+        </div>
+      </div>
+
+      <ErrorMessage
+        :tone="modalData.messageTone"
+        :icon="modalData.messageIcon"
+        variant="block"
+        class="rounded grow"
+      >
+        <template v-if="mode === 'requested'">
+          There are approvals that must be met before the change set can be
+          applied.
+        </template>
+        <template v-else-if="mode === 'approved'">
+          <p>
+            {{ requesterIsYou ? "Your" : "The" }} request to
+            <span class="font-bold">Apply</span> change set
+            <span class="font-bold">{{ changeSetName }}</span> has been
+            approved.
+          </p>
+        </template>
+        <template v-else>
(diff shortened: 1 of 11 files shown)


app/web/src/newhotness/InsetApprovalModal.vue
```diff
@@ -0,0 +1,452 @@
+<template>
+  <!-- 'xl:max-w-[50vw] xl:min-w-[680px] max-h-full', -->
```
jobelenus, line 2: whats this bit?

jobelenus (approved):

app/web/src/newhotness/InsetApprovalModal.vue
nickgerace: It was the original inset classes. I'll remove it when I fix the lints.

PR #6869, opened 2025-08-04
Redesign bulk edit & polish interactions

jobelenus:

- match figma designs
- if you start on HEAD, hitting bulk makes a new change set
- ensure each panel scrolls with contents correctly
- click checkboxes on the left and the component leaves and paths re-render
- hover values on the right, see which components have that value
- add `isSecret` to `ExternalSources` in the MV

#### Out of Scope:

- I don't yet have the "history" concept yet. But that is next! 
- Error reporting is still in toasts, need to move that.

#### Future:
we have a lot of copy & re-used styles and stacks of components... we can really make this nicer to work with

### DIFF
--- a/app/web/src/newhotness/explore_grid/AttributePanelBulk.vue
+++ b/app/web/src/newhotness/explore_grid/AttributePanelBulk.vue
@@ -1,10 +1,12 @@
 <template>
-  <div>
+  <div class="h-full flex flex-col">
     <div
       :class="
         clsx(
-          'flex flex-row items-center gap-xs px-xs mb-xs',
-          themeClasses('bg-neutral-300', 'bg-neutral-600'),
+          'flex-none',
+          'bulk-header flex flex-row items-center gap-xs',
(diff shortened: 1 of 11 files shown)


app/web/src/newhotness/explore_grid/AttributePanelBulk.vue
```diff
@@ -1,10 +1,12 @@
...
-  <div>
+  <div class="h-full flex flex-col">
     <div
       :class="
         clsx(
-          'flex flex-row items-center gap-xs px-xs mb-xs',
-          themeClasses('bg-neutral-300', 'bg-neutral-600'),
+          'flex-none',
+          'bulk-header flex flex-row items-center gap-xs',
+          'mx-[-12px]', // pull this banner beyond the margins of its container's style [&>div]:mx-[12px]
```
wendybujalski, line 8: clever

wendybujalski (approved):

PR #6871, opened 2025-08-05
fix(dal): give mgmt func state transitions a fresh connection state

zacharyhamm:
By providing the management func state transitions with a fresh connection state, they should have an empty job queue, without impacting the "parent" context.

Without this change, when we would update the func state, we were also dispatching (in this case) validation jobs, but as we hadn't actually rebased the work resulting from the management operations, the validation job had the old value still. 

Also added an integration test for this scenario.

### DIFF
--- a/lib/dal/src/context.rs
+++ b/lib/dal/src/context.rs
@@ -940,6 +940,16 @@ impl DalContext {
         Ok(())
     }
 
+    /// Start with a new connection state in this context, with new pg pool and
+    /// nats connections, and an empty job queue.
+    pub async fn restart_connections(&mut self) -> TransactionsResult<()> {
+        self.conns_state = Arc::new(Mutex::new(ConnectionState::new_from_conns(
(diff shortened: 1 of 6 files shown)


lib/dal/src/context.rs
```diff
@@ -940,6 +940,16 @@ impl DalContext {
...
     }
 
+    /// Start with a new connection state in this context, with new pg pool and
+    /// nats connections, and an empty job queue.
+    pub async fn restart_connections(&mut self) -> TransactionsResult<()> {
+        self.conns_state = Arc::new(Mutex::new(ConnectionState::new_from_conns(
+            self.services_context().connections().await?,
+        )));
+
+        Ok(())
```
nickgerace, line 950: Is it possible to "knock out" other valid jobs in flight that aren't management functions, like validation and DVU jobs?

nickgerace (approved):

PR #6874, opened 2025-08-05
Bulk Edit History (and errors)

jobelenus:

I updated some styles to match figma better. And I created some layout components to help encapsulate & reuse things. Two main things, History, and Errors.

### History
In order to create the history, we're reading the AV trees *once*. It is in a watcher, because it comes through a query, so the data doesn't exist on mount. So we stop the watcher after the trees exist, and we process them. Then we never touch the history again. History represents the _initial_ values that you opened the bulk editing window with. It doesn't change as you save things.

In order to know when to display things, after each save, we note that the specific path we saved should now display history. We push the history value of the path into the placeholder variable the template references.

We watch the tree data (MV data) in order to update the placeholder variable for the current selected path value.

I made a big change to our approach that made sure all the input boxes in the form were blank / empty. Previously, I would strip the `av.value` and `av.externalSources` to be blank, so the form wouldn't have anything to display. That was a bad approach—it broke the watchedForm watcher that would reset the form (since the MV data would never get back to the component that it had in fact changed).

To resolve this I added a new context that is provided by `AttributePanel`, `AttributeBulkPanel`, and `ResourcesValuesPanel`, all the panels which have `AttributeInput` as a child/grandchild. That context only has one field that represents "do i want blank inputs". That value is used in several places.

### Errors
In the API composable interface for calling an endpoint now has one more return argument—the specific arguments that you provided to the API call. Why is this useful? Almost all of our endpoints identify the object being acted on via the URL path. So, if one of the calls error out, we can return with the error the arguments, and thus the object that had the error to identify in the UI however we need.

There is a handful of opt-in type magic. To start, the `EndpointArgs` was always just a `Record<string, string>` because in terms of the API layer, it doesn't care what the path arguments are in order to do a string replacement. However, if you're the caller, you expect whatever you gave it to make the calls, comes back out the other side in the response:
```
type ApiArgs = { id: ComponentId };

...

    const saveApi = useApi<ApiArgs>(ctx);
    const call = saveApi.endpoint<SuccessResp>(
      routes.UpdateComponentAttributes,
      { id: componentId },
    );

...

upsertError(resp.endpointArgs.id, path, err);
```
Without this typing, in the last line, `endpointArgs.id` would throw a type error b/c there is no guarantee that it exists in a `Record<string, string>`. So now you're adding error handling code to your error handling. Not exactly ideal 😅 

Aside: I'm _slightly_ annoying that I need to put the endpoint arguments on the `useApi` and not the `.endpoint()`, but I need to type the return values of `useApi` with the generic... so thats just gotta be how it is, such is life.

### Screenshot

image

### DIFF
--- a/app/web/src/newhotness/AttributePanel.vue
+++ b/app/web/src/newhotness/AttributePanel.vue
@@ -157,6 +157,7 @@ import {
   computed,
   onBeforeUnmount,
   onMounted,
+  provide,
   reactive,
   ref,
   watch,
@@ -201,6 +202,9 @@ import {
   makeSavePayload,
 } from "./logic_composables/attribute_tree";
 import AttributeChildLayout from "./layout_components/AttributeChildLayout.vue";
+import { AttributeInputContext } from "./types";
+
+provide<AttributeInputContext>("ATTRIBUTEINPUT", { blankInput: false });
 
 const q = ref("");
 
--- a/app/web/src/newhotness/ResourceValuesPanel.vue
+++ b/app/web/src/newhotness/ResourceValuesPanel.vue
@@ -56,6 +56,7 @@ import {
   computed,
   onBeforeUnmount,
   onMounted,
+  provide,
   reactive,
   ref,
   watch,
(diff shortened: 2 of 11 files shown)


zacharyhamm:
The history panel shows some html chars when you go from an empty prop to a defined one: 
image

zacharyhamm:
a clearer example, for two docker images: 
image

zacharyhamm:
When you uncheck components for editing, they (1) disappear from the list (so you can't go back and change your mind), and (2) if you remove all but one, the attribute editor form disappears, so you have a screen that you can't do much with:
image

jobelenus:
> a clearer example, for two docker images

i fixed up the mucked greater-than symbol!

PR #6878, opened 2025-08-06
feat(dal): default subs for arrays and type checking

zacharyhamm:
Component::new will now create default subscriptions if a default subscription source is present that matches a prop suggestion for any prop in the component prop tree. No api routes create these default subscription sources yet, so no default subs will be created yet.

Default subs for maps do not currently make sense, since we need a way to specify the key for the value that has a default subscription.

### DIFF
--- a/lib/dal/src/component/new.rs
+++ b/lib/dal/src/component/new.rs
@@ -1,0 +1,405 @@
+use std::{
+    collections::{
+        BTreeMap,
+        VecDeque,
+    },
+    sync::Arc,
+};
+
+use si_events::{
+    ContentHash,
+    Timestamp,
+};
+use si_id::{
+    AttributeValueId,
+    SchemaVariantId,
+    ViewId,
+};
+use telemetry::prelude::*;
+
+use super::{
+    Component,
+    ComponentResult,
+};
+use crate::{
+    AttributeValue,
+    ComponentError,
+    DalContext,
+    EdgeWeightKind,
+    InputSocket,
+    OutputSocket,
+    Prop,
+    PropKind,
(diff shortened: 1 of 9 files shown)


lib/dal/src/prop.rs
```diff
@@ -1261,134 +1262,104 @@ impl Prop {
...
-                        source_schema_id: my_schema_id,
-                        source_prop_id: self.id(),
-                        dest_schema_id: schema.id(),
-                        dest_prop_id: prop_id,
-                    });
+                    let mut self_children =
+                        Prop::direct_child_props_ordered(ctx, self_prop.id).await?;
+                    self_queue.reserve(self_children.len());
+                    let mut other_children =
+                        Prop::direct_child_props_ordered(ctx, other_prop.id).await?;
```
nickgerace, line 1336: I think all fields needed (`id`, `name` and `kind`) are on the graph nodes. Wonder if we can make a version of this using those only. Thoughts?

lib/dal/src/component/new.rs
```diff
@@ -0,0 +1,405 @@
...
+        };
+
+        default_sub.subscribe(ctx).await?;
+
+        Ok(())
+    }
+}
+
+impl Component {
+    pub async fn new_component_inner(
```
nickgerace, line 99: nit: should this be public?

lib/dal/src/component/new.rs
nickgerace, line 1: I've seen debate on using `impl` in different files for the same object within the same crate, but I like this pattern. I think we should do it more often when it makes sense to.

PR #6879, opened 2025-08-06
Helper to run API calls against change sets other than the one in your URL

jobelenus:

The API interface defaults to using the `changeSetId` from the `Context` which is based on the URL. But, there are a few times where thats not actually the change set you want to operate against! So lets add a helper to get you the data you want, and clean up the expected fields for the API context (since it doesn't need everything hanging on the `Context` object.

We can think of the `Context` like we think of the DAL Context.

## How was it tested?

1. have multiple change sets
2. go to HEAD
3. open change set dropdown
4. click to rename a changeset
5. submit the name change
6. see the URL match the chosen change set ID

### DIFF
--- a/app/web/src/newhotness/api_composables/index.ts
+++ b/app/web/src/newhotness/api_composables/index.ts
@@ -8,6 +8,7 @@ import {
 } from "@/store/apis.web";
 import { changeSetExists, muspelheimStatuses } from "@/store/realtime/heimdall";
 import router from "@/router";
+import { ChangeSetId } from "@/api/sdf/dal/change_set";
 import { assertIsDefined, Context } from "../types";
 import * as rainbow from "../logic_composables/rainbow_counter";
 import { reset } from "../logic_composables/navigation_stack";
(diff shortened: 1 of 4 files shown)


app/web/src/newhotness/api_composables/index.ts
```diff
@@ -148,29 +149,44 @@ const setLabel = (obs: Obs, label: string): LabeledObs => {
...
+>;
+export const apiContextForChangeSet = (
+  ctx: Context,
+  changeSetId: ChangeSetId,
+): ApiContext => {
+  return {
+    workspacePk: ctx.workspacePk,
+    user: ctx.user,
+    changeSetId: computed(() => changeSetId),
+    onHead: computed(() => ctx.headChangeSetId.value === changeSetId),
```
nickgerace, line 164: This is what I was looking for! :)

app/web/src/newhotness/api_composables/index.ts
```diff
@@ -148,29 +149,44 @@ const setLabel = (obs: Obs, label: string): LabeledObs => {
...
+  };
+};
+
 export class APICall<Response> {
   workspaceId: string;
   changeSetId: string;
   path: string;
-  ctx: Context;
+  ctx: ApiContext;
```
nickgerace, line 172: Good idea to make it a new type

PR #6882, opened 2025-08-06
Re-align approval/apply language and types

nickgerace:
## Description

This change primarily re-aligns the apply/approval language and types to match the current paradigm. All types present in the old stores have been moved to "newhotness" for the eventual wrecking ball moment. As a result of these changes, "si_id" has been folded into "types".

This change also includes a reduction of some "assertIsDefined(ctx)" calls where "useContext" would be better suited. In the "component_actions" logic composable, this has resulted in the preferred pattern of passing in "ctx" into the composable rather than injecting it.


app/web/src/newhotness/logic_composables/action.ts
```diff
@@ -0,0 +1,32 @@
...
+  }[actionKind];
+};
+
+export const actionIconClass = (kind: ActionKind) => {
+  return {
+    Create: "text-success-600",
+    Destroy: "text-destructive-500 dark:text-destructive-600",
+    Refresh: "text-action-600",
+    Manual: "text-action-600",
+    Update: "text-warning-600",
```
wendybujalski: These should use `tw` like so to ensure Tailwind compliance -

```
Create: tw`text-success-600`,
Destroy: tw`text-destructive-500 dark:text-destructive-600`,
Refresh: tw`text-action-600`,
Manual: tw`text-action-600`,
Update: tw`text-warning-600`,
```

Also surprised to see only light/dark styles for Destroy, are we sure these are all right?

wendybujalski (dismissed):

app/web/src/newhotness/logic_composables/action.ts
nickgerace: I can fix `tw` in this PR if desired, but as far as this being "right" goes, this was copied 1:1 from its original location: https://github.com/systeminit/si/blob/56ca26ca7c7c4a0ab029446787585965f3ce99c3/app/web/src/store/actions.store.ts#L99-L127

(just to get us closer to not using these old stores)

wendybujalski: Ah, so they were already not compliant in the previous spot, fun! 🫠

nickgerace: Fixed the `tw` but kept the classes the same

app/web/src/newhotness/ActionCard.vue
```diff
@@ -127,15 +127,15 @@ import {
   DropdownMenuItem,
 } from "@si/vue-lib/design-system";
 import { ActionKind, ActionState } from "@/api/sdf/dal/action";
+import ConfirmHoldModal from "@/components/Actions/ConfirmHoldModal.vue";
```
jobelenus: this one we can move to `newhotness` b/c its not referenced by the old system

jobelenus (changes requested):
one nit!

app/web/src/newhotness/ActionCard.vue
nickgerace: ooooo good catch

jobelenus (approved):

jobelenus (approved):

PR #6885, opened 2025-08-06
upgraded Cypress, new tests

wendybujalski:

Updated to Cypress 14 to be able to test Tab interactions.
Added tests for the following -

- Get the log out test working again (most of the way there already)
- Create a change set (most of the way there already)
- Create components
- Open a component
- Set a value on the component
- Switch between change sets


app/web/src/newhotness/nav/ChangeSetPanel.vue
```diff
@@ -249,7 +250,7 @@ const waitForChangeSetExists = (
 
 async function onSelectChangeSet(newVal: string) {
   if (newVal === "NEW") {
-    createModalRef.value?.open();
+    openCreateModal();
```
britmyerss, line 253: ahh was this what was causing the double change sets??

britmyerss (approved):
great work!!

app/web/src/newhotness/nav/ChangeSetPanel.vue
wendybujalski: Ah, no, that's actually an unrelated bug fix 😅

jobelenus: what was the double change set cause? curious 😅

PR #6887, opened 2025-08-06
Annihilate old sdf routes: round one

nickgerace:
This change deletes unused change set voting process routes. They are not used in the old UI or the new UI in any capacity. It also deletes "get_actions" and "conflicts" unused routes.

The process for finding what is unused involves walking the "routes" module in the main "sdf-server" crate. I stopped after looking at the v1 component routes.

This change also deletes the unused routes from the API tests, if they appear (and they were also unused there).


lib/sdf-v1-routes-component/src/lib.rs
```diff
@@ -193,51 +189,49 @@ impl IntoResponse for ComponentError {
 
 pub fn routes() -> Router<AppState> {
     Router::new()
-        .route("/get_actions", get(get_actions::get_actions))
         .route(
-            "/get_property_editor_schema",
+            "/get_property_editor_schema", // USED IN OLD UI
```
jkeiser, line 193: Thank you for marking these as you notice them :)

jkeiser (approved):
So happy to see this process beginning!

nickgerace:
Removed and re-added just to double check SDF API tests... we are all good.

PR #6889, opened 2025-08-07
feat(sdf,dal): Default subs routes, reason edges and nodes

zacharyhamm:
Make default subscriptions creatable via sdf, and add "reason" nodes that tell us why a subscription was created.


lib/dal/src/workspace_snapshot/node_weight/reason_node_weight.rs
```diff
@@ -0,0 +1,50 @@
...
+    }
+}
+
+/// What reason is expressed by this Reason node?
+/// NOTE: this is postcard encoded, add new reasons to the end
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq, Hash)]
+pub enum Reason {
+    Custom(ContentAddress),
+    DefaultSubscription,
+    UserAdded,
```
britmyerss: I wonder if `UserAdded` should be `UserAdded(UserPk)`? Or even `UserAdded(UserPk, Timestamp)`?

nickgerace: Agreed with `UserPk`!

zacharyhamm: I thought about that... but shouldn't that be part of the auditing? If we want to add it we should add it now, so let me know and i'll whip up a commit to this pr

nickgerace (approved):

lib/dal/src/workspace_snapshot/node_weight/reason_node_weight.rs
zacharyhamm: we've added UserPk

nickgerace: We discussed offline, but while it would normally be part of auditing, keeping the `Reason` domain intact with all the basic information it needs seems reasonable for this PR.

In the future, perhaps we could rely _entirely_ on audit logs without even needing a `Reason` node. @johnrwatson is experimenting with this for "previous values" in the attributes panel.

nickgerace (dismissed):

nickgerace (approved):

PR #6890, opened 2025-08-07
BUG-932: ensure adding the first map key, and subsequent map keys

jobelenus:
I probably b0rked this since day 1 b/c maps are so rare in the system.


zacharyhamm (approved):
Confirmed working with Tags maps

PR #6895, opened 2025-08-08
Disable luminork tests due to flaking

nickgerace:
Example: https://github.com/systeminit/si/actions/runs/16820701301


nickgerace:
Re-running on `main` to double check: https://github.com/systeminit/si/actions/runs/16821245494

britmyerss (approved):

PR #6896, opened 2025-08-08
feat: rbe

sprutton1:
For reference: https://github.com/systeminit/si/pull/6251

**NOTE: this only enables RBE for CI. All builds outside of buildkite will default to local execution. This is a temporary measure as we roll BuildBarn out with public endpoints and authentication**

This PR enables Remote Build Execution. It does so by removing nix as the way that we provide toolchains to buck2 and instead grabs the build tools we need directly from their official sources. Think of it as a pseudo-hermetic build system as we don't rely on ambient tools to build the system (for the most part). All related work is to ensure that we can actually build in remote environments (as well as locally still).

* we will use [buildpack-deps](https://hub.docker.com/_/buildpack-deps/) as a base image for RBE. This lets us avoid having to build or source the common c libs and headers most tools expect to exist on any given system
* we have toolchains for rust, cxx, deno, and python that pull already-built tools down from official sources. These steps are generally slow, but highly cache-able, so the cost need not be paid often. We have toolchains covering linux arm/x64 and macos.
  * doing this directly instead of using tools like nix or mise is certainly clunkier, but also avoids a lot of the nuance and complexity those tools introduce. Here, all problems are our own (and claude's). It will be more difficult to update versions for these tools, but we should be considerate about these updates anyways
  * this also means that third-party tools that require specific cxx or c libs/headers to be available should do so via a fixup. See devicemapper for an example
* similarly, we can no longer rely on nix to provide tools to the rootfs for function execution, so the rootfs script has been updated to go grab these manually. It's much more difficult to add tools to the rootfs in this pattern, but I think we should move away from embedding these tools directly anyhow. Making Veritech responsible for sourcing and providing seems more appropriate
* we will no longer ship as many docker images, besides for the tools that currently run using them. Instead, we will package and ship the actual built binary (omnibus builds without the omnibus) and rely on the host os to have the required libs available
* some tools are still provided by nix. Cargo, cargo-sort, and taplo are needed for linting and formatting work, and these jobs must happen locally
* as stated, dev builds will be local-only to start. However, caching is enabled, so there may be some local benefits there. Cache speed is entirely network speed dependent. If the remote cache is making your life worse, setting `--no-remote-cache` on buck2 commands will disable it
* there are certain changes in `./prelude` that are required to make remote builds work in buildbarn. This is going to be painful the next time we update the prelude. We should consider upstreaming these

#### Out of Scope:

MacOs remote builds. In this PR, macs will build local-only by default. The cache will be enabled for dev-to-dev builds. The support for mac VMs is pretty limited, so this is difficult to manage correctly. We can address this with later work to either build remote mac workers or provide non-mac (cooler) dev environments for mac users. It should be noted that one of the benefits of rbe is remote caching, so your local builds can positively impact CI by allowing it to simply download the things you already built. This is entirely lost with mac builds where cache items only benefit mac users.

There's a not-small amount of follow-on work for CI. I think we may be able to move many of the build targets to use `ci-light` instead of `ci-base`, which is a several minute pull time difference. The linting bits require cargo and friends, but we may be able to strip down ci-base a bit more here. [This doc](https://docs.google.com/document/d/1vBWFIj1_7YJ5qe7SYzdlwbsuaIX0Kaxi4p0ukkg9Y6Y/edit?tab=t.0#heading=h.bghcjnjj342n) describes several changes that I think are still relevant. Pull times for images are going to continue to hamper us unless we switch to longer-lived buildkite agents.

## How was it tested?

So many times and in so many ways. Across CI and local dev, an x64, arm64, and macs


adamhjk:
I mean, lets be clear - this is heroic.

.buckconfig
```diff
@@ -30,3 +30,25 @@ ignore = \
...
+[buck2]
+digest_algorithms = SHA256
+restarter = true
+
+[buck2_re_client]
+action_cache_address = bb.systeminit.com:8980
+cas_address =  bb.systeminit.com:8980
+engine_address = bb.systeminit.com:8980
+respect_symlinks=true
+enabled = true
```
adamhjk: shouldn't this be disabled if we want local builds to not use it yet?

nickgerace:
Yeah this PR is INSANE. Great, great work Scott. Sheesh.

nickgerace (approved):
Approved with the knowledge that users will have to bring their own IDE config now. That was already a rough edge depending on your setup, since you needed `direnv` to play nicely with IDE toolchains. Not a big loss since the source of truth is `buck2`.

It's great to not see `nix` base images. I think we did the right thing since we wanted everything to be docker-based, but the size of the artifacts, everything we had to copy, and the builds were not taking advantage of NixOS caching benefits and using NixOS itself for builds. This was inevitable and I'm really happy to see not only RBE come in, but the images themselves make a lot more sense for what we desire.

This was awesome. Go party Scott!

.buckconfig
sprutton1: I manually toggle it off using the buildkite parameter below. It's a weird situation where I need it enabled, but then optionally disabled depending on the creation of a `.buckconfig.local` file via the nix shell. Setting this to `enabled = false` in the same way didn't work as expected.

nickgerace (dismissed):
Re-approved knowing that no code has changed according to @sprutton1.

prelude/rust/build.bzl
```diff
@@ -1032,7 +1032,10 @@ def _compute_common_args(
 
     # FIXME(JakobDegen): This computation is an awfully broad over-approximation
     emit_requires_linking = dep_metadata_kind == MetadataKind("link")
-    if compile_ctx.dep_ctx.advanced_unstable_linking or not crate_type_codegen(crate_type):
+    # NOTE(scott): This is necessary for remote execution where the advanced linking
+    # optimization doesn't work correctly. Without this, certain binaries will not build becaise
+    # they cannot find the .rlibs they require in the right place. Looking at you bin/pinga
+    if emit != Emit("link") and (compile_ctx.dep_ctx.advanced_unstable_linking or not crate_type_codegen(crate_type)):
```
capickett: Can you share a bit more information about this? e.g. what errors were you seeing and which rlib files were missing?

I work on buck and implemented advanced_unstable_linking. This change is technically fine, but it means you'll miss out on pipelining support for rust binary targets (that's what this `MetadataKind("full")` override enables). We have pipelining for all but the last step; I'm working to enable that last step.

Additionally if it's possible for me to reproduce on my end (and if you can share repro steps), I'd be interested in looking a bit closer at the problem.

sprutton1: The issue manifested as certain binaries not building because of missing deps, I believe because not all of the required files were materializing correctly between remote workers. For example, if I built `buck2 build bin/pinga:pinga` I would see an error like

```
Action failed: root//bin/pinga:pinga (rustc link [pic])
Remote command returned non-zero exit code 1
Remote action digest: `1f172339b048e24a47852feffccfde617a135b3421127e21ff7706e7562ba4e1:142`
stdout:
stderr:
error[E0463]: can't find crate for `pinga_server`
 --> bin/pinga/src/main.rs:4:5
  |
4 | use pinga_server::Server;
  |     ^^^^^^^^^^^^ can't find crate
```
I don't have a detailed analysis handy about which specific files were missing, unfortunately, but IIRC none of the relevant .rlibs (e.g. for pinga-server) were materialized.

All of that said, I undid my change here and forced `advanced_unstable_linking = True` in my toolchain and I am no longer getting this issue. I've hammered out quite a few details since I was having this issue originally, so it's possible it was some other combination of behavior that was giving me grief. Thanks for bringing it up, I'd much rather not maintain changes in prelude.

stack72 (approved):

PR #6902, opened 2025-08-11
Cached schemas MV

vbustamante:

- Adds a "deployment" context for mv creation, separate from the Changeset context we had, so we can build data and add recomputation triggers that exist outside of the graph 

#### Out of Scope:

This PR threads data from the start of a cached modules calculation on sdf to the webworker, but adds no treatment or storage of that data on the webworker itself. This will be dealt with later.

## How was it tested?

- [X] This change should present no UX behavior changes, and this was validated manually by building and using the system


lib/dal/src/cached_module.rs
```diff
@@ -264,7 +266,10 @@ impl CachedModule {
 
     /// Calls out to the module index server to fetch the latest module set, and
     /// updates the cache for any new builtin modules
-    pub async fn update_cached_modules(ctx: &DalContext) -> CachedModuleResult<Vec<CachedModule>> {
+    pub async fn update_cached_modules(ctx: &DalContext, edda_client: EddaClient) -> CachedModuleResult<Vec<CachedModule>> {
+        // FIXME this is for debugging only, sometimes the module index request fails
+        //  and i just want to test for this part.
+        edda_client.rebuild_for_deployment().await?;
```
jhelwig: Intentionally left in?

lib/dal/src/cached_module.rs
```diff
@@ -343,6 +349,10 @@ impl CachedModule {
             tokio::time::sleep(WAIT_BETWEEN_BATCHES).await;
         }
 
+        // Ask edda to rebuild the deployment MVs, which include the cached modules
+        // FIXME uncomment this for the actual PR
```
jhelwig: It's the actual PR. :)

lib/dal/src/entity_kind.rs
```diff
@@ -99,7 +100,7 @@ impl EntityKind {
                     .name()
                     .to_owned();
                 Some(view_name)
-            }
+            },
```
jhelwig: This probably won't pass rustfmt. There shouldn't be a trailing comma when using the block syntax for a match arm.

lib/edda-client/src/lib.rs
```diff
@@ -133,8 +133,7 @@ impl Client {
         let info = ContentInfo::from(&request);
 
         self.publish_inner(
-            workspace_id,
-            change_set_id,
+            Some((workspace_id, change_set_id)),
```
jhelwig: It feels like it'll be worth it pretty quickly to create a struct to represent the workspace ID & change set ID pair, instead of using a raw tuple everywhere.

lib/edda-server/src/change_set_processor_task.rs
```diff
@@ -419,25 +426,53 @@ mod handlers {
     }
 
     #[instrument(
-        name = "edda.change_set_processor_task.handle_request"
+        // Will be renamed to: `edda.requests.change_set.:workspace_id.:change_set_id process`
```
jhelwig: When?

lib/edda-server/src/handlers.rs
```diff
@@ -95,6 +100,26 @@ impl IntoResponse for HandlerError {
 }
 
 pub(crate) async fn default(State(state): State<AppState>, subject: Subject) -> Result<()> {
+    let subject_str = subject.as_str();
+
+    match parse_subject(state.nats.metadata().subject_prefix(), subject_str)? {
+        ParsedSubject::Deployment => run_deployment_processor_task(state, subject_str).await,
+        ParsedSubject::Workspace(_parsed_workspace_id) => {
+            unimplemented!("workspace tasks are not currently implemented!")
```
jhelwig: Panic seems a bit severe to leave in for a todo item. Log as an error, and continue, instead of causing the panic?

lib/edda-server/src/materialized_view.rs
```diff
@@ -411,11 +416,11 @@ pub async fn build_mv_for_changes_in_change_set(
     changes: &[Change],
 ) -> Result<(), MaterializedViewError> {
     let workspace_id = ctx.workspace_pk()?;
-    debug!("building for changes: {:?}", changes);
+    info!("building for changes: {:?}", changes);
```
jhelwig: Intentionally bumping from debug to info?

lib/edda-server/src/materialized_view.rs
```diff
@@ -481,30 +486,92 @@ pub async fn build_mv_for_changes_in_change_set(
...
+        return Ok(());
+    };
+
+
+    let index_reference = IndexReference::from(frontend_object.clone());
+
+    // Make an index, store it with the deployment function
+    let index_entries: Vec<_> = vec![index_reference];
+
+    info!("index_entries {:?}", index_entries);
```
jhelwig: Should this be bumped down to debug? Just want to make sure it won't blow through honeycomb limits.

lib/sdf-server/src/migrations.rs
```diff
@@ -197,9 +202,9 @@ impl Migrator {
     }
```
jhelwig, line 206: When should this TODO be resolved?

lib/sdf-v1-routes-ws/src/bifrost/proto.rs
```diff
@@ -363,17 +387,21 @@ impl BifrostStarted {
...
-                        Err(err) => {
+                        Some(Err(err)) => {
                             warn!(
                                 si.error.message = ?err,
                                 "encountered a recv error on NATS subscription; skipping",
                             );
                         }
+                        // We have a `RecvError`
+                        None => {
+                            dbg!("nats listener has closed; bifrost is probably shutting down");
```
jhelwig: Should be `debug!` when it goes in, instead of `dbg!`?

lib/si-frontend-mv-types-rs/src/cached_schemas.rs
```diff
@@ -0,0 +1,63 @@
...
+    SchemaId,
+    workspace_snapshot::EntityKind,
+};
+use si_id::ulid::Ulid;
+
+use crate::reference::ReferenceKind;
+
+const CACHED_SCHEMAS_STATIC_ID_STR: &str = "01K1Y078734YR40692NNSQRKG4";
+
+#[derive(
```
jhelwig, line 22: https://xkcd.com/221/

jhelwig (commented):
A few comments/questions, but nothing fundamental.

PR #6905, opened 2025-08-11
Re-work grid dependencies

jobelenus:

For [BUG-933](https://linear.app/system-initiative/issue/BUG-933/right-click-context-menu-bugs) leaving the bulk editing "mode" was causing the context modal to float at 0,0 because the template refs due to the show / hiding are getting lost. It seems the menu still has the _old_ template refs, which aren't in the DOM anymore, so the bounding box is all zeros (per spec).

The interdependent components (where `Explore` puts `ExploreGrid` into the Vue tree, but uses the `ExploreGrid` exposed functions in order for `Explore` to function correctly... became very trick because bulk editing _hides_ the `ExploreGrid`!

Unraveling the circular dependency and expanding the context object is a path to a resolution, both for the bug, and making this a bit easier to reason about.

## How was it tested?

Everything on the grid seems to work the same way it did before!
1. sort by
2. group by
3. searching
4. resizing
5. context menu selecting components
6. start bulk edit, close bulk edit, menu stays!

### DIFF
--- a/lib/vue-lib/src/design-system/menus/DropdownMenu.vue
+++ b/lib/vue-lib/src/design-system/menus/DropdownMenu.vue
@@ -140,7 +140,7 @@ interface SubmenuParent {
 const props = defineProps({
   // Set an anchorTo element if you want the DropdownMenu to be attached to a DOM element
   // If no anchorTo element is used, each open() event for this Dropdown will try to determine where to anchor based on the mouse position or event target
-  anchorTo: { type: Object }, // TODO: figure out right type to say "template ref / dom element"
+  anchorTo: { type: Object || HTMLElement },
 
   // You can add DropdownMenuItems via this prop or in a template
   items: {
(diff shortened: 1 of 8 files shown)


lib/vue-lib/src/design-system/menus/DropdownMenu.vue
```diff
@@ -297,7 +297,11 @@ function open(e?: MouseEvent, anchorToMouse?: boolean) {
 
   if (props.anchorTo) {
     // can anchor to a specific element via props
-    anchorEl.value = props.anchorTo.$el;
+    if (props.anchorTo instanceof HTMLElement) {
+      anchorEl.value = props.anchorTo;
+    } else {
+      anchorEl.value = props.anchorTo.$el;
+    }
```
wendybujalski, line 304: yayyy

wendybujalski (approved):
This is a much cleaner and tidier way to handle this data! Awesome.

nickgerace (approved):

PR #6908, opened 2025-08-11
fix(web): Ensure the required name of secret is consistent with the other attributes panel

stack72:
We threw a specific name validator - now we show it’s required with an 
asterix and a tooltip and stop form submission if it’s not present

This is much more in keeping with the AttributeInput


app/web/src/newhotness/layout_components/ComponentSecretAttribute.vue
```diff
@@ -234,6 +243,11 @@ const secretForm = wForm.newForm({
     const propId = props.attributeTree.prop?.id;
     if (!definition) throw new Error("Secret Definition Required");
     if (!propId) throw new Error("Secret Definition Prop Id required");
+    
+    // Stop submission if Name field is missing (no error message shown)
+    if (!value.Name || value.Name.trim() === "") {
+      throw new Error("Name is required");
```
jobelenus: This should be in the validator... was there a problem with that?

jobelenus (changes requested):

app/web/src/newhotness/layout_components/ComponentSecretAttribute.vue
stack72: yes, we show a random validator in the panel that we don't show any other props

stack72: Subnet has a required field validation and doesn't show a validator message - just a tooltip to say it's required

jobelenus: It's less about the visuals--the form shouldn't be allowed to submit without a name. 

The way to do that is by having a validator on the fork to prevent form submission. 

Not allow submission and throw an exception that may prevent other code from running and get reported in observability.

stack72: I made the change to push this back on the validator and not have the validator show the error

jobelenus (approved):

jobelenus (approved):

PR #6909, opened 2025-08-12
Revert PR 6902

vbustamante:
We saw very high disk usage and a bunch of "subject parse error"s on edda when deploying this to tools prod, so we're rolling it back


vbustamante:
Maybe we won't need to merge this actually!

PR #6914, opened 2025-08-12
When removing indexes, ensure you dont remove indexes on any other change set

jobelenus:
## How was it tested?

Changed a components name to get a patch lifecycle that would invalidate an old index and run this block of code.


app/web/src/workers/webworker.ts
```diff
@@ -280,7 +281,19 @@ const ensureTables = (testing: boolean) => {
...
+
+  // OUR FIRST MIGRATION! add a column, allow it to silently fail when the column exists
+  try {
+    sqlite.exec({
+      sql: `
+        ALTER TABLE indexes ADD COLUMN change_set_id TEXT;
+      `,
+    });
+  } catch (err) {
+    // N/A
```
nickgerace: For other reviewers: this to swallow the error and to not prevent further execution.

PR #6935, opened 2025-08-14
feat(luminork): Add the ability to generate a template from a list of components in Luminork

stack72:
Requires a payload similar to the following:

```
{
  "assetName": "My Cool Template",
  "category": "Templates",
  "componentIds": [
    "01K2MM02KNS0A4XM41RDZ1MHXE",
    "01K2MKZ99FTJEYQB249C9NFAF3"
  ],
  "funcName": "Generate My Template"
}
```

PLEASE NOTE: This PR also contains a refactor that moves some of the core of the generate_template work to happen in sdf_core which is shared between SDF and Luminork. Otherwise we would have had to duplicate the logic between both services

Testing:
* Created a template via Luminork
* Was able to extract a template using the UI

Screenshot 2025-08-14 at 18 52 46

### DIFF
--- a/lib/luminork-server/src/service/v1/components/generate_template.rs
+++ b/lib/luminork-server/src/service/v1/components/generate_template.rs
@@ -1,0 +1,108 @@
+use axum::response::Json;
+use dal::{
+    ComponentId,
+    FuncId,
+    SchemaVariantId,
+};
+use serde::{
+    Deserialize,
+    Serialize,
+};
+use si_events::audit_log::AuditLogKind;
+use utoipa::{
+    self,
+    ToSchema,
+};
+
+use crate::{
+    extract::{
+        PosthogEventTracker,
+        change_set::ChangeSetDalContext,
+    },
+    service::v1::ComponentsError,
+};
+
+#[utoipa::path(
+    post,
+    path = "/v1/w/{workspace_id}/change-sets/{change_set_id}/components/generate_template",
+    params(
+        ("workspace_id" = String, Path, description = "Workspace identifier"),
+        ("change_set_id" = String, Path, description = "Change Set identifier"),
+    ),
+    tag = "components",
+    request_body = GenerateTemplateV1Request,
+    summary = "Generate a template",
+    responses(
+        (status = 200, description = "Template generated successfully", body = GenerateTemplateV1Response),
+        (status = 401, description = "Unauthorized - Invalid or missing token"),
+        (status = 500, description = "Internal server error", body = crate::service::v1::common::ApiError)
+    )
+)]
+pub async fn generate_template(
+    ChangeSetDalContext(ref ctx): ChangeSetDalContext,
+    tracker: PosthogEventTracker,
+    payload: Result<Json<GenerateTemplateV1Request>, axum::extract::rejection::JsonRejection>,
+) -> Result<Json<GenerateTemplateV1Response>, ComponentsError> {
+    let Json(payload) = payload?;
+
+    let (new_variant, _, func, prototype_id) = sdf_core::generate_template::prepare_and_generate(
+        ctx,
+        payload.component_ids,
+        payload.asset_name.clone(),
+        payload.func_name.clone(),
+        payload.category.unwrap_or("Templates".to_string()),
+        "#aaaaaa".to_string(),
+    )
+    .await?;
+
+    tracker.track(
+        ctx,
+        "api_generate_template",
+        serde_json::json!({
+            "generated_schema_variant_id": new_variant.id,
+            "generated_prototype_id": prototype_id,
+            "generated_func_id": func.id,
+        }),
+    );
+
+    ctx.write_audit_log(
+        AuditLogKind::GenerateTemplate {
+            schema_variant_id: new_variant.id,
+            management_prototype_id: prototype_id,
+            func_id: func.id,
+            func_name: payload.func_name,
+            asset_name: payload.asset_name.to_owned(),
+        },
+        payload.asset_name,
+    )
+    .await?;
+
+    ctx.commit().await?;
+
+    Ok(Json(GenerateTemplateV1Response {
+        schema_variant_id: new_variant.id,
+        func_id: func.id,
+    }))
+}
+
+#[derive(Deserialize, Serialize, Debug, ToSchema)]
+#[serde(rename_all = "camelCase")]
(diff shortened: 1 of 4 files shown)


lib/luminork-server/src/service/v1/components/generate_template.rs
```diff
@@ -0,0 +1,108 @@
...
+        (status = 401, description = "Unauthorized - Invalid or missing token"),
+        (status = 500, description = "Internal server error", body = crate::service::v1::common::ApiError)
+    )
+)]
+pub async fn generate_template(
+    ChangeSetDalContext(ref ctx): ChangeSetDalContext,
+    tracker: PosthogEventTracker,
+    payload: Result<Json<GenerateTemplateV1Request>, axum::extract::rejection::JsonRejection>,
+) -> Result<Json<GenerateTemplateV1Response>, ComponentsError> {
+    let Json(payload) = payload?;
```
nickgerace, line 46: This is a neat pattern! So there isn't an extractor for the JSON, but we just take the raw "maybe JSON" type and then deserialize it accordingly?

lib/luminork-server/src/service/v1/components/generate_template.rs
```diff
@@ -0,0 +1,108 @@
...
+        "api_generate_template",
+        serde_json::json!({
+            "generated_schema_variant_id": new_variant.id,
+            "generated_prototype_id": prototype_id,
+            "generated_func_id": func.id,
+        }),
+    );
+
+    ctx.write_audit_log(
+        AuditLogKind::GenerateTemplate {
```
nickgerace, line 69: Thank you!

lib/luminork-server/src/service/v1/components/generate_template.rs
```diff
@@ -0,0 +1,108 @@
...
+            schema_variant_id: new_variant.id,
+            management_prototype_id: prototype_id,
+            func_id: func.id,
+            func_name: payload.func_name,
+            asset_name: payload.asset_name.to_owned(),
```
nickgerace, line 74: Once authoring moves off this naming, we can replace this audit log with one that doesn't use the asset language.

lib/luminork-server/src/service/v1/components/generate_template.rs
```diff
@@ -0,0 +1,108 @@
...
+
+    let (new_variant, _, func, prototype_id) = sdf_core::generate_template::prepare_and_generate(
+        ctx,
+        payload.component_ids,
+        payload.asset_name.clone(),
+        payload.func_name.clone(),
+        payload.category.unwrap_or("Templates".to_string()),
+        "#aaaaaa".to_string(),
+    )
+    .await?;
```
nickgerace, line 56: Why not put this in the dal? Just too high level?

lib/sdf-core/src/generate_template.rs
```diff
@@ -0,0 +1,204 @@
...
+    GenerateTemplateError,
+> {
+    let mut has_aws = false;
+    let mut template_components = Vec::new();
+    let mut component_id_to_variable_name: HashMap<ComponentId, String> = HashMap::new();
+    for component_id in &component_ids {
+        let schema = Component::schema_for_component_id(ctx, *component_id).await?;
+        if schema.name.starts_with("AWS") {
+            has_aws = true;
+        }
```
nickgerace, line 60: Will this special casing bite us in the butt?

lib/sdf-core/src/generate_template.rs
```diff
@@ -0,0 +1,204 @@
...
+        for (path, source) in sources {
+            let dest_path = path.path();
+            // We only want domain properties, and we want to deal with Region directly
+            if dest_path.starts_with("/domain/") && !dest_path.starts_with("/domain/extra/Region") {
+                let value = match source {
+                    Source::Value(json) => AttributeSource::value(json),
+                    Source::Subscription {
+                        component,
+                        path,
+                        keep_existing_subscriptions: _keep_existing_subs,
```
nickgerace, line 91: Prefixing underscore... do we want this to be the case?

lib/sdf-core/src/generate_template.rs
```diff
@@ -0,0 +1,204 @@
...
+    let mut sanitized = RE.replace_all(input, "").to_string();
+
+    if let Some(first_char) = sanitized.chars().next() {
+        if !first_char.is_ascii_alphabetic() && first_char != '_' && first_char != '$' {
+            sanitized.insert(0, '_');
+        }
+    }
+
+    sanitized
+}
```
nickgerace, line 180: ...I'm going to trust you that this works.

lib/sdf-server/BUCK
```diff
@@ -40,7 +40,6 @@ rust_library(
         "//lib/si-events-rs:si-events",
         "//lib/si-frontend-mv-types-rs:si-frontend-mv-types",
         "//lib/si-frontend-types-rs:si-frontend-types",
-        "//lib/si-generate-template:si-generate-template",
```
nickgerace, line 43: > This PR also contains a refactor that moves some of the core of the generate_template work to happen in sdf_core which is shared between SDF and Luminork

Got it

lib/luminork-server/src/service/v1/components/generate_template.rs
stack72: yeah we do this to be able to check that it's a parsing or validation error

lib/luminork-server/src/service/v1/components/generate_template.rs
stack72: yeah there's a lot of string templating that's there so it's a bit outside the scope of the dal IMO

lib/sdf-core/src/generate_template.rs
stack72: This is just a refactor of what is there currently - maybe is the answer

nickgerace: Okay great let's keep it then

lib/sdf-core/src/generate_template.rs
stack72: Just moving the code that was there :)

lib/sdf-core/src/generate_template.rs
stack72: Just moving the code that was there :)

nickgerace (approved):

PR #6938, opened 2025-08-14
Implement cached schema & variant materialized views for luminork

jhelwig:
## Summary

This PR implements materialized views (MVs) for cached access to schema and variant data from uninstalled packages, enabling efficient schema information retrieval without requiring full package installation.

- **Add trait-based function collection system** - HasUniqueId trait provides unified function ID access across all function spec types, eliminating repetitive patterns while maintaining zero-allocation semantics through Option<&str> returns

- **Implement cached schema MV infrastructure** - Uses the MV building pattern established by the change set MVs with deployment MV discovery system providing direct key lookup in content-addressable store vs O(nm) traversal through layers

- **Update Luminork API endpoints** - Maintain current behavior of looking up from the graph first, but instead of 404ing if the schema/variant isn't found, check to see if there is an MV. If there is one, return it. If there isn't one, check to see if there _should_ be one by consulting the cached_modules table. If there should be one, return a 202 and trigger a build. If not, return a 404.

- **Add parallel MV building** - JoinSet task scheduling in edda-server with function collection system providing error handling for variant data assembly

## Test plan
- [x] Unit tests cover all HasUniqueId trait implementations for each function type including edge cases for invalid UUIDs and error propagation paths
- [x] Integration tests validate cached schema variant MV assembly including error handling for non-existent schema variants
- [x] Manual testing of Luminork API endpoints through Swagger UI confirms MV-first lookups with DAL fallbacks work correctly


stack72:
@jhelwig on the luminork side this is great!

fnichol (approved):
Live code reviewed and looks great!

PR #6941, opened 2025-08-14
feat(web) - first draft of review page newhotness

wendybujalski:

A first draft of the new review page - if the feature flag is on, it can be accessed by pressing "R" on the Explore page.

This version of the review page only populates its data based on the current change set context. To get totally accurate data we will need to also pull info about Head and then eventually also check the audit log.

#### Screenshots:

image

### DIFF
--- a/app/web/src/newhotness/logic_composables/context.ts
+++ b/app/web/src/newhotness/logic_composables/context.ts
@@ -1,6 +1,8 @@
 import { inject } from "vue";
 import { assertIsDefined, Context } from "../types";
 
+// This will only work in Vue components which are descendants of the newhotness Workspace!
+// DO NOT USE THIS IN WORKSPACE OR INSIDE OF OTHER LOGIC COMPOSABLES
 export function useContext(): Context {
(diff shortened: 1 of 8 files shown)


app/web/src/newhotness/logic_composables/context.ts
```diff
@@ -1,6 +1,8 @@
 import { inject } from "vue";
 import { assertIsDefined, Context } from "../types";
 
+// This will only work in Vue components which are descendants of the newhotness Workspace!
+// DO NOT USE THIS IN WORKSPACE OR INSIDE OF OTHER LOGIC COMPOSABLES
```
jobelenus, line 5: You can use this inside logic composables that are _only_ called with a component's setup. Such logic composables should be named `use...`, so we can visually "know" that its "safe" to call `use...` functions within a `use...` function.

If a composable can be called anywhere (e.g. isn't required to be used during setup) then it should _not_ be named `use...`.

And yes, workspace cannot call `inject` on the same data it `provides`.

jobelenus (approved):

PR #6942, opened 2025-08-14
feat: hardcode AWS Account schema in the SI MCP server

aaron-dernley:
- updates the SI MCP server to be able to find the AWS Account asset

It hardcodes the AWS Account assets documentation and attributes so it can be used with the MCP server.

## How was it tested?

NEEDS TESTING


aaron-dernley:
@adamhjk not sure if this is what was needed, but it felt right and followed the pattern of the IAM assets

adamhjk (approved):
Looks great. Ship it

PR #6943, opened 2025-08-14
Fix: if the index already exists but isnt on the change set, and its a valid move forward, push the change set forward to the existing index

jobelenus:
Nick & Brit got into a state where:
1. browser made the new index checksum
1. browser wrote patch data for a new index
2. browser did not update the change set record to have the new index checksum

Because of the last bit, when they refreshed the page, or did a cold start, the change set wasn't getting set to the new index value. So the data that was being read was stale.

During a cold start, add the call to update the change set record to the new index checksum, even though in 99% of cases it will be a no-op.

## How was it tested?

1. browser A, make a new change set, make a change
2. open browser B (literally a different browser) load the change set.
3. see the change.

### DIFF
--- a/app/web/src/workers/webworker.ts
+++ b/app/web/src/workers/webworker.ts
@@ -2009,16 +2009,13 @@ const niflheim = async (
     // Use index checksum for validation - this is more reliable than snapshot addresses
     const indexChecksum = req.data.indexChecksum;
     const atoms = req.data.frontEndObject.data.mvList;
-    initIndexAndChangeSet(
-      db,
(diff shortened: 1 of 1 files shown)


jobelenus:
@nickgerace i actually changed where we fixed this. Rather than within the shared function for hammers, cold start, and patches... i just put the fix in the cold start, where the error actually happened.

Must less risky in terms of potential for doing bad things given that we have old indexes not being GC'd, etc...

britmyerss (approved):

PR #6944, opened 2025-08-14
fix(web): Normalize search string to prevent filterered components from losing reactivity

britmyerss:
This fixes a bug that can be reproduced as follows: 
- Create a change set
- Create a component
- Type something into the search bar
- backspace and delete the query
- click on the grid
- duplicate the component
- See that the count increases, but the component doesn't show up
- See this error in the console: 
``` 
main.ts:137 TypeError: Cannot read properties of null (reading '0')
    at SearchParser.consume (search.ts:380:26)
    at SearchParser.parseTerm (search.ts:296:14)
    at SearchParser.parseAndCondition (search.ts:281:30)
    at SearchParser.parseOrCondition (search.ts:265:30)
    at SearchParser.parseCondition (search.ts:254:17)
    at SearchParser.parse (search.ts:228:30)
    at parseSearchTerms (search.ts:189:35)
    at search.ts:59:25
    at async.ts:24:26
    at async.ts:48:13
mapStackTrace.cacheGlobally	@	main.ts:137
app.config.errorHandler	@	main.ts:130
Promise.then		
bustTanStackCache	@	heimdall.ts:220
``` 
- type something else in the search bar
- delete it
- see the component on the grid!

This happens because of the following chain of events:
1. Type in search
2. Delete all the text (making `searchString.value === ""`) 
3. Unfocus the search bar (which triggers `onBlur`)
4. `onBlur` calls `cleanValue("")`
5. `cleanValue("")` returns null 
6. This emits `update:modelValue` with `null`
7. The `searchString becomes `null` 
8. The search composable only expects `string | undefined ` 

By normalizing `searchString.value = ""` when `VormInput` sets it to `null` this fixes the bug!

#### Out of Scope:
Any changes to `VormInput` :) 

## How was it tested?
Manually! 
Many permutations of searching, deleting search terms, duplicating/deleting components (things that will impact the grid), and seeing all of those changes flow in as expected, regardless of the search value! 

<div><br/>via <a href="https://giphy.com/cbc/">CBC</a> on <a href="https://giphy.com/gifs/cbc-schitts-creek-QX6tXumZ2Q4rpvWIJ2">GIPHY</a></div>

### DIFF
--- a/app/web/src/newhotness/Explore.vue
+++ b/app/web/src/newhotness/Explore.vue
@@ -1390,8 +1390,12 @@ provide("EXPLORE_CONTEXT", exploreContext.value);
 
 // ================================================================================================
 // THE SEARCH BAR AND FILTERING
-const searchString = ref("");
-const filteredComponents = useComponentSearch(searchString, componentList);
+// searchString can be null because VormInput sets the value to null onBlur if it's an empty string
+const searchString = ref<string | null>("");
+const filteredComponents = useComponentSearch(
+  () => searchString.value ?? "",
+  componentList,
+);
 
 // Filtered components counter state
 const isScrolledToBottom = ref(false);
@@ -1401,7 +1405,7 @@ const shouldShowFilteredCounter = computed(() => {
     componentList.value.length > filteredComponents.value.length;
   return (
     hasFilteredComponents &&
-    (isScrolledToBottom.value || searchString.value.trim() !== "")
+    (isScrolledToBottom.value || (searchString.value ?? "").trim() !== "")
   );
 });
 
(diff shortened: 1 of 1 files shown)


jobelenus (dismissed):

app/web/src/newhotness/Explore.vue
```diff
@@ -1444,7 +1444,9 @@ watch(filteredComponents, () => {
 
 watch(searchString, (newValue, oldValue) => {
   if (oldValue === "" && newValue === null) {
```
britmyerss, line 1454: Not sure if we need both `oldValue === "" && newValue === ""` or if only checking `newValue` is sufficient.....

jkeiser (approved):

PR #6948, opened 2025-08-15
fix(dal): if conn state is invalid, error instead of panicking

zacharyhamm:
We're seeing panics in production in `txns`, which could only happen if the connection state is "invalid". We're investigating why the state is invalid after we try to run a transaction, but panicking is the wrong thing to do here, no matter what. So let's produce an error instead of panicking.

### DIFF
--- a/lib/dal/src/context.rs
+++ b/lib/dal/src/context.rs
@@ -319,6 +319,10 @@ impl ConnectionState {
         mem::replace(self, Self::Invalid)
     }
 
+    fn is_invalid(&self) -> bool {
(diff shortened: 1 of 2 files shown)


jkeiser:
Looks fine--where is the panic actually happening? This looks like it's trying to detect the situation that causes a later panic, and early-returning, is it possible to remove the panic itself or is that somebody else's library?

zacharyhamm:
@jkeiser the panic happens in the `txns` call just below. We panic there in a match

PR #6947, opened 2025-08-15
Fix apply button going off screen for approvals

nickgerace:
## Description

When the actions list is too long, the apply button goes off screen for approvals. This wasn't caught because the list has to be ~20 actions long and the original PR restoring approvals tested this with ~10 actions.

## Screenshots

Request with one action:

Screenshot 2025-08-15 at 11 57 07 AM

Awaiting approval with one action:

Screenshot 2025-08-15 at 11 57 03 AM

Awaiting approval with 72 actions:

Screenshot 2025-08-15 at 11 54 49 AM

PR #6932, opened 2025-08-13
Fix faded apply change set button

nickgerace:

PR #6949, opened 2025-08-15
Fix faded apply change set

nickgerace:
## Introduction

This change fixes the faded apply change set bug hopefully once and for all. I could not reproduce the bug locally, but it is now easy to see how (structurally) the bug could happen. In every instance of the bug happening, the spinner was not moving. We believe that because the status panel was not scoped to change sets, it would be very easy to get into a de-synced state.

## Primary Changes

- Make "status" a reactive record instead of a ref so that it can work for multiple change sets
- Make a super bucket that tracks the bucket of each change set
- Don't disable the button that brings up the modal pretty much ever (we already remove the button when on HEAD)
- Add deep=true for the rainbow watcher for increased UX

## Secondary Changes

- Split the request approval and apply buttons in the modal to reduce confusion (the original was flooded with conditionals)
- Add "useStatus" to the approval flow since it is in the standard apply flow
- Mark dead or deprecated change set statuses

## Bugs Fixed Along the Way

- #6947
- #6944 
- #6943

## Screenshots?

It looks exactly as it does today.

## Testing

- With two users in two browsers (owner and collaborator), apply change sets with each user involving a non-trivial number of components and actions
- See the faded button while the status indicator is on when the modal is open for each user (swap with each apply)
- Now, do the same test, but with approvals enabled
- The same functionality happens, but with the approval flow in the mix
- No faded change set apply button is observed when the status indicator isn't on

### DIFF
--- a/app/web/src/newhotness/ApplyChangeSetButton.vue
+++ b/app/web/src/newhotness/ApplyChangeSetButton.vue
@@ -15,7 +15,6 @@
       "
       loadingText="Applying Changes"
       :loading="applyInFlight"
-      :disabled="disallowApply"
       @click="openApplyChangeSetModal"
     >
       <template #iconRight>
@@ -80,7 +79,7 @@ watchEffect(() => {
   }
 });
 
-const { applyInFlight, disallowApply } = useApplyChangeSet(ctx);
+const { applyInFlight } = useApplyChangeSet(ctx);
 
 const key = useMakeKey();
 const args = useMakeArgs();
--- a/app/web/src/newhotness/ApplyChangeSetModal.vue
+++ b/app/web/src/newhotness/ApplyChangeSetModal.vue
@@ -77,7 +77,8 @@
             @click="closeModalHandler"
           />
           <VButton
-            :label="approvalsEnabled ? 'Request Approval' : 'Apply Change Set'"
+            v-if="approvalsEnabled"
+            label="Request Approval"
             :class="
               clsx(
                 'grow !text-sm !border !cursor-pointer !px-xs',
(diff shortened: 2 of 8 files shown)


app/web/src/newhotness/ApplyChangeSetButton.vue
```diff
@@ -15,7 +15,6 @@
       "
       loadingText="Applying Changes"
       :loading="applyInFlight"
-      :disabled="disallowApply"
```
nickgerace, line 18: Never disable the outer button

app/web/src/newhotness/logic_composables/change_set.ts
```diff
@@ -71,23 +70,16 @@ export const useChangeSets = (
...
     const { req } = await call.post({});
     return { success: api.ok(req) };
   };
-  const disallowApply = computed(
-    () =>
-      (ctx.changeSet.value?.status !== ChangeSetStatus.Open &&
-        ctx.changeSet.value?.status !== ChangeSetStatus.NeedsApproval) ||
-      ctx.onHead.value ||
-      status.value === "syncing",
-  );
```
nickgerace, line 88: Too messy as a composable and only two spots need it now. They both have different logic too.

app/web/src/newhotness/ApplyChangeSetModal.vue
```diff
@@ -80,11 +81,26 @@
...
               )
             "
-            :loadingText="approvalsEnabled ? 'undefined' : 'Applying Changes'"
-            :loading="approvalsEnabled ? false : applyInFlight"
-            :disabled="approvalsEnabled ? false : disallowApply"
             pill="Cmd + Enter"
-            @click="debouncedApplyOrRequestApproval"
+            @click="debouncedRequestApproval"
+          />
+          <VButton
```
nickgerace, line 94: Split into two buttons to reduce confusion. Yes, there's more cleanup and refactor work needing to be done in this domain, but it's piece-by-piece work in progress.

app/web/src/newhotness/ApplyChangeSetModal.vue
```diff
@@ -174,6 +191,24 @@ const approvalsEnabled = computed(
     approvalsEnabledWithoutSoloUserCheck.value && !isSoloUserWorkspace.value,
 );
 
+const status = useStatus();
+const allowedToApplyWithoutApprovals = computed(() => {
```
nickgerace: Let's be extremely explicit about when we disable this button.

app/web/src/newhotness/ApprovalFlow.vue
```diff
@@ -235,9 +236,20 @@ const props = defineProps<{
   user: User;
 }>();
 
-const status = computed(
+// FIXME(nick): remove all pre-ReBAC stuff from this component and only use "satisfied" below to
+// determine if something has been approved. This component should only be used if the change set
+// is in "NeedsApproval" state, but that will require a small refactor.
```
nickgerace, line 239: Originally in #6932, refactoring this stuff out helped uncover how these components truly worked. Distractions removed, etc. However, the end fixes ended up not being related, so they were not included in this PR.

app/web/src/newhotness/StatusPanel.vue
nickgerace, line 1: This is the main churn location

app/web/src/newhotness/StatusPanel.vue
```diff
@@ -58,96 +70,127 @@ watch(
...
+            ChangeSetStatus.Abandoned,
+            ChangeSetStatus.Applied,
+            ChangeSetStatus.Closed,
+          ].includes(data.changeSet.status) &&
+          data.changeSet.id !== ctx.headChangeSetId.value
+        ) {
+          if (status[data.changeSet.id]) {
+            delete status[data.changeSet.id];
+          }
+          delete superBucket[data.changeSet.id];
```
nickgerace, line 129: Garbage collection

PR #6950, opened 2025-08-15
fix(web): Fixup the approvals flow

stack72:
As an approver, the flow wasn’t quite working. Clicking on the buttons 
didn’t give them any feedback. And taking them to the change set didn’t 
pop the modal

https://jam.dev/c/1ae16b62-22c0-49ff-b59e-0b4d94fb1f6d
Light theme is collaborator, dark theme is approver

### DIFF
--- a/app/web/src/newhotness/ApplyChangeSetButton.vue
+++ b/app/web/src/newhotness/ApplyChangeSetButton.vue
@@ -36,11 +36,12 @@
 </template>
 
 <script lang="ts" setup>
-import { computed, ref } from "vue";
+import { computed, ref, watchEffect, nextTick } from "vue";
 import * as _ from "lodash-es";
 import { VButton, PillCounter, themeClasses } from "@si/vue-lib/design-system";
 import clsx from "clsx";
 import { useQuery } from "@tanstack/vue-query";
+import { useRoute, useRouter } from "vue-router";
 import {
   BifrostActionViewList,
   EntityKind,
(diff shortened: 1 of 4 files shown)


app/web/src/newhotness/ApplyChangeSetButton.vue
```diff
@@ -52,12 +53,33 @@ import { useContext } from "./logic_composables/context";
...
+      ...route,
+      query: { ...route.query, openApplyModal: undefined },
+    });
+
+    // Open the modal after navigation and context update are complete
+    setTimeout(() => {
+      nextTick(() => {
+        openApplyChangeSetModal();
+      });
+    }, 100); // Small delay to ensure context is updated
```
jobelenus, line 79: just curious, to ensure which context is updated?

app/web/src/newhotness/nav/ApprovalPendingModalCard.vue
```diff
@@ -72,27 +76,61 @@ defineProps<{
...
 
-  const call = api.endpoint(routes.ChangeSetApprove);
-  call.post({ status: "Rejected" });
+    const call = api.endpoint(routes.ChangeSetApprove);
+    const { req } = await call.post({ status: "Rejected" });
+    if (api.ok(req)) {
+      emit("closeModal");
+    }
+  } finally {
+    isRejecting.value = false;
```
jobelenus, line 116: interesting... i see why you can't use the `api.inFlight` here... I'll think about that a bit

app/web/src/newhotness/ApplyChangeSetButton.vue
stack72: What was happening without the delay:

  1. User clicks ApprovalPendingModalCard
  2. Navigation occurs (router.push to new change set)
  3. Route changes immediately
  4. Watcher fires immediately and detects openApplyModal=true
  5. Modal opens immediately using the old context data
  6. Context updates afterwards with the new change set data

app/web/src/newhotness/nav/ApprovalPendingModalCard.vue
stack72: yeah I am open for suggestions for this one! I *tried* to come up with a way I could at least track it - so if you have ideas I can change here for sure

PR #6957, opened 2025-08-15
feat(dal): Add the internals to create si/tags prop

stack72:
This enables us to build a tagging interface that sits agnostic of cloud
resource tagging and will work for any components

Putting this out there now so that as schema upgrades happen for users, they will be able to take advantage of this new prop that greats created on upgrade

enables us to do something like:

Screenshot 2025-08-16 at 00 57 55


zacharyhamm (approved):

zacharyhamm:
this may impact integration tests?

stack72:
> this may impact integration tests?

good shout! I will run a /try now to check!

zacharyhamm (dismissed):

aaron-dernley (approved):

PR #6959, opened 2025-08-15
feat: add delete context to mcp

aaron-dernley:

adds delete context to the MCP server so that it knows how to delete components.

## How was it tested?

NEEDS TESTING

### DIFF
--- a/bin/si-mcp-server/src/tools/componentDelete.ts
+++ b/bin/si-mcp-server/src/tools/componentDelete.ts
@@ -1,0 +1,76 @@
+import { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
+import { CallToolResult } from "@modelcontextprotocol/sdk/types.js";
+import { z } from "zod";
+import { ComponentsApi } from "@systeminit/api-client";
+import { apiConfig, WORKSPACE_ID } from "../si_client.ts";
+import {
+  errorResponse,
+  generateDescription,
+  successResponse,
+} from "./commonBehavior.ts";
+
+const name = "component-delete";
+const title = "Delete a component";
+const description =
+  `<description>Delete a component for a given componentId. Returns the deletion status on successful deletion. On failure, returns error details.</description><usage>Use this tool to delete a component in a change set. The component will be marked for deletion and removed from the workspace.</usage>`;
+
+const DeleteComponentInputSchemaRaw = {
+  changeSetId: z.string().describe(
+    "The change set to delete the component from; components cannot be deleted from the HEAD change set",
+  ),
+  componentId: z.string().describe("the component id to delete"),
+};
+
+const DeleteComponentOutputSchemaRaw = {
+  status: z.enum(["success", "failure"]),
+  errorMessage: z.string().optional().describe(
+    "If the status is failure, the error message will contain information about what went wrong",
+  ),
+  data: z.object({
+    status: z.string().describe("the deletion status (e.g., 'MarkedForDeletion')"),
+  }),
+};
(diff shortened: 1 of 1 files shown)


aaron-dernley:
@johnrwatson lets try test over the weekend

bin/si-mcp-server/src/tools/componentDelete.ts
```diff
@@ -0,0 +1,76 @@
...
+import {
+  errorResponse,
+  generateDescription,
+  successResponse,
+} from "./commonBehavior.ts";
+
+const name = "component-delete";
+const title = "Delete a component";
+const description =
+  `<description>Delete a component for a given componentId. Returns the deletion status on successful deletion. On failure, returns error details.</description><usage>Use this tool to delete a component in a change set. The component will be marked for deletion and removed from the workspace.</usage>`;
```
adamhjk: You should add to the description that this will cause any resource to also be deleted. That's setting the stage a little bit for adding an 'erase' endpoint that keeps the resource around.

This can merge without that change tho.

bin/si-mcp-server/src/tools/componentDelete.ts
```diff
@@ -0,0 +1,76 @@
...
+
+const DeleteComponentInputSchemaRaw = {
+  changeSetId: z.string().describe(
+    "The change set to delete the component from; components cannot be deleted from the HEAD change set",
+  ),
+  componentId: z.string().describe("the component id to delete"),
+};
+
+const DeleteComponentOutputSchemaRaw = {
+  status: z.enum(["success", "failure"]),
```
adamhjk: Add describe here; when does it return success or failure?

adamhjk (changes requested):
This looks like a reasonable implementation - you still need to add the tool itself to the server, and add a description to the output schema.

adamhjk:
The file you want is `server.ts` - you need to import the function, then add it to createServer and pass the context.

stack72:
we need to make a change to luminork here as status is both the status code response + the name of the parameter returned which is causing deno to get mad

stack72:
I took this for a test drive and made 1 slight change to it to make sure we capture whether it was a successful deletion or not

aaron-dernley:
thank you @stack72 legend!

stack72:
```
can you tell me what components are in my workspace
  ⎿ <response>{"status":"success","data":[{"componentId":"01K31GEG0RJ2F5ZAEJEXVFCFZF","componentName":"si-4958","schemaName"
    :"AWS::EC2::AMI"}]}</response>


⏺ You have 1 component in your workspace:

  - Name: si-4958
  - Type: AWS::EC2::AMI
  - Component ID: 01K31GEG0RJ2F5ZAEJEXVFCFZF

I want you to open a new change set and delete the AMI component

⏺ I'll create a new change set and then delete the AMI component from it.

  si-mcp-server - change-set-create (MCP)(changeSetName: "delete-ami-component")
  ⎿ <response>{"status":"success","data":{"id":"01K31GGE0BPJCHN27S2938ZM30","name":"delete-ami-component","status":"Open","i
    sHead":false}}</response>


⏺ Now I'll delete the AMI component from the new change set:

⏺ si-mcp-server - component-delete (MCP)(changeSetId: "01K31GGE0BPJCHN27S2938ZM30", componentId:
                                        "01K31GEG0RJ2F5ZAEJEXVFCFZF")
  ⎿  <response>{"status":"success","data":{"success":true}}</response>

⏺ The AMI component "si-4958" has been successfully deleted from the new change set "delete-ami-component". The component is
   now marked for deletion and will be removed when the change set is applied.
```

PR #6966, opened 2025-08-17
fix(si-mcp-server): Ensure MCP server prefers apply over approval

stack72:
I have changed the `change-set-update-status` tool. Now if a user passes
`apply` then it’s going to determine if it needs to request approval or
not the for the change set. A user can specify that they don’t care 
about approvals using `force-apply`.

We have also updated the luminork endpoint so that if a user requests 
approval and the workspace hasn’t gotten approvals enabled, then it will
error

### DIFF
--- a/app/web/src/newhotness/nav/Notifications.vue
+++ b/app/web/src/newhotness/nav/Notifications.vue
@@ -1,6 +1,6 @@
 <template>
   <button
-    v-if="numberICanApprove > 0"
+    v-if="numberICanApprove > 0 || currentChangeSetNeedsApproval"
     v-tooltip="{
       content: tooltipText,
       theme: 'notifications',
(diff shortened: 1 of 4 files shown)


app/web/src/newhotness/nav/Notifications.vue
```diff
@@ -106,11 +131,17 @@ const numberICanApprove = computed(() => {
...
+  return ctx.changeSet.value?.status === ChangeSetStatus.NeedsApproval;
+});
+
 const tooltipText = computed(() => {
   if (numberICanApprove.value === 1) {
     return "You have a Change Set to approve.";
   } else if (numberICanApprove.value > 1) {
     return `You have ${numberICanApprove.value} Change Sets to approve.`;
+  } else if (currentChangeSetNeedsApproval.value) {
+    return "This change set requires approval before it can be applied.";
```
nickgerace, line 144: Good idea

lib/luminork-server/src/service/v1/change_sets/request_approval.rs
```diff
@@ -44,7 +45,12 @@ use crate::extract::{
 )]
 pub async fn request_approval(
     ChangeSetDalContext(ref mut ctx): ChangeSetDalContext,
-    WorkspaceAuthorization { .. }: WorkspaceAuthorization,
+    WorkspaceAuthorization {
+        workspace_id,
```
nickgerace, line 49: Does this not just come off the ctx?

stack72: this is just a helper to get it off the ctx :)

PR #6967, opened 2025-08-18
You are in a change set simulation

jobelenus:
image

flex

### DIFF
--- a/app/web/src/newhotness/Explore.vue
+++ b/app/web/src/newhotness/Explore.vue
@@ -301,8 +301,7 @@
             id="footer"
             :class="
               clsx(
-                'flex-none h-12 px-xs border-t flex flex-row items-center',
-                'justify-between',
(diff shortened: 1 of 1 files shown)


mahirl:
I think we should say "You are in a simulation" rather than "..change set simulation." but I'll defer to the product folks, incluing victoria, on the verbiage. This is fine for now, we can refine later, if need be.

adamhjk:
"You are in simulated change set: {change set name}" - and make it not a button, but the same shade as the qualification text color

adamhjk:
image

No border, text that color, ship it

jobelenus:
image

adamhjk:
actually - do we need an 'a' in there?

stack72:
Is the final look for this turn:

Screenshot 2025-08-18 at 22 48 33

PR #6970, opened 2025-08-18
feat: Create change set CIA (create_initialize_apply) endpoint for onboarding flow

vbustamante:
Add the new endpoint so the onboarding flow can initialize the workspace with a region and a credential component, filled out and connected. Also adds a feature flagged modal for testing it out.

Extras: 
- move wait_for_dvu into dal code
- ignore empty secrets when redacting

#### Screenshots:

For testing only:

image
#### Out of Scope:

- The real onboarding UI will be implemented on a further PR.
- The existing modal does not have polish. This is fine because this is only for testing. It only works for a happy path test.
- The CIA endpoint does not publish all the audit logs it needs to, this will be left for another PR.

## How was it tested?

- [ ] Integration tests pass
- [X] Manual test: Enable the feature flag, copy a real credential, use the "Initialize workspace" button on an empty HEAD changeset, fill out the form fields and press Initialize. It should NOT leave an open changeset, and the credential and region components should be on HEAD
- [X] Reran the changed SensitiveString unit tests

### DIFF
--- a/lib/dal/src/change_set.rs
+++ b/lib/dal/src/change_set.rs
@@ -137,6 +137,8 @@ pub enum ChangeSetError {
     SiDb(#[from] Box<si_db::Error>),
     #[error("slow runtime error: {0}")]
     SlowRuntime(#[from] SlowRuntimeError),
+    #[error("timeout out waiting for dvu after {0}ms")]
+    TimedOutAwaitingDvu(u64),
     #[error("transactions error: {0}")]
     Transactions(#[from] Box<TransactionsError>),
     #[error("ulid decode error: {0}")]
@@ -1183,6 +1185,33 @@ impl ChangeSet {
             .await?
             .publish_on_commit(ctx)
             .await?;
+
+        Ok(())
+    }
+
+    /// Wait for the changeset's DVUs to be completely processed before continuing
+    /// This func also runs update_snapshot_to_visibility on the passed in ctx
+    /// with_timeout makes the function fail after a minute, since that's also the rebaser timeout.
+    /// On prod code, always pass `true`. `false` is for testing only.
+    pub async fn wait_for_dvu(ctx: &mut DalContext, with_timeout: bool) -> ChangeSetResult<()> {
+        let mut retry_count = 0;
+        const AWAIT_MS: u64 = 25;
+        const MAX_RETRIES: u64 = (60 * 1000) / AWAIT_MS;
+
+        loop {
+            ctx.update_snapshot_to_visibility().await?;
+            if !DependentValueRoot::roots_exist(ctx).await? {
+                break;
+            }
+            tokio::time::sleep(Duration::from_millis(AWAIT_MS)).await;
+
+            if with_timeout {
+                retry_count += 1;
+                if retry_count > MAX_RETRIES {
+                    return Err(ChangeSetError::TimedOutAwaitingDvu(AWAIT_MS * retry_count));
+                }
+            }
+        }
 
         Ok(())
     }
--- a/lib/sdf-server/src/service/v2/change_set/create_initialize_apply.rs
+++ b/lib/sdf-server/src/service/v2/change_set/create_initialize_apply.rs
@@ -1,0 +1,226 @@
+use axum::{
+    Json,
+    response::IntoResponse,
+};
+use dal::{
+    ChangeSet,
+    Component,
+    DalContext,
+    Prop,
+    Schema,
+    SchemaVariant,
+    Secret,
+    SecretAlgorithm,
+    SecretVersion,
+    diagram::view::View,
+    prop::PropPath,
+    property_editor::values::PropertyEditorValues,
+};
+use sdf_extract::{
+    PosthogEventTracker,
+    workspace::WorkspaceDalContext,
+};
+use serde::{
+    Deserialize,
+    Serialize,
+};
+use si_events::audit_log::AuditLogKind;
+use si_id::KeyPairPk;
+
+use super::Result;
+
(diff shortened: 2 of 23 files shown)


lib/sdf-server/src/service/v2/change_set/create_initialize_apply.rs
```diff
@@ -0,0 +1,233 @@
...
+
+    // Await the DVU before applying the changeset
+    ChangeSet::wait_for_dvu(ctx).await?;
+
+    ChangeSet::apply_to_base_change_set(ctx).await?;
+
+    ctx.write_audit_log(AuditLogKind::ApplyChangeSet, change_set.name.clone())
+        .await?;
+
+    Ok(())
```
nickgerace: @jhelwig @zacharyhamm would love your thoughts here in particular. This is intentionally a single-route design rather than `this new route --> status tracker in frontend for DVU roots --> apply route --> status tracker in frontend for DVU roots --> launch into regular grid view`.

nickgerace: To be clear, this is the desired design. I just pinged you both just to see if this flow will work without potential issues or side effects :)

From my POV, this will work well.

lib/dal/src/change_set.rs
```diff
@@ -1186,6 +1186,20 @@ impl ChangeSet {
...
+        loop {
+            ctx.update_snapshot_to_visibility().await?;
+            if !DependentValueRoot::roots_exist(&ctx).await? {
+                break;
+            }
+            tokio::time::sleep(Duration::from_millis(25)).await;
+        }
+
+        Ok(())
+    }
```
zacharyhamm: We should probably not wait for forever here. That's ok in the tests, but in the dal we should bail after a certain amount of time here. Maybe make it configurable (`retries: Option<usize>`). A minute is a good upper limit since the load balancer will time the request out around then anyway

lib/sdf-server/src/service/v2/change_set/create_initialize_apply.rs
zacharyhamm: Should probably refetch the snapshot after each commit here, to account for corrections and also anything that other users are doing.Otherwise I don't have any problem with this. I made a note about wait_for_dvu

lib/dal/src/change_set.rs
vbustamante: Ohh good call, will implement

lib/sdf-server/src/service/v2/change_set/create_initialize_apply.rs
vbustamante: @zacharyhamm wait_for_dvu does do `update_snapshot_to_visibility`. That's what you're talking about right?

zacharyhamm: oh yes i see that, that will be good enough, yeah

lib/dal/src/change_set.rs
vbustamante: done

nickgerace:
@vbustamante @zacharyhamm did we take a look at this (comment about re-fetching the graph)? I ask because I am got notified for the `/try` calls and comments 😄

https://github.com/systeminit/si/pull/6970?notification_referrer_id=NT_kwDOAlf8a7QxODMwMzg0MDAyNDozOTMyMDY4Mw#discussion_r2283231100

vbustamante:
@nickgerace the graph was already being refetched by wait_for_dvu. Dunno what the integration test errors are about though

vbustamante:
@nickgerace Turns out you were looking at the right direction! The problem was that my change made `commit_and_update_snapshot_to_visibility` not call update snapshot if there were no DVU roots, so a ton of tests that didn't enqueue DVU started failing. Took me a second to figure that out 😅

lib/sdf-server/src/service/v2/change_set/create_initialize_apply.rs
```diff
@@ -0,0 +1,233 @@
...
+                ctx,
+                sv_id,
+                &PropPath::new(["root", "secrets", "AWS Credential"]),
+            )
+            .await?;
+
+            let property_values = PropertyEditorValues::assemble(ctx, component.id()).await?;
+            let aws_credential_attribute_value_id = property_values
+                .find_by_prop_id(aws_credential_prop.id())
+                .expect("unable to find AWS Credential attribute value");
```
britmyerss: Should this expect be here?

vbustamante: oh god no let me fix this

stack72 (dismissed):

lib/sdf-server/src/service/v2/change_set/create_initialize_apply.rs
vbustamante: fixed!

lib/sdf-server/src/service/v2/change_set/create_initialize_apply.rs
```diff
@@ -0,0 +1,232 @@
...
+    ctx.write_audit_log(
+        AuditLogKind::ApproveChangeSetApply {
+            from_status: old_status.into(),
+        },
+        change_set.name.clone(),
+    )
+    .await?;
+
+    // commit again before apply so prepare_for_force_apply changes get saved
+    ctx.commit().await?;
```
britmyerss: @vbustamante we don't need this commit here actually - the prepare_for_force_apply checks if there are dvu roots then locks any unlocked variants/functions - which there won't be any here!

wendybujalski (approved):

PR #6971, opened 2025-08-18
Fix `onHead`, watch immediately

jobelenus:

Fix: ensure the watcher has `immediate: true` on it. My B.

## How was it tested?

About 1 in 10 times the `_headChangeSetId.value` field would be "empty string". Which tells me other things updated "fast enough" / within the mount loop that the watcher never fired in the first place, and thus was comparing a string value against an empty string for "am i on head?" which would always be `true`...

Changing the watcher to be immediate fixes this problem.

### DIFF
--- a/app/web/src/newhotness/Workspace.vue
+++ b/app/web/src/newhotness/Workspace.vue
@@ -281,7 +281,9 @@ const ctx = computed<Context>(() => {
     changeSet,
     approvers,
     user: authStore.user,
-    onHead: computed(() => changeSetId.value === _headChangeSetId.value),
+    onHead: computed(() => {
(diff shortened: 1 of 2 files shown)


app/web/src/newhotness/Workspace.vue
```diff
@@ -329,15 +331,30 @@ const changeSetsNeedingApproval = computed(() =>
...
+  activeChangeSet,
+  () => {
+    changeSet.value = activeChangeSet.value;
+  },
+  { immediate: true },
+);
+watch(
+  headChangeSetId,
+  () => {
+    // never assign a blank value over a truth-y value
```
nickgerace, line 351: Thanks!

PR #6972, opened 2025-08-18
feat(web,sdf): default subscriptions UI

zacharyhamm:
Adds the ability to mark attributes as default subscription sources, and to display components grouped by whether they are subscribed to those attributes.

Hidden behind the `default-subs` feature flag. Enable it to see the checkboxes and ability to group components by default subscription.

NOTE: Two parts of this are unfinished: it is not fully styled. And there is a bug in the interaction with the default subscription where the attribute subscription list will intercept the checkbox click and trigger blur, focusing the attribute input. But I would like to merge this behind the flag and follow up on finishing up the rest of it.

To test: mark a region and credential as a default subscription. Add AWS components, remove them, Ensure the grouping updates.

### DIFF
--- a/app/web/src/workers/webworker.ts
+++ b/app/web/src/workers/webworker.ts
@@ -35,7 +35,7 @@ import Axios, {
 import { ChangeSetId } from "@/api/sdf/dal/change_set";
 import { nonNullable } from "@/utils/typescriptLinter";
 import { DefaultMap } from "@/utils/defaultmap";
-import { ComponentId } from "@/api/sdf/dal/component";
+import { AttributePath, ComponentId } from "@/api/sdf/dal/component";
 import { WorkspacePk } from "@/api/sdf/dal/workspace";
 import { ViewId } from "@/api/sdf/dal/views";
 import {
@@ -84,6 +84,9 @@ import {
   UninstalledVariant,
   AttributeTree,
   ManagementConnections,
+  DefaultSubscriptions,
+  DefaultSubscription,
+  AttributeValue,
 } from "./types/entity_kind_types";
 import {
   bulkDone,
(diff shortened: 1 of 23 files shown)


app/web/src/workers/mjolnir_queue.ts
```diff
@@ -4,7 +4,7 @@ import { ChangeSetId } from "@/api/sdf/dal/change_set";
 import { BustCacheFn, Id } from "./types/dbinterface";
 import { EntityKind } from "./types/entity_kind_types";
 
-const _DEBUG = import.meta.env.VITE_SI_ENV === "local";
+const _DEBUG = false; // import.meta.env.VITE_SI_ENV === "local";
```
jobelenus: remove this sponge

zacharyhamm: whoops, yes

app/web/src/workers/webworker.ts
```diff
@@ -2433,23 +2583,43 @@ const postProcess = (
...
     }
+
     if (removed) {
       possibleConns.get(changeSetId).delete(id);
+      removeDefaultSubscriptionsForComponent(changeSetId, id);
     }
 
     // dont bust individually on cold start
     if (bust) {
+      bustCacheFn(
```
jobelenus, line 2605: shouldn't need this part since atoms bust themselves.

zacharyhamm: @jobelenus without this, the updates are not reactive! i meant to ask about it

jobelenus: `postProcess` for `AttributeTree` shouldn't need to have a `bust` call, b/c there _ought_ to be its own `bust` call for `AttributeTree` within the code path that called `postProcess`... no?

PR #6974, opened 2025-08-18
feat(si-mcp-server): Add a component upgrade tool

stack72:


bin/si-mcp-server/src/tools/componentUpgrade.ts
```diff
@@ -0,0 +1,71 @@
...
+import {
+  errorResponse,
+  generateDescription,
+  successResponse,
+} from "./commonBehavior.ts";
+import _ from "lodash";
+
+const name = "component-upgrade";
+const title = "Upgrade a component in a change set";
+const description = `<description>Upgrades a component in a change set. Returns success if component is upgraded. On failure, returns error details</description><usage>*Always* check that a component can be upgrades by using the 'get-component' tool first. Use this tool when the user wants to process the upgrade a component.</usage>`;
```
britmyerss: nit typo:
`check that a component can be upgrades` vs `check that a component can be upgraded`

stack72: fixed!

stack72:
Moved this to drafts because there's currently no way for the MCP server to get a list of what is actually upgradable or not and we need to expose that information to it

PR #6975, opened 2025-08-18
Fix map button by ignorig grid mode watcher in map

nickgerace:
## Description

When switching to map view, we need to ignore the new grid mode watcher introduced by the default subs UI. If not, the URL will be cleared out and set to the grid view.

## Testing

- Switch between views
- Create some components
- Switch between HEAD, the change set, and the component details

### DIFF
--- a/app/web/src/newhotness/Explore.vue
+++ b/app/web/src/newhotness/Explore.vue
@@ -918,6 +918,9 @@ const clearGroupBy = () => {
 watch(
   [hasSocketConnections, gridMode],
   ([newHasSocketConnections, newGridMode]) => {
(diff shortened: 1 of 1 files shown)


app/web/src/newhotness/Explore.vue
```diff
@@ -918,6 +918,9 @@ const clearGroupBy = () => {
 watch(
   [hasSocketConnections, gridMode],
   ([newHasSocketConnections, newGridMode]) => {
+    // Only new socket connections with incompatible components group by when in grid view.
+    if (!showGrid.value) return;
```
nickgerace, line 922: @jobelenus @zacharyhamm I added this one too

PR #6977, opened 2025-08-18
Fix: don't mutate data potentially returned from tan stack cache

jobelenus:

We were getting a warning about setting the `cs.isHead` value. That `cs` object can be returned from the readonly tan stack cache

## How was it tested?

You can still abandon a change set

### DIFF
--- a/app/web/src/newhotness/nav/AbandonChangeSetModal.vue
+++ b/app/web/src/newhotness/nav/AbandonChangeSetModal.vue
@@ -19,7 +19,7 @@
         icon="x"
         @click="closeModalHandler"
       />
-      <template v-if="!props.changeSet.isHead">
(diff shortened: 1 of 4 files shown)


app/web/src/newhotness/nav/AbandonChangeSetModal.vue
```diff
@@ -37,19 +37,28 @@
...
 import { routes, useApi } from "@/newhotness/api_composables";
+import { useContext } from "../logic_composables/context";
 
 const props = defineProps<{
   changeSet: ChangeSet;
 }>();
 
+const ctx = useContext();
+
+const notHead = computed(
```
nickgerace, line 52: I like this a lot

PR #6980, opened 2025-08-19
Display any errors from the component update attributes endpoint

jobelenus:
These errors are actually pretty rare!

`<AttributePanel>` provides a context with API failures against the `routes.UpdateComponentAttributes` endpoint

#### Screenshots:

The right end "bump" is there for a failing validation, so I'm leaving it there for this error case as well. Can of course adjust it, but it seems we made a decision that errors display this way?

image
image

## How was it tested?

It's actually really hard to generate one of these errors, now that @jkeiser has closed the one error he had with the string template.

1. Mock error data with a hardcoded value (see screenshot)
2. See it display correctly
3. Submit a new value
4. Watch the error disappear on API success

### DIFF
--- a/app/web/src/newhotness/AttributePanel.vue
+++ b/app/web/src/newhotness/AttributePanel.vue
@@ -170,6 +170,7 @@ import {
   onMounted,
   provide,
   reactive,
+  Ref,
   ref,
   watch,
 } from "vue";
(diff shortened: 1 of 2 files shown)


app/web/src/newhotness/AttributePanel.vue
```diff
@@ -309,6 +310,18 @@ const route = useRoute();
 
 const saveApi = useApi();
 
+const saveErrors = ref<Record<string, string>>({});
```
jkeiser, line 313: Is there a way to navigate to another component without reinitializing this (without re-running setup)? Since the context only uses the attribute path as the key, and is scoped to a reactive AttributePanel, I'm worried if you have an error on the Region field, and switch to another component, the Region field in the other component will show an error even though there isn't one.

wendybujalski (approved):

PR #6984, opened 2025-08-19
feat(dal,rebaser-server): Limit concurrent actions to 50

zacharyhamm:
In order to prevent the rebaser task for the head change set from being overwhelmed by a huge number of concurrent actions, we're limiting it to 50 at a time. This should help prevent rebaser deadline elapsed errors when running a large set of actions.


lib/dal/src/action.rs
```diff
@@ -316,18 +316,37 @@ impl Action {
...
-        // get a count of actions currently running/dispatched
+
+        // Only dispatch actions if capacity exists. Too many actions running at
+        // once can overwhelm the rebaser for the HEAD change set.
+        let dispatched_count = Action::dispatched_count(ctx).await?;
+        if dispatched_count >= concurrency_max {
+            return Ok(false);
+        }
+
         for dispatchable_ation_id in Action::eligible_to_dispatch(ctx).await? {
```
britmyerss, line 335: one thought - we might want to sort by action_id here so the oldest eligible gets dispatched first?

zacharyhamm: @britmyerss sure

jhelwig: It might be worth putting that all the way down in `independent_ids` in the base dependency graph implementation.

zacharyhamm: Good call! It is now sorted in dependency graph

PR #6993, opened 2025-08-20
feat(dal): Allow management functions to directly set a resource

britmyerss:

This change removes the root/resource tree from the list of paths management functions are not allowed to update. We recently made a change that allowed refresh actions to run in a change set immediately after import management functions ran, however on considering expanding this behavior to Discovery management functions as well, it became obvious that all of this could be avoided by simply allowing the Discover/Import functions to write to root/resource as well as root/si and root/domain.  I've updated the management func job to only dispatch the Refresh Action if the resource wasn't already set, so once we update Clover, we see the benefits. We can then rip out this logic in the management func job. 

After this change lands, we can update the management functions accordingly and remove the special action dispatching that was going on. 

#### Out of Scope:
- Clover changes

## How was it tested?

- [X] Integration tests pass - added a new one specifically for this 

<div><br/>via <a href="https://giphy.com/gifs/james-franco-same-but-different-C6JQPEUsZUyVq">GIPHY</a></div>


lib/dal/src/attribute/attributes.rs
```diff
@@ -192,10 +193,21 @@ pub async fn update_attributes(
                 counts.set_count += 1;
 
                 // Create the attribute at the given path if it does not exist
-                let target_av_id = av_to_set.vivify(ctx, component_id).await?;
+                let path = av_to_set.path();
+                let target_av_id = av_to_set.clone().vivify(ctx, component_id).await?;
 
                 match value {
                     Source::Value(value) => {
+                        // need to special case resource
```
britmyerss, line 223: calling this out - it felt strange to make the func author nest their response object under /resource/payload and also know to set a /resource/status so we're special casing the handling of setting /response here. All other paths already do this.

zacharyhamm: DId you extend this so you could do it in tests?

zacharyhamm: Ah, I see we use this in the management operator now. Do you think there's any reason to guard against setting resource in the SDF call that uses this?

lib/dal-test/src/helpers/change_set.rs
```diff
@@ -142,6 +145,42 @@ impl ChangeSetTestHelpers {
             .await?;
         Ok(())
     }
+    /// Wait for the management job for a given component to finish. Will wait for at least 10 seconds, checking every
```
zacharyhamm: Will it wait for up to 10 seconds or at least 10 seconds?

In any case, we might want to extend that (by a big margin) since tests can end up being very slow in CI, and this might make the test pretty flaky

PR #6995, opened 2025-08-20
fix(web): Ensure that when navigating to Map via VizualizeSubscriptions we hide unconnected components

stack72:


jobelenus (approved):
:shipit:

PR #7000, opened 2025-08-20
fix(deno): Deleting IAM Permissions check

stack72:
I generated the package locally and it installed perfectly without the 
qualification for the permissionsCheck

I intentionally left the permissionsMap in place incase we want to bring
this back in a different format


stack72:
You can watch this new discovery function in process here - https://jam.dev/c/994dd831-88ff-41a8-af3a-19cfa3a174aa

PR #7001, opened 2025-08-20
fix(web): Ensure default subs box is checkable, style the grouping

zacharyhamm:
Ensure you can click the checkbox and don't scroll to focus the input if the blur event fires because of checking the box.

Matching as closely as I can the design of the default subscriptions grouping. Now shows the value of the source (as the design intends).

Screenshot 2025-08-20 at 3 06 36 PM

### DIFF
--- a/app/web/src/newhotness/explore_grid/ExploreGridRow.vue
+++ b/app/web/src/newhotness/explore_grid/ExploreGridRow.vue
@@ -3,7 +3,7 @@
     v-if="row.type === 'header'"
     :class="
       clsx(
-        'flex flex-row items-center gap-xs px-xs',
+        'flex flex-row items-center justify-center gap-xs px-xs',
         themeClasses('bg-neutral-200', 'bg-neutral-800'),
       )
     "
@@ -24,7 +24,7 @@
     v-else-if="row.type === 'defaultSubHeader'"
     :class="
       clsx(
-        'flex flex-row items-center gap-xs px-xs',
+        'flex flex-row items-center gap-sm px-xs',
         themeClasses('bg-neutral-200', 'bg-neutral-800'),
       )
     "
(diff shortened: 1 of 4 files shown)


stack72 (dismissed):

wendybujalski (dismissed):

app/web/src/newhotness/explore_grid/ExploreGridRow.vue
```diff
@@ -276,10 +316,56 @@ const props = defineProps<{
...
+  { prop: string; value: string } | undefined
+>({
+  enabled: enableDefaultSubSourceValuesQuery,
+  queryKey: key(
+    EntityKind.AttributeTree,
+    props.row.type === "defaultSubHeader" ? props.row.componentId : "",
+    "DefaultSubHeader",
+  ),
+  queryFn: async () => {
+    if (props.row.type !== "defaultSubHeader") {
```
jobelenus, line 339: it feels to me like this should also be whether or not the query is enabled? it looks like a "we dont have data so just get out"?

app/web/src/newhotness/explore_grid/ExploreGridRow.vue
```diff
@@ -276,10 +316,56 @@ const props = defineProps<{
...
+      return undefined;
+    }
+    const attributeTree = await bifrost<AttributeTree>(
+      args(EntityKind.AttributeTree, props.row.componentId),
+    );
+    const value =
```
jobelenus, line 345: im wary of modifying data within the `queryFn`... its safe to do when the `queryFn` is idempotent for every data state & input of the `queryKey`...

What tips me off that it _may_ not be true is the inclusion of `props.row.path` in the `queryFn`, but not in the `queryKey`

jobelenus (changes requested):

app/web/src/newhotness/explore_grid/ExploreGridRow.vue
zacharyhamm: The query will only be enabled if prop.rows.type is defaultSubHeader, but we also need to type assert here to get access to the row fields. Hence the return.

jobelenus: ohhhh its for the typing OK 🫡 maybe just a comment then `// this is not reachable`

app/web/src/newhotness/explore_grid/ExploreGridRow.vue
zacharyhamm: Row path will never change (the attribute --> prop connection there is immutable). If that data is different it means we're computing an entirely different source and attribute.

jobelenus: Will never change... based on what? The _actual_ default sub?

jobelenus (approved):

PR #7003, opened 2025-08-20
feat: pretty review screen

jkeiser:
This PR presents @VictoriaMolgado's slick attribute-by-attribute review screen, driven by a new ComponentDiff MV in the backend.

#### Screenshots:

image

#### Out of Scope:

* Modified components sometimes show more fields than you'd expect. We'll bring that number down case by case.
* New components show too many fields! We'd like them to show fewer down the road.
* Removed components don't show up, because you can't generate an MV for something that doesn't exist. (We plan to solve this a different way.)

## How was it tested?

- [X] New integration tests for MVs
- [X] Integration tests pass
- [X] Manual test: review screen shows modified and added components
- [X] Manual test: review screen is reactive to changes

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8504,6 +8504,7 @@ dependencies = [
  "module-index-types",
  "remain",
  "serde",
+ "serde-tuple-vec-map",
  "serde_json",
  "si-events",
  "si-frontend-mv-types-macros",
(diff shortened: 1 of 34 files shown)


stack72 (approved):
There's a bunch of unnecessary churn as part of this PR that I had to skip over and just assume that it was well tested

The functionality of the reviews screen is looking solid - it's all net new so we can easily iterate as we go here and we know this is an another step toward our goal 

Nice work @wendybujalski and @jkeiser

stack72 (approved):

PR #7005, opened 2025-08-21
feat: Discover matching subscriptions after Discovery and Import

britmyerss:
This PR does a few things: 
- Create a new capability to discover matching subscriptions (“autosubscr ibe”) for a component
- Run this after each import/discovery management execution
- Expose this capability to the front end (currently behind a feature fla g )
- Adds a DashMap cache for Schema -> Prop Suggestions and Components by S schema to reuse when running consecutive autosubscribes (namely,  Discovery) 
- Adds a lot of tests for various scenarios 

The general autosubscribe algo is as follows: 

PreWork: Build the necessary data (schema name -> component id map, and schema v ariant id -> Prop Suggestions on that variant) 

Iterate over every attribute value in the component. 
If: 
1. It’s set manually
2. Has a prop suggestion to another component (or another component has  a `suggestAsSource` suggestion for this attribute value) 
3. The values match
4. There’s only 1 match across the workspace 

Then create the subscription, replacing the manually set value! 

#### Out of Scope:
Not included in this PR is propagating the subscriptions that couldn't be created when it runs in a management function. 

I've also realllllly held back on ripping out more code, but one step at a time

## How was it tested?

- [X] Integration tests pass - many new added!
- [X] Manual test: Tested with right click auto sub and with Discover management functions

<div><br/>via <a href="https://giphy.com/gifs/Filmeditor--halloween-ashley-olsen-l2YWs1NexTst9YmFG">GIPHY</a></div>


lib/dal/src/workspace_snapshot/split_snapshot.rs
```diff
@@ -1332,6 +1337,25 @@ impl SplitSnapshot {
         *inferred_connection_write_guard = None;
     }
 
+    /// Get the prop suggestions cache for autosubscribe optimization
+    pub async fn prop_suggestions_cache(
+        &self,
+        ctx: &DalContext,
+    ) -> WorkspaceSnapshotResult<&PropSuggestionsCache> {
+        if self.prop_suggestions.schema_suggestions.is_empty()
+            || self.prop_suggestions.schema_to_components.is_empty()
```
zacharyhamm: I wonder... should this be `&&`. Like is there any case where the suggestions have been calculated and one of these is empty but that's in fact the correct state?

lib/dal/src/component/suggestion.rs
```diff
@@ -0,0 +1,686 @@
...
+        for component_id in Component::list_ids(ctx).await? {
+            let comp_schema = Component::schema_for_component_id(ctx, component_id).await?;
+            let schema_name = comp_schema.name().to_string();
+
+            // Update or insert the component list for this schema
+            self.schema_to_components
+                .entry(schema_name)
+                .or_default()
+                .push(component_id);
+        }
```
zacharyhamm: Let's construct this map in the same pass over components that we make below for the other map

lib/dal/src/component/new.rs
```diff
@@ -343,6 +343,9 @@ impl Component {
         {
             Action::new(ctx, prototype_id, Some(component.id)).await?;
         }
+        ctx.workspace_snapshot()?
+            .clear_prop_suggestions_cache()
+            .await;
```
zacharyhamm, line 353: Do we have to clear the cache here, or can we just update the schema_to_components map with the new component?

lib/dal-test/src/helpers/schema/variant.rs
```diff
@@ -73,6 +82,27 @@ pub async fn update_asset_func(
     Ok(())
 }
 
+/// Create a qualification function for the given schema variant
+pub async fn create_qualification_func(
```
zacharyhamm, line 86: No harm in keeping this but it isn't used anywhere as far as I can tell

lib/dal/src/component/suggestion.rs
```diff
@@ -0,0 +1,686 @@
...
+
+        // Get all attribute values for this component
+        let attribute_value_tree = AttributeValue::tree_for_component(ctx, component_id).await?;
+        let mut flattened = Vec::new();
+        let mut queue = VecDeque::from_iter(attribute_value_tree.keys().copied());
+
+        while let Some(av_id) = queue.pop_front() {
+            flattened.push(av_id);
+            if let Some(children) = attribute_value_tree.get(&av_id) {
+                queue.extend(children);
```
zacharyhamm, line 378: As a bit of a perf benefit, we should do:

```
queue.reserve(children.len());
queue.extend(children)
```

lib/dal/src/component/suggestion.rs
```diff
@@ -0,0 +1,686 @@
...
+                schema: self.schema_name.clone(),
+                prop: prop_path.clone(),
+            };
+            let target_path = AttributePath::JsonPointer(prop_path);
+
+            // Check for explicit suggestions from the cache
+            if let Some(suggestions) = component_suggestion_map
+                .suggest_sources_map
+                .get(&prop_suggestion.prop)
+            {
```
zacharyhamm: This is a perfect case for `let else` to make the loop easier to read:

```
let Some(suggestions) = component_suggestion_map.suggest_sources_map.get(&prop_suggestion.prop) else {
    continue;
};
```

lib/dal/src/component/suggestion.rs
```diff
@@ -0,0 +1,686 @@
...
+                // This prop has explicit suggested sources, check them
+                for suggestion in suggestions {
+                    if let Some(source_components) =
+                        cache.schema_to_components.get(&suggestion.schema)
+                    {
```
zacharyhamm: Also a `let else` here.

lib/dal/src/component/suggestion.rs
```diff
@@ -0,0 +1,686 @@
...
+                        for &source_component_id in source_components.iter() {
+                            if source_component_id == self.component_id {
```
zacharyhamm: Good call here. I should make sure default subs are doing this

lib/dal/src/component/suggestion.rs
```diff
@@ -0,0 +1,686 @@
...
+                            }
+
+                            // Find the attribute value at the suggested prop path
+                            if let Some(source_av_id) = AttributeValueIdent::new(&suggestion.prop)
+                                .resolve(ctx, source_component_id)
+                                .await?
+                            {
+                                let source_value = AttributeValue::view(ctx, source_av_id).await?;
+                                // If values match, this is a potential subscription
+                                if source_value == Some(current_value.clone()) {
```
zacharyhamm: Let's avoid this clone. You should be able to do `source_value.as_ref() == Some(&current_value)`

lib/dal/src/component/suggestion.rs
```diff
@@ -0,0 +1,686 @@
...
+            let target_path = AttributePath::JsonPointer(prop_path);
+
+            for schema_variant_entry in cache.schema_suggestions.iter() {
+                let other_schema_variant_id = schema_variant_entry.key();
+                let other_suggestion_map = schema_variant_entry.value();
+
+                // Look for sources that suggest themselves for this schema and prop
+                if let Some(source_prop_paths) = other_suggestion_map
+                    .suggest_as_source_for_map
+                    .get(&prop_suggestion)
```
zacharyhamm: I also think `let else`s would make this much cleaner.

lib/dal/src/component/suggestion.rs
```diff
@@ -0,0 +1,686 @@
...
+            let schema_variant_id = Component::schema_variant_id(ctx, component_id).await?;
+
+            if processed_schema_variants.contains(&schema_variant_id) {
+                continue; // Already processed this schema variant
+            }
+            processed_schema_variants.insert(schema_variant_id);
+
+            // Build suggestion map for this schema variant
+            let mut suggest_sources_map = HashMap::new();
+            let mut suggest_as_source_for_map = HashMap::new();
```
zacharyhamm: Can we unify these two maps so we only have to do one pass when finding matches.

That is, if you detect a suggestAsSourceFor, transform that into a suggestSources?

lib/dal/src/component.rs
```diff
@@ -4036,7 +4039,9 @@ impl Component {
             )
             .await?;
         }
-
+        ctx.workspace_snapshot()?
+            .clear_prop_suggestions_cache()
+            .await;
         Ok(upgraded_component)
     }
```
zacharyhamm, line 4056: Shouldn't we update the cache on component removal?

nickgerace (approved):

PR #7011, opened 2025-08-21
feat(web) - review screen arrow controls + show only domain and name

wendybujalski:
Small turn to get these two changes in to better parallelize the rest of the work

- Arrow controls on the review screen
- AV diffs are filtered to show only ones that are part of `domain` and also `/si/name`

### DIFF
--- a/app/web/src/newhotness/Review.vue
+++ b/app/web/src/newhotness/Review.vue
@@ -38,6 +38,8 @@
         :borderBottom="false"
         @focus="() => (selectedComponentId = undefined)"
         @keydown.tab="onSearchTab"
+        @keydown.up="() => searchControl(true)"
+        @keydown.down="() => searchControl(false)"
       />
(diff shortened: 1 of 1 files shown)


app/web/src/newhotness/Review.vue
```diff
@@ -387,6 +390,24 @@ const selectedComponent = computed(() => {
   };
 });
 
+const selectedComponentDisplayDiffs = computed(() => {
+  if (selectedComponent.value?.diff?.attributeDiffs) {
+    const entries = Object.entries(selectedComponent.value.diff.attributeDiffs);
+    const output: [AttributePath, AttributeDiff][] = [];
+
+    entries.forEach((entry) => {
```
jkeiser, line 398: This is worth expressing as a single filter():
```ts
     const output = entries.filter(([path, _]) => path.startsWith("/domain") || path == "/si/name");
```

PR #7020, opened 2025-08-21
feat(web): Onboarding flow

vbustamante:
Implement the onboarding flow that initializes the workspace with a region and a credential on head, and gives the user an api token, with the ability to skip the whole thing 

#### Screenshots:

image

#### Out of Scope:

- When using the onboarding flow from the "Initialize your workspace" button, we don't get the initialization component showing on head reactively. This will be tackled on a further PR 
- Copy for the AI tutorial is stubbed. Will be filled in tomorrow

## How was it tested?

- [X] Manual test: Use an account that hasn't dismissed the old video modal, preferably a new one, load the workspace, see the onboarding flow  
- [X] Manual test: ON an empty HEAD changeset, click Initialize your workspace, follow the onboarding flow. You'll be brought to an empty HEAD. Do a full tab reload to see the components


app/web/src/newhotness/Workspace.vue
```diff
@@ -91,10 +80,26 @@
         :loadingSteps="_.values(muspelheimStatuses)"
       />
     </Transition>
-
-    <main v-if="lobby && tokenFail" class="grow min-h-0">
-      <div>Bad Token</div>
-    </main>
+    <template v-if="!lobby">
```
jobelenus, line 83: OK, so the expected behavior for how this is written is:
1. user comes to page
2. hits lobby
3. hits onboarding

Just wanted to confirm thats what we all decided on 😅

jobelenus: OK, i see how the computed `lobby`changed. Now it will be:
1. lobby = false, onboarding
2. lobby = true, in lobby
3. lobby = false, out of lobby

app/web/src/newhotness/Workspace.vue
```diff
@@ -387,6 +376,65 @@ startWindowResizeEmitter(window);
...
+    tokenFail.value = true;
+    return;
+  }
+
+  loadedOnboardingStateFromApi.value = true;
+  // firstTimeModal === true means that we should SHOW it, so we need to invert it to see if it's been dismissed.
+  onboardingCompleted.value = !data.firstTimeModal;
+};
+
+onMounted(checkOnboardingCompleteData);
```
jobelenus: i think this should go into the `onBeforeMount` with the muspelheim call? just to get it fired as early as possible

app/web/src/newhotness/api_composables/index.ts
```diff
@@ -219,19 +239,24 @@ export class APICall<Response, Args> {
   url(): string {
     if (
       [
-        _routes.Workspaces,
-        _routes.AbandonChangeSet,
-        _routes.CreateChangeSet,
-      ].includes(this.path)
```
jobelenus, line 229: why change from checking the path to checking the key?

jobelenus (changes requested):
Just a few Qs in there

app/web/src/newhotness/api_composables/index.ts
vbustamante: The routes that have ids get replaced before this step so matching route does not work for them

app/web/src/newhotness/Workspace.vue
vbustamante: First comment is right user hits page, gets lobby until:cold boot is done, feature flags are loaded And on-boarding completion data is loaded. That is so we don't blink anything. Lobby blink time was also updated so we don't blink the lobby if this takes less than 2 seconds

app/web/src/newhotness/Workspace.vue
vbustamante: Sounds like a good idea, yeah

jobelenus (approved):

PR #7022, opened 2025-08-21
Ensure Actions that are dispatched show as running 

britmyerss:
2 Changes: 
1. We missed one of the places where the front end was looking for actions in the `Running` state which is no longer in use
2. We recently modified the Actions Job to no longer modify the graph to capture the action’s now-running state. We planned on exposing `Dispatched` as the user indicator that the action is in flight, however because the dispatch happens after the rebase request has been processed (and Edda has been informed), we need to again, tell Edda about the 
change to the graph that the Rebaser does directly in post processing.

## How was it tested?

- [X] Manual test: Enqueue an action and see that it shows as running when it is dispatched!
- Also see that it successfully is removed on completion, and shows as failed when it fails

<div><br/>via <a href="https://giphy.com/gifs/reactiongifs-h2OLfcSKKthRK">GIPHY</a></div>

### DIFF
--- a/app/web/src/newhotness/Explore.vue
+++ b/app/web/src/newhotness/Explore.vue
@@ -1082,7 +1082,7 @@ const componentsHaveActionsWithState = computed(() => {
     if (!action.componentId) continue;
     if (action.state === ActionState.Failed) {
       results.failed.add(action.componentId);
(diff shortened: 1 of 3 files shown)


fnichol (approved):
Yep!

PR #7025, opened 2025-08-22
feat(si-mcp-server): Add initial posthog tracking to si-mcp-server and a Readme

britmyerss:
This change adds basic posthog tracking to the mcp server. We’ll have events for:
- Server start/end
- Each time a tool is invoked
- Each time an error occurs during tool invocation

We’re not threading much data through yet, but this gives us some initial visibility.

I've also added a readme with some basic info to help with the dev loop. 

#### Screenshots:
image

#### Out of Scope:
Enriched tracking- for now I've added simple events but the plumbing is there if we wanted to track more specific interactions or activity

## How was it tested?
Use the MCP and see the events show up in Posthog!

<div><br/>via <a href="https://giphy.com/gifs/beach-hedgehog-211e4CYjahC2Q">GIPHY</a></div>

### DIFF
--- a/bin/si-mcp-server/src/analytics.ts
+++ b/bin/si-mcp-server/src/analytics.ts
@@ -1,0 +1,84 @@
+import { PostHog } from "posthog-node";
+import { USER_ID, WORKSPACE_ID } from "./si_client.ts";
+
+export class Analytics {
+  private posthog: PostHog | null = null;
+  private sessionId: string;
+
+  constructor() {
+    this.sessionId = crypto.randomUUID();
+    this.initializePostHog();
+  }
+
+  private initializePostHog() {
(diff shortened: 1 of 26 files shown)


bin/si-mcp-server/src/analytics.ts
```diff
@@ -0,0 +1,84 @@
...
+  private posthog: PostHog | null = null;
+  private sessionId: string;
+
+  constructor() {
+    this.sessionId = crypto.randomUUID();
+    this.initializePostHog();
+  }
+
+  private initializePostHog() {
+    const apiKey = Deno.env.get("POSTHOG_API_KEY") || "phc_KpehlXOqtU44B2MeW6WjqR09NxRJCYEiUReA58QcAYK"; // Prod Posthog
```
stack72, line 14: this is 100% safe to check in as it's a public key!!

PR #7029, opened 2025-08-22
feat: improve map view experience

johnrwatson:
### This PR does the following
- **Adds component selection tracking to the map redraw:** If a user has component A selected, on map redraw that user will continue to see component A selected, previously it dropped the selection
- **Adds pan consistency to map redraw:** If a user is looking at component A, map redraws caused by external events will keep the current viewport offset from component A, previously it dropped the pan/reset
- **Adds animation to added/removed components and edges on map redraw:** If new components are added or components removed, sister sessions from other users (or the same user) show smooth redraws/smooth pans for map rejigging, previously it 1-framed / harshly added components which was impossible to follow

### Testing Steps
All the permutations!
- If a user has multiple components selected and  one of those is deleted by another user, the remaining components should remain selected
- If a user has `all` it's selected components by a third party, the map/pan position should remain static during redraw
- If a user has an incoming or outgoing edge added to a selected component the viewport must remain passive/no-impact for user's view
- Add a component while looking at the map
- Have the map open while switching change sets
- Delete components while looking at the map, with and without filters
- Add & remove subscriptions while looking at the map, and with filters.

<div></div>

### DIFF
--- a/app/web/src/newhotness/Explore.vue
+++ b/app/web/src/newhotness/Explore.vue
@@ -1720,16 +1720,6 @@ watch(searchString, () => {
   });
 });
 
-// Clear the selection when the filter or search string changes
-// TODO leave the selection as long as it is still one of the filtered components?
-watch(filteredComponents, () => {
-  if (mapRef.value && typeof mapRef.value.deselect === "function") {
-    mapRef.value.deselect();
-  }
-  // TODO: when the underlying componentList changes, the indexes will change
-  // which means the selected indexes could have moved in either direction
-});
-
 // this is so that when on the map view, if you have a component selected
 // and start searching, we clear the selected component
 // on grid view, this has no impact, because you can't focus on the search box if you've got component(s) selected
--- a/app/web/src/newhotness/Map.vue
+++ b/app/web/src/newhotness/Map.vue
@@ -284,6 +284,25 @@ const componentContextMenuRef =
   ref<InstanceType<typeof ComponentContextMenu>>();
 
 const selectedComponents = ref<Set<ComponentInList>>(new Set());
+// Store selected component IDs to maintain selection across component updates
+const selectedComponentIds = computed(
+  () => new Set(Array.from(selectedComponents.value).map((c) => c.id)),
+);
+
+// Watch for componentsById changes and update selectedComponents with new object references
+watch(componentsById, (newComponentsById) => {
+  if (selectedComponentIds.value.size > 0) {
+    const newSelectedComponents = new Set<ComponentInList>();
+    selectedComponentIds.value.forEach((componentId) => {
+      const component = newComponentsById[componentId];
+      if (component) {
+        newSelectedComponents.add(component);
+      }
+    });
+    selectedComponents.value = newSelectedComponents;
+  }
+});
+
 const showMinimap = ref((props.components?.length ?? 0) > 0);
 
 watch(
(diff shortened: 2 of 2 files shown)


app/web/src/newhotness/Map.vue
```diff
@@ -457,6 +474,283 @@ const smoothPan = (totalDx: number, totalDy: number, duration = 400) => {
   requestAnimationFrame(animate);
 };
 
+// Viewport preservation functions
+const preserveViewportCenter = () => {
+  const scale = transformMatrix[0] ?? 1;
+  const translateX = transformMatrix[4] ?? 0;
+  const translateY = transformMatrix[5] ?? 0;
```
nickgerace, line 484: Why `4` and `5`? I read the description but am curious to how this links to X and Y.

app/web/src/newhotness/Map.vue
```diff
@@ -457,6 +474,283 @@ const smoothPan = (totalDx: number, totalDy: number, duration = 400) => {
...
+
+// Smooth node transition functions
+const animateNodePositions = (oldNodes: layoutNode[], newNodes: layoutNode[], duration = 600) => {
+  const startTime = performance.now();
+  const nodeMap = new Map(oldNodes.map(n => [n.id, n]));
+  
+  const animate = (currentTime: number) => {
+    const elapsed = currentTime - startTime;
+    const progress = Math.min(elapsed / duration, 1);
+    const easeProgress = 1 - (1 - progress) ** 3; // ease-out cubic
```
nickgerace: Smart as hell

app/web/src/newhotness/Explore.vue
```diff
@@ -1707,14 +1707,38 @@ watch(searchString, () => {
   });
 });
 
-// Clear the selection when the filter or search string changes
-// TODO leave the selection as long as it is still one of the filtered components?
-watch(filteredComponents, () => {
```
jobelenus, line 1725: i was like 99% sure we don't need this watcher at all... so this gives us a good reason to better understand that we can remove it

app/web/src/newhotness/Map.vue
```diff
@@ -283,6 +283,23 @@ const componentContextMenuRef =
   ref<InstanceType<typeof ComponentContextMenu>>();
 
 const selectedComponents = ref<Set<ComponentInList>>(new Set());
+// Store selected component IDs to maintain selection across component updates
+const selectedComponentIds = ref<Set<string>>(new Set());
+
+// Watch for componentsById changes and update selectedComponents with new object references
+watch(componentsById, (newComponentsById) => {
```
jobelenus, line 293: i think the better way to do this is:
1. instead of mutating `selectedComponentIds`, mutate `selectedComponents`
2. `selectedComponentIds` should be computed based on `selectedComponents`

app/web/src/newhotness/Map.vue
jobelenus: its the format for the matrix math transformation

app/web/src/newhotness/Map.vue
```diff
@@ -457,6 +474,283 @@ const smoothPan = (totalDx: number, totalDy: number, duration = 400) => {
...
+    const componentId = Array.from(element.classList)
+      .find(cls => cls.startsWith("id-"))
+      ?.substring(3);
+    
+    if (componentId) {
+      const hasFailedActions = props.componentsWithFailedActions.has(componentId);
+      if (hasFailedActions) {
+        element.classList.add("failed-actions");
+      } else {
+        element.classList.remove("failed-actions");
```
jobelenus, line 719: glad to see you found a spot to get this add/remove working on update 🫡

app/web/src/newhotness/Map.vue
johnrwatson: @nickgerace - It's to do with the 2D transformation matrix.

The indices 4 and 5 are the translation components in a 2D transformation matrix.

A 2D transformation matrix is represented as a 6-element array that maps to this structure:

[a, b, c, d, e, f]

Which represents the mathematical transformation matrix:
| a  c  e |
| b  d  f |
| 0  0  1 |

Where:
- a (index 0) = horizontal scaling
- b (index 1) = vertical skewing
- c (index 2) = horizontal skewing
- d (index 3) = vertical scaling
- e (index 4) = horizontal translation (translateX)
- f (index 5) = vertical translation (translateY)

So basically transformMatrix[4] gets the X translation and transformMatrix[5] gets the Y translation.

app/web/src/newhotness/Map.vue
johnrwatson: ahhhh this is smart, yes absolutely the better approach

stack72 (approved):

app/web/src/newhotness/Map.vue
britmyerss: This looks like it'd be helpful docs to keep in here somewhere if we don't have it already!

PR #7032, opened 2025-08-22
Separate publish errors for pending events

nickgerace:
## Description

This change separates publish error bubbling based on the kind of publish attempt made.

## Testing

The compiler and integration test suite covers any possible failure this would cause.


fnichol (approved):
I see, yes, this should help us to understand if there's an error, which of the 4 cases we have, nice!

PR #7040, opened 2025-08-22
continued work on the review screen

wendybujalski:
- Show secrets on the review screen - subscriptions show as usual, secret values are hidden
- A filter to prevent components from appearing on the review screen if they have no AV diffs.
- Not showing hidden AVs
- Hiding parents whose children are all hidden
- Hiding parent diff values when they are shown within children
- Removing the revert button for children of a parent with a subscription
- Fixed some styles and overflow/truncation issues

See ENG-3197 to see remaining work on the review screen.


nickgerace (approved):

stack72:
Broken test in this one

PR #7044, opened 2025-08-23
Ignore Specific Luminork MVs

jobelenus:

Make a list of `EntityKind` (MV) that the front end will not interact with. Don't patch it, don't throw hammers for it, don't allow lookups (e.g. `get`).


app/web/src/workers/types/entity_kind_types.ts
```diff
@@ -44,6 +44,9 @@ export enum EntityKind {
   View = "View",
   ViewComponentList = "ViewComponentList",
   ViewList = "ViewList",
+  // IGNORING THESE
+  LuminorkDefaultVariant = "LuminorkDefaultVariant",
```
stack72, line 48: Nice work! We can extend that easily now :)

PR #7047, opened 2025-08-23
feat(edda): Additional Diff MVs and Secret Data

britmyerss:

In order to correctly populate the Review screen, we need additional pieces of information:

`ErasedComponents` provides a map of components that exist on head but do not exist in this change set. Because we replay changes to head on open change sets, any change to components on head will trigger this MV to build. If you’ve erased a component in this change set, but it still exists on head, we’ll provide the same ComponentDiff, with all data populated on head. 

`ActionDiffList` provides a list of Actions and their diff status: 
- If the state has been changed in this change set (for example, putting an action on hold)
- If the action is new in this change set
- If the action has been removed in this change set

`ComponentDiff` now includes information specific to Secrets so we can correctly show when changes have been made regarding secrets. Note: We aren’t diffing the actual encrypted secret data, but whenever the secret data changes, we always end up with a new secret so we can show when non-encrypted data changes (name/description) as well as show when subscriptions to secrets change

I also fixed a bug where we were using the wrong `DalContext` when computing the `ComponentDiff`

## How was it tested?

Manual testing - all MVs look as expected on the front end 
I need to add automated tests still but wanted to get this out there to build upon

<div><br/>via <a href="https://giphy.com/gifs/reaction-different-im-hWl65BEtMP5DO">GIPHY</a></div>

### DIFF
--- a/lib/dal-materialized-views/src/component/component_diff.rs
+++ b/lib/dal-materialized-views/src/component/component_diff.rs
@@ -5,6 +5,7 @@ use dal::{
     DalContext,
     Func,
     Prop,
+    Secret,
     attribute::{
         prototype::argument::{
             AttributePrototypeArgument,
@@ -31,6 +32,7 @@ use si_id::{
     AttributeValueId,
     ComponentId,
 };
+use telemetry::prelude::*;
 
 /// Generates a [`ComponentDiff`] MV.
 pub async fn assemble(new_ctx: DalContext, id: ComponentId) -> crate::Result<ComponentDiff> {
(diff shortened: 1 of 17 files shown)


lib/dal-materialized-views/src/component/component_diff.rs
```diff
@@ -429,7 +482,7 @@ async fn child_av_pairs(
         // not detect whether two fields are the same.
         let old_av_id = match old_children.peek() {
             Some(&old_av_id) => {
-                let old_key = AttributeValue::key_for_id(new_ctx, old_av_id).await?;
+                let old_key = AttributeValue::key_for_id(old_ctx, old_av_id).await?;
```
britmyerss, line 495: this is what caused the bug where removing a subscription and adding an manual entry wouldn't show correctly

lib/dal-materialized-views/src/component/erased_components.rs
```diff
@@ -0,0 +1,47 @@
...
+use dal::{
+    Component,
+    ComponentId,
+    DalContext,
+};
+use si_frontend_mv_types::component::erased_components::ErasedComponents;
+use telemetry::prelude::*;
+
+#[instrument(
+    name = "dal_materialized_views.component_list",
```
nickgerace, line 15: We should change this. It doesn't have to be in this PR, but IF IT ISN'T!!!... you are on the hook 😄

lib/si-events-rs/src/action.rs
```diff
@@ -44,6 +54,7 @@ pub enum ActionState {
     OnHold,
     /// Action is available to be dispatched once all of its prerequisites have succeeded, and been
     /// removed from the graph.
+    #[default]
```
nickgerace, line 57: Likely fine, but let's keep an eye out.

PR #7062, opened 2025-08-24
feat(web): Add a header to the reviews page

stack72:
That header will have left and right buttons and show the selected 
component number

That header will have left and right buttons and show the selected 
component number

Also ensure that we bind the left and right arrow keys to navigate 
between the components and then that we can deselect the components and 
remove from the url

Screenshot 2025-08-25 at 00 28 43

### DIFF
--- a/app/web/src/newhotness/Review.vue
+++ b/app/web/src/newhotness/Review.vue
@@ -1,145 +1,101 @@
 <template>
-  <section
-    v-if="ctx.onHead.value"
-    class="p-lg flex flex-col items-center gap-sm"
-  >
-    <EmptyState
-      icon="x"
-      text="You are on HEAD"
-      secondaryText="There are no changes to review"
-    />
-    <VButton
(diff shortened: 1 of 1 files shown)


wendybujalski (dismissed):

app/web/src/newhotness/Review.vue
```diff
@@ -1,271 +1,360 @@
...
-    />
-    <VButton
-      label="Exit"
-      tone="neutral"
-      icon="chevron--left"
-      @click="exitReview"
-    />
-  </section>
-  <section v-else class="grid review w-full h-full">
+  <div class="w-full h-full flex flex-col">
```
wendybujalski, line 2: Would be better to do with grid instead of flex, but this works.

stack72: We can circle back around afterward - good shout!

wendybujalski (approved):

PR #7068, opened 2025-08-25
fix: Show that a component is ghosted in review

jkeiser:
This shows a header disclaiming that a component is set to be removed. It also sets the icon type on the side to "Removed."

image

## Out of Scope

* "Restore" button will follow, but likely not tonight.
* If the component is also ghosted on head (like, with a failed action), we still show the component, even though its "ghosted" state was not changed in the changeset. @britmyerss's upcoming mv will make this possible to show.

## How was it tested?

- [X] Manual test: ghosted components (with and without changes) show disclaimer/icon
- [X] Manual test: other components work as before.

### DIFF
--- a/app/web/src/newhotness/Review.vue
+++ b/app/web/src/newhotness/Review.vue
@@ -185,6 +185,22 @@
           </template>
 
           <div class="flex flex-col gap-xs p-xs">
+            <div
+              v-if="selectedComponent.toDelete"
+              :class="
+                clsx(
+                  'flex flex-col gap-xs p-sm border',
+                  themeClasses(
+                    'text-neutral-800 border-neutral-400 bg-neutral-100',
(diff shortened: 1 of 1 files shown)


app/web/src/newhotness/Review.vue
```diff
@@ -443,22 +459,19 @@ const componentList = computed(() => {
...
         componentDiff,
         attributeDiffTree,
       };
     })
-    .filter(
-      (component) =>
-        component.attributeDiffTree.children &&
-        Object.keys(component.attributeDiffTree.children).length > 0,
-    );
+    .filter((component) => component.diffStatus !== "None");
```
britmyerss, line 474: one thing we might want to check here @jkeiser is if you've got a component with no attribute changes, but only a newly enqueued action, that it's not filtered out.

britmyerss (approved):

PR #7072, opened 2025-08-25
fix(web): Ensure we have consistent spacing in the Review AttributePanel

stack72:


jobelenus (dismissed):

nickgerace (approved):

app/web/src/newhotness/ReviewAttributeItemSourceAndValue.vue
```diff
@@ -50,7 +50,7 @@
       v-else
       :class="
         clsx('py-2xs min-w-0', {
-          'line-through': !!($secret || $value),
+          'line-through': !!old && !!($secret || $value),
```
wendybujalski, line 53: old is already a boolean so the double ! is unnecessary

PR #7078, opened 2025-08-25
review screen polish items pt 1

wendybujalski:
- Values inside of each AV diff can now display on multiple lines instead of truncating
- ComponentHistory now has a loader and a better empty state
- Fixed consistency of headers across `Review` and `ComponentDetails`
- Added truncation to the header of the middle section of `Review`
- Fixed truncation balance in a few places

### DIFF
--- a/app/web/src/newhotness/ComponentListItem.vue
+++ b/app/web/src/newhotness/ComponentListItem.vue
@@ -16,11 +16,20 @@
     <TextPill
       tighter
       variant="component"
-      :class="themeClasses('text-green-light-mode', 'text-green-dark-mode')"
+      :class="
+        clsx(
+          themeClasses('text-green-light-mode', 'text-green-dark-mode'),
+          'max-w-fit flex-1',
+        )
+      "
     >
       <TruncateWithTooltip>{{ component.schemaName }}</TruncateWithTooltip>
     </TextPill>
-    <TextPill tighter variant="component" class="text-purple">
+    <TextPill
+      tighter
+      variant="component"
+      :class="clsx('text-purple max-w-fit flex-1')"
+    >
       <TruncateWithTooltip>{{ component.name }}</TruncateWithTooltip>
     </TextPill>
     <StatusIndicatorIcon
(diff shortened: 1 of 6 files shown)


app/web/src/newhotness/ComponentListItem.vue
```diff
@@ -16,11 +16,20 @@
     <TextPill
       tighter
       variant="component"
-      :class="themeClasses('text-green-light-mode', 'text-green-dark-mode')"
+      :class="
+        clsx(
+          themeClasses('text-green-light-mode', 'text-green-dark-mode'),
+          'max-w-fit flex-1',
```
wendybujalski, line 22: `max-w-fit` prevents this item from growing larger than its contents and `flex-1` allows it to grow or shrink.

Together these two classes allow multiple TruncateWithTooltip elements to share a flex row and distribute the space between them properly as needed.

app/web/src/newhotness/ReviewAttributeItemSourceAndValue.vue
```diff
@@ -29,27 +29,25 @@
...
-        </TruncateWithTooltip>
+        </div>
       </div>
     </AttributeValueBox>
-    <TruncateWithTooltip
+    <div
       v-else
       :class="
-        clsx('py-2xs min-w-0', {
+        clsx('py-2xs min-w-0 break-words', {
```
jkeiser, line 50: Just so I understand, what will happen if there are really big words? I can see this happening for text fields that are base 64, for example.

app/web/src/newhotness/ReviewAttributeItemSourceAndValue.vue
```diff
@@ -82,8 +80,4 @@ const props = defineProps<{
 const $source = computed(() => props.sourceAndValue.$source);
 const $value = computed(() => props.sourceAndValue.$value);
 const $secret = computed(() => props.sourceAndValue.$secret);
-
-const emit = defineEmits<{
```
jkeiser, line 86: Yay for removing things!

app/web/src/newhotness/ReviewAttributeItemSourceAndValue.vue
wendybujalski: they will break!

PR #7081, opened 2025-08-25
fix(dal): treat "empty" values as None in dependent_values_update

zacharyhamm:
1. When comparing the values before and after execution, use `unprocessed_value` to ensure we are looking at full arrays or objects.

2. Ensure that if the value before function execution was None, it is not treated as "changed" if the value afterwards is the empty object, empty array, or blank string. Otherwise just compare the value.

3. Also makes it so we do not write the audit log unless the value has changed, according to the above rule.

Addresses BUG-981


stack72 (dismissed):

zacharyhamm:
had to fix up the tests to reflect the new behavior

stack72 (approved):

PR #7085, opened 2025-08-25
fix(frigg): Temporarily bypass deployment MV index definition checksum validation

jhelwig:
Currently, deployment-level MVs (which take 2+ hours to rebuild) and change set-level MVs (which change frequently) share the same definition checksum, causing unnecessary rebuilds of deployment indexes.

This is a temporary workaround until we can decouple the checksums between these two levels.


jhelwig:
@stack72 This should only need to be deployed to luminork as it should be the only thing using the `get_deployment_index` method.

PR #7092, opened 2025-08-25
fix(web): terminate webworker if app is out of date

zacharyhamm:
If we are showing the CachedAppNotification, the user needs to refresh. This could mean that the webworker or shared worker is out of date. But the out of date webworker might be holding the lock! If it is, new windows that are not out of date are unlikely to work. So just terminate the webworker, which will release the lock and let a healthy tab pick it up.

To make this work, this also makes it impossible to dismiss the CachedAppNotification modal without refreshing the page.

I tested this by having two tabs open. In the tab with the lock, I forced CachedAppNotification to trigger (with code changes). The lock was released in the tab with the modal, and acquired by the other tab.


jobelenus (approved):
Elegant

PR #7094, opened 2025-08-25
fix(sdf-server): Ensure we thread through the correct componentId to action audit log events

stack72:


lib/sdf-server/src/service/v2/action.rs
```diff
@@ -302,6 +330,7 @@ pub async fn retry(
             func_id,
             func_display_name: func.display_name,
             func_name: func.name.clone(),
+            component_id: Some(component_id),
```
jkeiser: Confused how this one is compiling? Action::component_id() (correctly) returns an Option already :)

stack72: You are correct!!

jkeiser (dismissed):

jkeiser (approved):

PR #7101, opened 2025-08-26
Ensure patches wait for cold starts. Tighten up logging

jobelenus:

When cold starting many change sets, any one of them finishing doesn't mean we can start processing patches.
**PSA:** Putting all web worker logs behind the `_DEBUG` (false everywhere, true for local) flag so we don't flood users with information

## How was it tested?

General app usage

### DIFF
--- a/app/web/src/workers/webworker.ts
+++ b/app/web/src/workers/webworker.ts
@@ -126,8 +126,6 @@ registerInstrumentations({
 
 const tracer = trace.getTracer("bifrost");
 // eslint-disable-next-line no-console
-const log = console.log;
-// eslint-disable-next-line no-console
 const error = console.error;
 const _DEBUG = import.meta.env.VITE_SI_ENV === "local";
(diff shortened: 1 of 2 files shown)


app/web/src/workers/webworker.ts
```diff
@@ -1205,176 +1213,189 @@ const applyWorkspacePatch = async (
...
-    // Check if we actually have the atom data, not just the MTM relationship
-    const upToDateAtomIndexes = workspaceAtomExistsOnIndexes(
-      db,
-      atom.kind,
-      atom.id,
-      atom.toChecksum,
-    );
-    if (upToDateAtomIndexes.length > 0) {
-      patchRequired = false;
+    try {
```
jobelenus, line 1216: **PSA**: all this diff stuff is just simply indenting for a try/catch/finally

britmyerss (approved):

PR #7107, opened 2025-08-26
feat: Serialize object/array when setting JSON fields

jkeiser:
Claude is sometimes setting policy documents as objects (the deserialized form of the JSON). This makes them fail, because policy documents are supposed to be JSON strings. This is a fairly intuitive thing to try when using the API:

```json
{
  "/domain/AssumeRolePolicyDocument": {
    "Version": "2012-10-17",
    "Statement": [
      {
        "Effect": "Allow",
        "Principal": { "Service": "elasticmapreduce.amazonaws.com" },
        "Action": "sts:AssumeRole"
      }
    ]
  }
}
```

This PR will accept these inputs, serializing objects or arrays to JSON when the user tries to set them to a JSON field.

## How was it tested?

- [X] Integration tests pass
- [X] Manual test: can still read and write Bucket and Role policy documents in the UI
- [X] New integration tests for object/array/string
- [X] Integration test validating that setting JSON field to string works as before (expects serialized JSON and does not serialize it itself)

### DIFF
--- a/lib/dal/src/attribute/attributes.rs
+++ b/lib/dal/src/attribute/attributes.rs
@@ -238,9 +238,9 @@ async fn get_before_prop_value_source(
                 Some(resolved_av_id) => {
                     let source_component_id =
                         AttributeValue::component_id(ctx, resolved_av_id).await?;
-                    let view = AttributeValue::view(ctx, resolved_av_id).await?;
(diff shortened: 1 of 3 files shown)


zacharyhamm (approved):

zacharyhamm:
@stack72 keiser is fixing a test issue

PR #7118, opened 2025-08-26
bugs 990 and 949

wendybujalski:

Solves BUG-990 by adding splash screens for various scenarios where the user cannot access the workspace.

Solves BUG-949 by fixing the code which accounts for scrolling in the Grid view.

#### Screenshots:

### DIFF
--- a/app/web/src/newhotness/EmptyState.vue
+++ b/app/web/src/newhotness/EmptyState.vue
@@ -12,15 +12,46 @@
     >
       <Icon :name="icon" :size="iconSize" />
     </div>
+    <!-- Primary text should be a concise title -->
     <div :class="themeClasses('text-neutral-800', 'text-neutral-200')">
-      {{ text }}
+      <slot>
+        {{ text }}
+      </slot>
     </div>
+    <!-- Secondary text can be a longer explanation and/or a link -->
     <slot name="secondary">
       <div
         v-if="secondaryText"
-        :class="themeClasses('text-neutral-600', 'text-neutral-400')"
+        :class="
+          clsx(
+            secondaryClickable && [
+              'cursor-pointer hover:underline',
+              themeClasses('hover:text-action-500', 'hover:text-action-300'),
+            ],
+            themeClasses('text-neutral-600', 'text-neutral-400'),
+          )
+        "
+        @click="secondaryClick"
       >
         {{ secondaryText }}
+      </div>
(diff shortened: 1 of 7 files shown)


app/web/src/newhotness/EmptyState.vue
wendybujalski, line 1: Cleaned up EmptyState a bit for this PR.

app/web/src/newhotness/Explore.vue
```diff
@@ -286,14 +286,10 @@
...
-              clsx(
-                'grow',
-                bulkEditing && 'min-h-0',
-                !bulkEditing && 'scrollable',
-              )
-            "
+            :class="clsx('grow', bulkEditing ? 'min-h-0' : 'scrollable')"
             :style="!bulkEditing && 'overflow-anchor: none;'"
+            @scroll="onScroll"
+            @scrollend="fixContextMenuAfterScroll"
```
wendybujalski, line 292: These two event handlers were in the wrong spot, causing necessary side effects of scrolling on the grid to not work!

app/web/src/newhotness/Explore.vue
```diff
@@ -1851,7 +1845,12 @@ const fixContextMenu = async () => {
 const clearSelection = () => {
   selectedComponentIndexes.clear();
   bulkEditing.value = false;
-  unfocus();
+  if (focusedComponentIdx.value === -1) {
+    setFocusedComponentIdx(-1);
+    componentContextMenuRef.value?.close();
+  } else {
+    unfocus();
+  }
```
wendybujalski, line 1853: This adjustment prevents the focusedComponentIdx from being set to undefined when the search is clicked. This caused a bug where the Tab key would not work the first time after clicking the search.

app/web/src/newhotness/explore_grid/ExploreGrid.vue
```diff
@@ -194,7 +194,7 @@ const scrollCurrentTileIntoView = () => {
   );
 };
 
-watch([() => exploreContext.focusedComponentIdx], scrollCurrentTileIntoView);
+watch([exploreContext.focusedComponentIdx], scrollCurrentTileIntoView);
```
wendybujalski, line 197: This watcher was not firing properly.

jobelenus (approved):

PR #7122, opened 2025-08-26
Ensure both v1 and v2 SDF API endpoints consistently trigger and wait for index update on change set creation

jhelwig:
While the v1 endpoint was triggering the new change set message for Edda, only the v2 endpoint was actually waiting for the index update to happen.

The method used by v2 to do the trigger-and-wait is now in sdf-core where it can be used by both v1 and v2 endpoints.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7484,6 +7484,7 @@ dependencies = [
  "edda-client",
  "edda-core",
  "frigg",
+ "futures-lite",
  "hyper 0.14.32",
  "nats-multiplexer-client",
  "once_cell",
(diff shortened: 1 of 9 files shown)


Cargo.lock
```diff
@@ -7484,6 +7484,7 @@ dependencies = [
  "edda-client",
  "edda-core",
  "frigg",
+ "futures-lite",
```
nickgerace, line 7487: This did not add a new dep to the codebase so this does not require a full `support/update-cargo-deps` (or whatever it is called). This and the deps changes look good!

nickgerace (approved):

PR #7127, opened 2025-08-27
Display disconnected WS status for a given workspace

jobelenus:
## Screenshots:

Note: it pulses to indicate retrying.

image

## How was it tested?

1. Mocked up a timeout after the WS connects, that closes the WS. Displays the message.
2. Open up two tabs, make sure the one web worker communicates to both tabs.

Unfortunately the dev tools "offline" doesn't close down WS connections 🤷


stack72 (approved):
This is great! thank you

stack72 (approved):

PR #7132, opened 2025-08-27
feat(web): Link to review from audit when not on HEAD

stack72:


app/web/src/components/layout/navbar/NavbarPanelCenter.vue
```diff
@@ -24,6 +24,16 @@
     />
 
     <NavbarButton
+      v-if="!onHead"
+      tooltipText="Audit"
+      icon="eye"
+      :selected="route.matched.some((r) => r.name === 'new-hotness-review')"
```
jobelenus: i think you're in the wrong component. There is `newhotness/nav/NavbarPanelCenter`

jobelenus: this will never be `true` in the old navbar

PR #7141, opened 2025-08-27
Application level keep-alive on the bifrost websocket

jobelenus:

In tools-prod/prod the LBs enforce an idle time. Every minute. So each websocket was opening, and if there were no MVs going across the wire, it would be shut down a minute later. And the client would reconnect.

By sending data, 3 times a minute, at least, we won't get shut down by the LB.

## How was it tested?

Locally we can see with log messages that the SDF web socket is receiving data.

### DIFF
--- a/lib/sdf-v1-routes-ws/src/bifrost/proto.rs
+++ b/lib/sdf-v1-routes-ws/src/bifrost/proto.rs
@@ -305,12 +305,8 @@ impl BifrostStarted {
                                 handle: self.handle,
                             });
                         }
-                        // Received unexpected web socket message type
-                        Some(Ok(unexpected_message)) => {
-                            warn!(
-                                message = ?unexpected_message,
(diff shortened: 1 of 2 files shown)


lib/sdf-v1-routes-ws/src/bifrost/proto.rs
```diff
@@ -305,12 +305,8 @@ impl BifrostStarted {
                                 handle: self.handle,
                             });
                         }
-                        // Received unexpected web socket message type
-                        Some(Ok(unexpected_message)) => {
```
britmyerss, line 309: Do we want to remove this?

jobelenus: `cargo check` told me that was unreachable... apparently `Binary` was the only one missing from the match arms!

britmyerss (approved):

PR #7145, opened 2025-08-27
Restore luminork API tests with retry loop

nickgerace:
## Description

This change restores the luminork API tests, but does so with a retry loop with backoff semantics. This is not ideal as we would perfect to retry individual steps, but we've invested an abundance of time there to little effect. This retry loop may be a big hammer, but it should help the original spirit of the tests: "did we regress luminork" and "is it still online"?


nickgerace:
Note the ~5m run. The retry logic worked!

Screenshot 2025-08-27 at 5 21 07 PM

PR #7146, opened 2025-08-27
chore(cleanup): Cleanup old SDF Routes no longer used

stack72:
This is thanks to Nick’s great investigation work!


nickgerace (approved):
Approved provided tests pass!

lib/sdf-server/src/service/v2/admin/migrate_connections.rs
```diff
@@ -260,7 +260,6 @@ async fn migrate_connection(
             // inferred connections.
             if migration.explicit_connection_id.is_none() {
                 info!("Making inferred socket connection explicit due to issues");
```
nickgerace, line 262: We should just delete this block right?

stack72: No we should keep it for now as information about any sockets that needed migrated

PR #7148, opened 2025-08-27
chore: Make it impossible to create new socket connections

jkeiser:
This PR removes the ability to create socket connections from System Initiative (even through APIs or management functions). This will allow us to safely migrate all remaining connections and be sure they no longer exist. Summary of changes:

* Management functions no longer honor connect or disconnect operations
* Luminork create/update component no longer create connections (but update can remove them still)
* Component upgrade and regenerate variant remove socket connections instead of carrying them over
* Old UI: override_connection and create_connection endpoints removed. UI not updated, and will return an error if you try.
* Old UI: copy/paste no longer copies socket connections.

## Approach

Removed `Component::connect()` and stuff that used it. (Actually renamed to `Component::connect_for_tests()`, because I still want to be able to *remove* connections, and that means tests have to be able to create them first!

## How was it tested?

- [X] Integration tests pass
- [X] Manual test: can still create, delete, update, duplicate, subscribe to components
- [X] Manual test: management functions still work as long as you don't pass connect/disconnect operations


lib/dal/src/management/mod.rs
```diff
@@ -277,7 +264,6 @@ pub struct ManagementUpdateOperation {
     properties: Option<serde_json::Value>,
     attributes: Option<AttributeSources>,
     geometry: Option<HashMap<String, ManagementGeometry>>,
-    connect: Option<ManagementUpdateConnections>,
```
zacharyhamm, line 280: Probably fine, but this will make old mgmt funcs fail if they used connections. Dunno if we have any in the wild

lib/dal-test/src/test_exclusive_schemas/legos/small.rs
```diff
@@ -210,171 +210,6 @@ pub(crate) async fn migrate_test_exclusive_schema_small_odd_lego(
         update_in_view_mgmt_func_name,
     )?;
 
-    let create_and_connect_from_self_func_code = r#"
-    async function main({ thisComponent, components }: Input): Promise<Output> {
```
britmyerss, line 214: do we have coverage for this same scenario only using subscriptions? I think so, but wasn't sure if you confirmed

lib/dal/src/component.rs
```diff
@@ -2146,7 +2146,7 @@ impl Component {
     }
 
     #[instrument(level = "debug", skip(ctx))]
-    async fn create_new_connection(
+    async fn create_new_connection_for_tests(
```
britmyerss, line 2149: this is a great approach!

lib/dal/src/component.rs
```diff
@@ -2964,342 +2966,6 @@ impl Component {
         Ok(None)
     }
 
-    pub async fn autoconnect(
```
britmyerss, line 2967: 🥹

zacharyhamm:
This looks pretty safe to me. Just one quick question: Your PR description says the create_connection and override_connection endpoints are removed in this PR, but I can't find that to review it

britmyerss (commented):
This looks good! Let's confirm the timing on merging this, but otherwise it's good!

lib/dal/src/component.rs
jkeiser: Heh, this one was unused anywhere :)

lib/dal-test/src/test_exclusive_schemas/legos/small.rs
jkeiser: Yep, we have coverage for subscription scenarios in management.rs; I've been writing management functions in among the actual tests so you can see what's being tested easier.

lib/dal/src/management/mod.rs
jkeiser: Yeah, I tested that connect operations fail (the biggest risk to me is *silent* failures--if the user tells us to do something, we shouldn't just ignore it, we should fail). Failures show up reasonably in the UI, and the github action reports failure (though you have to click into the UI to see the logs).

image

jkeiser: ... actually, I think this will silently ignore `connect` when it hangs off an `update` operation, so that one must have slipped by my test! Unless you put `#[serde(deny_unknown_fields)]` on your struct, it will happily let you misspell or pass extra stuff, and just not deserialize it.

britmyerss (approved):

PR #7153, opened 2025-08-27
Chore: make it impossible to set a component's parent

jkeiser:
In preparation for removing inferred connections entirely, this prevents components from having their parents set. It does not remove parents from old components; that's a migration. But we need to lock this down, to guarantee that the migration will get rid of the last ones! This PR:

* Removes support for `parent` operations in management functions. Management functions will fail when this happens.
* Removes the /v2/views/:id/paste_components, convert_to_view, and set_component_parent endpoints, which are not used in any UI or typescript code.
* Does not preserve parentage when upgrading components or regenerating variants

Methodology: removed Frame::upsert_parent() and Component::add_edge_to_frame(). Fixed dependents.

## How was it tested?

- [X] Integration tests pass (includes management function tests)
- [X] Manual test: can create, delete, update components


lib/sdf-server/src/service/v2/view.rs
```diff
@@ -165,7 +162,6 @@ pub fn v2_routes() -> Router<AppState> {
             "/create_and_move",
             post(create_view_and_move::create_view_and_move),
         )
-        .route("/convert_to_view", post(convert_to_view::convert_to_view))
```
britmyerss, line 168: are we not using `convert_to_view` in newhotness?

jkeiser: The only .ts that has the text `convert_to_view` in it is viewsStore.CONVERT_TO_VIEW(), and that's never called anywhere. (source: ctrl+shift+f convert_to_view, case insensitive)

jkeiser: (That's why this PR removes CONVERT_TO_VIEW)

jkeiser: I just updated the PR to remove CONVERT_TO_VIEW, which I should have done in the first place!

britmyerss (approved):
left one question, but otherwise looks good to me!

jkeiser:
The integration test failure doesn't appear to be a test failure but an OOM timeout. Retrying.

PR #7161, opened 2025-08-29
Store mv index on the client. Patch it from IndexUpdate. Delete MVs not in the index.

jobelenus:

1. On a cold start, store the MvIndex data as an atom
2. After patches complete, we get an IndexUpdate message
3. That message now has patch data for the MvIndex itself
4. Apply that patch
5. Remove MTMs that aren't on the index

Added a fn to `heimdall` that will allow you to ask whether a kind & id exist on the index. This way, we can prompt the user that: "Yes—this component doesn't exist" vs "it may exist but we dont have it yet"

## How was it tested?

1. Use the rune menu to call Odin
2. See the MvIndex data stored
3. Go to a component detail page, remove characters from the component ID, so it will not find one, see the existence check fire without failure.
4. Make a change
5. Odin again
6. see the index checksum on the MvIndex atom is updated.


app/web/src/workers/webworker.ts
```diff
@@ -957,6 +942,124 @@ const initIndexAndChangeSet = (
...
+          index_checksum,
+          kind,
+          args,
+          checksum,
+        )
+      NOT IN
+        (
+          (${placeholders.join("), (")})
+        );
+    `;
```
zacharyhamm: Won't this delete all the index atom mtms that we have in place for *other* change sets than the one that we just got an index patch for?

jobelenus: hows about that change 🤣

zacharyhamm (approved):

stack72 (approved):

PR #7162, opened 2025-08-29
feat: Add "Executed AI Agent" flag to database

vbustamante:
Add a luminork endpoint (`/v1/w/{workspace_id}/user/set_ai_agent_executed`) to set a flag that indicates the user has used the AI agent once. 

Thread it to the auth store in web. Make that change reactive with a websocket event

#### Out of Scope:

There'll be another PR that uses this data on the MCP and the frontend

## How was it tested?

- [X] Run system, make sure web ui still works, hit new endpoint on luminork, check if the reconnect endpoint has the flags data

### DIFF
--- a/lib/sdf-v1-routes-session/src/auth_connect.rs
+++ b/lib/sdf-v1-routes-session/src/auth_connect.rs
@@ -71,6 +71,7 @@ pub struct AuthConnectResponse {
     pub user: User,
     pub workspace: Workspace,
     pub token: String,
+    pub user_workspace_flags: serde_json::Value,
 }
 
 #[derive(Debug, Serialize, Deserialize)]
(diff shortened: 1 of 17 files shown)


lib/sdf-v1-routes-session/src/auth_connect.rs
```diff
@@ -281,7 +283,7 @@ pub async fn auth_connect(
 
     let ctx = builder.build_default(request_ulid).await?;
 
-    let (user, workspace) = find_or_create_user_and_workspace(
+    let (ctx, user, workspace) = find_or_create_user_and_workspace(
```
zacharyhamm, line 286: what's going on with the dal context here?

zacharyhamm: ok i understand it, it's workable!

vbustamante: When I tried just making the passed argument a mutable ref, it broke because the &mut does not implement the SiDbCtx trait that an inner function needs. So i just let it swallow a ctx and return the modified one. Clunky but functional

PR #7163, opened 2025-08-29
feat: Remove all parents and socket connections in migration

jkeiser:
This removes all remaining socket connections (if it can't migrate, it just sets its static value), as well as parents.

## How was it tested?

This is entirely local to the admin-only `/migrate_connections` endpoint, so it won't affect anything unless that's run. Plan currently is to test it in dry runs against workspaces that actually have parents and socket connections.

### DIFF
--- a/lib/sdf-server/src/service/v2/admin/migrate_connections.rs
+++ b/lib/sdf-server/src/service/v2/admin/migrate_connections.rs
@@ -5,6 +5,7 @@ use std::collections::{
 
 use axum::extract::Path;
 use dal::{
+    AttributePrototype,
     AttributeValue,
     ChangeSetId,
     Component,
@@ -16,6 +17,7 @@ use dal::{
         prototype::argument::AttributePrototypeArgument,
         value::subscription::ValueSubscription,
     },
+    component::frame::Frame,
     slow_rt,
     workspace_snapshot::{
         graph::validator::connections::{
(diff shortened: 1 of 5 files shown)


lib/sdf-server/src/service/v2/admin/migrate_connections.rs
```diff
@@ -323,38 +327,80 @@ async fn add_prop_connection(
...
+            if let Some(prop_id) = AttributePrototype::prop_id(ctx, prototype_id).await? {
+                let av_id =
+                    Component::attribute_value_for_prop_id(ctx, to_component_id, prop_id).await?;
+                // Only set its value if it isn't already set!
+                if AttributeValue::component_prototype_id(ctx, av_id)
+                    .await?
+                    .is_none()
+                {
+                    // Pull the current value from the AV, and set it explicitly to that value
+                    // so it remains the same, but doesn't rely on the subscription anymore
```
zacharyhamm: do you mean connection here, not subscription?

zacharyhamm (approved):

lib/sdf-server/src/service/v2/admin/migrate_connections.rs
jkeiser: I do!

jkeiser: fixed :)

zacharyhamm (approved):

PR #7164, opened 2025-08-29
no more VButton, unified NewButton component

wendybujalski:

All instances of VButton have been replaced with NewButton, which aligns with Victoria's designs and eliminates variants that we will not be using in the new UI.

#### Screenshots:

Screenshot 2025-08-29 at 6 31 17 PM

#### Out of Scope:

There are a small few places where we still have inconsistent buttons, and we also need to revamp or replace IconButton as well.

### DIFF
--- a/app/web/src/newhotness/AddViewModal.vue
+++ b/app/web/src/newhotness/AddViewModal.vue
@@ -36,23 +36,12 @@
     </form>
 
     <div class="flex gap-sm mt-sm">
-      <VButton
-        label="Cancel"
-        tone="shade"
-        variant="ghost"
-        @click="() => modalRef?.close()"
-      />
-      <VButton
-        :class="
-          clsx(
-            'flex-grow !text-sm !border !cursor-pointer !px-xs',
-            themeClasses(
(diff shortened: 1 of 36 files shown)


stack72 (commented):
ok so @VictoriaMolgado and I went through this and we have the following feedback - THIS ISN'T ALL DUE TO THS PULL REQUEST! It's just a good time to fix the inconsistencies when we are in here :)

1. Change Set Create / Abandon doesn't need to change because the old nav still uses them
2. CollapsingFlexIteem now has the expanding icon by default - that's not correct - that should only show when they are expanded
3. All button text should be 14px - we have Vizualise Subscriptions as 12px
4. Connect button on arrays and maps should be neutral and text 14px
5. When there's an icon or a counter in the button, we need to make sure that the padding and styling is adjusted accordingly - @VictoriaMolgado can provide the necessary info here

stack72 (dismissed):

aaron-dernley (approved):

jobelenus (approved):

PR #7165, opened 2025-08-29
feat: Set "Executed AI Agent" flag from MCP and add warning to web

vbustamante:
Adds a banner to the web UI that only goes UI after user has started the AI Agent

image

### DIFF
--- a/bin/si-mcp-server/src/user_state.ts
+++ b/bin/si-mcp-server/src/user_state.ts
@@ -1,0 +1,13 @@
+import axios from "npm:axios";
+import { apiConfig, WORKSPACE_ID } from "./si_client.ts";
+import { logger } from "./logger.ts";
+
+export async function setAiAgentUserFlag(){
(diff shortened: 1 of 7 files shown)


bin/si-mcp-server/src/user_state.ts
```diff
@@ -0,0 +1,13 @@
+import axios from "npm:axios";
+import { apiConfig, WORKSPACE_ID } from "./si_client.ts";
+import { logger } from "./logger.ts";
+
+export async function setAiAgentUserFlag(){
+  const url = `${apiConfig.basePath}/v1/w/${WORKSPACE_ID}/user/set_ai_agent_executed`;
```
jobelenus, line 6: So this endpoint already exists?

jobelenus (approved):

bin/si-mcp-server/src/user_state.ts
vbustamante: It was added in this previous PR: https://github.com/systeminit/si/pull/7162

I'm testing the deployment to make sure it works before merging this one

stack72 (approved):

PR #7173, opened 2025-09-02
feat(luminork): adding the authoring endpoints for Luminork

stack72:
The following endpoints are available:

* POST /schemas to create a schema and it's variant
* POST /schemas/{schema_id}/unlock
* POST /schemas/{schema_id}/variant/{schema_variant_id}/funcs/action
* POST /schemas/{schema_id}/variant/{schema_variant_id}/funcs/authentication
* POST /schemas/{schema_id}/variant/{schema_variant_id}/funcs/codegen
* POST /schemas/{schema_id}/variant/{schema_variant_id}/funcs/qualification 
* POST /schemas/{schema_id}/variant/{schema_variant_id}/funcs/management
* PUT /funcs/{func_id} to update a func contents

### DIFF
--- a/lib/luminork-server/src/service/v1.rs
+++ b/lib/luminork-server/src/service/v1.rs
@@ -108,6 +108,10 @@ pub use funcs::{
     FuncsResult,
     get_func::GetFuncV1Response,
     get_func_run::GetFuncRunV1Response,
+    update_func::{
+        UpdateFuncV1Request,
+        UpdateFuncV1Response,
+    },
 };
 pub use management_funcs::{
     ManagementFuncJobStateV1RequestPath,
@@ -122,11 +126,38 @@ pub use schemas::{
     SchemaError,
     SchemaV1RequestPath,
     SchemaVariantV1RequestPath,
+    create_action::{
+        CreateVariantActionFuncV1Request,
+        CreateVariantActionFuncV1Response,
+    },
+    // create_attribute::{
+    //     CreateVariantAttributeFuncV1Request,
+    //     CreateVariantAttributeFuncV1Response,
+    // },
+    create_authentication::{
+        CreateVariantAuthenticationFuncV1Request,
+        CreateVariantAuthenticationFuncV1Response,
+    },
+    create_codegen::{
+        CreateVariantCodegenFuncV1Request,
+        CreateVariantCodegenFuncV1Response,
+    },
+    create_management::{
+        CreateVariantManagementFuncV1Request,
+        CreateVariantManagementFuncV1Response,
+    },
+    create_qualification::{
+        CreateVariantQualificationFuncV1Request,
+        CreateVariantQualificationFuncV1Response,
+    },
+    create_schema::CreateSchemaV1Request,
     find_schema::{
         FindSchemaV1Params,
         FindSchemaV1Response,
     },
     list_schemas::ListSchemaV1Response,
+    unlock_schema::UnlockedSchemaV1Response,
+    // update_schema_variant::UpdateSchemaVariantV1Request,
 };
 pub use workspaces::WorkspaceError;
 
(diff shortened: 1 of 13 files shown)


lib/luminork-server/src/service/v1/funcs/update_func.rs
```diff
@@ -0,0 +1,130 @@
...
+}
+
+#[derive(Deserialize, Serialize, Debug, ToSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct UpdateFuncV1Request {
+    #[schema(value_type = String, example = "Updated Display Name")]
+    pub display_name: Option<String>,
+    #[schema(value_type = String, example = "Updated Description")]
+    pub description: Option<String>,
+    #[schema(value_type = String, example = "<!-- String escaped Typescript code here -->")]
```
nickgerace, line 121: Can a Vue template comment work for TypeScript code with this example?

lib/luminork-server/src/service/v1/schemas/create_qualification.rs
```diff
@@ -0,0 +1,161 @@
...
+    let schema_variant = SchemaVariant::get_by_id(ctx, schema_variant_id).await?;
+    if schema_variant.is_locked() {
+        return Err(SchemaError::LockedVariant(schema_variant_id));
+    }
+
+    let mut locations = payload.locations;
+    if locations.is_empty() {
+        locations.push(LeafInputLocation::Domain);
+        locations.push(LeafInputLocation::Code);
+    }
```
nickgerace, line 85: We always insert both and not just one? Is this to ensure that the user's first experience with the qualification is a good one and THEN they can scope it down to just one leaf input?

lib/luminork-server/src/service/v1/schemas/unlock_schema.rs
```diff
@@ -0,0 +1,109 @@
...
+                AuditLogKind::UnlockSchemaVariant {
+                    schema_variant_id: unlocked.id(),
+                    schema_variant_display_name: unlocked.display_name().to_owned(),
+                },
+                schema.name().to_owned(),
+            )
+            .await?;
+            unlocked.id()
+        }
+    };
```
nickgerace, line 83: I like this

lib/luminork-server/src/service/v1.rs
```diff
@@ -122,11 +126,38 @@ pub use schemas::{
...
     SchemaV1RequestPath,
     SchemaVariantV1RequestPath,
+    create_action::{
+        CreateVariantActionFuncV1Request,
+        CreateVariantActionFuncV1Response,
+    },
+    // create_attribute::{
+    //     CreateVariantAttributeFuncV1Request,
+    //     CreateVariantAttributeFuncV1Response,
+    // },
```
nickgerace, line 136:
```suggestion
```

lib/luminork-server/src/service/v1.rs
```diff
@@ -122,11 +126,38 @@ pub use schemas::{
...
+        CreateVariantQualificationFuncV1Response,
+    },
+    create_schema::CreateSchemaV1Request,
     find_schema::{
         FindSchemaV1Params,
         FindSchemaV1Response,
     },
     list_schemas::ListSchemaV1Response,
+    unlock_schema::UnlockedSchemaV1Response,
+    // update_schema_variant::UpdateSchemaVariantV1Request,
```
nickgerace, line 160:
```suggestion
```

lib/luminork-server/src/service/v1.rs
```diff
@@ -176,8 +207,18 @@ pub use crate::api_types::func_run::v1::{
         schemas::get_schema::get_schema,
         schemas::get_variant::get_variant,
         schemas::get_default_variant::get_default_variant,
+        schemas::create_schema::create_schema,
+        schemas::unlock_schema::unlock_schema,
+        schemas::create_action::create_variant_action,
+        // schemas::create_attribute::create_variant_attribute,
```
nickgerace, line 213: Wait as I am suggesting to delete these... do you intentionally want them there in case we want attributes back? I'd attach a comment explaining if so

lib/luminork-server/src/service/v1/schemas/create_qualification.rs
stack72: On qualifications - always both by default in our function editor

lib/luminork-server/src/service/v1/funcs/update_func.rs
stack72: This is what the sample request has in it:

```
{
  "code": "<!-- String escaped Typescript code here -->",
  "description": "Creates an EC2 Instance",
  "displayName": "Create EC2 Instance",
  "kind": "Create",
  "name": "awsEC2InstanceCreate"
}
```

lib/luminork-server/src/service/v1.rs
stack72: yes, these are coming back - just in a follow up PR :)

PR #7176, opened 2025-09-02
fix: we need deno in the rootfs

sprutton1:

lang-js is compiled with deno and runs deno so it needs deno in the rootfs

## How was it tested?

I will have to test this one manually after a try

### DIFF
--- a/bin/cyclone/BUCK
+++ b/bin/cyclone/BUCK
@@ -69,12 +69,6 @@ docker_image(
     ],
 )
 
-nix_omnibus_pkg(
-    name = "omnibus",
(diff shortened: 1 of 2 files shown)


stack72 (approved):

bin/cyclone/BUCK
```diff
@@ -69,12 +69,6 @@ docker_image(
     ],
 )
 
-nix_omnibus_pkg(
```
stack72, line 72: Expected removal?

sprutton1: yep, we don't need this anymore

PR #7180, opened 2025-09-02
feat: Component Details skeleton

vbustamante:
Skeleton will be show until the component query and the attribute tree query have completed loading

image


stack72:
I have a feeling that label check will fail but that's fine - merge away!

PR #7184, opened 2025-09-02
Fix: watched form will stay up to date with data mutations, until...

jobelenus:

Watched form will stay up to date with data mutations, until a user dirties the form.

- Reset on form submission, and resets dirty.
- Correct any watch fns in implementations.
- Secrets needs additional love because its a two-step resolution from DVU.

#### Out of Scope:

I'll have a follow on PR to deal with observability, its not where it needs to be.

## How was it tested?

- adding a view
- renaming a view

Ensure that these are reactive across two browser tabs:
- changing component name
- prop editing
- adding & editing keys to a prop object
- subscription editing (including subscribing to secrets)
- secret creation

**Note**: for secret creation we have a `subscribing...` state that is manually implemented that picks up the distance between "the API call succeeding, where `you are set from a fn value`" and DVU running and filling in the secret value. Without that, the UX is a bit jarring.


app/web/src/newhotness/logic_composables/watched_form.ts
```diff
@@ -88,8 +84,23 @@ export const useWatchedForm = <Data>(
     const opts = formOptions({
       defaultValues: toValue(data),
     });
+
+    // keep the form up to date with values until the user
```
jkeiser, line 99: I suspect we want to set dirty = false on reset too. It looks like `<AttributeInput>` calls `reset()` directly on the form when the input is opened back up, and we would want it to track changes until it's dirty again.

app/web/src/newhotness/logic_composables/watched_form.ts
```diff
@@ -106,51 +117,42 @@ export const useWatchedForm = <Data>(
         // Mark submission as complete and remove the rainbow spinner
         const markComplete = () => {
           bifrosting.value = false;
+          dirty = false;
           rainbow.remove(ctx.changeSetId.value, label);
-          if (!wForm.state.canSubmit)
-            wForm.reset(resetBlank ? undefined : props.value);
+          wForm.reset(resetBlank ? undefined : props.value);
```
jkeiser, line 136: Nice factoring the two calls into one!

wendybujalski (dismissed):

app/web/src/newhotness/logic_composables/watched_form.ts
jobelenus: we do set `dirty=false` on reset (?)

wendybujalski (approved):

PR #7187, opened 2025-09-02
Implement intelligent materializedview rebuild system on schema changes

jhelwig:
Establishes comprehensive type schema change detection system enabling selective MaterializedView rebuilds when only type definitions change, avoiding expensive full data rebuilds.

**Core Infrastructure**:
- Add DefinitionChecksum trait with static compilation-time checksum resolution
- Create comprehensive DefinitionChecksum implementations for all MV-used types
- Build DefinitionChecksum derive macro supporting recursive type handling
- Add RebuildChangedDefinitionsRequest API type with versioned structure

**MaterializedView Integration**:
- Extend MaterializedView derive macro to generate definition checksums
- Update MV infrastructure to detect and rebuild only definition-changed views
- Implement compressed request handling for selective rebuild operations

**Code Organization**:
- Refactor frigg into deployment, kv_history, and workspace modules
- Split si_frontend_mv_types::index into change_set and deployment submodules

**Compression Algorithm Enhancements**:
- Support RebuildChangedDefinitions requests in compression pipeline
- Maintain contiguous update sequences while filtering definition rebuilds
- Prioritize full rebuilds when mixed with selective definition rebuilds

The system now distinguishes between data changes requiring full rebuilds and schema changes needing only definition updates, significantly improving rebuild efficiency for type system modifications.

The change set/workspace level MvIndex will be considered "missing" until it has been upgraded to the V2 format, which will trigger full rebuilds for the change sets in question.

The deployment level MvIndex will continue to ignore the checksum information for now, until we have had time to make sure it has been upgraded to the V2 format.


lib/edda-server/src/materialized_view.rs
```diff
@@ -1115,45 +228,70 @@ async fn build_deployment_mv_inner(
     let mut build_slowest_mv_kind: &str = "N/A";
 
     loop {
-        // If there aren't any queued builds waiting for the concurrency limit then
+        // If there there aren't any queued builds waiting for the concurrency limit then
```
nickgerace:
```suggestion
        // If there aren't any queued builds waiting for the concurrency limit then
```

lib/si-frontend-mv-types-rs/src/definition_checksum.rs
```diff
@@ -0,0 +1,93 @@
+use std::collections::HashMap;
+
+use si_events::workspace_snapshot::Checksum;
+
+/// Trait for computing definition checksums of types based on their schema/shape.
+/// Unlike FrontendChecksum which handles data values, DefinitionChecksum focuses
+/// on the structure and type information of the data.
```
nickgerace, line 7: Nice! Structural vs. data values. 👍

lib/si-frontend-mv-types-rs/src/reference.rs
```diff
@@ -159,7 +167,7 @@ where
     T: Eq + PartialEq + Clone + std::fmt::Debug + Serialize + std::fmt::Display,
 {
     fn checksum(&self) -> Checksum {
-        todo!()
```
nickgerace, line 162: Oops

lib/si-frontend-mv-types-rs/src/checksum.rs
```diff
@@ -396,3 +397,421 @@ impl FrontendChecksum for &[u8] {
...
+        });
+        *CHECKSUM
+    }
+}
+
+impl DefinitionChecksum for Timestamp {
+    fn definition_checksum() -> Checksum {
+        static CHECKSUM: ::std::sync::LazyLock<Checksum> = ::std::sync::LazyLock::new(|| {
+            let mut hasher = ChecksumHasher::new();
+            hasher.update(b"Timestamp");
```
nickgerace, line 562: How does this protect against the underlying type changing? What if someone changes `Timestamp`?

lib/si-frontend-mv-types-macros/src/materialized_view.rs
```diff
@@ -46,20 +58,45 @@ pub fn derive_materialized_view(
...
+                hasher.update(#field_type.as_bytes());
+            });
+        } else {
+            // For non-recursive fields: use field name + recursive DefinitionChecksum
+            // This provides full schema change detection for nested types
+            // CRITICAL: No fallback - missing DefinitionChecksum implementations cause compilation errors
+            let field_ty = &field.ty;
+            definition_checksum_updates.push(quote! {
+                hasher.update(#field_name.as_bytes());
+                hasher.update(<#field_ty as crate::definition_checksum::DefinitionChecksum>::definition_checksum().as_bytes());
```
nickgerace, line 98: 👍

lib/si-frontend-mv-types-rs/src/checksum.rs
jhelwig: Yeah, there are a few like this (`Checksum`, all of the `FooId` types, `Ulid`). It's definitely not perfect, but I think that if any of these types end up changing to the point where it would justify rebuilding the MVs, we're probably not silently replacing the inner-workings of the types since it would also fundamentally break the graphs, too.

That said, I don't think it would take much to convince me that there's a reasonable more robust option for these types.

nickgerace (dismissed):

lib/frigg/src/deployment.rs
```diff
@@ -0,0 +1,386 @@
...
+        // MvIndex was built for, then the MvIndex is out of date and should not be used at all.
+        if index_pointer_value.definition_checksums != materialized_view_definition_checksums() {
+            debug!(
+                "deployment index pointer is out of date: index checksums: {:?}, expected checksums: {:?}",
+                index_pointer_value.definition_checksums,
+                materialized_view_definition_checksums()
+            );
+            return Ok(None);
+        }
+        */
```
nickgerace, line 337: Calling this out in a comment for reviewers

fnichol (approved):

PR #7195, opened 2025-09-03
feat(luminork): API Test to avoid a 202 on schema find in Luminork

stack72:
I am using an obscure schema that we don’t test here so we can check it’s as expected and that we don’t get a 202

https://github.com/systeminit/si/actions/runs/17441680235/job/49525954402#step:4:364

### DIFF
--- a/bin/si-luminork-api-tests/src/api/schemas.ts
+++ b/bin/si-luminork-api-tests/src/api/schemas.ts
@@ -1,10 +1,10 @@
 /**
  * Schemas API Client
- * 
+ *
  * Client for interacting with Schemas endpoints
  */
 
-import { LuminorkClient, ApiResponse } from '../client.ts';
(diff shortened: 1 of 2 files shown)


bin/si-luminork-api-tests/src/api/schemas.ts
```diff
@@ -56,94 +56,113 @@ export interface GetSchemaVariantResponse {
...
   /**
    * Find schemas by name or category
    */
   async findSchema(
     workspaceId: string,
     params: FindSchemaParams,
-    changeSetId?: string
-  ): Promise<ApiResponse<ListSchemasResponse>> {
+    changeSetId?: string,
+  ): Promise<ApiResponse<SchemaView>> {
```
nickgerace, line 99: I like this response type naming better anyway

nickgerace (approved):

PR #7207, opened 2025-09-04
chore: Remove incoming_connections / outgoing_connections (DO NOT MERGE UNTIL MIGRATION COMPLETE)

jkeiser:
This PR is the first wave of removing socket connections from the codebase:

* Removes ArgumentTargets and target() from AttributePrototypeArgument
* Removes all downstream code, notably anything involving incoming or outgoing connections
* Removes tests intended to test socket connections
* Reworks DVU and secret tests to use subscriptions when they were using sockets before

This *should not be merged* until the Final Socket Connection Migration is complete. After that point, this PR should provide 100% equivalent functionality

#### Out of Scope:

* The graph-level code has not yet been lifted: we still need to remove ArgumentTargets from NodeWeight, but doing that alongside all this seemed too big to digest.
* Typescript code receives empty arrays for serialized incomingConnections, which is correct because this will not be checked in until all connections have been migrated away. We should dump connections from typescript types as well.
* Inferred connections and parentage remain, even though those are also going away in the migration. Coming in a followup
* It turns out downstream subscriptions aren't being zeroed out when the upstream is ghosted--I updated the test to use subscriptions, and #[ignore]d it with a TODO. This issue predates this PR.

There are 4 tests that are commented out, that seem like they *might* still need to exist, which are commented out with a TODO because they involve complex socket functions but may be testing generic functionality around dynamic functions on schema. I'll look at these in the near term, but don't want to hold up the main PR in the meantime.

## How was it tested?

- [X] Integration tests pass
- [ ] Manual test: discovery / analysis loop with Claude works (management functions) 
- [ ] Manual test: components, subscriptions, map and grid view all look right with subscriptions


lib/dal/src/diagram.rs
```diff
@@ -467,7 +468,6 @@ pub struct Diagram {
 
 pub struct DiagramComponentViews {
     pub component_views: Vec<DiagramComponentView>,
-    pub diagram_edges: Vec<SummaryDiagramEdge>,
```
jkeiser, line 470: This is not a serializable type, so I removed it entirely (only preserving empty edge lists that are serialized right now)

app/web/src/store/components.store.ts
```diff
@@ -170,6 +170,7 @@ export type ConnectionMenuData = {
 
 export type ComponentsAndEdges = {
   components: RawComponent[];
+  // TODO REMOVE this is empty now
```
jkeiser, line 173: This will go away in a followup that removes the serialized type (and JS's use of it). For now, this is always an empty array; in the future, it won't be sent at all.

lib/dal-materialized-views/src/component/attribute_tree.rs
```diff
@@ -46,19 +41,6 @@ pub async fn assemble(ctx: DalContext, component_id: ComponentId) -> crate::Resu
         Component::attribute_value_for_prop(ctx, component_id, &["root", "secrets"]).await?;
     let secret_ids_by_key = Secret::list_ids_by_key(ctx).await?;
 
-    let component_has_socket_connection = ctx
```
jkeiser, line 49: All this was used to calculate has_socket_connection, which is always `false` now.

lib/dal/src/action/dependency_graph.rs
```diff
@@ -112,26 +112,10 @@ impl ActionDependencyGraph {
         // feed data into other things must have their actions run before the actions for the
         // things they are feeding data into.
         while let Some(component_id) = components_to_process.pop_front() {
-            let component = Component::get_by_id(ctx, component_id).await?;
             let component_index = component_dependencies_index_by_id
                 .entry(component_id)
                 .or_insert_with(|| component_dependencies.add_node(component_id))
                 .to_owned();
-            for incoming_connection in component.incoming_connections(ctx).await? {
```
jkeiser, line 120: incoming_connections is always empty, so can't affect the DVU graph anymore.

lib/dal/src/attribute/prototype/argument.rs
```diff
@@ -127,7 +123,6 @@ pub type AttributePrototypeArgumentResult<T> = Result<T, AttributePrototypeArgum
 #[derive(Deserialize, Serialize, Debug, Clone, PartialEq, Eq)]
 pub struct AttributePrototypeArgument {
     id: AttributePrototypeArgumentId,
-    targets: Option<ArgumentTargets>,
```
jkeiser, line 130: This is the principal change: removed targets here, and all code that uses it (which is our socket connections code).

lib/dal/src/attribute/prototype/debug.rs
```diff
@@ -131,178 +131,157 @@ impl AttributePrototypeDebugView {
...
                 .await?
                 .is_some();
-        let destination_component_id =
-            AttributeValue::component_id(ctx, attribute_value_id).await?;
+        let component_id = AttributeValue::component_id(ctx, attribute_value_id).await?;
         let mut func_binding_args: HashMap<String, Vec<FuncArgDebugView>> = HashMap::new();
         let attribute_prototype_arg_ids =
             AttributePrototypeArgument::list_ids_for_prototype(ctx, prototype_id).await?;
         for attribute_prototype_arg_id in attribute_prototype_arg_ids {
-            let attribute_prototype_argument =
```
jkeiser, line 140: is_used was only ever `false` if socket connections were involved. This will get set a different way when we make ghosting affect downstream subscriptions.

lib/dal/src/attribute/prototype/debug.rs
```diff
@@ -131,178 +131,157 @@ impl AttributePrototypeDebugView {
             AttributeValue::component_prototype_id(ctx, attribute_value_id)
                 .await?
                 .is_some();
-        let destination_component_id =
-            AttributeValue::component_id(ctx, attribute_value_id).await?;
+        let component_id = AttributeValue::component_id(ctx, attribute_value_id).await?;
```
jkeiser, line 134: We no longer have `expected_source_component_id`, so we don't need to mark this as a `destination_`.

lib/dal/src/attribute/value.rs
```diff
@@ -802,81 +802,58 @@ impl AttributeValue {
         let mut func_binding_args: HashMap<String, Vec<Value>> = HashMap::new();
         let apa_ids = AttributePrototypeArgument::list_ids_for_prototype(ctx, prototype_id).await?;
         for apa_id in apa_ids {
-            let apa = AttributePrototypeArgument::get_by_id(ctx, apa_id).await?;
```
jkeiser, line 805: This code handles cases where the upstream is a ghosted component, and treats the socket connection as if it wasn't there. Subscriptions have no equivalent functionality yet; that will be added as its own fix.

lib/dal/src/component/socket.rs
```diff
@@ -391,23 +355,6 @@ impl ComponentInputSocket {
         self,
         ctx: &DalContext,
     ) -> ComponentResult<Option<ComponentId>> {
-        let maybe_explicit_connection_source = {
```
jkeiser, line 394: This will end up going away entirely when inferred connections get removed.

lib/dal/src/management/prototype.rs
```diff
@@ -326,14 +326,8 @@ async fn build_incoming_connections(
 ) -> ManagementPrototypeResult<HashMap<String, SocketRefsAndValues>> {
     let mut incoming_connections = HashMap::new();
     for input_socket in ComponentInputSocket::list_for_component_id(ctx, component_id).await? {
-        // Collect explicit connections for this input socket
```
jkeiser, line 329: This method will disappear when inferred connections disappears.

lib/dal/src/workspace_snapshot/node_weight/attribute_prototype_argument_node_weight.rs
```diff
@@ -37,20 +37,11 @@ pub struct AttributePrototypeArgumentNodeWeight {
 
 impl AttributePrototypeArgumentNodeWeight {
     pub fn new(id: Ulid, lineage_id: Ulid) -> Self {
-        Self::new_with_targets_for_tests(id, lineage_id, None)
-    }
-
-    // Only for use in tests (and called by other test-only functions)
```
jkeiser, line 43: Can't make socket connections even in tests anymore!

lib/dal/src/component.rs
```diff
@@ -2975,28 +2477,6 @@ impl Component {
         let output_sockets: Vec<ComponentOutputSocket> =
             ComponentOutputSocket::list_for_component_id(ctx, self.id()).await?;
         for output_socket_match in output_sockets {
-            let output_socket =
```
jkeiser, line 2978: Method will disappear when inferred connections do.

lib/dal/src/diagram.rs
jkeiser, line 470: Not serialized, so we remove it and fix any Rust callers.

lib/dal/tests/integration_test/component/duplicate.rs
```diff
@@ -53,101 +53,6 @@ async fn duplicate_component_with_value(ctx: &mut DalContext) -> Result<()> {
     Ok(())
 }
 
-#[test]
```
jkeiser, line 56: duplicate_components_with_subscriptions below handles this.

lib/dal/tests/integration_test/component/paste.rs
```diff
@@ -66,100 +66,12 @@ async fn paste_component_with_value(ctx: &mut DalContext) -> Result<()> {
     Ok(())
 }
 
-#[test]
```
jkeiser, line 69: paste_component_with_subscriptions handles this.

lib/dal/tests/integration_test/component/upgrade.rs
```diff
@@ -624,140 +624,11 @@ async fn upgrade_array_of_objects(ctx: &mut DalContext) -> Result<()> {
     Ok(())
 }
 
-#[test]
```
jkeiser, line 627: This was a multiplayer scenario where socket connections didn't always get removed during schema upgrades. Subscriptions don't have this issue. Corresponding CTs no longer do anything, and will be removed when ArgumentTargets is removed from the NodeWeight (next turn).

lib/dal/tests/integration_test/deserialize/mod.rs
```diff
@@ -124,14 +126,16 @@ fn make_me_one_with_everything(graph: &mut WorkspaceSnapshotGraphVCurrent) {
...
             NodeWeightDiscriminants::AttributePrototypeArgument => {
-                NodeWeight::new_attribute_prototype_argument_with_targets_for_tests(
-                    Ulid::new(),
-                    Ulid::new(),
-                    Some(ArgumentTargets {
+                NodeWeight::AttributePrototypeArgument(AttributePrototypeArgumentNodeWeight {
+                    id: Ulid::new(),
+                    lineage_id: Ulid::new(),
+                    merkle_tree_hash: MerkleTreeHash::default(),
+                    targets: Some(ArgumentTargets {
```
jkeiser, line 133: This is the one place we *can* create an ArgumentTargets; leaving it in here until we literally remove it from the NodeWeight, so that we know we can load old graphs with this in it.

lib/dal/tests/integration_test/func/authoring/binding.rs
```diff
@@ -776,24 +773,13 @@ async fn return_the_right_bindings(ctx: &mut DalContext, nw: &WorkspaceSignup) {
...
-        .await
-        .expect("could not perform find with name")
-        .expect("output socket not found by name");
-    let destination_input_socket =
-        InputSocket::find_with_name(ctx, "dummy", destination_schema_variant_id)
-            .await
-            .expect("could not perform find with name")
-            .expect("input socket not found by name");
-    Component::connect_for_tests(
+    value::subscribe(
```
jkeiser, line 775: Secrets are now communicated with subscriptions, not sockets!

lib/dal/tests/integration_test/node_weight/attribute_value.rs
```diff
@@ -13,39 +13,40 @@ use dal_test::{
...
-    assert!(
-        DependentValueRoot::get_dependent_value_roots(ctx)
-            .await
-            .expect("able to get dvu values")
-            .contains(&DependentValueRoot::Unfinished(
-                image.attribute_value(ctx).await.id().into()
-            ))
-    );
-}
+// TODO use subscriptions?
```
jkeiser, line 16: I had trouble tracking the purpose of this test, so commented it out for now. Will ask around and decide whether to remove or follow up with a fix.

lib/dal/tests/integration_test/node_weight/component.rs
```diff
@@ -55,225 +52,138 @@ async fn deleting_a_component_deletes_component_in_other_change_sets(ctx: &mut D
...
-    let incoming_sources: HashSet<ComponentId> = butane
-        .incoming_connections(ctx)
-        .await
-        .expect("able to get incoming connections")
-        .iter()
-        .map(|conn| conn.from_component_id)
-        .collect();
-    assert!(incoming_sources.is_empty());
-}
+// TODO restore this using subscriptions!
```
jkeiser, line 48: Maybe could be useful with subscriptions? It does an awful lot and isn't trivially convertible. Will follow up with a remove-or-fix.

lib/dal/tests/integration_test/schema/variant/authoring/regenerate.rs
```diff
@@ -163,57 +162,16 @@ async fn update_socket_data_on_regenerate(ctx: &mut DalContext) -> Result<()> {
...
         "small odd lego",
         "small odd",
     )
     .await?;
 
-    let output_comp =
+    let _output_comp =
         create_component_for_default_schema_name_in_default_view(ctx, "small even lego", "even")
             .await?;
-    // let mut input_output_socket_avs = input_comp
```
jkeiser, line 176: Still tests sockets on regenerate! Just doesn't test connections.

lib/dal/tests/integration_test/secret/with_actions.rs
```diff
@@ -63,24 +64,13 @@ async fn create_action_using_secret(ctx: &mut DalContext, nw: &WorkspaceSignup)
...
-        .await
-        .expect("could not perform find with name")
-        .expect("output socket not found by name");
-    let destination_input_socket =
-        InputSocket::find_with_name(ctx, "dummy", destination_schema_variant_id)
-            .await
-            .expect("could not perform find with name")
-            .expect("input socket not found by name");
-    Component::connect_for_tests(
+    value::subscribe(
```
jkeiser, line 61: This tests secrets, not socket connections, so it becomes a subscription instead.

lib/dal/tests/integration_test/attributes.rs
```diff
@@ -639,117 +639,3 @@ async fn update_attribute_child_of_subscription(ctx: &mut DalContext) -> Result<
 
     Ok(())
 }
-
-// Test that updating attributes sets them (and their parents) correctly, but leaves default
-// values and other values alone.
-#[test]
```
jkeiser, line 645: Equivalent subscription test already exists.

lib/dal/tests/integration_test/dependent_values_update.rs
```diff
@@ -1,509 +1,167 @@
...
+        attribute::value,
+        change_set,
+        component,
     },
     test,
 };
 use serde_json::json;
 use veritech_client::ResourceStatus;
 
+// TODO implement this! It worked for connections but does not work for subscriptions
```
jkeiser, line 19: Subscriptions still work when the upstream component is ghosted; we probably want them to stop yielding a value. Adding features/fixing actual bugs is out of scope for this PR.

lib/dal/tests/integration_test/dependent_values_update.rs
```diff
@@ -1,509 +1,167 @@
...
-
-    let view = AttributeValue::view(ctx, units_value_id)
-        .await
-        .expect("able to get units view")
-        .expect("units has a view");
-    let units_json_string = serde_json::to_string(&view).expect("Unable to stringify JSON");
-    assert!(units_json_string.contains("docker.io/library/oysters in my pocket\\n"));
-
-    royel_component
+async fn normal_to_marked_for_deletion_flows(ctx: &mut DalContext) -> Result<()> {
```
jkeiser, line 72: Same test, but with subscriptions and using the newer, more readable test methods.

lib/dal/tests/integration_test/dependent_values_update.rs
```diff
@@ -1,509 +1,167 @@
...
+    Ok(())
 }
 
 /// Until we have a better system for signalling that a DVU has run and
 /// finished, we can't actually verify that it executed these per-component. But
 /// we can ensure that with a concurrency limit: (1) the job finishes and (2) it
 /// produces the correct data
 #[test]
-async fn component_concurrency_limit(ctx: &mut DalContext) {
+async fn component_concurrency_limit(ctx: &mut DalContext) -> Result<()> {
```
jkeiser, line 118: Uses subscriptions now, but does the same thing.

lib/dal/tests/integration_test/property_editor.rs
```diff
@@ -557,176 +556,179 @@ async fn prop_can_be_set_by_socket(ctx: &mut DalContext) {
...
-            "isControlledByDynamicFunc": true,
-            "overridden": false
-        }], // expected
-        PropEditorTestView::for_component_id(ctx, pirate_component.id())
-            .await
-            .expect("could not get property editor test view")
-            .get_value(parrots_path)
-            .expect("could not get value")
-    );
+    // Removed: canBeSetBySocket may still be true, but socket connections no longer exist
```
jkeiser, line 559: This test will go away entirely when sockets are, but for now we just remove any actual connections.

lib/dal/tests/integration_test/property_editor.rs
```diff
@@ -557,176 +556,179 @@ async fn prop_can_be_set_by_socket(ctx: &mut DalContext) {
...
-            }], // expected
-            PropEditorTestView::for_component_id(ctx, pirate_component.id())
-                .await
-                .expect("could not get property editor test view")
-                .get_value(&["root", "domain", "parrot_names", "0"])
-                .expect("could not get value")
-        );
-    }
-}
+// TODO restore this using subscriptions!
```
jkeiser, line 601: Complicated test requiring a lot of rewrites, probably using prop <-> prop attribute functions instead of socket -> prop / prop -> socket.

lib/dal/tests/integration_test/resource_metadata.rs
```diff
@@ -27,65 +31,26 @@ use si_events::{
...
-            .await
-            .expect("could not create component");
-    let destination_schema_variant_id =
-        Component::schema_variant_id(ctx, destination_component.id())
-            .await
-            .expect("could not get schema variant id for component");
-    ChangeSetTestHelpers::commit_and_update_snapshot_to_visibility(ctx)
-        .await
-        .expect("could not commit and update snapshot to visibility");
+async fn list(ctx: &mut DalContext, nw: &WorkspaceSignup) -> Result<()> {
```
jkeiser, line 26: We'd still like to be able to list components! I rewrote this using modern stuff before I realized connections are really only there for secrets.

lib/dal/tests/integration_test/secret.rs
```diff
@@ -282,23 +284,24 @@ async fn update_metadata_and_encrypted_contents(ctx: &DalContext, nw: &Workspace
 }
 
 #[test]
-async fn copy_paste_component_with_secrets_being_used(ctx: &mut DalContext, nw: &WorkspaceSignup) {
+async fn copy_paste_component_with_secrets_being_used(
```
jkeiser, line 284: Uses subscriptions for the secrets, and uses new-style test helpers.

lib/dal/tests/integration_test/secret.rs
```diff
@@ -309,93 +312,58 @@ async fn copy_paste_component_with_secrets_being_used(ctx: &mut DalContext, nw:
...
-        .expect("paste");
-    expected::commit_and_update_snapshot_to_visibility(ctx).await;
+    Ok(())
 }
 
+// TODO it's unclear how / whether this should work this way with subscriptions
+#[ignore]
 #[test]
-async fn consumed_secrets_work_when_dvu_not_up_to_date(ctx: &mut DalContext, nw: &WorkspaceSignup) {
+async fn consumed_secrets_work_when_dvu_not_up_to_date(
```
jkeiser, line 346: Modified this for subscriptions, but it doesn't do the same thing as secrets did. I don't know what the purpose of this was in the first place, though. Will ask around and remove or update the test in a follow up.

lib/dal/tests/integration_test/validations.rs
```diff
@@ -230,91 +229,91 @@ async fn validation_pre_post_mgmt_func(ctx: &mut DalContext) -> Result<()> {
...
-        }),
-        source_result
-    );
-
-    let destination_result = extract_value_and_validation(destination_prop_view)?;
-    assert_eq!(source_result, destination_result,);
-
-    Ok(())
-}
+// #[ignore]
```
jkeiser, line 232: Was ignored; now doesn't compile because connections. Probably could be updated to use subscriptions, but may as well let that happen when it actually gets unignored.

lib/dal/tests/integration_test/node_weight/attribute_value.rs
britmyerss: This was testing the following: 
1. Component A exists on Head
2. Create change set with Component B
3. Connect A output socket -> B input socket in that change set
4. Create ANOTHER change set, modify A's output socket, then apply
5. Ensure that the Component B gets updated appropriately on the replay

lib/dal/tests/integration_test/secret.rs
britmyerss: Hmm I'm surprised about this one - maybe because with sockets it requires more of a double hop, we have a longer transient 'Null' secret?

lib/dal/tests/integration_test/property_editor.rs
britmyerss: This looks like it's MORE testing the property editor responses, so it can probably be ported to use MVs instead or just nuke it :)

lib/dal/tests/integration_test/node_weight/component.rs
britmyerss: I think this was testing a specific bug, where we were enqueueing too many AVs in DVU because of how attribute prototype arguments are/were...... so it might not be too valuable to recreate in the new world.

lib/dal/tests/integration_test/frame.rs
britmyerss, line 1: I'm surprised this whole file isn't nuked yet :)

lib/dal/tests/integration_test/component.rs
```diff
@@ -515,63 +510,6 @@ async fn through_the_wormholes_child_value_reactivity(ctx: &mut DalContext) -> R
     Ok(())
 }
 
-#[test]
-async fn through_the_wormholes_dynamic_child_value_reactivity(ctx: &mut DalContext) -> Result<()> {
```
britmyerss, line 519: this one not worth porting?

lib/dal/src/attribute/value.rs
britmyerss: let's make sure this is a fast follow - it can cause pain for the user.

britmyerss (approved):
This is phenomenal work - and the comments were so helpful in review! Let's make sure we don't lose track of the todos

PR #7209, opened 2025-09-04
feat(luminork): Add ability to be able to update a schema variant code and regenerate

stack72:


jhelwig (approved):
Pretty much exactly the pattern we use in some SDF endpoints to return new MV data from the web request endpoint.

PR #7212, opened 2025-09-04
Add component-generate-template MCP tool

nickgerace:
## Description

This change adds a tool for generating templates from components. The tool not only generates the template, but it also finds the resulting schema. Along the way to making this change, the following was also changed:

- Format all "si-mcp-server" files using the buck2 target "fix-format" (side note: we should be checking for this in CI)
- Add an additional example finding buck2 targets locally to the dev docs (this was used to find the "fix-format" command)
- Return the schema ID in addition to the existing data in the luminork route for generating a template
- Try to find the schema on the graph first before going to cached modules when finding schemas in luminork

The primary change was tested by creating a VPC with two subnets and then generating a template from it. The template was then ran to confirm that it worked.

## Screenshot

Screenshot 2025-09-04 at 4 59 37 PM

### DIFF
--- a/bin/si-mcp-server/src/tools/componentGenerateTemplate.ts
+++ b/bin/si-mcp-server/src/tools/componentGenerateTemplate.ts
@@ -1,0 +1,141 @@
+import { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
+import { CallToolResult } from "@modelcontextprotocol/sdk/types.js";
+import { z } from "zod";
+import { ComponentsApi, SchemasApi } from "@systeminit/api-client";
+import { apiConfig, WORKSPACE_ID } from "../si_client.ts";
+import {
+  errorResponse,
+  generateDescription,
+  successResponse,
+  withAnalytics,
+} from "./commonBehavior.ts";
+
+const name = "component-generate-template";
+const title = "Generate a template from a list a components";
+const description =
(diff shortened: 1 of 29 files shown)


bin/si-mcp-server/src/tools/componentGenerateTemplate.ts
```diff
@@ -0,0 +1,141 @@
...
+    z.string().describe(
+      "the component id to be included in the generated template",
+    ),
+  ).describe(
+    "the list of component ids to be included in the generated template",
+  ),
+  templateName: z.string().describe(
+    "the name of the template that will be generated",
+  ),
+  secrets: z.boolean().optional().describe(
```
stack72, line 32: 🙆‍♂️

lib/luminork-server/src/service/v1/components/generate_template.rs
```diff
@@ -45,15 +46,16 @@ pub async fn generate_template(
...
-        ctx,
-        payload.component_ids,
-        payload.asset_name.clone(),
-        payload.func_name.clone(),
-        payload.category.unwrap_or("Templates".to_string()),
-        "#aaaaaa".to_string(),
-    )
-    .await?;
+    let (new_variant, schema_id, func, prototype_id) =
+        sdf_core::generate_template::prepare_and_generate(
```
stack72: Nice addition!

lib/luminork-server/src/service/v1/schemas/find_schema.rs
```diff
@@ -77,39 +78,73 @@ pub async fn find_schema(
         ));
     };
 
-    let module = match schema_ref {
-        SchemaReference::ById { schema_id } => {
-            if let Some(module) = CachedModule::find_latest_for_schema_id(ctx, schema_id).await? {
-                module
-            } else {
-                return Err(SchemaError::SchemaNotFound(schema_id));
+    let (schema_name, schema_id, category, installed) = match schema_ref {
```
stack72: great set of changes!!

aaron-dernley:
🚀

nickgerace:
This requires SDK generation.

PR #7215, opened 2025-09-04
feat(web): Add more tracking events to startup and onboarding

vbustamante:

- Adds hedgehogs for: finished cold boot, started onboarding (automatically or via the get started button), and copied api token.
- Make credential fields required
- Remove skip button and closing the onboarding until after the AI Agent has been used. This is behind a feature flag that is only true for my user right now.

#### Screenshots:
image


stack72 (approved):
🥳

app/web/src/newhotness/CopyableTextBlock.vue
wendybujalski, line 1: This might be a good component to promote into the `vue-lib` eventually if we use it more.

PR #7217, opened 2025-09-04
chore: Remove sockets from clover

jkeiser:
This removes all sockets from all Clover assets.

Fixes ENG-3222.

#### Out of Scope:

This does not bring Clover assets' prop suggestions to parity with the old socket annotations, but it does leave TODO comments where significant prop suggestions could be added. These will take a bit of time to research what schemas the prop suggestions should be for.

## How was it tested?

- [X] Integration tests pass
- [X] Manual test: Clover assets upgrade, subscribe, and discover


stack72 (approved):

stack72:
I rebuilt my local assets based on this PR, I uploaded them to my local instance of SI and run the VPC Template and things were good! Nice work @jkeiser

PR #7222, opened 2025-09-05
fix: use isOriginSecret for subscription

aaron-dernley:
FIXES: BUG-1010

Only show subscribing when theres an active subscription, not just when isControlledByDynamicFunc is true. Hopefully fixes issue where new components incorrectly displayed subscribing... on creation.

I think this preserves the original intent of https://github.com/systeminit/si/pull/7184 but uses isOriginSecret to determine the correct behaviour when not an AWS Credential.

#### Screenshots:

JAM - https://jam.dev/c/1af78257-9c84-4049-a00c-3cc5f9492882

## How was it tested?

- [X] Creating a new component in the UI no longer shows 'subscribing...' dy default

### DIFF
--- a/app/web/src/newhotness/layout_components/ComponentSecretAttribute.vue
+++ b/app/web/src/newhotness/layout_components/ComponentSecretAttribute.vue
@@ -347,7 +347,10 @@ const secretFormOpen = ref(false);
 const interstitialSpinner = computed(
   () =>
     !props.attributeTree.secret &&
-    props.attributeTree.attributeValue.isControlledByDynamicFunc,
+    props.attributeTree.attributeValue.isControlledByDynamicFunc &&
+    (props.attributeTree.prop?.isOriginSecret ||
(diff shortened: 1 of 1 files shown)


jobelenus (changes requested):
I’m pretty sure that’s not the fix. Here’s the situation:
The API call doesn’t set “secret”, it sets “controlledByFunction”So when the form finishes and bifrosting.value is false, secret is false, and “controlled” is true“Controlled” and “bifrosting” will never be true at the same time, so this computed value will always be false.DVU sets “secret”“Subscribing…” is only the interstitial display for “I’ve saved the secret and set the AV up”, once DVU finishes the job, it goes away

Without the change in my PR the experience is: “User hits the add secret button, big form goes away, little secret field shows up, it’s empty, the spinner at the top of the page stops… it’s still empty.. why is it empty… DVU runs, and now it has the secret name”

PR #7225, opened 2025-09-05
fix(sdf-server): fix cargo check by changing editor typescript path

zacharyhamm:


stack72:
@zacharyhamm I think #7246 took care of this and can be now closed?

PR #7227, opened 2025-09-05
Add instructions on how to update crates

nickgerace:
## Description

This change adds instructions on how to update crates to the dev docs. It is based on PR #7226.


nickgerace:
@jkeiser agreed! Those are good additions :) I set this to merge, but yes, good ideas!

nickgerace:
@jkeiser looks like it actually has the former already:
https://github.com/systeminit/si/blob/161ec6d1161f91b7ee58738008493366ed614eb6/DEV_DOCS.md?plain=1#L745

PR #7228, opened 2025-09-05
unified new and old navbar

wendybujalski:

Changes to the new navbar need to be ported to the old navbar until we are ready to remove all uses of stores from the old navbar. This PR unifies the designs between the two so that they look the same for now while we are using both.

#### Screenshots:

Screenshot 2025-09-05 at 1 32 10 PM

#### Out of Scope:

Eventually we will want to remove the old navbar entirely and replace it with the new navbar.

### DIFF
--- a/app/web/src/components/layout/navbar/ChangeSetPanel.vue
+++ b/app/web/src/components/layout/navbar/ChangeSetPanel.vue
@@ -42,22 +42,22 @@
       </DropdownMenuButton>
     </label>
 
-    <VButton
+    <NewButton
       v-tooltip="{
         content: 'Create Change Set',
       }"
+      data-testid="create-change-set-button"
       icon="git-branch-plus"
-      size="sm"
       tone="action"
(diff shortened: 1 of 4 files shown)


app/web/src/components/layout/navbar/ChangeSetPanel.vue
```diff
@@ -42,22 +42,22 @@
       </DropdownMenuButton>
     </label>
 
-    <VButton
+    <NewButton
       v-tooltip="{
         content: 'Create Change Set',
       }"
+      data-testid="create-change-set-button"
```
jkeiser, line 49: This is so e2e tests can find it?

jkeiser (approved):

app/web/src/components/layout/navbar/ChangeSetPanel.vue
wendybujalski: yep!

PR #7236, opened 2025-09-07
feat(clover): Adding an action we can use to trigger an instance refresh

stack72:
We often have to rotate nodes in an AutoScaling group - this is the best way to do that as we have autoscaling groups we can ask to do it for us

HOLD BACK ON MERGING THIS! I WANT TO TEST A FEW MORE SCENARIOS!

### DIFF
--- a/bin/clover/src/cloud-control-funcs/overrides/AWS::AutoScaling::AutoScalingGroup/actions/instanceRefresh.ts
+++ b/bin/clover/src/cloud-control-funcs/overrides/AWS::AutoScaling::AutoScalingGroup/actions/instanceRefresh.ts
@@ -1,0 +1,61 @@
+async function main(component: Input): Promise<Output> {
+  const resource = component.properties.resource?.payload;
+  if (!resource) {
+    return {
+      status: component.properties.resource?.status ?? "ok",
+      message: component.properties.resource?.message,
+    };
+  }
+
+  if (!resource.AutoScalingGroupName) {
+    return {
+      status: "error",
+      payload: resource,
+      message: "No AutoScaling Group name found",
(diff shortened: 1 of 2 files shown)


aaron-dernley (approved):

bin/clover/src/cloud-control-funcs/overrides/AWS::AutoScaling::AutoScalingGroup/actions/instanceRefresh.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  const child = await siExec.waitUntilEnd("aws", [
+    "autoscaling",
+    "start-instance-refresh",
+    "--auto-scaling-group-name",
+    resource.AutoScalingGroupName,
+    "--strategy",
+    "Rolling",
+    "--preferences",
+    '{"MinHealthyPercentage":100,"InstanceWarmup":300}',
```
johnrwatson: Usually it's a capacity constraint system like, you need 25% throughput at all times, for us these are what we need, but that's not what I'd consider "normal"

We should have these values driven from the attribute tree, if they're not set we should fail this action explaining they're missing

(Rather than needing to rotate one at a time like that with such a small number)

stack72:
Ok, I was a little worried about the semantics about replacing instances via the ASG as I thought it didn't interact well with the ALBs. But it seems the following is the flow:

```
You don’t manually deregister anything. The ASG handles registration of new instances to the target groups / ELB and deregisters old ones automatically when it terminates them.
```

So this is ready to ship now!

johnrwatson (approved):

PR #7245, opened 2025-09-08
Fix: api & form observability

jobelenus:

Getting both `watchedForm` and `watchedApi` with http request data squared away.

#### Out of Scope:

One tiny missed form reset from last PR

## How was it tested?

Make an API call

### DIFF
--- a/app/web/src/newhotness/api_composables/index.ts
+++ b/app/web/src/newhotness/api_composables/index.ts
@@ -289,14 +289,29 @@ export class APICall<Response, Args> {
     data: D,
     params?: URLSearchParams,
   ): Promise<DoResponse<Response, Args>> {
+    const start = performance.now();
     this.obs.requested.value = true;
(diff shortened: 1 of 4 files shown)


app/web/src/newhotness/api_composables/index.ts
```diff
@@ -327,9 +342,19 @@ export class APICall<Response, Args> {
       data: formattedData,
       validateStatus: (_status) => true, // don't throw exception on 4/5xxx
     });
+    const end = performance.now();
+    this.obs.span.setAttributes({
+      "http.status_code": req.status,
+      // "watched" API "duration" will contain "how long it took for the data to update"
+      // this will just be the http call time + latency, good to have both
+      "http.duration": end - start,
```
britmyerss, line 350: nice!!

britmyerss (approved):

PR #7249, opened 2025-09-08
feat: add the ability to run a template management function to the mcp

aaron-dernley:

- Refactored componentGenerateTemplate.ts to match new Template tools, this is now templateGenerate.ts
- Adds Template List and Template Run functionality to the SI MCP Server.

#### Videos:

Template Generate - https://jam.dev/c/04f426e6-5c39-45f6-bfb4-35feb831bbde
Template List and Template Run - https://jam.dev/c/d8072225-45cb-4bbd-91f5-e9b19cd03810

## How was it tested?

- [X] Manual test: used a local dev version of the MCP server to run the new tools against the Production UI.


bin/si-mcp-server/src/tools/templateFind.ts
```diff
@@ -0,0 +1,121 @@
...
+import {
+  errorResponse,
+  generateDescription,
+  successResponse,
+  withAnalytics,
+} from "./commonBehavior.ts";
+import { ChangeSet } from "../data/changeSets.ts";
+
+const name = "template-find";
+const title = "Find template schemas";
```
nickgerace: Addressing Adam's feedback, I believe the guts of this are actually correct. However, we need to be explicit that this is a LIST call and not a FIND call. This is very similar to the component list and change set list tools. We should replace all "find" language with "list" language.

bin/si-mcp-server/src/tools/templateFind.ts
```diff
@@ -0,0 +1,121 @@
...
+const description =
+  `<description>Find template schemas in a change set. Templates are special schemas in the 'Template' category that can be used to generate multiple component instances. Returns an array of template schemas with their IDs, names, categories, and installation status. On failure, returns error details.</description><usage>Use this tool to discover available templates that can be run using the component-run-template tool. Templates are schemas in the 'Template' category that have management functions for generating component instances. Use the returned schemaId with component-create to instantiate template components.</usage>`;
```
nickgerace:
```suggestion
  `<description>Find template schemas in a change set. Templates are special schemas in the 'Templates' category that can be used to generate multiple component instances. Returns an array of template schemas with their IDs, names, categories, and installation status. On failure, returns error details.</description><usage>Use this tool to discover all available templates that can be ran using the component-run-template tool. Templates are schemas in the 'Templates' category that can be ran to generate component instances.</usage>`;
```

bin/si-mcp-server/src/tools/templateFind.ts
```diff
@@ -0,0 +1,121 @@
...
+                  error instanceof Error ? error.message : String(error)
+                }`,
+            });
+          }
+        }
+
+        try {
+          const schemasApi = new SchemasApi(apiConfig);
+
+          // Search for schemas in the Template category
```
nickgerace:
```suggestion
```

bin/si-mcp-server/src/server.ts
```diff
@@ -51,6 +53,8 @@ export function createServer(): McpServer {
   componentEraseTool(server);
   componentRestoreTool(server);
   componentGenerateTemplateTool(server);
+  componentRunTemplateTool(server);
```
nickgerace:
```suggestion
runTemplateTool(server);
```

bin/si-mcp-server/src/server.ts
```diff
@@ -24,6 +24,8 @@ import { componentRestoreTool } from "./tools/componentRestore.ts";
 import { generateSiUrlTool } from "./tools/generateSiUrl.ts";
 import { componentGenerateTemplateTool } from "./tools/componentGenerateTemplate.ts";
 import { upgradeComponentsTool } from "./tools/upgradeComponents.ts";
+import { componentRunTemplateTool } from "./tools/componentRunTemplate.ts";
```
nickgerace:
```suggestion
import { runTemplateTool } from "./tools/runTemplate.ts";
```

bin/si-mcp-server/src/tools/componentRunTemplate.ts
```diff
@@ -0,0 +1,193 @@
...
+  errorResponse,
+  generateDescription,
+  successResponse,
+  withAnalytics,
+} from "./commonBehavior.ts";
+import { AttributesSchema } from "../data/components.ts";
+import { buildAttributesStructure } from "../data/schemaAttributes.ts";
+
+const name = "component-run-template";
+const title = "Run a template from a template component";
```
nickgerace: Addressing Adam's feedback, running a template should not directly name components as part of the process. In fact, in a future iteration, we should probably delete the component after the template has ran (unless the user wants to keep it around for a factory).

For this PR, I don't think we should delete the template after it has ran because some templates can take a long time to run and blocking forward progress for `claude` is a concern we need to think about (e.g. building a big VPC with an EKS cluster inside it).

So that being said... `run-template` is the move rather than `component-run-template`.

bin/si-mcp-server/src/tools/componentRunTemplate.ts
```diff
@@ -0,0 +1,193 @@
...
+const description =
+  `<description>Execute a management function on a template component to run/instantiate the template and create new components based on the template definition. This tool is used to run templates that were previously generated using component-generate-template or found using template-find. The template execution will create new components according to the template's definition prefixed with a Name Prefix which can be automatically genereated or chosen by the user.</description><usage>Use this tool to run a template component by executing its management function. First, use template-find to discover available templates, and finally use this tool to run the template.</usage>`;
```
nickgerace: We should remove all references to management functions. There's three main things the LLM should be aware of:

1. I can generate templates (DONE in a prior PR)
2. I can list templates
3. I can run templates

Yes, running templates results in generated components. Yes, listing templates is really listing schemas under a certain category. Yes, a template function is really just a spicy management function. However, to a user, this all falls under the "template" domain.

nickgerace: Also we should make sure `template-find` and relevant language is replaced with `template-list`, etc.

In fact, we may want to unify the three tools: `template-list`, `template-generate`, `template-run` perhaps?

bin/si-mcp-server/src/tools/componentRunTemplate.ts
```diff
@@ -0,0 +1,193 @@
...
+
+const ComponentRunTemplateInputSchemaRaw = {
+  changeSetId: z.string().describe(
+    "The change set to run the template in",
+  ),
+  schemaName: z.string().describe(
+    "The schema name for the template component to create and run",
+  ),
```
nickgerace: This should probably now be the `templateName` because `template-list` should return a list templates that have `templateNames`.

bin/si-mcp-server/src/tools/componentRunTemplate.ts
```diff
@@ -0,0 +1,193 @@
...
+  attributes: AttributesSchema.describe(
+    "attributes to set on the component before running a template; this *must* include setting a raw value for /domain/Name Prefix which ends with a '-'; for AWS resources, this *must* include setting a subscription (usually from /secrets/AWS Credential on an AWS Credential component) for the AWS Credential, and *must* include setting a subscription (usually from /domain/region on a Region component) for /domain/extra/Region as well",
+  ),
+  componentTemplateName: z.string().describe(
+    "The name of the template component to create and run",
+  ),
```
nickgerace: This is interesting. We need a name for that template component we create internally, but we don't want to expose that concept to the user... let's think on it but I think removing the "component" bit of this language is good.

bin/si-mcp-server/src/tools/componentRunTemplate.ts
```diff
@@ -0,0 +1,193 @@
...
+        description,
+        "componentRunTemplate",
+        ComponentRunTemplateOutputSchema,
+      ),
+      inputSchema: ComponentRunTemplateInputSchemaRaw,
+      outputSchema: ComponentRunTemplateOutputSchemaRaw,
+    },
+    async (
+      { changeSetId, schemaName, attributes, componentTemplateName },
+    ): Promise<CallToolResult> => {
```
nickgerace: Otherwise, I think the guts of both tools at the function level are good, barring renaming to match the language that we desire.

bin/si-mcp-server/src/tools/componentRunTemplate.ts
nickgerace: @aaron-dernley ^^ I think we should start here as a guiding light: unifying under `template-list`, `template-generate`, `template-run`

PR #7258, opened 2025-09-09
Vue unit tests

jobelenus:
# How does this PR change the system?

Set up `vitest` with `@vue/test-utils`

Locally run tests with `pnpm run test --silent=false`, that way you see console.logs you put in there, without the flag, no console logs!

I made a single test that would have caught the bug that this PR fixed: https://github.com/systeminit/si/pull/7222

You can create a `XXX.test.ts` file anywhere inside `/newhotness/` and it will run! I prefer having test files next to source files. That makes the test file easy to see, and a reminder that if you're editing the file, odds are you should be adding a test 😉 

# Oh, And.. Uh...

Some of our TS types were not correct based on the MV data that was coming back. Most of this was that `Option<T>`  the "None" value is coming over the wire as `null`, not `undefined` as one might expect, since `foo?: T` is the TS "optional" and resolves to `undefined`.

It wasn't very feasible to get `serde` to match `tsc` expectations, so I changed the typing; which of course required a few touch ups in the code base.

## How do we work with this?!

There are **four** things you need to get a test running:

### First
Every single test needs the plugin data and a tiny bit of set up. Import this from `testing/index`. It goes in the mount call:
```
    global: {
	  plugins,
```

### Second
The context you want to operate in (e.g. `useWithContext()`). I made a `context1` file. If you can re-use these, please do. If you need it to be different—make a new one and use that one! You _will at least need_ CONTEXT, you may need more. The context data gets put in the mount call:
```
    global: {
      provide: {
        ATTRIBUTEINPUT,
        CONTEXT,
        ATTRIBUTE_ERRORS,
      },
```

### Third
The "fixture data" that you'll either send in as a prop (see how easy that is?!) or set in tan stack query client so the component tan stack call can look it up. I made a `fixture1` file. We *definitely* want this fixture data to _ONLY_ be MV data (or whatever data is coming over the wire). If that data is mutated, like in this case it is, perform that mutation within the test. This is the only way to stay sane and ensure that we can _always_ regenerate the fixture data at a later time without wondering how we broke a test.

### Fourth
You need to mock `heimdall`. Due to the fact that this the mock is intercepting imports, we can't extract this to a one-liner import that does all the work in the other file. So, its a ~10 line copypasta instead. In order to get this code reusable for both the test and the app, I moved a bit into `heimdall_inner`. Let's continue to follow that pattern as needed. **_Note:_** I _didn't_ need the getPossibleConnections _actually_ implemented, I just needed it defined. If I did need it implemented, it should be returning fixture data!

And then, finally, write your test 😄 

## Out of Scope:

Many many tests 😮‍💨 

**We are going to want rust code generation and fixture data generation rather soon**. If we've manually created dozens of fixture files, and we start changing types, we're in for a world of pain!

## How was it tested?

1. `pnpm run test`, clean tests
2. Take away the fix in that PR, with a diff like this:
```
--- a/app/web/src/newhotness/layout_components/ComponentSecretAttribute.vue
+++ b/app/web/src/newhotness/layout_components/ComponentSecretAttribute.vue
@@ -347,10 +347,10 @@ const secretFormOpen = ref(false);
 const interstitialSpinner = computed(
   () =>
     !props.attributeTree.secret &&
-    props.attributeTree.attributeValue.isControlledByDynamicFunc &&
-    (props.attributeTree.prop?.isOriginSecret ||
-      (props.attributeTree.attributeValue.externalSources &&
-        props.attributeTree.attributeValue.externalSources.length > 0)),
+    props.attributeTree.attributeValue.isControlledByDynamicFunc // &&
+    // (props.attributeTree.prop?.isOriginSecret ||
+    //   (props.attributeTree.attributeValue.externalSources &&
+    //     props.attributeTree.attributeValue.externalSources.length > 0)),
 );
```
3. Run the test, see failure:
```
src/newhotness/layout_components/ComponentSecretAttribute.test.ts > testing component secret attributes
AssertionError: expected 'subscribing...subscribing...' to be '' // Object.is equality

- Expected
+ Received

+ subscribing...subscribing...

 ❯ src/newhotness/layout_components/ComponentSecretAttribute.test.ts:105:16
    103|     .join("");
    104|   // ec2 instance should have a secret field that is blank
    105|   expect(text).toBe("");

```
App continues to work!


app/web/src/newhotness/testing/fixture1.ts
wendybujalski, line 1: Yeah I can see why fixture data generation will be essential down the line!

app/web/package.json
```diff
@@ -21,7 +21,7 @@
     "dev": "pnpm run start",
     "start": "vite",
     "serve": "vite preview --port 8080",
-    "test": "echo 'Sorry, no tests yet!'",
```
fnichol, line 24: 💥

app/web/src/api/sdf/dal/prop.ts
```diff
@@ -14,6 +16,8 @@ export enum PropKind {
 export interface Prop {
   id: PropId;
   kind: PropKind;
+  childKind?: null | PropKind;
+  widgetKind?: null | PropertyEditorPropWidgetKind;
```
fnichol, line 20: Ah darn it, that is unfortunate! But excellent catch

PR #7260, opened 2025-09-09
feat(luminork): Allow searching for components that match a schema variant category

stack72:


britmyerss (approved):
we could consider using or developing other MVs for this - but good idea to start here

stack72:
> we could consider using or developing other MVs for this - but good idea to start here

yes, this will be replaced by the search mechanism we talked about earlier! I need to allow the MCP Server to do something before we get to that work though

PR #7264, opened 2025-09-10
fix: git_info sha hash length not guarenteed

sprutton1:

This ensures that the hashes we generate are always 8 characters long. In some cases, we were generating 9 character long hashes, causing mismatched during promote vs. publish steps 

This also swaps out `git` for `gitMinimal` in the flake. This is what we used to use and this change was unintentional. It's possible the minimal version of git was less prone to generating 9 characters short hashes?

## How was it tested?

CI will be a good test for this. Locally, this works for me

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -66,7 +66,7 @@
           b3sum
           buck2
           deno
-          git
+          gitMinimal
           makeWrapper
           nodePkgs.pnpm
           nodejs
--- a/prelude-si/git/git_info.py
+++ b/prelude-si/git/git_info.py
@@ -70,6 +70,7 @@ def parse_git_show() -> Dict[str, str]:
         "git",
         "show",
         "--no-patch",
+        "--abbrev=8",
         f"--format=format:{format_str}",
     ]
     result = subprocess.run(git_show_cmd, capture_output=True)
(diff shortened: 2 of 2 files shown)


zacharyhamm:
The description of abbrev says:

> Instead of using the default number of hexadecimal digits (which will vary according to the number of objects in the repository with a default of 7) of the abbreviated object name, use <n> digits, or as many digits as needed to form a unique object name. An <n> of 0 will suppress long format, only showing the closest tag.


This means it would have generated a 9 no matter what (even with abbrev=8). Is it possible to make the place we parse these flexible?

I think what happened was there was a conflict with another hash prefix and it had to extend it to 9 characters to be unique.

sprutton1:
> The description of abbrev says:
> 
> > Instead of using the default number of hexadecimal digits (which will vary according to the number of objects in the repository with a default of 7) of the abbreviated object name, use  digits, or as many digits as needed to form a unique object name. An  of 0 will suppress long format, only showing the closest tag.
> 
> This means it would have generated a 9 no matter what (even with abbrev=8). Is it possible to make the place we parse these flexible?
> 
> I think what happened was there was a conflict with another hash prefix and it had to extend it to 9 characters to be unique.

There's probably a more elegant solution, but I made it smoosh the resultant string down to 8 characters just to ensure we cover this case. Version strings include dates, so I don't think this will ever be an issue for us.

PR #7266, opened 2025-09-10
NewButton no longer changes size on loading

wendybujalski:

This PR fixes the spacing on NewButton so that it does not change size if it gains or loses an icon. The DebugPage has a button garden showing all the possible variations.

It also replaces the non-standard pill on the ChangeSetButton with the standard one.

#### Screenshots:

Screenshot 2025-09-10 at 1 36 55 PM

#### Out of Scope:

This PR does not implement the secondary number pill shown in Figma as we are not using it anywhere yet.

This PR also does not finish eradicating VButton or IconButton from the few places it is left in the vue-lib and old UI.

## How was it tested?

Lots of manual testing, the DebugPage confirms that all possible variants match Figma.

### DIFF
--- a/app/web/src/newhotness/ComponentHistory.vue
+++ b/app/web/src/newhotness/ComponentHistory.vue
@@ -6,15 +6,9 @@
     @scrollend="handleScrollEnd"
   >
     <div class="flex flex-row justify-between items-center gap-xs pb-xs">
+      <NewButton class="grow" label="Expand All" @click="handleAll('expand')" />
       <NewButton
         class="grow"
-        size="xs"
-        label="Expand All"
-        @click="handleAll('expand')"
-      />
-      <NewButton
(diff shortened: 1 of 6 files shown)


app/web/src/newhotness/ComponentHistory.vue
```diff
@@ -6,15 +6,9 @@
     @scrollend="handleScrollEnd"
   >
     <div class="flex flex-row justify-between items-center gap-xs pb-xs">
+      <NewButton class="grow" label="Expand All" @click="handleAll('expand')" />
       <NewButton
         class="grow"
-        size="xs"
```
wendybujalski, line 11: We want to avoid non-standard NewButton sizes as much as possible!

jobelenus (approved):

PR #7273, opened 2025-09-11
fix(rootfs): Add Azure CLI back in via rootfs

JakeGinnivan:

This installs the Azure CLI during the rootfs prelude script. As far as I can tell, https://github.com/systeminit/si/pull/6896 moved the CLI tools which are available to functions from the `langJsExtraPkgs` list in `flake.nix` to prelude-si/rootfs/rootfs_build.sh.

## How was it tested?

As far as I can see, there is no precedent for testing a CLI tool is available to functions, but if there is, point me at it.

Draft PR until I test manually.

- [X] Integration tests pass
- [ ] Manual test: new functionality works in UI


britmyerss:
Thanks @JakeGinnivan for the contribution!! We needed to add a bit more to support this, PR is here: https://github.com/systeminit/si/pull/7281 and will be shipping tomorrow!

PR #7283, opened 2025-09-12
fixed first time user default workspace redirect

wendybujalski:

This PR fixes an issue where first time users were not being redirected to their default workspace after setting up their account.

It also updates the UI/UX so that users whose email has not been verified get better feedback and are redirected properly when they are locked on the user profile step.

Manually tested a bunch with the local auth stack for every possible scenario I could think of!

### DIFF
--- a/app/auth-portal/src/pages/ProfilePage.vue
+++ b/app/auth-portal/src/pages/ProfilePage.vue
@@ -128,16 +128,15 @@
 
             <VButton
               iconRight="chevron--right"
-              :disabled="validationState.isError"
+              :disabled="validationState.isError || failedEmailVerify"
               :requestStatus="updateUserReqStatus"
               loadingText="Saving your profile..."
               successText="Updated your profile!"
               tone="action"
               variant="solid"
+              :label="saveButtonLabel"
               @click="saveHandler"
-            >
-              Save
-            </VButton>
+            />
           </Stack>
         </form>
       </div>
@@ -175,6 +174,8 @@ const DISCORD_TAG_REGEX =
 const { validationState, validationMethods } = useValidatedInputGroup();
 const authStore = useAuthStore();
 const workspacesStore = useWorkspacesStore();
+// Reactively load the workspaces so that the workspace redirect happens for first time users
+workspacesStore.refreshWorkspaces();
 const router = useRouter();
 
 const loadUserReqStatus = authStore.getRequestStatus("LOAD_USER");
@@ -184,6 +185,10 @@ const updateUserReqStatus = authStore.getRequestStatus("UPDATE_USER");
 const storeUser = computed(() => authStore.user);
 const draftUser = ref<User>();
 const isOnboarding = ref<boolean>();
+
+const refreshAuth0ReqStatus = authStore.getRequestStatus(
+  "REFRESH_AUTH0_PROFILE",
+);
 
 useHead({ title: "Profile" });
 
@@ -201,6 +206,12 @@ function checkUserOnboarding() {
 
 watch(storeUser, checkUserOnboarding, { immediate: true });
 
+watch([refreshAuth0ReqStatus, storeUser], () => {
+  if (storeUser.value?.emailVerified) {
+    failedEmailVerify.value = false;
+  }
+});
+
 onBeforeMount(() => {
   // normally when landing on this page, we should probably make sure we have the latest profile info
   // but we already load user info with CHECK_AUTH so can skip if it was just loaded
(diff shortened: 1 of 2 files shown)


app/auth-portal/src/pages/ProfilePage.vue
```diff
@@ -128,16 +128,15 @@
 
             <VButton
               iconRight="chevron--right"
-              :disabled="validationState.isError"
+              :disabled="validationState.isError || failedEmailVerify"
```
wendybujalski, line 131: Disable the profile save button if the user's email is not verified and they tried to save their profile once already.

app/auth-portal/src/pages/ProfilePage.vue
```diff
@@ -128,16 +128,15 @@
...
               :requestStatus="updateUserReqStatus"
               loadingText="Saving your profile..."
               successText="Updated your profile!"
               tone="action"
               variant="solid"
+              :label="saveButtonLabel"
```
wendybujalski, line 137: Change the save button to tell the user they need to verify their email if they try to move on without verifying their email.

app/auth-portal/src/pages/ProfilePage.vue
```diff
@@ -175,6 +174,7 @@ const DISCORD_TAG_REGEX =
 const { validationState, validationMethods } = useValidatedInputGroup();
 const authStore = useAuthStore();
 const workspacesStore = useWorkspacesStore();
+workspacesStore.refreshWorkspaces();
```
wendybujalski, line 178: This is required to load the available workspaces for the user so that the default workspace redirect works!

app/auth-portal/src/pages/ProfilePage.vue
```diff
@@ -216,7 +216,15 @@ onBeforeMount(() => {
...
   if (validationMethods.hasError()) return;
 
-  // if this is first time, we will take them off profile page after save
+  if (!authStore.user?.emailVerified) {
+    await verifyEmail();
+    if (!authStore.user?.emailVerified) {
+      youMustVerifyYourEmailAddress();
+      return;
+    }
+  }
```
wendybujalski, line 237: If we reach this point without confirming the user's email is verified, send the check one more time and if it fails, we block the user here until they verify.

app/auth-portal/src/pages/ProfilePage.vue
```diff
@@ -244,11 +252,14 @@ const saveHandler = async () => {
     }
 
     const completeProfileReq = await authStore.COMPLETE_PROFILE({});
+
     if (completeProfileReq.result.success) {
-      if (authStore.user?.emailVerified && workspacesStore.defaultWorkspace) {
+      if (workspacesStore.defaultWorkspace) {
```
wendybujalski, line 270: This was not working because the workspacesStore did not have the workspaces data loaded (see line 177)

app/auth-portal/src/pages/ProfilePage.vue
```diff
@@ -244,11 +252,14 @@ const saveHandler = async () => {
...
+        // all new users should hit here and go to their first workspace
         tracker.trackEvent("workspace_launcher_widget_click");
         window.location.href = `${API_HTTP_URL}/workspaces/${workspacesStore.defaultWorkspace.id}/go`;
       } else {
+        // last resort if the new user doesn't have a default workspace
```
wendybujalski, line 275: This should never happen unless a brand new user is not provisioned a workspace for some reason, probably because something is broken on our end!

app/auth-portal/src/router.ts
```diff
@@ -64,6 +64,13 @@ export const routerOptions: RouterOptions = {
         return { name: "workspaces" };
       },
     },
+    {
+      path: "/download",
+      name: "download",
+      redirect() {
+        return { name: "workspaces" };
+      },
+    },
```
wendybujalski: Saw a few users hit this URL and 404 in the last few days in Posthog, this is where we used to have the SI launcher way back when, I changed it to a redirect to avoid more 404s

wendybujalski:
Ah I'll fix the lint when I get back home 😅

app/auth-portal/src/pages/ProfilePage.vue
wendybujalski: The reason we don't just disable the button entirely if they have not verified is because we want to be able to check verification on clicking the button since it takes time to update. The user may have already verified but the auth portal does not know yet.

stack72 (approved):

PR #7285, opened 2025-09-13
Only rebuild MVs for added/removed modules

nickgerace:
This change ensures that we only rebuild deployment MVs for modules that have been added or removed.

## How was it tested?

- [X] Integration tests pass


nickgerace:
Closing in favor of #7286

PR #7286, opened 2025-09-13
Only rebuild MVs for added/removed modules

nickgerace:
This change ensures that we only rebuild deployment MVs for modules that have been added or removed.


nickgerace:
Putting this on hold until #7187 is merged and operationalized

nickgerace:
Closing since @britmyerss is going to take a different approach by using the module index client in edda and having a thin request payload

PR #7294, opened 2025-09-16
chore(deps): Upgrade fuser to 0.16.0 to fix CVE

stack72:
Fixes: https://github.com/systeminit/si/security/dependabot/434


fnichol (approved):
Looking good, and looks like the merge pipeline passed

PR #7310, opened 2025-09-22
allow users to enter names with a wider set of characters

wendybujalski:
This PR changes the regex used in the auth portal to allow for users to enter more characters. This is to address the common user problem of not being able to enter their name properly if it contains an accent or other uncommon character.

The old regex only allowed for a-z, A-Z, 0-9, and -,_@/+

The new regex allows for those characters as well as a wider range of uppercase and lowercase latin characters with accents and other diacritics. This should allow users to enter any name written with latin characters without any security risks.

### DIFF
--- a/bin/auth-api/src/lib/validation-helpers.ts
+++ b/bin/auth-api/src/lib/validation-helpers.ts
@@ -14,6 +14,6 @@ export function validate<Z extends Zod.Schema>(obj: any, schema: Z) {
   }
 }
 
-export const ALLOWED_INPUT_REGEX = /^[a-zA-Z0-9-.,_@/+ ]*$/;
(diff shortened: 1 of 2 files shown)


jobelenus (approved):

bin/auth-api/src/lib/validation-helpers.ts
```diff
@@ -14,6 +14,6 @@ export function validate<Z extends Zod.Schema>(obj: any, schema: Z) {
   }
 }
 
-export const ALLOWED_INPUT_REGEX = /^[a-zA-Z0-9-.,_@/+ ]*$/;
```
nickgerace, line 17: 👍

nickgerace (approved):

PR #7317, opened 2025-09-23
fix: allow bindings to be retrieved for func used in subscriptions

jkeiser:
The asset editor does not show attribute functions if said attribute functions are being used as transformation functions in subscriptions. This is because we cannot generate the binding during the `/funcs` endpoint. The error `UnexpectedValueSource(ValueSubscription(...))` is reported and the function is silently omitted from the `/funcs` response.

This only happens right now on workspaces that were migrated with transformation functions, or (theoretically) workspaces PUT /attributes to luminork or sdf created a subscription using an attribute function.

This adds the ability to serialize subscriptions in frontend types, preventing the error from occurring and allowing the function to be seen in the asset editopr. However, since we don't actually *send* the binding (we only send bindings attached to schema variants, not components), we don't need to adjust the frontend.

Fixes BUG-1029.

**Factor Budget:** converted bindings tests to use modern dal test helpers.

## How was it tested?

- [X] Integration tests pass
- [X] New integration test fails before PR, succeeds after
- [X] Manual test: asset editor shows attribute function even if it has been used in a subscription via PUT /attributes


lib/dal/src/func/binding/attribute_argument.rs
```diff
@@ -107,6 +115,14 @@ impl AttributeArgumentBinding {
...
                     AttributeFuncArgumentSource::StaticArgument(static_value.value)
                 }
+                ValueSource::Secret(secret_id) => AttributeFuncArgumentSource::Secret(secret_id),
+                ValueSource::ValueSubscription(subscription) => {
+                    let (component_id, path) = subscription.path_from_component(ctx).await?;
+                    AttributeFuncArgumentSource::ValueSubscription { component_id, path }
+                }
+                // jkeiser 2024-06-03: We should serialize all the value sources here, because
+                // functions can now be bound to AVs, which means any value you can use in an
+                // AV, you may see in a function binding.
```
nickgerace, line 125: What's going on here?

PR #7318, opened 2025-09-23
chore(auth-api): Upgrade the Auth API deps

stack72:
The version of TypeScript we used was below 5.x. A recent change to a version of posthog-core brought in TypeScript 5.x changes so we needed to upgrade

Why switch to tsx
Needed to move to TypeScript ≥5 (to support export type *) and our old ts-node/esm + Node flags path was brittle and broke ESM resolution.
tsx runs TS/ESM without experimental flags and avoids resolver quirks from ts-node/esm.

What changed
Dev scripts now use tsx:
dev / dev:run: tsx watch src/main.ts
boot-ts: tsx src/main.ts
Tests load with --loader tsx (no more --experimental-specifier-resolution=node).

Benefits
More reliable ESM + TS 5 execution (no loader dance).
Faster startup/reloads (esbuild under the hood).
Simpler scripts and fewer version/peer-dep landmines.

Impact / risk
Dev/test only; no prod runtime change.
Minimal risk: tsx compiles on the fly, but build artifacts still come from tsc.

Validation
Local dev and boot-ts run cleanly.
Test suite runs with the tsx loader.
No changes required to tsc build or runtime node entrypoint.


stack72:
I need to test local web app flow - bit the auth api works well locallu here

johnrwatson:
Added `vulnerability-fix` label as this is the same kind of direction as the suggested change

PR #7323, opened 2025-09-23
docs: adding what is SI, mermaid diagrams

adamhjk:
Adds a page to the docs site that describes what system initiative is to a technical audience. Adds the ability to render mermaid diagrams to the docs site.

Adds a new 'What is System Initiative' page to the docs site.

#### Screenshots:
image

## How was it tested?

By hand, as god intended it.


adamhjk:
Okay, so the issue is that 'widdershins' which gets used to generate the api documentation has the dep on from-data 3.0.0

PR #7325, opened 2025-09-23
chore: Remove ArgumentTargets support from the graph

jkeiser:
Now that all socket connections are gone everywhere, we can now proceed to remove the ability to represent socket connections in the graph.

Because #7207 removed most *uses* of socket connections, this only removes some CTs and the migration code. All other endpoints are untouched.

#### Out of Scope:

* inferred connections: there is still code related to parentage and inferred connections.
* connections/incomingConnections[]/etc. in component frontend payloads and management functions: these will get fully removed after inferred connection code is gone.

## How was it tested?

- [X] Integration tests pass (particularly deserialization)
- [X] Manual test: can still load snapshots created before this PR and manipulate them in the UI
- [X] Manual test: can still create components make subscriptions and see DVU flow
- [X] Manual test: can still run the graph validator in the admin interface


lib/dal/src/workspace_snapshot/node_weight/attribute_prototype_argument_node_weight.rs
```diff
@@ -46,30 +50,37 @@ impl AttributePrototypeArgumentNodeWeight {
...
 
     pub fn content_store_hashes(&self) -> Vec<ContentHash> {
         vec![]
     }
 
     pub fn node_hash(&self) -> ContentHash {
-        self.content_hash()
+        let mut content_hasher = ContentHash::hasher();
+        // This makes the node_hash() the same as before (we used "" when we had targets = None)
+        content_hasher.update("".as_bytes());
```
zacharyhamm, line 82: Not to be fixed here but this does remind me of a wart in the whole system which is that we should have included IDs in the node hashes

zacharyhamm (approved):
Fantastic stuff. Goodbye argument targets!

PR #7326, opened 2025-09-23
Fix: queue the index update fn, add some more obs, use the patched doc dont refetch

jobelenus:

> If you "do nothing" and sit in a change set in the UI while multiple applies/replays are happening, you may hit an infinite lobby or an infinite skeleton (e.g. sitting in the grid and seeing the skeleton forever).

I *believe* this is happening for a few reasons:
1. IndexMvPatches aren't queued until after all the PatchBatches complete
2. Edda is serving up IndexMvPatches duplicated & out order.
    3. We are seeing patch1, patch2, patch1 in honeycomb

**Resolution:**
1. IndexMv patches should be queued just like normal patches
2. Don't re-fetch the doc, just use the returned patched doc
3. Don't delete if we don't have a patched MV list (covering off on bad things that shouldn't happen)
4. Adding some more obs for honeycomb.

### DIFF
--- a/app/web/src/workers/webworker.ts
+++ b/app/web/src/workers/webworker.ts
@@ -982,9 +982,7 @@ const handleIndexMvPatch = async (db: Database, msg: WorkspaceIndexUpdate) => {
     if (indexExists) span.setAttribute("indexExists", indexExists?.toString());
     else {
       span.end();
-      debug(
-        `${msg.meta.toIndexChecksum} doesn't exist, ignoring index patch WTF`,
-      );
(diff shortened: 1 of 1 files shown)


app/web/src/workers/webworker.ts
```diff
@@ -998,6 +998,7 @@ const handleIndexMvPatch = async (db: Database, msg: WorkspaceIndexUpdate) => {
     );
     span.setAttribute("previousIndexes", JSON.stringify(previousIndexes));
     if (previousIndexes.length === 0) {
+      span.setAttribute("ragnaork", true);
```
zacharyhamm: ragnarok?

PR #7327, opened 2025-09-24
chore: clover schema update

systeminitbot:


stack72:
can't merge this as it bumps the version of the deno lockfile version which breaks local dev

PR #7330, opened 2025-09-24
chore: remove FrameContains (parentage) from codebase

jkeiser:
Now that parents have been removed from all graphs in prod and tools, we can remove parentage from the codebase / graph entirely.

This PR remove *all* references to the `FrameContains` edge in the codebase, and any methods that use them which are related to parents / children. It also re-types parentId as "always undefined" in the frontend (only diagram code uses it, which will get deleted when stores go out of fashion).

#### Out of Scope:

* Removing the edge entirely awaits a new graph version (probably split graph). If you resurrect an ancient snapshot or localdev workspace, it probably *won't* produce unexpected results, but it *could.*
* Removing inferred connections will come next, but this seemed complete enough by itself.

## How was it tested?

- [X] Integration tests pass


app/web/src/api/sdf/dal/component.ts
```diff
@@ -57,7 +57,7 @@ export interface RawComponent {
   resourceId: string;
   id: ComponentId;
   componentType: ComponentType;
-  parentId?: ComponentId;
+  parentId?: undefined;
```
zacharyhamm, line 60: Why not just remove the property?

jkeiser: I started with that; but I didn't want to have to surgically edit the code that uses the properties (it's all unused code, but it's a separate project to dump it all). Honestly didn't even need to do this, but it felt a little more honest.

PR #7334, opened 2025-09-24
Clover: fetching Hetzner schema from public URL and saving to a file

jobelenus:

On the tin

#### Out of Scope:

Everything that is next

## How was it tested?

`deno task run fetch-schema --provider=hetzner`

That is how we got the JSON file in there.
Now we want to test it in github CI


sprutton1:
https://github.com/systeminit/si/actions/runs/17989277673

sprutton1 (approved):

stack72 (commented):
@jobelenus do we want to store the full .json file or add it to .gitignore?

sprutton1:
> @jobelenus do we want to store the full .json file or add it to .gitignore?

@stack72 

It's equivalent to the `cfdb.json` file for the [aws bits](https://github.com/systeminit/si/blob/main/lib/jsr-systeminit/cf-db/cf-schema.json). We need it so we can generate the diffs if the api changes.

stack72:
> > @jobelenus do we want to store the full .json file or add it to .gitignore?
> 
> @stack72
> 
> It's equivalent to the `cfdb.json` file for the [aws bits](https://github.com/systeminit/si/blob/main/lib/jsr-systeminit/cf-db/cf-schema.json). We need it so we can generate the diffs if the api changes.

gotcha! I was asking the question if it was intentional :)

PR #7340, opened 2025-09-25
fix(luminork): Fix the API types for Luminork

stack72:
We are making sure that if they are an option that we mark it as such otherwise we get validation errors in our SDKs


nickgerace (approved):

nickgerace:
CI failed due to the Docker auth outage. We'll look at this PR after #7187 is merged _and_ deployed.

PR #7349, opened 2025-09-25
fix for bug 1040 - smarter selectedWorkspaceToken

wendybujalski:

This PR fixes [bug 1040](https://linear.app/system-initiative/issue/BUG-1040/the-ai-agent-banner-is-not-in-the-correct-state) by changing the `auth-store` to more intelligently select the correct `selectedWorkspaceToken` even in scenarios where the typical approach which uses `urlSelectedWorkspaceId` from the `workspacesStore` can't work because the `routerStore` is not up yet.

Now if you hit the `/go` endpoint on the auth-api the correct `selectedWorkspaceToken` will be used for that workspace. This includes when users select a workspace from the dropdown in the app. The correct `selectedWorkspaceToken` will also be selected when the user reloads the browser window.

#### Screenshots:

Screenshot 2025-09-25 at 5 05 41 PM

#### Out of Scope:

We may eventually want to change `selectedOrDefaultAuthToken` to never send a default token if a selected workspace token is not found.

## How was it tested?

The auth api change which sends the workspaceId along with the `/go` redirect is live [due to this PR](https://github.com/systeminit/si/pull/7329). This bug fix has been manually tested against that live auth stack in addition to being tested against a local auth stack. Testing included hitting the '/go' endpoint from a variety of locations in and outside of the app, reloading the page, and throwing malformed URLs at the fix as well.

### DIFF
--- a/app/web/src/store/auth.store.ts
+++ b/app/web/src/store/auth.store.ts
@@ -62,7 +62,6 @@ export const useAuthStore = () => {
         !_.isEmpty(state.tokens) && state.user?.pk,
       selectedWorkspaceToken: (state) => {
         const workspacesStore = useWorkspacesStore();
-        // console.log("urlSelectedWorkspaceId: ", workspacesStore.urlSelectedWorkspaceId);
         if (workspacesStore.urlSelectedWorkspaceId) {
           // this case works in most scenarios except if we are asking for the
           // selectedWorkspaceToken before useRouterStore is ready (like on page refresh)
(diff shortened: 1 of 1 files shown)


app/web/src/store/auth.store.ts
```diff
@@ -71,12 +71,21 @@ export const useAuthStore = () => {
...
-          // console.log(path);
-          // console.log(path.split("/"));
-          // console.log("WORKSPACE ID: ", workspaceId);
-          if (workspaceId) {
-            return state.tokens[workspaceId];
+
+          if (path.includes("auth-connect")) {
+            // this case handles when we are arriving from the /go endpoint of the auth api
+            const pathQueryData = new URLSearchParams(window.location.search);
+            const workspaceId = pathQueryData.get("workspaceId");
```
wendybujalski: The workspaceId is sent as a query parameter from the `/go` endpoint redirect. We parse it out here.

app/web/src/store/auth.store.ts
```diff
@@ -71,12 +71,21 @@ export const useAuthStore = () => {
...
+            if (workspaceId) {
+              return state.tokens[workspaceId];
+            }
+          } else if (path.startsWith("/n/") || path.startsWith("/w/")) {
+            // this case attempts to handle standard workspace urls for the old and new ui
+            const pathParts = path.split("/");
+            const workspaceId = pathParts[2];
```
wendybujalski, line 93: The workspaceId _should_ be in this position in a properly formed URL. Malformed urls can pass an invalid id here, but that will just result in the return on line 87 being `undefined`.

app/web/package.json
```diff
@@ -150,6 +150,7 @@
     "unplugin-icons": "^0.17.4",
     "vite": "^5.4.19",
     "vite-plugin-checker": "^0.6.4",
+    "vite-plugin-node-polyfills": "^0.24.0",
```
wendybujalski: Needed to use node-polyfills to enable using `import { URLSearchParams } from "url";`

aaron-dernley (approved):

app/web/package.json
jobelenus: looks like its everything except old IE (non-edge)? https://caniuse.com/urlsearchparams

jobelenus (dismissed):

app/web/package.json
wendybujalski: Yeah, the polyfill is to deal with this issue - https://vite.dev/guide/troubleshooting.html#module-externalized-for-browser-compatibility

app/web/src/store/auth.store.ts
```diff
@@ -71,12 +70,30 @@ export const useAuthStore = () => {
...
+            const queryString = window.location.search.replace("?", ""); // grab the whole query string, removing the starting ?
+            const queryParts = queryString.split("&"); // split each of the individual parts
+            const workspacePartString = queryParts.find((part) =>
+              part.includes("workspaceId="),
+            ); // find the part we care about
+            if (workspacePartString) {
+              const workspaceId = workspacePartString.replace(
+                "workspaceId=",
+                "",
+              ); // strip out unnecessary data
```
wendybujalski, line 87: The string parsing here will always work for this very specific case. Would be better to eventually use URLSearchParams but there's too much lift on getting it working for now.

stack72 (approved):

app/web/src/store/auth.store.ts
wendybujalski: Any malformed url here will simply not succeed it getting a token just like the section below.

britmyerss (approved):

PR #7350, opened 2025-09-25
App & Web Worker are querying & firing requests way too much

jobelenus:

1. quieting down the bust logging
2. but the real fix, i made a dedupe queue and didnt fill it ☹️ 
3. fix to wrongly deleting the MvIndex atom MTM

I'm _somewhat_ confident that the "my whole grid started going skeletons" problems is related to MvIndex deletion. Here is the theory:
1. the IndexMv patches
2. then its MTM gets deleted
3. on the next index patch
4. we try to patch, but cannot find the previous atom
5. so we run niflheim
6. which replaces all the MTMs in a giant sql statement
7. and calls to bust all the tanstack caches
8. which re-runs all the UI queries

On a large enough workspace, those queries will contend and run slowly enough, as well as pushing all the data across the thread interface, you'll see skeletons.

## How was it tested?

1. Seeing that on load only one getSchemaMembers query fires.
10. On a change set, duplicate a component, wait for patches to come through
11. Then delete that component, wait for patches to come through
12. You should not see any calls to `/index` or the log message "Cannot patch" regarding the MvIndex

Updated the web worker test case to catch this problem.

image

image

### DIFF
--- a/app/web/src/workers/mjolnir_queue.ts
+++ b/app/web/src/workers/mjolnir_queue.ts
@@ -67,6 +67,7 @@ export const bustQueueAdd = (
 ) => {
   const key = `${workspaceId}-${changeSetId}-${kind}-${id}`;
   if (!_bustQueue.has(key)) {
+    _bustQueue.add(key);
     bustQueue.add(() => {
       fn(workspaceId, changeSetId, kind, id);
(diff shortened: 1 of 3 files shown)


britmyerss (approved):
Let's get this to tools!

PR #7354, opened 2025-09-26
feat: Allow Claude to define asset function, category, etc. on new schema

jkeiser:
This gives Claude the ability to define asset function, color, documentation, category, etc. when it creates a schema.

Adds props for all things that can be defined on a schema.

## Screenshots:

> Create a schema representing fruits with properties representing color, number and type of fruit

image

## How was it tested?

- [X] Claude: "Create a schema representing fruits with properties representing color, number and type of fruit"


wendybujalski (dismissed):

wendybujalski (dismissed):

bin/si-mcp-server/src/tools/schemaCreate.ts
```diff
@@ -11,32 +10,56 @@ import {
...
-const SchemaCreateInputSchemaRaw = {
-  changeSetId: z.string().describe(
-    "The change set to create a schema in; schemas cannot be created on HEAD",
-  ),
+const DEFAULT_SCHEMA_DEFINITION_FUNCTION = `function main() {
+    const asset = new AssetBuilder();
+    return asset.build();
+}`;
+
+const InputSchemaRaw = {
```
stack72: We usually call these types after their tool e.g. what it was before - that's the convention of the app

stack72:
@jkeiser for the second iteration of this we should be careful about how that asset code looks - this is not what we use internally so we should teach it what we want the schema definition to look like

jkeiser:
@stack72 100% agreed! We didn't tell Claude anything that should have led it to create the code it did, and we were surprised it worked. Refining will take time.

stack72 (approved):

bin/si-mcp-server/src/tools/commonBehavior.ts
aaron-dernley, line 1: I'm curious on the changes in this file if you could talk me through later please @jkeiser, looks great though nice work!

PR #7355, opened 2025-09-26
Add a retry on getting mv index

jobelenus:

When the system is deployed both the websocket and SDF connections are lost, so if an index is requested then, and we get a 500/503, and there will be no ws event telling the client that it was (eventually) completed and it should request it

## Out of Scope

We ought to consider, in the future, to move web socket connections *off* SDF, onto a very simple service that rarely / never needs restarts because it doesn't ever have to change. It just shuffles messages off NATs over to clients.

## How was it tested?

Loading work spaces, flipping the status code check so it does retry even on success...

### DIFF
--- a/app/web/src/workers/webworker.ts
+++ b/app/web/src/workers/webworker.ts
@@ -2096,6 +2096,27 @@ const getSdfClientForWorkspace = (workspaceId: string, span?: Span) => {
   return sdf;
 };
 
+const sleep = async (ms: number) => {
+  return new Promise((resolve) => {
+    setTimeout(resolve, ms);
+  });
+};
+
+type AxiosFn<T> = () => Promise<AxiosResponse<T>>;
+const ONE_MIN = 1000 * 60;
+const retry = async <T>(
+  fn: AxiosFn<T>,
+  retryNum?: number,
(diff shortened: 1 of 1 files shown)


britmyerss (approved):

app/web/src/workers/webworker.ts
```diff
@@ -2096,6 +2096,30 @@ const getSdfClientForWorkspace = (workspaceId: string, span?: Span) => {
...
+  retryNum?: number,
+): Promise<AxiosResponse<T>> => {
+  if (retryNum && retryNum >= MAX_RETRY)
+    throw new Error("Index Retrieval Failed on Retry");
+
+  const r = await fn();
+
+  if (r.status >= 500) {
+    retryNum = retryNum ? retryNum + 1 : 1;
+    await sleep(retryNum ** 2 * 2000); // 2s, 4s, 8s, 16s, etc
```
zacharyhamm: isn't this: 1^2, 2^2, 3^2, 4^2, 5^2, 6^2 (so 1, 4, 9, 16, 25, 36) * 2 seconds.

zacharyhamm (commented):
Why not retry indefinitely?

PR #7356, opened 2025-09-26
draft: test mcp server

jkeiser:
testing si-mcp-server rebuild typecheck


stack72:
We can close this as you fixed it in a follow up PR

PR #7359, opened 2025-09-26
fix: Adjust si-mcp-server to use latest OpenAPI types

jkeiser:
Many Luminork response fields were made optional in a recent OpenAPI spec and si-mcp-server no longer compiles. This PR adjusts si-mcp-server to accept (and sometimes pass through to Claude) these undefined or null values.

## Out of Scope

Some of these things (e.g. componentId on action) seem like they really *aren't* optional, so I wonder whether Luminork's API should be changed instead for those things.

## How was it tested?

- [x] Claude can still use these endpoints

### DIFF
--- a/bin/si-mcp-server/src/data/schemaAttributes.ts
+++ b/bin/si-mcp-server/src/data/schemaAttributes.ts
@@ -1,4 +1,7 @@
-import { GetSchemaVariantV1Response } from "@systeminit/api-client";
+import {
+  GetSchemaVariantV1Response,
+  PropSchemaV1,
+} from "@systeminit/api-client";
 
 // ---------- Your existing types ----------
 type PropNode = {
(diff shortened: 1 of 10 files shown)


bin/si-mcp-server/src/data/schemaAttributes.ts
```diff
@@ -80,7 +78,7 @@ function addLeaf(
  * Optionally also builds a docs index { path -> {description, docLink} } in the same pass.
  */
 function collectFlatAttributes(
-  nodes: PropNode[] | null | undefined,
+  nodes: PropSchemaV1[] | null | undefined,
```
jkeiser, line 81: This appears to be the new typename for `domainProp.children`. Has all the same stuff on it as `PropNode` did.

bin/si-mcp-server/src/data/schemaAttributes.ts
```diff
@@ -92,8 +90,8 @@ function collectFlatAttributes(
 
     switch (node.propType) {
       case "object": {
-        const hasChildren = Array.isArray(node.children) &&
-          node.children.length > 0;
+        const hasChildren =
```
jkeiser, line 93: !@$!@#$ format-on-save. I don't know if we have a formatter for these, it's using the rules that we use for our other .ts files. I can change this back if the current style is intentional, just means re-editing a couple files and save-without-format.

stack72 (approved):

PR #7360, opened 2025-09-26
feat(dal): schema level actions, overlays

zacharyhamm:
Adds the ability to add an ActionPrototype as an outgoing target of a Schema instead of a Schema Variant. When picking the action to run, the schema level action for a specific kind (Create, Refresh, Destroy, Update) will be chosen instead of the schema variant level action.

Schema level action prototypes will also be made a child of a new category node: Overlays. This will allow us to find "overlay" action prototypes and to relate them to more than one schema.

There is currently no way to create these overlays via sdf/luminork so this functionality is only exercised in a test.


lib/dal/src/action/prototype.rs
```diff
@@ -313,6 +394,14 @@ impl ActionPrototype {
         result: ActionPrototypeResult,
     );
 
+    implement_add_edge_to!(
+        source_id: Ulid,
```
jkeiser, line 398: At some point we should get a `CategoryId` type going :)

lib/dal/src/action/prototype.rs
```diff
@@ -362,10 +451,56 @@ impl ActionPrototype {
         Ok(action_prototype_ids)
     }
 
+    pub async fn parentage(
+        ctx: &DalContext,
+        id: ActionPrototypeId,
+    ) -> ActionPrototypeResult<ActionPrototypeParent> {
+        if let Some(schema_ids) = Self::schema_ids(ctx, id).await? {
```
jkeiser, line 458: Should this error if an ActionPrototype has both Schemas and a SchemaVariant as parents?

lib/dal/src/action/prototype.rs
```diff
@@ -362,10 +451,56 @@ impl ActionPrototype {
...
+        }
+
+        if result.is_empty() {
+            Ok(None)
+        } else {
+            Ok(Some(result))
+        }
+    }
+
+    /// If this Action Prototype is parented by a SchemaVariant, returns the SchemaVariant Id.
```
jkeiser, line 499: Thinking out loud here: might be less future confusion if schema_variant_id() and schema_ids didn't exist, and you had to call `parentage()` for everything. Calls to schema_variant_id() would look like

```rust
let ActionPrototypeParent::SchemaVariant(schema_variant_id) = ActionPrototype::parentage(ctx, ap_id) else { ... };
```

Not needed to change before checkin, what you have will work! But this seems like it'll be roughly the same amount of work for the caller (who has to check for `Some()` anyway, and exposes them to the real reason for the `Option` (the fact that prototypes can have other parents).

lib/dal/src/action/prototype.rs
zacharyhamm: you know i don't have a really good answer here. It's not actually a problem if the same prototype has multiple "parents", one a variant, the other a schema. We only use parentage in authoring and "binding". But we also don't expose the ability to take an existing prototype and connect it to something else (like another schema variant, or two schemas) yet.

jkeiser: Yeah, it's a nit for sure, the system shouldn't be doing this. I'd at least add a comment in parentage() though, especially since `AttributePrototypeParent::Schemas` implies that there can be multiple parents for a thing.

lib/dal/src/action/prototype.rs
```diff
@@ -513,24 +708,37 @@ impl ActionPrototype {
 
     pub async fn get_prototypes_to_trigger(
         ctx: &DalContext,
-        id: ActionPrototypeId,
+        action_id: ActionId,
+        prototype_id: ActionPrototypeId,
```
jkeiser, line 712: Can an action have more than one prototype? Might not need both of these params?

lib/dal/src/attribute/value/default_subscription.rs
```diff
@@ -253,7 +216,10 @@ impl AttributeValue {
         ctx: &DalContext,
         id: AttributeValueId,
     ) -> AttributeValueResult<bool> {
-        let category_id = get_or_create_default_subscription_category(ctx).await?;
+        let category_id = ctx
+            .workspace_snapshot()?
+            .get_or_create_static_category_node(CategoryNodeKind::DefaultSubscriptionSources)
```
jkeiser, line 221: Do these need to do `create`? i.e. if `is_default_subscription_source()` is called on main, should it write to the graph?

lib/dal/src/func/binding/action.rs
```diff
@@ -124,7 +124,7 @@ impl ActionBinding {
             action_kind,
             func.name.to_owned(),
             func.description.to_owned(),
-            schema_variant_id,
+            ActionPrototypeParent::SchemaVariant(schema_variant_id),
```
jkeiser, line 127: I think you can just pass schema_variant_id here--it at least ought to have a `From`

lib/dal/src/func/binding/attribute.rs
```diff
@@ -371,7 +371,7 @@ impl AttributeBinding {
             validate_intrinsic_inputs(
                 ctx,
                 func_id,
-                eventual_parent,
+                eventual_parent.clone(),
```
jkeiser, line 374: Does the .clone() really need to be added? I don't see changes further down that ought to require it ...

lib/dal/src/func/binding/leaf.rs
```diff
@@ -283,6 +290,12 @@ impl LeafBinding {
                     let schema_variant_id = Component::schema_variant_id(ctx, component_id).await?;
                     schema_variant_ids.push(schema_variant_id);
                 }
+                EventualParent::Schemas(schemas) => {
+                    for schema_id in schemas {
+                        let default_variant = Schema::default_variant_id(ctx, schema_id).await?;
```
jkeiser: This one seems tricky--shouldn't we push all variants, e.g. if there is an unlocked one? (At the very least, I'd expect us to pick the unlocked one instead of the default if it exists, since this is authoring.)

lib/dal/src/attribute/value/default_subscription.rs
zacharyhamm: It's mostly harmless to do so, and i'd prefer to get new category nodes on the graph asap. (if you don't want to mutate don't commit!). Also this code is old, just being changed to use the new method here.

lib/dal/src/func/binding/attribute.rs
jkeiser: Oh I see why .clone(), EventualParent is no longer `Copy` because of the Vec. In this case at least, the method you call can just take `&EventualParent` and not need to clone the Vec needlessly.

lib/dal/src/func/binding.rs
```diff
@@ -305,8 +306,8 @@ impl From<FuncBinding> for si_frontend_types::FuncBinding {
             FuncBinding::Attribute(attribute) => si_frontend_types::FuncBinding::Attribute {
                 func_id: Some(attribute.func_id),
                 attribute_prototype_id: Some(attribute.attribute_prototype_id),
-                component_id: attribute.eventual_parent.into(),
-                schema_variant_id: attribute.eventual_parent.into(),
+                component_id: attribute.eventual_parent.clone().into(),
```
jkeiser: Making the `From` use `&` (or maybe better now, adding a `fn component_id(&self) -> Option<ComponentId>`) would avoid cloning (and allocating) here.

lib/dal/src/workspace_snapshot/node_weight/category_node_weight.rs
```diff
@@ -39,9 +42,11 @@ pub enum CategoryNodeKind {
     View,
     DiagramObject,
     DefaultSubscriptionSources,
+    Overlays,
 }
 
 const DEFAULT_SUBSCRIPTION_SOURCE_CATEGORY_ID_STR: &str = "0000DNP8N22X0A8S4CV2APWX3A";
+const OVERLAY_CATEGORY_ID_STR: &str = "0000BQJPTG3PZ3XQG9M3W7BSDR";
```
jkeiser, line 49: Hadn't noticed this before, <3 the static node id

PR #7361, opened 2025-09-26
basic funcCreate MCP tool

wendybujalski:

This PR adds a new tool for the MCP server which can create functions. This first version of the tool is very basic but it can create a function of a few types. It can create Action functions of each of the four action kinds.

The tool is not yet added to the `server.ts` file so it will not yet be available to users.

#### Screenshots:

Screenshot 2025-09-26 at 4 42 01 PM

#### Out of Scope:

This PR just scaffolds the basics, there are many possible improvements and additional features we will want in the future!

## How was it tested?

Used the tool with the AI Agent to successfully create one function of each type, and one action function of each kind.

### DIFF
--- a/bin/si-mcp-server/src/tools/funcCreate.ts
+++ b/bin/si-mcp-server/src/tools/funcCreate.ts
@@ -1,0 +1,153 @@
+import { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
+import { z } from "zod";
+import { SchemasApi } from "@systeminit/api-client";
+import { apiConfig, WORKSPACE_ID } from "../si_client.ts";
+import {
+  errorResponse,
+  generateDescription,
+  successResponse,
+  withAnalytics,
+} from "./commonBehavior.ts";
+
+const name = "func-create";
+const title = "Create a new function for a schema.";
+const description = `
+<description>
+Create a new function of the given type for a given schema.
+</description>
+<usage>
+Use this tool to create a new function. If the user only gives a schema name, use the schema-find tool to find the schemaId. If the user does not say which type of function they want to make, ask them which they want. If the user does not specify the actionKind for an action function, ask them which kind they want. Do not mention any schemaVariantId to the user.
+</usage>
+`;
+
+const DEFAULT_QUALIFICATION_FUNCTION = `function main() {
+    return { result: "success", message: "All good!" };
+}`;
+
+const DEFAULT_CODEGEN_FUNCTION = `async function main(component: Input): Promise<Output> {
+  return {
+    format: "json",
+    code: JSON.stringify(component),
+  };
+}`;
+
+const DEFAULT_MANAGEMENT_FUNCTION = `async function main({ thisComponent, components }: Input): Promise<Output> {
+  throw new Error("unimplemented!");
+}`;
+
+const DEFAULT_ACTION_FUNCTION = `async function main(component: Input): Promise<Output> {
+  throw new Error("unimplemented!");
+}`;
+
+const FuncCreateInputSchemaRaw = {
+  changeSetId: z
+    .string()
+    .describe(
+      "The change set to create a function in; functions cannot be created on HEAD",
+    ),
+  schemaId: z.string().describe("The schema id the function is for"),
+  name: z.string().describe("The name of the function").min(1),
+  description: z.string().describe("A description for the function"),
+  functionType: z.enum(["qualification", "codegen", "management", "action"]),
+  functionCode: z
+    .string()
+    .describe(
+      `A typescript function definition. Documentation on functions can be found at https://docs.systeminit.com/reference/asset/function`,
+    )
+    .optional(),
+  actionKind: z.enum(["Create", "Destroy", "Refresh", "Update"]).optional(),
+};
+
+const FuncCreateOutputSchemaRaw = {
+  status: z.enum(["success", "failure"]),
+  errorMessage: z
+    .string()
+    .optional()
+    .describe(
+      "If the status is failure, the error message will contain information about what went wrong",
+    ),
+  data: z.object({}),
+};
+const FuncCreateOutputSchema = z.object(FuncCreateOutputSchemaRaw);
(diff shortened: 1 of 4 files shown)


bin/si-mcp-server/src/tools/funcCreate.ts
```diff
@@ -0,0 +1,136 @@
...
+            await siSchemasApi.createVariantAction({
+              ...request,
+              createVariantActionFuncV1Request: {
+                ...createFuncRequest,
+                code: functionCode,
+                kind: actionKind,
+              },
+            });
+          } else {
+            console.log("idk how to make that type of function yet!");
```
stack72: ideally this would be an error that we return to the user now a console.log and fall through to the response

bin/si-mcp-server/src/tools/funcCreate.ts
```diff
@@ -0,0 +1,136 @@
...
+  schemaId: z.string().describe("The schema id the function is for"),
+  name: z.string().describe("The name of the qualification").min(1),
+  description: z.string().describe("A description for the qualification"),
+  functionType: z.enum(["qualification", "codegen", "management", "action"]),
+  functionCode: z
+    .string()
+    .describe(
+      `A typescript function definition. Documentation on functions can be found at https://docs.systeminit.com/reference/asset/function`,
+    )
+    .default(DEFAULT_QUALIFICATION_FUNCTION),
```
stack72: I think asking for an action func and getting default qualification isn't correct - maybe we just default in the code below rather than a zod default

nickgerace: This is just a mistake I think

bin/si-mcp-server/src/tools/funcCreate.ts
nickgerace: Agreed!

bin/si-mcp-server/src/tools/funcCreate.ts
wendybujalski: ah yeah we forgot to change the default after implementing other types of funcs than quals

bin/si-mcp-server/src/tools/funcCreate.ts
wendybujalski: yeah that makes sense, will do

aaron-dernley (approved):

bin/si-mcp-server/src/tools/funcCreate.ts
```diff
@@ -0,0 +1,153 @@
...
+} from "./commonBehavior.ts";
+
+const name = "func-create";
+const title = "Create a new function for a schema.";
+const description = `
+<description>
+Create a new function of the given type for a given schema and schema variant.
+</description>
+<usage>
+Use this tool to create a new function. If the user only gives a schema name, use the schema-find tool to find the schemaId. If the user does not say which type of function they want to make, ask them which they want. If the user does not specify the actionKind for an action function, ask them which kind they want. Do not mention any schemaVariantId to the user.
```
wendybujalski, line 19: For now we tell the AI to prompt the user for info it needs - later we may want it to infer that info based on what the user wants.

bin/si-mcp-server/src/tools/funcCreate.ts
```diff
@@ -0,0 +1,153 @@
...
+  generateDescription,
+  successResponse,
+  withAnalytics,
+} from "./commonBehavior.ts";
+
+const name = "func-create";
+const title = "Create a new function for a schema.";
+const description = `
+<description>
+Create a new function of the given type for a given schema and schema variant.
```
nickgerace:
```suggestion
Create a new function of the given type for a given schema.
```

bin/si-mcp-server/src/tools/funcCreate.ts
```diff
@@ -0,0 +1,153 @@
...
+}`;
+
+const FuncCreateInputSchemaRaw = {
+  changeSetId: z
+    .string()
+    .describe(
+      "The change set to create a schema in; schemas cannot be created on HEAD",
+    ),
+  schemaId: z.string().describe("The schema id the function is for"),
+  name: z.string().describe("The name of the qualification").min(1),
```
nickgerace:
```suggestion
  name: z.string().describe("The name of the function").min(1),
```

bin/si-mcp-server/src/tools/funcCreate.ts
```diff
@@ -0,0 +1,153 @@
...
+  description: z.string().describe("A description for the qualification"),
```
nickgerace:
```suggestion
  description: z.string().describe("A description for the function"),
```

bin/si-mcp-server/src/tools/funcCreate.ts
```diff
@@ -0,0 +1,153 @@
...
+  functionType: z.enum(["qualification", "codegen", "management", "action"]),
+  functionCode: z
+    .string()
+    .describe(
+      `A typescript function definition. Documentation on functions can be found at https://docs.systeminit.com/reference/asset/function`,
+    )
+    .optional(),
+  actionKind: z.enum(["Create", "Destroy", "Refresh", "Update"]).optional(),
```
nickgerace, line 58: We should add a description here. Can wait for the next PR though.

nickgerace: To be clear, if the tool was registered, we would do that here :)

bin/si-mcp-server/src/tools/schemaFind.ts
```diff
@@ -163,7 +163,7 @@ export function schemaFindTool(server: McpServer) {
           }
 
           const responseData: NonNullable<FindSchemaOutput["data"]> = {
-            schemaId: schemaNameOrId,
```
nickgerace, line 166: @stack72 I believe this is the right motion since `schemaName` is returned below. this is a bug fix that was needed for func create and did not have adverse effects from what we observed.

stack72: Great find!!

PR #7367, opened 2025-09-29
fix: filter build if only changes made to mcp server

aaron-dernley:

Currently when making changes to the MCP server artifacts are being built for other services despite no changes being made. This aims to filter out other services when only changes to the MCP server are made.

Previously, a PR in the MCP server repo resulted in 3 artifacts being built for other untouched services, here's a link to the pipeline run - https://buildkite.com/system-initiative/si-merge-queue/builds/7261 - this is the PR that was merged to trigger this pipeline - https://github.com/systeminit/si/pull/7348

## How was it tested?

I'm not sure it can be tested before it is merged.

### DIFF
--- a/bxl/dependent_targets.bxl
+++ b/bxl/dependent_targets.bxl
@@ -273,7 +273,13 @@ def _dependent_modified_file_targets(ctx: bxl.Context) -> bxl.UnconfiguredTarget
     for results in results_set.values():
         targets = targets + results
 
-    return _rdeps_for_targets(ctx, map(lambda e: "{}".format(e.label), targets))
+    rdeps = _rdeps_for_targets(ctx, map(lambda e: "{}".format(e.label), targets))
+
+    # If only si-mcp-server files changed, exclude other bin/* services
+    if all([f.startswith("bin/si-mcp-server/") for f in ctx.cli_args.modified_file]):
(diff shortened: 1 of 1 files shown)


bxl/dependent_targets.bxl
```diff
@@ -273,7 +273,39 @@ def _dependent_modified_file_targets(ctx: bxl.Context) -> bxl.UnconfiguredTarget
...
-    return _rdeps_for_targets(ctx, map(lambda e: "{}".format(e.label), targets))
+    rdeps = _rdeps_for_targets(ctx, map(lambda e: "{}".format(e.label), targets))
+
+    # Apply service isolation filtering
+    return _filter_cross_service_deps(ctx, rdeps, ctx.cli_args.modified_file)
+
+# Filters out cross-service dependencies for isolated services like si-mcp-server
+def _filter_cross_service_deps(ctx: bxl.Context, targets: bxl.UnconfiguredTargetSet, modified_files: list[str]) -> bxl.UnconfiguredTargetSet:
+
+    # Check if all modified files are in si-mcp-server
```
stack72: I don't believe we should do this - this is WAY too specific for this script

aaron-dernley: @stack72 For sure, my initial stab didn't feel quite right - going to chat to JW and Fletcher today about it

PR #7368, opened 2025-09-29
feat(clover): refactor for shared pipeline usage, pipeline for hetzner

sprutton1:

Significant overhaul of the clover pipelines to allow for more reuse of existing bits. Includes Hetzner assets with basic Refresh, Discovery, and Import funcs.

## How was it tested?

Locally importing and validating. The PR will generate changes to ensure we haven't mucked with the AWS bits.


bin/clover/src/pipelines/hetzner/funcs/management/import.ts
```diff
@@ -0,0 +1,78 @@
...
+      },
+    },
+  );
+
+  if (!response.ok) {
+    throw new Error(`API Error: ${response.status} ${response.statusText}`);
+  }
+
+  const data = await response.json();
+  const resource = data[endpoint.slice(0, -1)] || data; // Handle singular vs plural endpoint names
```
jobelenus, line 35: :shake fist:

bin/clover/src/pipelines/hetzner/spec.ts
```diff
@@ -0,0 +1,373 @@
...
+      switch (method) {
+        case "get": {
+          getOperation = op;
+          const opId = op.operationId as string;
+          handlers[opId.startsWith("list_") ? "list" : "read"] = defaultHandler;
+          break;
+        }
+        case "put": {
+          putOperation = op;
+          handlers["update"] = defaultHandler;
```
jobelenus, line 140: a nit to come back to... we might want to destructure `defaultHandler`, just so each handler isn't a referenced copy, b/c I suspect that each would eventually have different data. but we can do that at the time!

PR #7369, opened 2025-09-29
Add event for created funcs in luminork

nickgerace:
## Description

This change adds a WsEvent for created and attached funcs in luminork. This fixes an issue where authoring with the MCP server resulted in not seeing the new funcs appear in the customization UI. Publishing the event is now shared logic between sdf and luminork.

In addition to the above, this change adds a README for luminork.

## Screenshot

The authoring UI after it automatically drops you into the function without clicking anything.

image

## Known Issue

The counters to the left remain broken. This is fixed when using the UI to add a function. The `func.store.ts` pinia store works around this by adding func bindings upon creation.

### DIFF
--- a/lib/luminork-server/src/service/v1/schemas/create_action.rs
+++ b/lib/luminork-server/src/service/v1/schemas/create_action.rs
@@ -104,6 +104,8 @@ pub async fn create_variant_action(
     )
     .await?;
 
+    FuncAuthoringClient::publish_func_create_event(ctx, &func).await?;
+
     tracker.track(
         ctx,
         "api_create_action_func",
--- a/lib/luminork-server/src/service/v1/schemas/create_authentication.rs
+++ b/lib/luminork-server/src/service/v1/schemas/create_authentication.rs
@@ -97,6 +97,8 @@ pub async fn create_variant_authentication(
     )
     .await?;
 
+    FuncAuthoringClient::publish_func_create_event(ctx, &func).await?;
+
     tracker.track(
         ctx,
         "api_create_authentication_func",
--- a/lib/luminork-server/src/service/v1/schemas/mod.rs
+++ b/lib/luminork-server/src/service/v1/schemas/mod.rs
@@ -98,7 +98,7 @@ pub enum SchemaError {
     Transactions(#[from] TransactionsError),
     #[error("validation error: {0}")]
     Validation(String),
-    #[error("variant authuring error: {0}")]
+    #[error("variant authoring error: {0}")]
     VariantAuthoring(#[from] VariantAuthoringError),
     #[error("workspace snapshot error: {0}")]
     WorkspaceSnapshot(#[from] dal::WorkspaceSnapshotError),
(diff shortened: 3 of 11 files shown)


lib/luminork-server/src/service/v1/schemas/create_action.rs
```diff
@@ -104,6 +105,18 @@ pub async fn create_variant_action(
     )
     .await?;
 
+    // Publish an event for the authoring UI to have ready.
+    let summary = func
```
stack72: I really think we should try and push these back to the DAL so we have the same logic as we do in SDF - I don't think building a frointend type here is the right thing to do

nickgerace: I started that way. It turned out a bit funky.

I tried to make an authoring client func to publish the event, but in turn that will always generate a summary. That's not a bad abstraction, but I think this will require a longer pull to move all that logic into the dal. The good news is that I think you can do it. I think it'll just be bit a bit churn-y to pull the whole thread in this PR.

Let try it again and I'll re-request.

nickgerace: @stack72 thoughts on the new shape?

lib/luminork-server/src/service/v1/schemas/mod.rs
```diff
@@ -98,7 +98,7 @@ pub enum SchemaError {
     Transactions(#[from] TransactionsError),
     #[error("validation error: {0}")]
     Validation(String),
-    #[error("variant authuring error: {0}")]
+    #[error("variant authoring error: {0}")]
```
zacharyhamm, line 101: what about an arthuring error

lib/luminork-server/src/service/v1/schemas/create_authentication.rs
```diff
@@ -97,6 +97,8 @@ pub async fn create_variant_authentication(
     )
     .await?;
 
+    FuncAuthoringClient::publish_func_create_event(ctx, &func).await?;
```
britmyerss, line 100: why not put these in the FuncAuthoringClient directly for now?

zacharyhamm (approved):

lib/luminork-server/src/service/v1/schemas/create_authentication.rs
nickgerace: This comment is on a `FuncAuthoringClient` call, so what do you mean?

lib/luminork-server/src/service/v1/schemas/mod.rs
nickgerace: Not a bad idea

lib/luminork-server/src/service/v1/schemas/create_authentication.rs
britmyerss: Sorry, I'm wondering why we wouldn't have the various `FuncAuthoringClient::save_code` or FuncAuthoringClient::create_new_auth` methods enqueue the event, vs. having each route handler do it in luminork + sdf

nickgerace: Because in sdf, there is one route that handles it all. We should absolutely do that! It is just a larger refactoring pull

PR #7372, opened 2025-09-29
longer timeouts for get-tiles

wendybujalski:

### DIFF
--- a/app/web/cypress/ux/tile-view/get-tiles.cy.ts
+++ b/app/web/cypress/ux/tile-view/get-tiles.cy.ts
@@ -24,7 +24,7 @@ describe("web", () => {
     cy.wait(5000);
 
     // Find the URL for the synthetic workspace and go there
-    cy.get('a[href="' + AUTH_API_URL + '/workspaces/' + SI_WORKSPACE_ID + '/go"]')
+    cy.get('a[href="' + AUTH_API_URL + '/workspaces/' + SI_WORKSPACE_ID + '/go"]', { timeout: 60000 })
       .should('be.visible')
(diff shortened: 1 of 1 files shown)


app/web/cypress/ux/tile-view/get-tiles.cy.ts
```diff
@@ -24,7 +24,7 @@ describe("web", () => {
     cy.wait(5000);
 
     // Find the URL for the synthetic workspace and go there
-    cy.get('a[href="' + AUTH_API_URL + '/workspaces/' + SI_WORKSPACE_ID + '/go"]')
+    cy.get('a[href="' + AUTH_API_URL + '/workspaces/' + SI_WORKSPACE_ID + '/go"]', { timeout: 60000 })
```
wendybujalski, line 27: The default timeout here was 4 seconds, this lengthens it to 1 minute.

app/web/cypress/ux/tile-view/get-tiles.cy.ts
```diff
@@ -46,7 +46,7 @@ describe("web", () => {
     function checkTileCount() {
       // For virtual scrolling, we need to scroll through the entire list
       // First scroll to top, then gradually scroll down to ensure all items are rendered
-      cy.get('[data-testid="explore-grid"]').parent().scrollTo("top");
+      cy.get('[data-testid="explore-grid"]', { timeout: 60000 }).parent().scrollTo("top");
```
wendybujalski, line 49: Same here. 1 minute might not be the exact length we want, but 4 seconds is definitely too short!

nickgerace (approved):

PR #7374, opened 2025-09-29
fix: match mgmt func kind by starts_with

sprutton1:
Sets the mgmt func kind "matching" by starts_with instead of hard equals so I can set import correctly for Hetzner assets.

#### Out of Scope:

We should really be threading this through as a distinct type. 

## How was it tested?

- [X] Integration tests pass
- [] Manual test: new functionality works in UI


lib/dal/src/management/prototype.rs
```diff
@@ -386,17 +386,26 @@ impl ManagementPrototype {
     ) -> ManagementPrototypeResult<ManagementFuncKind> {
         let func_id = Self::func_id(ctx, id).await?;
         let func = Func::get_by_id(ctx, func_id).await?;
-        // TODO: Make Management Func Kinds a real thing
+        // TODO: (Brit) Make Management Func Kinds a real thing
```
nickgerace, line 389: 💀

britmyerss (approved):

PR #7377, opened 2025-09-29
Bump get public IP timeout in deploy stack

nickgerace:
## Description

We need a bit more time depending on both the system and AWS for the public IP to be available. This bumps it from 60 seconds to 120 seconds.


johnrwatson (approved):
Thanks for fixing @nickgerace

PR #7384, opened 2025-09-30
Apply a batch of patches in batches

jobelenus:

We are seeing `PatchBatch` data come over the wire from Edda to the web worker with hundreds of patches in a single batch. As a result we are sending too much work directly at SQLite—and too much work without allowing the UI to refresh. (Note: we started to see slow downs above 50+ patches in a batch)

1. chunk the patches into sets of 45, which will bust the cache for the UI to remain responsive.
2. lower the amount of queries being run by doing the `kind, args, checksum` check for an `indexChecksum` all at once into a map that is provided to each `applyPatch` run. This removes 2 queries per patch.
3. Removed an unnecessary idempotent call that the parent caller runs. This removes 2 queries per patch.
4. Add performance timings to individual DB queries

#### Out of Scope:

Edda shouldn't allow hundreds of patches to get compressed into a single batch. It ought to read and compress at smaller batch sizes for a more responsive experience.

## How was it tested?

1. New change set, add components, delete components, erase components. Apply. General basic functionality.
5. Also the web worker test covers this behavior.

### DIFF
--- a/app/web/src/workers/webworker.ts
+++ b/app/web/src/workers/webworker.ts
@@ -454,7 +454,9 @@ const removeAtom = (
   kind: EntityKind,
   id: string,
   checksum: string,
+  span?: Span,
 ) => {
+  const start = performance.now();
   db.exec({
     sql: `
     DELETE FROM index_mtm_atoms
@@ -462,6 +464,8 @@ const removeAtom = (
     `,
     bind: [indexChecksum, kind, id, checksum],
   });
+  const end = performance.now();
+  span?.setAttribute("performance.removeAtom", end - start);
 };
 
 const createAtomFromPatch = async (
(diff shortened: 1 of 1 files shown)


app/web/src/workers/webworker.ts
```diff
@@ -1166,140 +1192,149 @@ const handleWorkspacePatchMessage = async (
...
-        connectionAtoms.map(async (atom) => {
-          return await applyWorkspacePatch(db, atom, indexChecksum);
-        }),
-      );
+      // apply 45 changes at time and bust
+      // after each in order for the UI
+      // to remain responsive...
+      const chunkedAtoms = chunkArray<Required<WorkspaceAtom>>(atoms, 45);
+
+      await Promise.all(
```
jobelenus, line 1209: NOTE: this is just a white space change to put this within the map() fn

PR #7385, opened 2025-09-30
core(rebaser): allow certain errors to trigger request retries

fnichol:
This change is primarily focused on the code called in the
`ChangeSetProcessorTask` for each request that is being processed.

Prior to this change any error that was encountered that used the
question mark operator (i.e. Rust's `?`) would trigger the `default`
Naxum handler to return a `Result::Err(_)`. The `JetstreamPostProcess`
middleware had a custom `OnFailure` function which would copy the
message from the original stream subscription to the dead letter queue
stream (i.e. "DLQ"), treating all errors as unrecoverable. With more
time and experience seeing the service deployed in production, it's
clear that not all errors are unrecoverable, but rather some errors at
certain points in the business logic *should* be retried (as not
retrying some of these requests would lead to dropped graph changes).

This change updates the `OnFailure` implementation and uses a new
wrapping error type--still called `HandlerError` which only has 4
possible variants:

- `PostCommit`: a message processing error *after* committing to a
  graph change
- `PreCommit`: a message processing error *before* committing to a graph
  change
- `TransientService`: an error occurring due to an error from an
  external service (this could be Postgres, NATS, etc.)
- `Unprocessable`: a request message that is not processable in a
  deterministic way (such as a failure to deserialize a NATS header, for
  example)

Importantly this error type implements the `naxum::IntoResponse` trait,
meaning that each of these variants will map to a Response with a
discrete `StatusCode`. The `OnFailure` function can use the `StatusCode`
values to determine which failures should be retried and which should
not.

The "retry" logic is not apparent in the code change as the retry
activity is an outcome of the task scheduling of the Rebaser. Each
subscription for a `ChangeSetProcessorTask` is an push-based ordered
consumer, meaning the consumer is essentially a core NATS subscriber
with messages being pushed to the subsciber as fast as the messages can
be read. One side effect of this style of consumer is that there is no
Ack/Nack'ing of message.

A successful processing of a message is generally deleted, if the
message stream is expected to be a "work queue". With the
`ChangeSetProcessorTask`, a successful handler result (that is, the
`async fn default(...) -> Result<(), _>` function returns `Ok(()`) will
invoke the `OnSuccess` middleware callback function and the message will
be deleted before the Naxum app pulls the next message from the
subscription stream.

If the handler returns an `Err(_)` however, the `OnFailure` middleware
callback function would have copied the message onto the DLQ and then
also delete the message. With this change, the `OnFailure` function uses
the `StatusCode` to determine if a message should be retried. There is a
new "internal token" (i.e. another `CancellationToken`) which this
`OnFailure` function uses to stop the execution of the
`ChangeSetProcessorTask`'s Naxum app so that no further messages will be
pulled for processing. This is important as the Rebaser should *not*
process any further messages before first attempting to process the
just-failed request at least one more time. In shutting down the
`ChangeSetProcessorTask` in such a way that it is re-scheduled to run
again, the just-failed request is the first request message to consider
when a new `ChangeSetProcessorTask` is spun up.

### DIFF
--- a/lib/naxum/src/middleware/jetstream_post_process/future.rs
+++ b/lib/naxum/src/middleware/jetstream_post_process/future.rs
@@ -1,6 +1,7 @@
 use std::{
     future::Future,
     pin::Pin,
+    sync::Arc,
     task::{
         Context,
         Poll,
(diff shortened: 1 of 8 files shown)


britmyerss (approved):
👏 👏 👏 👏

PR #7402, opened 2025-10-01
chore(sdf,luminork): skip loading snapshot when abandoning a change set

fnichol:
This change attempts to reduce the chances error or long waits when abandoning a change set. Prior to this change the workspace snapshot for the target change set was loaded via the
`DalContext::update_visibility_and_snapshot_to_visibility` method. No code in either the SDF or Luminork API handler is using the loaded workspace snapshot, and in the case of another bug, this load led to an error which resulted in not being able to abandon a change set. Finally, the workspace snapshot load can take some time if it is not already in cache, and will take up to 10 seconds if the snapshot is no longer available in the layer cache database (this is due to the current layer cache spin lock retry logic).

### DIFF
--- a/lib/luminork-server/src/service/v1/change_sets/delete.rs
+++ b/lib/luminork-server/src/service/v1/change_sets/delete.rs
@@ -33,22 +33,15 @@ pub async fn abandon_change_set(
     ChangeSetDalContext(ref mut ctx): ChangeSetDalContext,
     tracker: PosthogEventTracker,
 ) -> ChangeSetResult<Json<DeleteChangeSetV1Response>> {
-    let maybe_head_changeset = ctx.get_workspace_default_change_set_id().await?;
-    if maybe_head_changeset == ctx.change_set_id() {
+    if ctx.is_head().await? {
         return Err(ChangeSetError::CannotAbandonHead);
     }
 
     let mut change_set = ChangeSet::get_by_id(ctx, ctx.change_set_id()).await?;
     let old_status = change_set.status;
-    ctx.update_visibility_and_snapshot_to_visibility(change_set.id)
-        .await?;
+    // Skipping the load of the snapshot here as it is not required and be expensive
+    ctx.update_visibility_deprecated(change_set.id.into());
     change_set.abandon(ctx).await?;
-
-    tracker.track(
-        ctx,
-        "api_create_change_set",
-        json!({"abandoned_change_set": ctx.change_set_id()}),
-    );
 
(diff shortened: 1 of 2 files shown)


lib/luminork-server/src/service/v1/change_sets/delete.rs
```diff
@@ -58,6 +51,12 @@ pub async fn abandon_change_set(
     )
     .await?;
 
+    tracker.track(
+        ctx,
+        "api_create_change_set",
```
stack72, line 56: I just noticed this :/ the event is wrong - I will fix that up after

lib/luminork-server/src/service/v1/change_sets/delete.rs
```diff
@@ -58,6 +51,12 @@ pub async fn abandon_change_set(
...
+        json!({"abandoned_change_set": ctx.change_set_id()}),
+    );
+
     ctx.commit_no_rebase().await?;
```
nickgerace, line 60: @fnichol do the audit logs end up in the right place? Commit no rebase works too?

stack72: he didn't add this - he just moved it - if they doesn't work, we can follow up

nickgerace: I didn't ask about the moved tracker call. I asked about the `ctx.write_audit_log` and `ctx.commit_no_rebase` calls :)

fnichol: I changed the ordering to exactly match the calling timings of SDF. Unless there was a good reason to track, then audit I assumed that copy/paste drift.

fnichol: I also checked the audit and track code to see what side effects they were relying on from the `DalContext` state--it was primarily change set which I why I used the `DalContext::update_visibility_deprecated` method.

lib/luminork-server/src/service/v1/change_sets/delete.rs
fnichol: Ha, that's a good catch, I wasn't even reading that tag--only ensuring that it didn't change!

lib/luminork-server/src/service/v1/change_sets/delete.rs
nickgerace: Great!

PR #7406, opened 2025-10-01
Enable PgPool connection culling by default

jhelwig:
Previously this was opt-in behavior while we were experimenting on things, but we want this on everywhere, and 2-hours seems like a reasonable enough default interval.


fnichol (dismissed):
Love it, all hail the production default config!

fnichol (approved):

PR #7415, opened 2025-10-02
feat: remove resource if remote resource is not found

johnrwatson:
Removes the resource from the component if the upstream is not found via a ResourceNotFoundException during Refresh funcs. 

Tested against 5 SG's, which it correctly removes the resource when I deleted it out of the console separately from the model like a savage.


aaron-dernley (approved):

johnrwatson:
Conducted final check on asset using refresh from this branch, both regular refreshes can sync over and over, as soon as you delete the resource in the provider and refresh the resource is removed from SI. Beaut.

PR #7416, opened 2025-10-02
feat: Component search api

jkeiser:
This adds the ability to search for components using the same query syntax as the frontend. Right now, this is just an additional option on top of the Luminork component search, but we will quickly create a new API endpoint so we can deprecate this one.

```
POST /v1/w/:workspaceId/change-sets/:changeSetId/components/search
{ queryString: "Instance region:us-east-1" }

{ components: ['01K4T90DB8FKV6PTD9D85NV2T1', '01K4T90DDMQC17HG3JQ72KJKYB'] }
```

There are three major differences:
* Instead of fuzzy search, searching for bare words like `Instance` will only return things that *contain* the word Instance.
* Attribute path search is now case-insensitive
* Searching for `AWS::EC2::Instance` will now return EC2 instances (this did not work because it was considered an attr search for attributes named "AWS" with value ":EC2::Instance" in the frontend search)

### How This PR Changes The System

* Adds `queryString` parameter to /components/search in Luminork, runs that to get the initial list of component IDs, and then runs the existing schemaName/category/componentName filters (no change to those).
* New `SearchQuery` module provides generic search string parsing and query utilities, with an eye toward applying search to other things in the future (such as schemas)
* Searches through AttributeTree MVs rather than the graph

#### Out of Scope:

* Wildcards: we support partial matches and exact matches, but we don't use the wildcards like the frontend does right now: you can't do `region:us-*-1`
* Unicode case comparison: right now, we only honor ASCII lower<->upper comparisons when searching for attributes and values.
* New endpoint: we are considering a new endpoint that can return more than just components, but this is a single-step to get the search infrastructure into an existing endpoint.
* Query syntax: no new syntax except the minor changes mentioned above. There is more we'd like to do, including possibilities tag search, schema: attribute search and upgradeable:true search, * in attribute paths, among others
* Perf: there is reason to believe that overhead (possibly from grabbing snapshots we don't use) accounts for most of the 500ms; the difference between searching 2 components vs. 1000 components is about 50ms. We can probably get this *well* within typing time.

## How was it tested?

- [X] Integration tests pass
- [X] New unit tests for syntax and value matching
- [X] Manual test: attribute and non-attribute searches against real components
- [X] Manual test: endpoint works the same when passing existing parameters
- [X] Manual test: endpoint produces correct results when passing a combination of existing parameters and queryString
- [X] Perf test: searching 1000 components takes ~500ms

### DIFF
--- a/lib/frigg/src/workspace.rs
+++ b/lib/frigg/src/workspace.rs
@@ -2,6 +2,11 @@ use kv_history::{
     History,
     Keys,
 };
+use serde::{
+    Deserialize,
+    Serialize,
+    de::DeserializeOwned,
+};
 use si_data_nats::{
     Subject,
     async_nats::jetstream::{
@@ -99,6 +104,41 @@ impl FriggStore {
         {
             Some((bytes, _)) => Ok(Some(
                 serde_json::from_slice(bytes.as_ref()).map_err(Error::Deserialize)?,
+            )),
+            None => Ok(None),
+        }
+    }
+
+    #[instrument(
+        name = "frigg.get_workspace_object",
+        level = "debug",
+        skip_all,
+        fields(
+            si.workspace.id = %workspace_id,
+            si.frontend_object.id = %id,
+            si.frontend_object.kind = %kind,
+            si.frontend_object.checksum = %checksum,
+    ))]
+    pub async fn get_workspace_object_data<T: DeserializeOwned>(
+        &self,
+        workspace_id: WorkspacePk,
+        kind: &str,
+        id: &str,
+        checksum: &str,
+    ) -> Result<Option<T>> {
+        match self
+            .get_object_raw_bytes(&Self::workspace_object_key(
+                workspace_id,
+                kind,
+                id,
+                checksum,
+            ))
+            .await?
+        {
+            Some((bytes, _)) => Ok(Some(
+                serde_json::from_slice::<FrontendObjectData<T>>(bytes.as_ref())
+                    .map_err(Error::Deserialize)?
+                    .data,
             )),
             None => Ok(None),
         }
@@ -492,4 +532,12 @@ impl FriggStore {
             Scope::ChangeSet.as_ref()
         ))
     }
+}
+
+// Used for deserializing the data field directly to a type.
+// This is just a subset of FrontendObject with a different type on the data field.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct FrontendObjectData<T> {
+    data: T,
 }
--- a/lib/luminork-server/src/search/component.rs
+++ b/lib/luminork-server/src/search/component.rs
@@ -1,0 +1,196 @@
+use std::{
+    collections::HashMap,
+    sync::Arc,
+};
+
+use futures::future::try_join_all;
+use itertools::Itertools as _;
+use serde::{
+    Deserialize,
+    Serialize,
+};
+use si_frontend_mv_types::{
+    index::change_set::ChangeSetMvIndexVersion,
+    reference::IndexReference,
+};
+use si_id::{
+    AttributeValueId,
+    ChangeSetId,
+    ComponentId,
+    WorkspacePk,
+};
+use telemetry::prelude::*;
+
+use crate::search::{
+    Error,
+    Result,
+    SearchQuery,
+    SearchTerm,
+};
+
+/// Search for components matching the given query in the given workspace and change set.
+pub async fn search(
+    frigg: &frigg::FriggStore,
+    workspace_id: WorkspacePk,
+    change_set_id: ChangeSetId,
+    query: &Arc<SearchQuery>,
+) -> Result<Vec<ComponentId>> {
+    // Grab the index first
+    let attribute_trees = attribute_tree_mv_index(frigg, workspace_id, change_set_id).await?;
+
+    // Spawn parallel fetch+match tasks for each component
+    let match_tasks = try_join_all(attribute_trees.map(|index_ref| {
+        tokio::spawn(match_component(
+            frigg.clone(),
+            workspace_id,
+            index_ref,
+            query.clone(),
+        ))
+    }))
+    .await?;
+
+    // Wait for them all to complete and collect the results
+    match_tasks.into_iter().flatten_ok().try_collect()
+}
+
+/// Fetch the AttributeTree MV for a component and match it against the query.
+#[instrument(level = "debug", skip_all, fields(id))]
+async fn match_component(
+    frigg: frigg::FriggStore,
+    workspace_pk: WorkspacePk,
+    index_ref: IndexReference,
+    query: Arc<SearchQuery>,
+) -> Result<Option<ComponentId>> {
+    let attribute_tree = attribute_tree_mv(frigg, workspace_pk, index_ref).await?;
+    if match_attribute_tree(&attribute_tree, &query) {
+        Ok(Some(attribute_tree.id))
+    } else {
+        Ok(None)
+    }
+}
+
+/// Match a component against a query or sub-query.
+///
+/// This is called once for each term in the query, and the results are combined according to
+/// query rules (AND, OR, NOT).
+fn match_attribute_tree(attribute_tree: &AttributeTreeForSearch, query: &SearchQuery) -> bool {
+    match query {
+        SearchQuery::MatchValue(term) => {
+            term.match_str(&attribute_tree.component_name)
+                || term.match_str(&attribute_tree.schema_name)
+                || term.match_ulid(attribute_tree.id)
+        }
+        // TODO support schema:, category:, component: and id:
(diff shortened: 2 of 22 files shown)


lib/frigg/src/workspace.rs
```diff
@@ -104,6 +109,41 @@ impl FriggStore {
         }
     }
 
+    #[instrument(
```
jkeiser, line 112: This is an optimization: `frigg.get_workspace_object()` currently deserializes the MV data to `serde_json::Value`, which means if you want it in another format (such as the actual `AttributeTree` MV you are essentially doing a two-hop deserialize of the entire tree from bytes -> Value -> AttributeTree. This method lets you avoid that by using serde to deserialize directly from bytes -> AttributeTree.

In the case of search this is even more pertinent, because search only needs a small subset of the AttributeTree to do its work. By deserializing directly to a smaller `AttributeTreeForSearch` struct, serde doesn't even allocate Strings, HashMaps or Vecs for attributes we don't want. This translated to a roughly 3x speedup of the search endpoint during testing of 1,000 components.

lib/luminork-server/src/search/component.rs
```diff
@@ -0,0 +1,196 @@
...
+use telemetry::prelude::*;
+
+use crate::search::{
+    Error,
+    Result,
+    SearchQuery,
+    SearchTerm,
+};
+
+/// Search for components matching the given query in the given workspace and change set.
```
jkeiser, line 31: Once you have a parsed `SearchQuery`, this is the endpoint that retrieves the components and matches them to the query. It pulls the index, then pulls / process each AttributeTree MV in a parallel tokio task (which is another massive speedup).

lib/luminork-server/src/search/component.rs
```diff
@@ -0,0 +1,196 @@
...
+) -> Result<Option<ComponentId>> {
+    let attribute_tree = attribute_tree_mv(frigg, workspace_pk, index_ref).await?;
+    if match_attribute_tree(&attribute_tree, &query) {
+        Ok(Some(attribute_tree.id))
+    } else {
+        Ok(None)
+    }
+}
+
+/// Match a component against a query or sub-query.
```
jkeiser, line 72: This is the meat of where we process the actual query. If you have a query like `Instance region:us-east-1`, it will look like And(MatchValue("Instance"), MatchAttr("region", "us-east-1")). The And() processor first checks ithe component matches MatchValue("Instance"), and if it does, then it checks MatchAttr("region", "us-east-1").

If there are multiple attr:value matches, each MatchAttr will be processed separately, which means it will read through all the attributes and do the match separately. There might be ways to batch them and be a little more performant, but this is by far the most straightforward way to write it (especially when you take full boolean logic into account).

lib/luminork-server/src/search/component.rs
```diff
@@ -0,0 +1,196 @@
...
+    // If it was attr: with no value, match anything with that attr.
+    if terms.len() == 0 {
+        return true;
+    }
+
+    // Otherwise, match if any of the values match.
+    terms.iter().any(|term| term.match_value(value))
+}
+
+/// A pared-down version of the AttributeTree MV, containing only the fields we need for searching.
```
jkeiser, line 131: This is the stripped-down version of `AttributeTree` we're using to avoid deserializing stuff we don't care about. In synchronous tests (doing the MVs serially so we can really see how much work we're doing), the endpoint body went from roughly 700ms -> 200ms by adding this optimization. (I'm stripping the extractor overhead from these numbers to get a clearer picture of how much it's actually helping the search.)

lib/luminork-server/src/search/parser.rs
```diff
@@ -0,0 +1,555 @@
+use super::{
+    Error,
+    Result,
+    SearchQuery,
+    SearchTerm,
+};
+
+/// Parses a search query.
```
jkeiser, line 8: The parser is written to be a direct copy of the frontend `search.ts` parser. I think there are better ways to do this that will be more robust, but it seemed prudent to *start* by replicating the logic as closely as possible, to maximize the chance we've got it right.

lib/luminork-server/src/search/parser.rs
```diff
@@ -0,0 +1,555 @@
...
+    // If there are a bunch of conditions with stray parens between, AND them all together.
+    if queries.len() > 1 {
+        Ok(SearchQuery::And(queries))
+    } else {
+        Ok(queries.into_iter().next().unwrap_or(SearchQuery::All))
+    }
+}
+
+/// Holds state for parsing a search query string.
+struct SearchQueryParser<'a> {
```
jkeiser, line 43: SearchQueryParser keeps track of what we're parsing next in the query string, and the code exclusively uses the `consume(<char>)` and `consume_until(<ch1> | <ch2> | ...)` methods to read the next bit of the query string, one at a time, inexorably marching forward and consuming every character as we go.

The parser *structure* is recursive descent: for example, `parse_or_condition()` calls `parse_and_condition()` in a loop and joins them with `|`; parse_and_condition() calls `parse_term()` and joins them with `&` or ' ' (two terms separated by spaces is `&`. This is one of the simpler ways to write a robust parser with precedence (i.e. a&b | c&d does a&b first, then c&d, then |'s them together). This is how the typescript parser works as well (which makes sense, as this is a nearly line-for-line translation of the TS parser).

lib/luminork-server/src/search/parser.rs
```diff
@@ -0,0 +1,555 @@
...
+        consumed
+    }
+
+    /// Whether we're at the end of the query string.
+    fn finished(&self) -> bool {
+        self.remaining_query_string.is_empty()
+    }
+}
+
+#[cfg(test)]
```
jkeiser, line 268: Unit tests for anything that can be reasonably tested! In this case, we try to test all the syntax variants we can think of and make sure they produce the right structure. These tests do *not* test whether the parsed query is actually processed correctly to match components, but this handles a lot of complex variants so that the downstream component matching tests can be simpler.

lib/luminork-server/src/search/query.rs
```diff
@@ -0,0 +1,317 @@
...
+
+impl FromStr for SearchQuery {
+    type Err = Error;
+
+    fn from_str(s: &str) -> Result<Self> {
+        parser::parse(s)
+    }
+}
+
+impl SearchTerm {
```
jkeiser, line 47: These utilities are designed to handle case sensitivity, number/boolean matching, etc. in a consistent way. That way we can test this stuff in one place, and the actual component-specific code is smaller and easier to test in isolation. This also will let us reuse a lot of this logic when we implement search for other things (schemas, functions, etc.)

lib/luminork-server/src/search/query.rs
```diff
@@ -0,0 +1,317 @@
...
+        assert!(!SearchTerm::Match("1".to_string()).match_bool(false));
+    }
+
+    #[test]
+    fn match_number() {
+        assert!(SearchTerm::Match("0".to_string()).match_number(&0.into()));
+        // assert!(SearchTerm::Match("-0".to_string()).match_number(&0.into()));
+
+        assert!(SearchTerm::Match("1".to_string()).match_number(&1.into()));
+        // assert!(SearchTerm::Match("1.0".to_string()).match_number(&1.into()));
```
jkeiser, line 259: serde_json does some crazy stuff with numbers--it doesn't consider 1 == 1.0. (1.00 == 1.0 though, and 1 == 1). I'm opting not to fix this for now, but it needs love.

lib/luminork-server/src/service/v1/components/search_components.rs
```diff
@@ -46,12 +57,21 @@ use crate::{
 )]
 pub async fn search_components(
     ChangeSetDalContext(ref ctx): ChangeSetDalContext,
+    FriggStore(ref frigg): FriggStore,
+    ChangeSetAuthorization {
```
jkeiser, line 61: ChangeSetAuthorization gives us a way to get the workspace_id and change_set_id independent of the DalContext (because very soon, we are not going to need the full DalContext for search--frigg is all you need).

lib/sdf-extract/src/change_set.rs
```diff
@@ -46,11 +49,48 @@ impl FromRequestParts<AppState> for ChangeSetDalContext {
         parts: &mut Parts,
         state: &AppState,
     ) -> Result<Self, Self::Rejection> {
+        // Get the workspace we are accessing (and authorized for)
+        let ChangeSetAuthorization { ctx, .. } = parts.extract_with_state(state).await?;
```
jkeiser, line 53: This code has literally just moved down into ChangeSetAuthorization, which is a direct analogue of WorkspaceAuthorization.

lib/sdf-server/Cargo.toml
```diff
@@ -9,7 +9,6 @@ rust-version.workspace = true
 publish.workspace = true
 
 [dependencies]
-async-openai = { workspace = true }
```
jkeiser, line 12: I noticed we still had this in here, when we don't use it--so I removed it. Cargo-only change.

PR #7420, opened 2025-10-03
feat(edda): Deployment MVs can be built incrementally

britmyerss:
### New Edda Request for incrementally building Deployment MVs
This new request kind is used by SDF only.  When we update cached modules, we record the `schema_id`s (if any) that have been removed as well as which `module_id`s have been added and send them in the request. This allows `Edda` to only build the MVs impacted by any newly rejected/published builtins.

If necessary, these requests get compressed down as with the Change Set update requests and the compression logic follows 'last write wins'. For example, if we reject, then re-promote the same module, and both of those requests get compressed, it'll get rebuilt and we'll ignore the initial rejection

### Stop building `CachedSchemaVariant` MV as we only ever want to expose the latest variant, which is the default
This has a **significantly positive** impact on performance, but was not the driving reason. When dealing with uninstalled schemas, we only ever want to expose the latest default variant, so no reason to even iterate and see if there are any more.  A few supporting notes: when we contribute a module, we only ever ship the default anyways and cached modules assumes modules are 1:1 with schemas, which means our Deployment MVs also become 1:1:1 for Module:Schema:Variant. 

I also updated `Luminork` to only find/return the `CachedDefaultSchemaVariant` as opposed to using the `CachedSchemaVariant`. 

I didn't fully rip out `CachedSchemaVariant` as I don't want to couple this change with MV changes. For now, we're not building them, and when building the `CachedSchema` we're still returning a list of `variant_id`s, but the list will only include the default. 

###  Expose a button in the admin panel to force rebuild all deployment MVs
There are now two buttons available in the admin panel (screenshot below).  One will only send the request to edda if we've removed or added cached modules, the other will force a full rebuild regardless.  Hopefully we don't have to use this, but it made testing easier, so might as well ship it. 

### When rebuilding only specific Deployment MVs, always check if any of the shapes have changed, and also rebuild those
Following on @jhelwig 's work here: https://github.com/systeminit/si/pull/7187 - we are now taking advantage of this mechanism for Deployment MVs.  This means, every time we rebuild only changed MVs, we will also check if we need to rebuild any other kinds because their shape has changed (`DefinitionChecksum`).  

### Telemetry
I've added much more instrumentation to this loop compared to what we have for building Change Set Mvs. It runs far less often, and we want to know about errors immediately. I can pair it back if it becomes too noisy. 
1. `info` messages when the processes kicks off/ends
2. wrapped each individual MV build with telemetry so we can better see the timings
3. SDF logs the update report so if things go haywire, we can compare what SDF thought it was doing, to what Edda did. 

#### Screenshots:
image

#### Out of Scope:

Refactoring this whole flow....... despite all of the temptation......

## How was it tested?

- [X] Added new unit tests to Edda for decompression logic
- [X] Manual tests: 
 -  Locally filter down to only 10 builtins being sent from the module index, see that the rebuild works as expected, and subsequent updates result in no messages as there are no changes
 - Update the filter to remove a few, and add a few, see that the update works as expected

<div><br/>via <a href="https://giphy.com/abcnetwork/">ABC Network</a> on <a href="https://giphy.com/gifs/abcnetwork-the-good-doctor-abc-doctorabc-Vf4Tu8M2flVAiOe2A1">GIPHY</a></div>

### DIFF
--- a/app/web/src/components/Workspace/WorkspaceAdminDashboard.vue
+++ b/app/web/src/components/Workspace/WorkspaceAdminDashboard.vue
@@ -21,6 +21,18 @@
               @click="updateModuleCache"
             />
           </div>
+          <div class="flex flex-row-reverse gap-sm">
+            <VButton
+              :disabled="adminStore.updatingModuleCacheOperationRunning"
+              class="flex-grow"
+              icon="plus-circle"
+              label="Update module cache and FORCE REBUILD"
+              loadingText="Updating module cache"
+              :loading="adminStore.updatingModuleCacheOperationRunning"
+              tone="success"
+              @click="updateAndRebuildModuleCache"
+            />
+          </div>
         </Stack>
         <Stack class="max-w-xl">
           <h2 class="font-bold text-lg">CLEAR INNIT PARAMETER CACHE</h2>
@@ -114,6 +126,10 @@ onBeforeMount(async () => {
 
 const updateModuleCache = async () => {
   await adminStore.UPDATE_MODULE_CACHE();
+};
+
+const updateAndRebuildModuleCache = async () => {
+  await adminStore.UPDATE_MODULE_CACHE(true);
 };
 
 const clearInnitCache = async () => {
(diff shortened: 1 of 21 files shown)


britmyerss:
Closing - this is no longer as urgent and not worth the complexity given removing building CachedSchemaVariant MVs brings the pipeline down to less than 20 seconds: https://github.com/systeminit/si/pull/7421

PR #7421, opened 2025-10-03
fix(edda): Stop building CachedSchemaVariant MVs

britmyerss:
Currently, when modules are promoted as builtins, they only contain a single schema, and a single variant (the default variant, which is the latest).  When exposing builtins as uninstalled, we only ever want to show the latest variant as we don’t want users to install old builtins.

We decided to stop building these while working on supporting incremental builds of the Deployment MVs, and at that time, we found that if we simply stop building the CachedSchemaVariantMV alltogether, it brings the total build time down to 16s (locally) for all current builtins. 

Given this, we likely don’t need to support incremental builds for some time, so we’ll ship this as a stand alone improvement and wait for our needs to grow again to revist incremental builds

## How was it tested?

Locally ran the sync, see that it finishes very quickly! (14s!)

<div><br/>via <a href="https://giphy.com/cbc/">CBC</a> on <a href="https://giphy.com/gifs/cbc-schitts-creek-J1M6o5U4M7LmoSYEYl">GIPHY</a></div>

### DIFF
--- a/lib/edda-server/src/materialized_view/deployment.rs
+++ b/lib/edda-server/src/materialized_view/deployment.rs
@@ -4,10 +4,7 @@ use std::{
     time::Duration,
 };
 
-use dal::{
-    DalContext,
-    cached_module::CachedModuleError,
-};
+use dal::DalContext;
 use frigg::FriggStore;
 use si_frontend_mv_types::{
     index::deployment::DeploymentMvIndexV2,
@@ -61,6 +58,7 @@ pub async fn build_all_mvs_for_deployment(
     reason_message: &'static str,
 ) -> Result<(), MaterializedViewError> {
     let span = current_span_for_instrument_at!("info");
+    info!("Started building all deployment MVs");
 
     // Discover all deployment MVs that need to be built
     let deployment_tasks = discover_deployment_mvs(ctx).await?;
@@ -116,7 +114,7 @@ pub async fn build_all_mvs_for_deployment(
     edda_updates
         .publish_deployment_index_update(index_update)
         .await?;
-
+    info!("Started building all deployment MVs");
     Ok(())
 }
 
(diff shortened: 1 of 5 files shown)


lib/edda-server/src/materialized_view/deployment.rs
```diff
@@ -359,12 +360,12 @@ async fn discover_deployment_mvs(
...
     });
 
-    // Discover individual CachedSchema and CachedSchemaVariant objects from cached modules
+    // Discover individual CachedSchema and CachedDefaultVariant objects from cached modules
     let modules = CachedModule::latest_modules(ctx).await?;
     for mut module in modules {
         let si_pkg = module.si_pkg(ctx).await?;
         let schemas = si_pkg.schemas().map_err(CachedModuleError::SiPkg)?;
-        for schema in schemas {
+        for _schema in schemas {
```
jhelwig: If `schema` is no longer used, do we still need this loop at all?

lib/edda-server/src/materialized_view/deployment.rs
```diff
@@ -507,61 +494,68 @@ async fn spawn_deployment_mv_task(
...
-
-            build_tasks.spawn(build_mv_for_deployment_task(
-                ctx.clone(),
-                frigg.clone(),
-                mv_id,
-                mv_kind,
-                dal_materialized_views::cached::schema::variant::assemble(ctx.clone(), variant_id),
-                maybe_deployment_mv_index,
-            ));
+            // Note(Brit): Currently, we only want to expose the latest builtin for a schema
```
jhelwig: `latest -> default`

lib/luminork-server/src/service/v1/schemas/get_variant.rs
```diff
@@ -133,16 +133,16 @@ pub async fn get_variant(
         }
     }
 
-    // Phase 1 Fallback: Try CachedSchemaVariant MV (deployment-level, for uninstalled modules)
+    // Phase 1 Fallback: Try CachedDefaultVariant MV (deployment-level, for uninstalled modules)
     match frigg
         .get_current_deployment_object(
-            ReferenceKind::CachedSchemaVariant,
+            ReferenceKind::CachedDefaultVariant,
             &schema_variant_id.to_string(),
```
jhelwig: The `CachedDefaultVariant` is keyed by `SchemaId`, not the `SchemaVariantId`.

jhelwig (changes requested):
One needs to change, one question, and one nit that isn't a blocker.  Definitely a good find that we can optimize the process by "Just don't do that work". 👍🏻

PR #7422, opened 2025-10-03
Prevent eviction of recently-created workspace snapshots

jhelwig:
Add configurable grace period to protect newly-created snapshots from
premature eviction. Previously,
snapshots could be evicted immediately after creation if no change sets
referenced them, but there is a race condition, especially around
creating new change sets while updates are happening.

If there was a single change set (`A`) referencing a snapshot with
updates queued for the rebaser, when we go to create a new change set
(`B`), it was possible for the new change set entry to reference the
snapshot that had been referenced by `A`, but had already been evicted
when the update went through before `B` is actually created to make a
second reference to the snapshot.

Modify evict_unused_snapshot() to check both usage status and recency
before evicting. A snapshot is only evicted if it's both unused AND
outside the grace period window. Default grace period is 5 minutes,
matching the existing quiescent period default.

This does not address the issue if the snapshot referenced by `A` in the
above scenario was originally created outside of the grace period, but
this tends to be a problem during busy periods in a workspace where the
likelihood of the snapshot being an older one is much less likely. We
would need recency-of-reference tracking to address this more fully, but
that is a much larger change, and we already know that we want/need to
make much larger changes to the snapshot garbage collection logic.

## How was it tested?

- [X] Integration tests pass

### DIFF
--- a/lib/dal/src/change_set.rs
+++ b/lib/dal/src/change_set.rs
@@ -1141,6 +1141,55 @@ impl ChangeSet {
         if count > 0 { Ok(true) } else { Ok(false) }
     }
 
+    /// Checks if a workspace snapshot was created within the specified grace period.
+    /// Returns true if the snapshot is recent (within grace period), false otherwise.
+    #[instrument(
+        name = "change_set.snapshot_is_recently_created",
+        level = "debug",
+        skip_all,
+        fields(
+            si.workspace_snapshot_address = %workspace_snapshot_address,
+            si.grace_period_secs = grace_period.as_secs(),
+        ),
+    )]
+    pub async fn snapshot_is_recently_created(
+        ctx: &DalContext,
+        workspace_snapshot_address: &WorkspaceSnapshotAddress,
+        grace_period: std::time::Duration,
+    ) -> ChangeSetResult<bool> {
+        // Query the workspace_snapshots table in si_layer_db
+        // PostgreSQL performs the age comparison directly
+        let key = workspace_snapshot_address.to_string();
+        let grace_period_secs = grace_period.as_secs_f64();
+
+        let row = ctx
+            .layer_db()
+            .workspace_snapshot()
+            .cache
+            .pg()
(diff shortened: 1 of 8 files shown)


lib/dal/src/change_set.rs
```diff
@@ -1141,6 +1141,55 @@ impl ChangeSet {
...
+        let row = ctx
+            .layer_db()
+            .workspace_snapshot()
+            .cache
+            .pg()
+            .query_opt(
+                "SELECT (CLOCK_TIMESTAMP() - created_at) < make_interval(secs => $2) AS is_recent FROM workspace_snapshots WHERE key = $1",
+                &[&key, &grace_period_secs],
+            )
+            .await?;
```
fnichol, line 1174: TIL about `make_interval()`

lib/rebaser-server/src/rebase.rs
```diff
@@ -502,17 +510,32 @@ pub(crate) async fn send_updates_to_edda_legacy_snapshot(
     Ok(())
 }
 
-pub(crate) async fn evict_unused_snapshots(
+pub(crate) async fn evict_unused_snapshot(
```
fnichol, line 513: Well, the singular form of this function name is a lot more accurate than the plural form

lib/rebaser-server/src/rebase.rs
```diff
@@ -502,17 +510,32 @@ pub(crate) async fn send_updates_to_edda_legacy_snapshot(
...
+        ?grace_period,
+        "Checking snapshot for potential eviction",
+    );
+    if !in_use && !is_recent {
         ctx.layer_db().workspace_snapshot().evict(
             workspace_snapshot_address,
             ctx.events_tenancy(),
             ctx.events_actor(),
         )?;
     }
```
fnichol, line 537: Yeah, I like it!

fnichol (approved):
Nicely done!

PR #7427, opened 2025-10-03
feat: Luminork top-level search endpoint

jkeiser:
This attaches search to a top-level search endpoint, which currently only returns components but which may return other things like schemas and functions in the future.

```
GET /v1/w/:workspaceId/change-sets/:changeSetId/search?q=region:us-east-1
{ "components": [
  {'id': '01K4T90DDMQC17HG3JQ72KJKYB', 'name': 'us-east-1', 'schema': {'name': 'Region'}},
  {'id': '01K6BV9HMB56T33N2Z0QZP8SV2', 'name': 'test-component-994', 'schema': {'name': 'AWS::EC2::Instance'}},
  {'id': '01K6BWP9X9B1Q0HSZAYVMYF3C1', 'name': 'another us-east-1', 'schema': {'name': 'Region'}}
] }
```
**Factor Budget:** because this endpoint uses the Posthog track_no_ctx, this lightly factors the method so callers pass `workspace_id` instead of all of them doing `Some(workspace_id.to_string())`. There are only a few callers of this method.

## How was it tested?

- [X] Integration tests pass
- [X] Manual test: new API endpoint works

### DIFF
--- a/lib/luminork-server/src/service/v1/search.rs
+++ b/lib/luminork-server/src/service/v1/search.rs
@@ -1,0 +1,90 @@
+use std::sync::Arc;
+
+use axum::{
+    Router,
+    extract::Query,
+    response::Json,
+    routing::get,
+};
+use sdf_core::app_state::AppState;
+use sdf_extract::{
+    FriggStore,
+    change_set::ChangeSetAuthorization,
+};
+use serde::{
+    Deserialize,
+    Serialize,
+};
+use serde_json::json;
+use utoipa::{
+    self,
+    ToSchema,
+};
+
+use crate::{
+    extract::PosthogEventTracker,
+    search::{
+        self,
+        component::ComponentSearchResult,
+    },
+    service::v1::ComponentsError,
+};
+
+pub fn routes() -> Router<AppState> {
+    Router::new().route("/", get(search))
+}
+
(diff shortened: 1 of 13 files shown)


lib/luminork-server/src/service/v1/search.rs
```diff
@@ -0,0 +1,90 @@
...
+    },
+    service::v1::ComponentsError,
+};
+
+pub fn routes() -> Router<AppState> {
+    Router::new().route("/", get(search))
+}
+
+#[utoipa::path(
+    post,
```
stack72: this needs to be a GET to match the route

lib/luminork-server/src/service/v1/search.rs
```diff
@@ -0,0 +1,90 @@
...
+    path = "/v1/w/{workspace_id}/change-sets/{change_set_id}/search",
+    params(
+        ("workspace_id" = String, Path, description = "Workspace identifier"),
+        ("change_set_id" = String, Path, description = "Change Set identifier"),
+    ),
+    request_body = SearchV1Request,
```
stack72: if we are not passing a request body then we need to move this to params

jkeiser:
Updated all, thanks for catching that!

PR #7428, opened 2025-10-03
feat(web): use bulk queries for workspace patches

zacharyhamm:
This changes the workspace patch handling to insert all new patches in bulk, and to be sure not to apply patches if the atom already exists.

This should make large patch batches much faster than inserting atoms one query at a time. It also ensures we do not run any sqlite3 ops in parallel, which can slow sqlite3 down.

Also fixes a bug in handleIndexMvPatch where we were not correctly checking for NOROW.

## How was it tested?

- [X] Integration tests pass


wendybujalski (approved):

app/web/src/workers/webworker.ts
```diff
@@ -307,8 +308,10 @@ const exec = (
 /**
  * A few small utilities
  */
+const textEncoder = new TextEncoder();
 const encodeDocumentForDB = async (doc: object) => {
-  return await new Blob([JSON.stringify(doc)]).arrayBuffer();
+  const docJson = JSON.stringify(doc);
+  return textEncoder.encode(docJson);
```
jobelenus, line 314: we're fairly confident there isn't any functional difference here? And we see speed gains?

zacharyhamm: i'm 100% confident there's no functional difference, but not sure where it caused a speed up or not, i'd have to measure again. it does mean we can remove the async form encodeDocumentForDB

jobelenus (approved):

britmyerss (approved):

PR #7431, opened 2025-10-03
feat: wildcard support in search API

jkeiser:
When you search for `AWS::EC2::*`, it should get you all EC2 components. When you search for region:us-*-1, it should give you us-east-1 and us-west-1.

## How was it tested?

- [X] Integration tests pass
- [X] Manual test: searched for AWS::EC2::* and region:us-*-1

### DIFF
--- a/lib/luminork-server/src/search/query.rs
+++ b/lib/luminork-server/src/search/query.rs
@@ -68,34 +68,18 @@ impl SearchTerm {
         // TODO handle international case comparison as well. Library? icu? case_sensitive_string?
         match self {
             SearchTerm::Match(term) => {
-                // Rust doesn't have a `contains_ignore_ascii_case()` function, so we do it
-                // ourselves by checking if the any subset of bytes in the value *equal* the
-                // query string (ignoring case).
-                let term = term.as_bytes();
-                let value = value.as_bytes();
-                value
-                    .windows(term.len())
-                    .any(|window| window.eq_ignore_ascii_case(term))
-            }
-            SearchTerm::Exact(term) => term.eq_ignore_ascii_case(value),
-            SearchTerm::StartsWith(term) => {
-                // Rust doesn't have a `starts_with_ignore_ascii_case()` function, so we do it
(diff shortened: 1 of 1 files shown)


lib/luminork-server/src/search/query.rs
```diff
@@ -68,34 +68,18 @@ impl SearchTerm {
...
-                // checks equality, because it's a byte-by-byte comparison.
-                let term = term.as_bytes();
-                let value = value.as_bytes();
-                if value.len() >= term.len() {
-                    term.eq_ignore_ascii_case(&value[..term.len()])
-                } else {
-                    false
+                // If we have a*b*c, find a, then b, then c ...
+                let mut value = value;
+                for part in term.split('*') {
```
jkeiser, line 73: To answer a question nobody asked: in Rust `split()` doesn't allocate, it's an iterator that scans through the string.

lib/luminork-server/src/search/query.rs
```diff
@@ -68,34 +68,18 @@ impl SearchTerm {
...
+                    let Some((_, remaining)) = value.split_once_ignore_ascii_case(part) else {
```
jkeiser, line 74: When you try to match `abc*def*ghi` against `_abc_def_ghi_`, we:

1. `_abc_def_ghi_`: search anywhere in the string for the first `abc` (case insensitive). value = remaining
2. `_def_ghi_`: search for first `def`. value = remaining
3. `ghi_`: search for first `ghi`. value = remaining
4. `_`: nothing else to search! Matches!

lib/luminork-server/src/search/query.rs
```diff
@@ -146,6 +130,41 @@ impl SearchTerm {
     }
 }
 
+/// String methods that ignore ascii case (like eq_ignore_ascii_case for other pattern matches)
```
jkeiser, line 133: I moved the methods that ignore ASCII case down into a little trait, to make it clearer that there was a correspondence between these methods and other Rust `str` methods.

PR #7436, opened 2025-10-05
chore(deps): remove unused deps

xuhongxu96:

We've identified many redundant dependencies in bazel build scripts and refactored them, as part of a research project on dependency reduction.

Feel free to leave feedback, suggest improvements, or directly edit this PR.

Thanks for your support!


fnichol:
Thanks for the PR! I noticed that you mentioned Bazel build scripts in your summary, however we're using the [Buck2](https://buck2.build/) build system. There is a very strong overlap in approach *and* in the format and syntax so it's possible that your tooling happened to work with our build definitions.

I'd be curious as to your approach to determine the unused dependencies in the Buck2 packages in this repo. Are you using any existing tools/products or is this is result of some research processes? In eyeballing some of the changes, some of these dependency removals are very plausible where others would not trivially compile.

I think without understanding more about the process that led to these changes, it will be hard for us to start pairing down build-able combinations that would yield a smaller, working set of changes. Thank you!

xuhongxu96:
> Thanks for the PR! I noticed that you mentioned Bazel build scripts in your summary, however we're using the [Buck2](https://buck2.build/) build system. There is a very strong overlap in approach _and_ in the format and syntax so it's possible that your tooling happened to work with our build definitions.
> 
> I'd be curious as to your approach to determine the unused dependencies in the Buck2 packages in this repo. Are you using any existing tools/products or is this is result of some research processes? In eyeballing some of the changes, some of these dependency removals are very plausible where others would not trivially compile.
> 
> I think without understanding more about the process that led to these changes, it will be hard for us to start pairing down build-able combinations that would yield a smaller, working set of changes. Thank you!

Sorry for the typo and thanks for your feedback. Yes, we noticed that si uses Buck2 and our tool supports Buck2 as well. The result comes from our research project (a new tool).

During reduction, we validated the changes by executing the following:
```
buck2 uquery 'kind("rust_(binary|library|test)", set("//bin/..." "//lib/..."))' | xargs buck2 build
```
So I believe the project should be able to compile.

PR #7438, opened 2025-10-06
feat(luminork): Error when a user tries to update a readonly property

stack72:
FIXES: BUG-1080

When a user tries to update a read-only prop, we should throw a 412.

Trying to change the KeyName of an AWS::EC2::KeyPair as the following request:

```
curl -X 'PUT' \
  'http://localhost:5380/v1/w/01HPJ3K7RC486W9RPR3T0B6HS2/change-sets/01K6WQ2YG46GJYSXFA4S58F2EX/components/01K6WPWQX5VCJ2NETCDQ433VEP' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "attributes": {
    "/domain/KeyName": "updated-key-name
  }
}'
```
now throws an error so that the user knows what the issue is:

```
{
  "message": "attribute error: cannot update create-only property at path '/domain/KeyName' when component has a resource attached",
  "statusCode": 412,
  "code": null
}
```

### DIFF
--- a/lib/dal/src/attribute/attributes.rs
+++ b/lib/dal/src/attribute/attributes.rs
@@ -33,10 +33,12 @@ use crate::{
     Component,
     DalContext,
     Func,
+    Prop,
     PropKind,
     WsEvent,
     component::resource::ResourceData,
     func::intrinsics::IntrinsicFunc,
+    prop::PropError,
     workspace_snapshot::node_weight::{
         NodeWeight,
         reason_node_weight::Reason,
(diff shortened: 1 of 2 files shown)


aaron-dernley (approved):

lib/dal/src/attribute/attributes.rs
```diff
@@ -278,6 +297,19 @@ async fn update_attributes_inner(
                 // Create the attribute at the given pa th if it does not exist
                 let path = av_to_set.path();
                 let target_av_id = av_to_set.clone().vivify(ctx, component_id).await?;
+
+                // Check if this is a create-only property and component has a resource
+                let component = Component::get_by_id(ctx, component_id).await?;
+                if component.resource(ctx).await?.is_some() {
+                    let prop_id = AttributeValue::prop_id(ctx, target_av_id).await?;
+                    let prop = Prop::get_by_id(ctx, prop_id).await?;
```
britmyerss: let's switch this - first check if it's a create_only, and then, use `Component::resource_by_id` function as we don't need the full component here.

stack72: rgr!

britmyerss:
@stack72 we should add coverage of this in an integration test too if we don't have it

britmyerss (approved):

PR #7439, opened 2025-10-06
fix(si-mcp-server): Capture when trying to update a read-only prop

stack72:
Depends-on #7438

When a user tries to update a READ-ONLY prop, we need to capture the error and hint the user:

```
can you update the KeyPair component and set the KeyName to be "updated-name"

⏺ I'll update the KeyPair component's KeyName to "updated-name". Let me first find the KeyPair component and then update it.
  ⎿  <response>{"status":"success","data":[{"id":"01K6DS9XXCXW5EDW22920QRAQE","name":"HEAD","status":"Open","isHead":true}]}</response>
     <hints>Apply, Force-Apply, and Abandon change sets using their 'id' and the change-set-status-update tool.</hints>

⏺ si-mcp-server - component-list (MCP)(changeSetId: "01K6DS9XXCXW5EDW22920QRAQE", filters: {"filterGroups":[{"responseField":"schemaName","regularExpressions":["KeyPair"]}]})
  ⎿  <response>{"status":"success","data":[{"componentId":"01K6WPWQX5VCJ2NETCDQ433VEP","componentName":"si-2225","schemaName":"AWS::EC2::KeyPair"}]}</response>

⏺ I found a KeyPair component but I can't update it in HEAD. Let me create a change set for you first.

⏺ si-mcp-server - change-set-create (MCP)(changeSetName: "Update KeyPair name")
  ⎿  <response>{"status":"success","data":{"id":"01K6WSGTRGM7P5FP6FA5R60R03","name":"Update KeyPair name","status":"Open","isHead":false}}</response>

⏺ si-mcp-server - component-update (MCP)(changeSetId: "01K6WSGTRGM7P5FP6FA5R60R03", componentId: "01K6WPWQX5VCJ2NETCDQ433VEP", attributes: {"/domain/KeyName":"updated-name"})
  ⎿  Error: <response>{"status":"failure","errorMessage":"Error Status: 412\nError Data: {\"message\":\"attribute error: cannot update create-only property at path '/domain/KeyName' when component has
      a resource attached\",\"statusCode\":412,\"code\":null}\nMessage:Request failed with status code 412"}</response>
     <hints>The property '/domain/KeyName' is read-only and cannot be updated once a resource is attached. Consider duplicating this component with the new value instead.</hints>

⏺ The KeyName property is read-only once the KeyPair resource exists. You can't update it. You would need to create a new KeyPair component with the desired name instead.
```

### DIFF
--- a/bin/si-mcp-server/deno.json
+++ b/bin/si-mcp-server/deno.json
@@ -12,7 +12,7 @@
     "@onjara/optic": "jsr:@onjara/optic@^2.0.3",
     "@std/assert": "jsr:@std/assert@1",
     "@std/encoding": "jsr:@std/encoding@^1.0.10",
-    "@systeminit/api-client": "jsr:@systeminit/api-client@^1.1.8",
+    "@systeminit/api-client": "jsr:@systeminit/api-client@^1.3.0",
     "axios": "npm:axios@^1.6.1",
     "jwt-decode": "npm:jwt-decode@^4.0.0",
     "lodash": "npm:lodash@^4.17.21",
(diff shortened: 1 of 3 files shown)


aaron-dernley:
zing!

PR #7445, opened 2025-10-06
chore(luminork): Regenerate docs based on new Search API

stack72:


app/docs/src/reference/public-api.md
```diff
@@ -2775,6 +2776,46 @@ Management functions endpoints
...
+|change_set_id|path|string|true|Change Set identifier|
+|q|query|string|true|Query string. See https://docs.systeminit.com/explanation/search-syntax for details.|
+
+> Example responses
+
+> 200 Response
+
+```json
+{
+  "components": [
```
jkeiser, line 2803: Oops, looks like I messed up the search response! I also thought I added an example search but don't see it here.

stack72: this is autogenerated so we should ship this now and then fix the API up

jkeiser (approved):

PR #7449, opened 2025-10-06
Web worker patching priority

jobelenus:
Give the patching queue a priority based on the user's activity on what change sets they've selected.

Also we needed a type fix to the last PR


zacharyhamm (approved):
Works! We may need to play with priorities a bit more since even with the higher priority it's not emphasizing HEAD as much as I expected, but this should still be an improvement

PR #7452, opened 2025-10-07
chore(luminork): Ensure we capture the specific errors when luminork returns non 2xx responses

britmyerss:
Discovered while investigating something, we’ve got the default tracing for luminork but not the actual error messages which can be helpful to see.

Adds an error/warn event when we handle `ApiError`s in Luminork

#### Screenshots:
image

## How was it tested?

Manually triggered a few expected errors and saw the events accordingly.

### DIFF
--- a/lib/luminork-server/src/service/v1/common.rs
+++ b/lib/luminork-server/src/service/v1/common.rs
@@ -11,6 +11,7 @@ use serde::{
     Deserialize,
     Serialize,
 };
+use telemetry::prelude::*;
 use thiserror::Error;
 use utoipa::ToSchema;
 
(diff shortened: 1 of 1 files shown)


lib/luminork-server/src/service/v1/common.rs
```diff
@@ -62,7 +63,17 @@ impl IntoResponse for ApiError {
     fn into_response(self) -> Response {
         let status =
             StatusCode::from_u16(self.status_code).unwrap_or(StatusCode::INTERNAL_SERVER_ERROR);
+
+        // Store message before moving self
+        let message = self.message.clone();
```
stack72, line 68: Is there any reason why we don't do something similar in SDF or does this point to a difference between how the projects are setup?

britmyerss: it's there too - but across a few spots in sdf. The default extractor is shared, but converting errors to responses past the extractors is not

PR #7457, opened 2025-10-07
get_variant endpoint now has from_builtin boolean

wendybujalski:

Adds the boolean value `from_builtin` to the response for the `get_variant` Luminork endpoint.

This boolean is calculated by checking if we find a `CachedSchema` for the variant's `schema_id` - to do this, the `schema_id` was added to the `luminork_schema_variant` MV. 

We need this boolean so that we can warn users of the SI AI Agent about the possibility of losing their customizations of builtins when upgrading.

#### Screenshots:

Screenshot 2025-10-07 at 1 14 17 PM

#### Out of Scope:

We may want this boolean elsewhere on other Luminork endpoints in the future, but for now it is just on the one.


lib/luminork-server/src/service/v1/schemas/mod.rs
```diff
@@ -286,6 +286,8 @@ pub struct GetSchemaVariantV1Response {
     pub color: String,
     #[schema(value_type = bool, example = false)]
     pub is_locked: bool,
+    #[schema(value_type = bool, example = false)]
+    pub from_builtin: bool,
```
stack72: I don't think we should call it 'from_builtin` - we wll need to add vocab about what a builtin is otherwise. This is a public consumable endpoint so we need to make it a wee bit more descriptive

stack72:
We also need to add this to get_default_Variant

lib/luminork-server/src/service/v1/schemas/mod.rs
wendybujalski: We waffled on the name a bit - do you have a suggestion?

lib/dal/tests/integration_test/component.rs
```diff
@@ -1,16 +1,31 @@
 use dal::{
-    AttributeValue, AttributeValueId, Component, DalContext, Schema, SchemaVariant, Ulid,
+    AttributeValue,
```
zacharyhamm, line 2: This will fail lints for formatting (run `cargo fmt`)

zacharyhamm: or maybe not! but how did it change so much i wonder

zacharyhamm (approved):

lib/dal/tests/integration_test/component.rs
wendybujalski: 🤷‍♀️

wendybujalski: I am pretty sure that file only changed upon running `cargo fmt` - not sure why 🤔

zacharyhamm:
Good solution to just use the MV for the cached modules. Ship it!

PR #7466, opened 2025-10-07
feat(web): Skip onboarding flow for invited users

stack72:
If a user has onboard to a workspace and then invites another user and that user is new to SI, we should skip the onboarding flow for that invited user.

The way we determine if the workspace has been onboarded is that there are components on HEAD. These components would be the credential and the region as part of the onboarding

If a user deletes them, then it’s fine for an invited user to go through onboarding as the aim of the game is to make that workspace ready for action!

### DIFF
--- a/app/web/src/newhotness/Workspace.vue
+++ b/app/web/src/newhotness/Workspace.vue
@@ -489,11 +489,36 @@ const checkOnboardingCompleteData = async () => {
   onboardingCompleted.value = !data.firstTimeModal;
 };
 
+const componentsOnHeadApi = useApi(ctx.value);
+const componentsOnHeadQuery = useQuery<boolean | undefined>({
+  queryKey: ["componentsOnHead", workspacePk.value],
+  queryFn: async () => {
+    const call = componentsOnHeadApi.endpoint<{ componentsOnHead: boolean }>(
+      routes.ComponentsOnHead,
+    );
+    const response = await call.get();
+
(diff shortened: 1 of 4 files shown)


app/web/src/newhotness/Workspace.vue
```diff
@@ -496,12 +496,25 @@ const lobby = computed(
...
+  componentsOnHead.value = response.data.componentsOnHead;
+};
+
 const showOnboarding = computed(() => {
   // Force Onboarding2 if this debug boolean is set to true
   if (DEBUG_MODE) return true;
 
   if (!featureFlagsStore.INITIALIZER_ONBOARD) return false;
 
+  if (componentsOnHead.value === true) return false;
```
jobelenus, line 534: If we're at all concerned with this endpoint not returning (token failure, SDF down / deploying, etc) we should wrap this with tan stack and use a "undefined" / true / false pattern, so in the case of failure (we know we tried the endpoint) we can let "undefined" skip the onboarding.

jobelenus (approved):

app/web/src/newhotness/Workspace.vue
stack72: I can check the status of the endpoint if that would be better?

stack72: I will follow on with changes to make this better

jobelenus (approved):

PR #7468, opened 2025-10-08
feat: enable authoring mcp tools

aaron-dernley:

Enables the Schema and Function authoring tools in the SI MCP server.

## How was it tested?

- [X] Integration tests pass
- [X] Manual test: Each iteration of this work was manually tested using the MCP server with Claude.


aaron-dernley:
@stack72 ready for when we want to enable the authoring tools!

stack72:
@aaron-dernley you should run `deno task build` from bin/si-mcp-server to update the deno.lock then we are god to ship!

PR #7469, opened 2025-10-08
fix: bump esbuild version in api example docs repo

aaron-dernley:
Fixes - https://github.com/systeminit/si/security/dependabot/436

### DIFF
--- a/app/docs/public-api-examples/typescript/package-lock.json
+++ b/app/docs/public-api-examples/typescript/package-lock.json
@@ -13,6 +13,7 @@
       "devDependencies": {
         "@types/node": "^20.0.0",
         "@vitest/coverage-v8": "^2.0.0",
+        "esbuild": "^0.25.10",
         "typescript": "^5.0.0",
         "vitest": "^2.0.0"
       },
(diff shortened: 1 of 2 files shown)


app/docs/public-api-examples/typescript/package-lock.json
```diff
@@ -8,6 +8,7 @@
       "name": "@si/public-api-examples-typescript",
       "version": "1.0.0",
       "dependencies": {
+        "esbuild": "^0.25.10",
```
johnrwatson: @stack72 Does this look right? I'd expect this dep to not be here to be honest, but maybe it was missing originally for the examples

app/docs/public-api-examples/typescript/package.json
```diff
@@ -10,6 +10,7 @@
     "test:coverage": "vitest run --coverage"
   },
   "dependencies": {
+    "esbuild": "^0.25.10",
```
stack72: I believe this should be a devDependency not an actual dependency

aaron-dernley: Updated that @stack72

stack72 (approved):

PR #7477, opened 2025-10-09
[feat] Hetzner: add name to Locations component

jkeiser:
Right now you can't actually specify the location in Hetzner::Location. This adds the ability to set it and a suggestion to subscribe to it.

* Adds /domain/name to Hetzner::Location
* Adds suggestions to all fields named "/domain/location" in other Hetzner objets
* Makes Locations.name and all location fields dropdowns with valid values including all Hetzner regions 

#### Screenshots:

image

#### Out of Scope:

* Qualifications validating that the location is valid for the account

## How was it tested?

- [X] Manual test: Suggestions and subscriptions work from Server.location -> Locations.name
- [ ] Manual test: Server is actually created when name is specified from dropdown

NOTE: no reason to think create action will fail (Locations domain and create action haven't changed) but I haven't got a credential set up yet to do so--will make sure in the morning.

### DIFF
--- a/bin/clover/src/pipelines/hetzner/overrides.ts
+++ b/bin/clover/src/pipelines/hetzner/overrides.ts
@@ -1,20 +1,38 @@
+import { createScalarProp } from "../../spec/props.ts";
+import { widget, suggest } from "../generic/overrides.ts";
 import { PropOverrideFn, SchemaOverrideFn } from "../types.ts";
+
+const HETZNER_LOCATIONS = ["fsn1", "nbg1", "hel1", "ash", "hil", "sin"];
 
 // Hetzner-specific property overrides (empty for now)
 export const HETZNER_PROP_OVERRIDES: Record<
   string,
   Record<string, PropOverrideFn | PropOverrideFn[]>
 > = {
-  // Add Hetzner-specific property overrides here as needed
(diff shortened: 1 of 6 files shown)


bin/clover/src/pipelines/hetzner/overrides.ts
```diff
@@ -1,20 +1,38 @@
...
   Record<string, PropOverrideFn | PropOverrideFn[]>
 > = {
-  // Add Hetzner-specific property overrides here as needed
-  // Example:
-  // "Hetzner::Cloud::Server": {
-  //   LocationsProp: makeDropdownOverride,
-  // },
+  ".*": {
+    // Add location dropdown and suggestion to "location" prop on all resources
+    location: [
```
britmyerss, line 14: not related to this change - but do we have prop suggestions for the Hetzner cred too?

sprutton1: [I think so ](https://github.com/systeminit/si/blob/f5d4fd9990934e507a83a8bedd5bf37dd40399d4/bin/clover/src/pipelines/hetzner/pipeline-steps/addDefaultProps.ts#L165-L168)

PR #7479, opened 2025-10-09
fix: upgrade version of tmp in pnpm

aaron-dernley:

Fixes - https://github.com/systeminit/si/security/dependabot/427

## How was it tested?

- [X] Integration tests pass
- [X] Manual test: UI works as expected


pnpm-lock.yaml
```diff
@@ -18648,7 +18640,7 @@ snapshots:
     dependencies:
       chardet: 0.7.0
       iconv-lite: 0.4.24
-      tmp: 0.0.33
+      tmp: 0.2.4
```
johnrwatson, line 18643: @aaron-dernley I don't think we should change it like this. We should pull the updated version of `external-editor` if possible, which would transitively pull the newer version of `tmp`. The change suggested here is a bit risky as external-editor @ 3.1.0 might not be compatible completely with tmp @ 0.2.4 even if all the cases we are `currently` use it still work.

aaron-dernley:
Marking as no fix in dependabot for the time being.

PR #7480, opened 2025-10-09
feat(web): Alter how GridTiles are shown when in hover or selected states

stack72:
The selection checkbox should not show until the Tile is either in hovered or selected. We need to show a better onHover state with a border and a different background and when selected we show a slightly thicker border

This is as per figma and will look as follows: https://jam.dev/c/20d304d9-ac61-47e8-8c56-0b643dcbc669

### DIFF
--- a/app/web/src/newhotness/explore_grid/ExploreGridTile.vue
+++ b/app/web/src/newhotness/explore_grid/ExploreGridTile.vue
@@ -4,16 +4,22 @@
     :class="
       clsx(
         'component tile',
-        'cursor-pointer rounded-sm overflow-hidden relative select-none',
-        hasRunningActions ? 'spinning-border' : 'border',
+        'cursor-pointer rounded-sm overflow-hidden relative select-none border',
+        hasRunningActions && 'spinning-border',
         !hasRunningActions && [
           selected || focused
-            ? themeClasses('border-action-500', 'border-action-300')
+            ? [
+                themeClasses('border-neutral-600', 'border-neutral-400'),
+                themeClasses(
+                  'outline outline-1 outline-neutral-600',
+                  'outline outline-1 outline-neutral-400',
+                ),
+              ]
             : hasFailedActions
             ? themeClasses('border-destructive-500', 'border-destructive-400')
             : [
-                hovered
-                  ? themeClasses('border-black', 'border-white')
+                isHovering
+                  ? themeClasses('border-action-700', 'border-action-600')
                   : themeClasses('border-neutral-400', 'border-neutral-600'),
               ],
         ],
@@ -21,13 +27,18 @@
         component.toDelete && 'opacity-70',
       )
     "
+    @mouseenter="handleMouseEnter"
+    @mouseleave="handleMouseLeave"
   >
     <header
       :class="
         clsx(
           'p-xs pr-sm',
-          !component.toDelete &&
-            themeClasses('bg-neutral-200', 'bg-neutral-800'),
+          !component.toDelete && [
+            isHovering
+              ? themeClasses('bg-neutral-300', 'bg-neutral-600')
+              : themeClasses('bg-neutral-200', 'bg-neutral-800'),
+          ],
         )
       "
     >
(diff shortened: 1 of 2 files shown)


app/web/src/newhotness/explore_grid/ExploreGridTile.vue
```diff
@@ -5,29 +5,35 @@
       clsx(
         'component tile',
         'cursor-pointer rounded-sm overflow-hidden relative select-none',
-        hasRunningActions ? 'spinning-border' : 'border',
+        hasRunningActions || selected ? 'border-2' : 'border',
```
wendybujalski: Right now, switching from `border` to `border-2` here causes the tile to shift in position/layout very slightly. Normally I would suggest using `outline` or `box-shadow` to increase the border instead of adjusting the actual border width, but because of `spinning-border` this may be a place to use `box-border`

app/web/src/newhotness/explore_grid/ExploreGridTile.vue
```diff
@@ -5,29 +5,35 @@
...
+        hasRunningActions && 'spinning-border',
         !hasRunningActions && [
           selected || focused
-            ? themeClasses('border-action-500', 'border-action-300')
+            ? themeClasses('border-action-500', 'border-neutral-400')
```
wendybujalski: Is having an `action` style for light mode and a `neutral` style for dark mode intentional here?

app/web/src/newhotness/explore_grid/ExploreGridTile.vue
```diff
@@ -5,29 +5,35 @@
...
             : hasFailedActions
             ? themeClasses('border-destructive-500', 'border-destructive-400')
             : [
-                hovered
-                  ? themeClasses('border-black', 'border-white')
+                isHovering
+                  ? themeClasses('border-action-500', 'border-action-600')
```
wendybujalski: I'm surprised to see 500/600 `action` styles here, as the standard for the rest of the UI has been 500/300 for selection borders. Something to consider in terms of consistency perhaps? We can always set a new standard or have this one spot be different if we want of course!

app/web/src/newhotness/explore_grid/ExploreGridTile.vue
stack72: it's not - I will be chosing a better style when Victoria

app/web/src/newhotness/explore_grid/ExploreGridTile.vue
stack72: this is from Figma

app/web/src/newhotness/explore_grid/ExploreGridTile.vue
```diff
@@ -5,29 +5,35 @@
...
                   : themeClasses('border-neutral-400', 'border-neutral-600'),
               ],
         ],
         themeClasses('bg-shade-0', 'bg-neutral-900'),
         component.toDelete && 'opacity-70',
       )
     "
+    @mouseenter="handleMouseEnter"
+    @mouseleave="handleMouseLeave"
```
wendybujalski, line 31: Handling hover this way works but is best to avoid when possible. Based on the effects of hover here, we could just use group-hover to handle everything instead.

app/web/src/newhotness/explore_grid/ExploreGridTile.vue
```diff
@@ -5,29 +5,35 @@
...
   >
     <header
       :class="
         clsx(
           'p-xs pr-sm',
-          !component.toDelete &&
-            themeClasses('bg-neutral-200', 'bg-neutral-800'),
+          !component.toDelete && [
+            isHovering
+              ? themeClasses('bg-action-100', 'bg-neutral-600')
```
wendybujalski: https://github.com/systeminit/si/pull/7480/files#r2416681939

app/web/src/newhotness/explore_grid/ExploreGridTile.vue
```diff
@@ -165,10 +171,14 @@
...
-                'border-neutral-400 hover:border-black',
-                'border-neutral-600 hover:border-white',
-              ),
+            : [
+                isHovering
+                  ? themeClasses('border-black', 'border-white')
+                  : themeClasses('border-neutral-400', 'border-neutral-600'),
+              ],
+          isHovering
+            ? themeClasses('bg-action-100', 'bg-neutral-600')
```
wendybujalski: https://github.com/systeminit/si/pull/7480/files#r2416681939

stack72:
This won't be merged until I get updated styles for light mode

PR #7483, opened 2025-10-09
feat: iterate on management funcs via mcp

aaron-dernley:

This PR adds outputting the list of functions to the schemaCreateEditGet tool. That allows an MCP user to iterate on an existing or newly created  Run Template management function on a Template Schema.

This PR also fixes the funcCreateEditGet tool so that it can handle locked functions properly, automatically ensuring that the function and its schema are unlocked.

This work completes [ENG-3252](https://linear.app/system-initiative/issue/ENG-3252/ability-to-iterate-on-run-template-management-functions)!

## Screenshot

Screenshot 2025-10-09 at 6 57 40 PM

## How was it tested?

- [X] Integration tests pass
- [X] Manual tests: all the demos [in the Linear](https://linear.app/system-initiative/issue/ENG-3252/ability-to-iterate-on-run-template-management-functions) work!

### DIFF
--- a/bin/si-mcp-server/src/tools/funcCreateEditGet.ts
+++ b/bin/si-mcp-server/src/tools/funcCreateEditGet.ts
@@ -920,7 +920,7 @@ const funcCreateEditGetInputSchemaRaw = {
     .describe(
       "The change set to create, update, get information about a function in; functions cannot be manipulated on HEAD",
     ),
-  schemaId: z.string().optional().describe("The schema id the function is for. Required for creating a new function, not needed for updating or getting information about one."),
+  schemaId: z.string().describe("The schema id the function is for."),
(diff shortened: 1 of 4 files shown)


bin/si-mcp-server/src/tools/funcCreateEditGet.ts
```diff
@@ -920,7 +920,7 @@ const funcCreateEditGetInputSchemaRaw = {
     .describe(
       "The change set to create, update, get information about a function in; functions cannot be manipulated on HEAD",
     ),
-  schemaId: z.string().optional().describe("The schema id the function is for. Required for creating a new function, not needed for updating or getting information about one."),
+  schemaId: z.string().describe("The schema id the function is for."),
```
wendybujalski, line 923: schemaId is now required, to ensure that if you are editing or getting info about a function it can be unlocked if it is locked

PR #7489, opened 2025-10-09
feat(dal): management prototype overlays

zacharyhamm:
Adds the ability to create a mangement prototype that is parented by a schema instead of a schema variant. These will be listed in addition to schema variant level management prototypes when prototypes are listed for a component.

### DIFF
--- a/lib/dal/src/management/prototype.rs
+++ b/lib/dal/src/management/prototype.rs
@@ -10,6 +10,7 @@ use serde::{
     Serialize,
 };
 use si_events::FuncRunId;
+use si_id::ulid::Ulid;
 use strum::Display;
 use telemetry::prelude::*;
 use thiserror::Error;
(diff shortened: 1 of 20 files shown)


lib/dal/src/management/prototype.rs
```diff
@@ -296,6 +301,24 @@ pub enum ManagementFuncKind {
     RunTemplate,
 }
 
+#[derive(Debug, Clone)]
+pub enum ManagementPrototypeParent {
+    Schemas(Vec<SchemaId>),
```
britmyerss, line 306: Why multiple Schemas here?

lib/dal/src/management/prototype.rs
```diff
@@ -476,6 +503,26 @@ impl ManagementPrototype {
         })
     }
 
+    /// Make this management prototype an "overlay" action prototype.
+    pub async fn promote_to_overlay(
```
britmyerss, line 507: nice!

lib/dal/src/management/prototype.rs
zacharyhamm: We decided before to make it possible for overlays to be portable to multiple schemas (although we don't have a UX story for this yet)

britmyerss: ah yes ok!

britmyerss:
Looks great! One thing that came to mind, is we might want to be able to pass along whether a prototype is an overlay or not, but we can solve this later when we learn we need it. Great work!

PR #7490, opened 2025-10-09
chore(ci): Put luminork into maintenance mode during deploy

stack72:


lib/luminork-server/config/default.toml
```diff
@@ -17,7 +17,7 @@ ws_url = "ws://localhost:5157"
 http_url = "http://localhost:5157"
 
 [service]
-port = 5156
+port = 5380
```
johnrwatson, line 20: Is this just a typo from the sdf->luminork copy? If the file has no value maybe just deleting the file is better

stack72: yes, it was set to the same port by default - which was wrong

nickgerace:
@stack72 @johnrwatson do we still want this? No rush, just curious.

PR #7493, opened 2025-10-10
[feat] Hetzner: Add SSH key suggestion

jkeiser:
This adds a suggestion for Hetzner::Servers.ssh_keys to use Hetzner::SshKey.name (resource_value).

It also factors the prop override to create "name" a bit to be more readable and less redundant; we should be able to reuse this in the AWS schemas as well.

## How was it tested?

- [X] Integration tests pass
- [ ] Manual test: Servers suggest SshKey.name

### DIFF
--- a/bin/clover/src/pipelines/generic/overrides.ts
+++ b/bin/clover/src/pipelines/generic/overrides.ts
@@ -1,10 +1,13 @@
 import { ExpandedPkgSpec } from "../../spec/pkgs.ts";
 import {
   addPropSuggestSource,
+  createScalarProp,
   ExpandedPropSpec,
   ExpandedPropSpecFor,
   findPropByName,
+  PropPath,
   propPathStr,
+  toPropPathArray,
 } from "../../spec/props.ts";
 import { PropUsageMap } from "../aws/pipeline-steps/addDefaultPropsAndSockets.ts";
 import { ActionFuncSpecKind } from "../../bindings/ActionFuncSpecKind.ts";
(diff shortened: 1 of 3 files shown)


bin/clover/src/pipelines/generic/overrides.ts
```diff
@@ -282,3 +285,59 @@ export function widget(
     }
   };
 }
+
+export function addScalarProp(
```
sprutton1, line 289: I didn't get this until I saw the call site. Nice, I like that you can just use the path like that.

bin/clover/src/spec/props.ts
```diff
@@ -787,10 +783,33 @@ export function propSuggestionFor(
...
+  return ["root", ...propPath.split("/").slice(1)];
+}
+
+export function toPropPath(
+  propPathArray: ExpandedPropSpec["metadata"]["propPath"],
+): PropPath {
+  if (propPathArray[0] !== "root") {
+    throw new Error(`propPath array must start with root: ${propPathArray}`);
+  }
+  if (propPathArray[1] !== "domain" && propPathArray[1] !== "resource_value") {
```
sprutton1, line 816: Technically this could start with `secrets` or `secrets_definition` but it seems like that isn't an issue any where at the moment.

bin/clover/src/spec/props.ts
```diff
@@ -120,6 +120,8 @@ export type CreatePropQueue = {
 
 const MAX_PROP_DEPTH = 30;
 
+export type PropPath = `/${"domain" | "resource_value"}/${string}`;
```
sprutton1, line 124: I did not know you could define types like this. Typescript is wild sometimes.

bin/clover/src/spec/props.ts
jkeiser: Yeah, I figure we'll add that when we need it--it just needs to be specific to a small subset of possible paths, since we have to look these top level ones up differently from object children.

bin/clover/src/spec/props.ts
jkeiser: I love it :)

PR #7494, opened 2025-10-10
[feat] Hetzner: qualification for location name

jkeiser:
Qualifies the location name in Hetzner::Locations.

Also adds some affordances for easily adding qualification function overrides.

image

## How was it tested?

- [X] Integration tests pass
- [X] Manual test: qualification function passes
- [x] Manual test: qualification function from uploaded asset passes

### DIFF
--- a/bin/clover/src/pipelines/generic/overrides.ts
+++ b/bin/clover/src/pipelines/generic/overrides.ts
@@ -286,6 +286,27 @@ export function widget(
   };
 }
 
+export function addQualificationFunction(
+  funcPath: string,
+  name: string,
+  uniqueId: string,
+): SchemaOverrideFn {
+  return (spec) => {
+    const variant = spec.schemas[0].variants[0];
+    if (!variant.domain.uniqueId) {
+      throw new Error("Expected domain.uniqueId");
+    }
+    const { func, leafFuncSpec } = attachQualificationFunction(
+      funcPath,
+      name,
+      uniqueId,
+      variant.domain.uniqueId,
+    );
+
+    spec.funcs.push(func);
+    variant.leafFunctions.push(leafFuncSpec);
+  };
+}
 export function addScalarProp(
   propPath: PropPath,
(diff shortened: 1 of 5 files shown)


bin/clover/src/pipelines/generic/overrides.ts
```diff
@@ -282,3 +285,80 @@ export function widget(
     }
   };
 }
+
+export function addQualificationFunction(
```
sprutton1, line 289: I wonder if we should hoist this up (and maybe path walking bits for adding props) into props.ts and funcs.ts and just use them all over. It's definitely a cleaner interface.

bin/clover/src/pipelines/hetzner/overrides.ts
```diff
@@ -19,20 +23,27 @@ export const HETZNER_PROP_OVERRIDES: Record<
...
+  Object.entries({
+    //
+    // Add Hetzner::Cloud::Locations.name so it can be selected and filled in
+    //
+    "Hetzner::Cloud::Locations": [
+      addScalarProp("/domain/name", "string", true),
+      addQualificationFunction(
+        "./src/pipelines/hetzner/funcs/overrides/Hetzner::Cloud::Locations/qualifications/locationQualification.ts",
+        "Hetzner::Cloud::Locations Check Valid Location Name",
+        undefined, // TODO MAKE AN ACTUAL ID
```
sprutton1: This just needs to be a sha256 of sufficient uniqueness that it won't clash with any other func ids. We do this for the intrinsic and all clover funcs so we can always keep track of function changes.

jkeiser: In the future, maybe we should consider having addQualificationFunction (and friends) hash the actual contents of the file? Or does it need to remain stable even if you change the contents of the function?

jkeiser:
There shouldn't be changes to AWS assets--I'll check main to make sure I haven't regressed something there.

PR #7499, opened 2025-10-10
Persist Deployment MVs in WASM SQLite

jobelenus:
# Storing Global Data
## MVs not associated with a workspace and change set

We make a new table, `global_atoms`, with the same columns as the `atoms` table. But the PK & constraint is slightly different. We aren't allowing the storage of multiple `checksum` with the same `args` and `kind`.

There is no MTM table, since this is one for the world.

The patching operations follow the same pattern. For modify patches, get the doc, apply, persist. Removals and creations are straightforward. Use the INSERT ON CONFLICT sql approach to push everything into a single statement.

### First scenario:
1. Cold start per normal
2. See hammers thrown for all the deployment MVs
3. Refresh the page
4. See zero hammers—the data has been stored

**Note**: 3k takes roughly 4 seconds

### Second scenario:
1. Delete the deployment MVs from SQLite
2. Empty Frigg of its deployment MVs and index
3. Cold start
4. The index isn't ready yet
6. The deployment MV data comes over the wire as a patch
7. Refresh the page
8. Cold start happens
9. See zero hammers—the data has been stored from the patches

**Note:** this is ~3k patches, happens really fast. So when we bring on new schemas for providers—it will be fast!

### Third scenario:
This is a slightly different code path, since it pulls the existing docs out of the DB and will the most experienced code path for users.
1. Covered in unit tests, patching an individual deployment MV.

### DIFF
--- a/app/web/src/workers/webworker.ts
+++ b/app/web/src/workers/webworker.ts
@@ -70,6 +70,8 @@ import {
   DeploymentPatchBatch,
   ConnStatusFn,
   StoredMvIndex,
+  AtomWithData,
+  AtomOperation,
 } from "./types/dbinterface";
 import {
   BifrostComponent,
@@ -89,6 +91,8 @@ import {
   DefaultSubscriptions,
   DefaultSubscription,
   AttributeValue,
+  GLOBAL_ENTITIES,
+  GlobalEntity,
 } from "./types/entity_kind_types";
 import {
   bulkDone,
@@ -199,6 +203,7 @@ const dropTables = () => {
   const sql = `
   DROP TABLE IF EXISTS index_mtm_atoms;
   DROP TABLE IF EXISTS atoms;
+  DROP TABLE IF EXISTS global_atoms;
   DROP TABLE IF EXISTS indexes;
   DROP TABLE IF EXISTS changesets;
   DROP TABLE IF EXISTS weak_references;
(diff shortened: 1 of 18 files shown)


app/web/src/workers/webworker.ts
```diff
@@ -2179,6 +2197,174 @@ const retry = async <T>(
   } else return r;
 };
 
+const vanaheim = async (
```
zacharyhamm, line 2420: maybe a comment describing what vanaheim does :)

lib/frigg/src/deployment.rs
```diff
@@ -91,7 +87,7 @@ impl FriggStore {
     ))]
     pub async fn get_current_deployment_object(
         &self,
-        kind: ReferenceKind,
+        kind: &str,
```
zacharyhamm, line 90: Why was this change necessary?

jobelenus: Yes!, so I'm following the pattern of the _other_ `impl Frigg` where `kind` is a `str`. The HTTP request accepts a `FrontendObject` which has `kind` as a `str`.

jobelenus:
> After rebasing this on main, `luminork` fails to build:

Yep! I've grabbed main, rebased, and fixed this up

PR #7504, opened 2025-10-10
chore: stop using `#[error(transparent)]` in Error variants

fnichol:
This change fixes an old mistake (made by this author) where the use of `transparent` was mis-understood. The `transparent` usage wouldn't add an error message and makes it harder to track error chains.


fnichol:
> you beat me to it! 😄

Well.... Claude beat you to it!

PR #7509, opened 2025-10-13
feat(luminork): Rebuild MV when unlocking schema

wendybujalski:

Does as it says in the title. Fixes bugs in the `schemaCreateEditGet` tool. Thank you Paul!


lib/luminork-server/src/service/v1/schemas/unlock_schema.rs
```diff
@@ -99,6 +100,13 @@ pub async fn unlock_schema(
 
     ctx.commit().await?;
 
+    let _luminork_variant =
+        slow_rt::spawn(dal_materialized_views::luminork::schema::variant::assemble(
```
zacharyhamm, line 112: doesn't this just generate the MV and then throw it away?

PR #7513, opened 2025-10-14
Add persistent retry queue for transient PostgreSQL failures

jhelwig:
## Why This Change Is Needed

The layer cache architecture assumes PostgreSQL writes will succeed immediately. When a write is requested, the data becomes immediately available in the in-memory cache before the PostgreSQL write completes. This design works well for performance, but creates a critical problem: any PostgreSQL write failure (network blips, connection pool exhaustion, brief database unavailability) results in permanent data loss despite the data being successfully cached in memory.

### Why Disk-Backed Instead of In-Memory Queue or NATS

We chose disk persistence over an in-memory queue and NATS because:

1. **Survive Process Restarts**: If the application crashes or is restarted during a PostgreSQL outage, in-memory queues would be lost along with all pending writes
2. **Bounded Memory Usage**: During extended PostgreSQL outages, an in-memory queue would grow indefinitely and potentially cause OOM conditions
3. **No External Dependencies**: If the PostgreSQL problems are because of networking issues, there is a significant chance that NATS is also going to have similar problems.

The tradeoff is disk I/O overhead on enqueue, but since this only happens during failures (not the happy path), the performance impact is acceptable.

### Why Exponential Backoff (100ms → 5s)

The backoff parameters balance several concerns:

- **Initial 100ms**: Quick recovery from momentary glitches (single dropped packet, brief connection reset)
- **2x Multiplier**: Rapid but not aggressive growth - reaches 5s max in 6 attempts
- **5s Maximum**: Prevents excessive delays while still being gentle on recovering databases
- **Per-Queue Reset**: Success in one cache type resets only that cache's backoff, allowing fast recovery without thundering herd

### Why Biased tokio::select! Priority

The PersisterTask event loop prioritizes: shutdown > new writes > retries. This
ordering ensures:

1. **Clean Shutdown**: Always respected, even during heavy retry load
2. **Fresh Data Priority**: New writes aren't blocked by retry backlog (prevents write amplification during recovery)
4. **Retry Progress**: Retries still make progress when new writes are idle

### Why ULID Filenames

Using ULID for pending file names provides:

1. **Chronological Ordering**: Retries are attempted in the order writes failed (BTreeSet iteration over filenames)
2. **Uniqueness**: No filename collisions even with high write rates

## Architecture Overview

**RetryQueueManager** manages all retry queues as a single component owned by PersisterTask. When a write fails with a retryable error (PostgreSQL connection/network errors), the event is serialized to disk using postcard in `{cache_disk_path}/{cache_name}_retries/{ulid}.pending`. Each cache's queue maintains independent backoff state that increases on failure and resets on success.

On startup, existing retry queues are scanned and set to retry immediately, ensuring queued writes survive process restarts.

The PersisterTask event loop processes ready retries alongside new writes, attempting the PostgreSQL write again. On success, the pending file is deleted. On failure, backoff is updated for the next attempt.

### Disk Space Exhaustion

If PostgreSQL stays down for extended periods while writes continue, retry queues will grow until disk space is exhausted. Operators should monitor `layer_cache_retry_queue_depth` metrics and investigate if queues grow beyond expected thresholds.

### Startup Performance

On process startup, all retry queue directories are scanned to load pending files. With thousands of pending retries, startup could be delayed. The implementation loads filenames only (not deserializing events), minimizing overhead, but very large queues may still cause noticeable startup delays.

### Non-Retryable Errors

Not all PostgreSQL errors are transient. Schema mismatches, constraint violations, or serialization failures are permanent. These are logged as errors and removed from the queue to prevent infinite retry loops. However, this means such writes are permanently lost - operators must monitor `layer_cache_retry_queue_permanent_failure` metrics to detect systemic problems.

## Testing Approach

Unit tests verify the retry queue mechanics in isolation (enqueue, retrieve, backoff, scanning). Integration tests verify the retry queue is transparent to existing functionality - all existing tests pass without modification, confirming backward compatibility.

## Metrics! 📈 📉 🎉 

These are from my local testing where I applied the following patch to simulate a 20% (retryable) failure rate for all `LayerDb` PG write & delete operations.

```diff
diff --git a/lib/si-layer-cache/src/pg.rs b/lib/si-layer-cache/src/pg.rs
index d59888d3d..31cd1cae6 100644
--- a/lib/si-layer-cache/src/pg.rs
+++ b/lib/si-layer-cache/src/pg.rs
@@ -3,6 +3,7 @@ use std::{
     sync::Arc,
 };

+use rand::Rng;
 use si_data_pg::{
     PgPool,
     PgPoolConfig,
@@ -205,6 +206,16 @@ impl PgLayer {
         sort_key: impl AsRef<str>,
         value: &[u8],
     ) -> LayerDbResult<()> {
+        // LOCAL TESTING ONLY: Simulate 20% failure rate for retry queue testing
+        if rand::thread_rng().gen_range(0.0..1.0) < 0.20 {
+            return Err(crate::error::LayerDbError::PgPool(
+                si_data_pg::PgPoolError::CreateCertificate(std::io::Error::new(
+                    std::io::ErrorKind::ConnectionAborted,
+                    "simulated connection failure for retry queue testing",
+                )),
+            ));
+        }
+
         let client = self.pool.get().await?;
         let sort_key = sort_key.as_ref();
         client
@@ -224,6 +235,16 @@ impl PgLayer {
     }

     pub async fn delete(&self, key: &str) -> LayerDbResult<()> {
+        // LOCAL TESTING ONLY: Simulate 20% failure rate for retry queue testing
+        if rand::thread_rng().gen_range(0.0..1.0) < 0.20 {
+            return Err(crate::error::LayerDbError::PgPool(
+                si_data_pg::PgPoolError::CreateCertificate(std::io::Error::new(
+                    std::io::ErrorKind::ConnectionAborted,
+                    "simulated connection failure for retry queue testing",
+                )),
+            ));
+        }
+
         let client = self.pool.get().await?;
         client.query(&self.delete_query, &[&key]).await?;
         Ok(())
```

image

### DIFF
--- a/dev/docker-compose.platform.yml
+++ b/dev/docker-compose.platform.yml
@@ -49,6 +49,7 @@ services:
       - "3000:3000"
     volumes:
       - ./config/grafana/provisioning/datasources:/etc/grafana/provisioning/datasources
+      - ./config/grafana/provisioning/dashboards:/etc/grafana/provisioning/dashboards
 
   jaeger:
     #<<: *loki-logged-service
@@ -156,6 +157,8 @@ services:
   prometheus:
     #<<: *loki-logged-service
     image: prom/prometheus
+    ports:
+      - "9090:9090"
     volumes:
       - "./config/prometheus/config.yml:/etc/prometheus/prometheus.yml"
     depends_on:
--- a/dev/config/grafana/provisioning/dashboards/dashboards.yml
+++ b/dev/config/grafana/provisioning/dashboards/dashboards.yml
@@ -1,0 +1,12 @@
+apiVersion: 1
+
+providers:
+  - name: "SI Service Metrics"
+    orgId: 1
+    folder: ""
+    type: file
+    disableDeletion: false
+    updateIntervalSeconds: 10
+    allowUiUpdates: true
+    options:
+      path: /etc/grafana/provisioning/dashboards
--- a/dev/config/grafana/provisioning/dashboards/layer-cache-metrics.json
+++ b/dev/config/grafana/provisioning/dashboards/layer-cache-metrics.json
@@ -1,0 +1,1265 @@
+{
+  "id": null,
+  "uid": "layer-cache-metrics",
+  "title": "LayerCache Metrics",
+  "tags": [
+    "layercache",
+    "cache",
+    "postgres",
+    "retry",
+    "persistence"
+  ],
+  "timezone": "browser",
+  "schemaVersion": 16,
+  "version": 0,
+  "refresh": "5s",
+  "annotations": {
+    "list": [
+      {
+        "builtIn": 1,
+        "datasource": {
+          "type": "grafana",
+          "uid": "-- Grafana --"
+        },
+        "enable": true,
+        "hide": true,
+        "iconColor": "rgba(0, 211, 255, 1)",
+        "name": "Annotations & Alerts",
+        "type": "dashboard"
+      }
+    ]
+  },
+  "templating": {
+    "list": [
+      {
+        "name": "service",
+        "type": "query",
+        "label": "Service",
+        "datasource": {
+          "type": "prometheus",
+          "uid": "PBFA97CFB590B2093"
+        },
+        "query": "label_values(layer_cache_persister_write_attempted, exported_job)",
+        "refresh": 2,
+        "sort": 1,
+        "multi": true,
+        "includeAll": true,
+        "allValue": ".*"
+      },
+      {
+        "name": "cache",
+        "type": "query",
+        "label": "Cache",
+        "datasource": {
+          "type": "prometheus",
+          "uid": "PBFA97CFB590B2093"
+        },
+        "query": "label_values(layer_cache_persister_write_attempted{exported_job=~\"$service\"}, cache_name)",
+        "refresh": 2,
+        "sort": 1,
+        "multi": true,
+        "includeAll": true,
+        "allValue": ".*"
+      }
+    ]
+  },
+  "panels": [
+    {
+      "id": 2,
+      "gridPos": {
+        "x": 0,
+        "y": 0,
+        "w": 12,
+        "h": 8
+      },
+      "type": "timeseries",
+      "title": "Retry Queue - Enqueue Rate",
+      "datasource": {
+        "type": "prometheus",
+        "uid": "PBFA97CFB590B2093"
+      },
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "PBFA97CFB590B2093"
+          },
+          "expr": "rate(layer_cache_retry_queue_enqueued{exported_job=~\"$service\", cache_name=~\"$cache\"}[1m])",
+          "legendFormat": "{{exported_job}}/{{cache_name}} - enqueued",
+          "refId": "A"
+        }
+      ],
+      "fieldConfig": {
+        "defaults": {
+          "unit": "ops",
+          "color": {
+            "mode": "palette-classic"
+          },
+          "custom": {
+            "lineWidth": 2,
+            "fillOpacity": 10,
+            "showPoints": "never"
+          }
+        }
+      },
+      "options": {
+        "tooltip": {
+          "mode": "multi",
+          "sort": "desc"
+        },
+        "legend": {
+          "displayMode": "table",
+          "placement": "right",
+          "calcs": [
+            "lastNotNull",
+            "max",
+            "mean"
+          ]
+        }
+      }
+    },
+    {
+      "id": 3,
+      "gridPos": {
+        "x": 12,
+        "y": 0,
+        "w": 12,
+        "h": 8
+      },
+      "type": "timeseries",
+      "title": "Retry Queue - Success Rate",
+      "datasource": {
+        "type": "prometheus",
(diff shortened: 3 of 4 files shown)


dev/docker-compose.platform.yml
```diff
@@ -156,6 +157,8 @@ services:
   prometheus:
     #<<: *loki-logged-service
     image: prom/prometheus
+    ports:
+      - "9090:9090"
```
jhelwig, line 161: Being able to query Prometheus directly made creating/debugging the Grafana dashboard much easier.

lib/si-layer-cache/src/persister.rs
```diff
@@ -146,12 +159,46 @@ impl PersisterTask {
...
+            "func_run_log",
+            "rebase_batch",
+            "workspace_snapshot",
+            "split_snapshot_subgraph",
+            "split_snapshot_supergraph",
+            "split_snapshot_rebase_batch",
+        ];
+        retry_queue_manager
+            .scan_existing_queues(CACHE_NAMES)
+            .await?;
```
jhelwig, line 203: I have some work in progress to make this not hard-coded, but it requires threading more information through in a few places, and it seemed more important to not delay this further.

lib/si-layer-cache/src/persister.rs
```diff
@@ -160,19 +221,33 @@ impl PersisterTask {
...
+                }
+
+                // Priority 2: New messages from channel
+                Some(msg) = self.messages.recv() => {
+                    self.spawn_persist_task(msg);
+                }
+
+                // Priority 3: Ready retries from RetryQueueManager
+                Some((event, handle)) = self.ready_rx.recv() => {
+                    self.spawn_retry_task(event, handle);
```
zacharyhamm, line 245: Is there any risk that a massive queue of persist tasks will prevent retry tasks from running?

lib/si-layer-cache/src/persister.rs
```diff
@@ -146,12 +173,46 @@ impl PersisterTask {
...
+        tracker.spawn(retry_queue_manager.run(retry_rx, ready_tx, shutdown_token.clone()));
+
         Ok(Self {
             messages,
             pg_pool,
             layered_event_client,
             tracker,
             shutdown_token,
+            retry_tx,
+            ready_rx,
```
zacharyhamm: The names of these queues are a little confusing, especially what the ready_tx/rx is vs the retry_tx/rx

lib/si-layer-cache/src/retry_queue.rs
```diff
@@ -0,0 +1,609 @@
...
+    pub(crate) filename: OsString,
+}
+
+/// Calculate the next backoff duration using exponential backoff with a maximum cap
+fn calculate_next_backoff(current: Duration, max: Duration, multiplier: f64) -> Duration {
+    std::cmp::min(
+        Duration::from_secs_f64(current.as_secs_f64() * multiplier),
+        max,
+    )
+}
```
zacharyhamm: Not a deal breaker, but this https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/ has me convinced backoff should always be "jittered" (e.g: https://github.com/systeminit/si/blob/f2d11382d0d9fa73b9efbe75d9e0c5e600ae8e81/lib/dal/src/job/consumer.rs#L293)

lib/si-layer-cache/src/retry_queue.rs
```diff
@@ -0,0 +1,609 @@
...
+async fn write_retry_file(path: &std::path::Path, event: &LayeredEvent) -> LayerDbResult<()> {
+    use tokio::fs;
+
+    let bytes = postcard::to_stdvec(event)?;
+
+    // Write to temp file first, then atomic rename
+    let temp_path = path.with_extension("tmp");
+    fs::write(&temp_path, bytes)
+        .await
+        .map_err(LayerDbError::RetryQueueFileWrite)?;
```
zacharyhamm: My only thought here is that we might want to call an `fsync` equivalent to prevent the loss of a retry queue file if the system fails before the page buffers have been flushed to disk. (there's `sync_all`, but you need a `File` handle for that). But that might be extra paranoid (and will impact performance)

zacharyhamm (commented):
This looks solid to me! I do worry a bit about filling up the disks since it's competing with foyer disk usage, but that's a failure case. I'm just going to build it with the random failures from your readme and see how it goes before approving.

A few comments in here, but nothing blocking.

zacharyhamm (approved):
This is great work. We should verify with Scott or John about how much headroom we have on the disks, but since that's only relevant if we have a ton of writes that fail, I'm not tremendously worried

lib/si-layer-cache/src/persister.rs
jhelwig: Some, but it does mean that any new data being generated by the system has priority for being persisted, and the newest data should be the most likely to be needed, so that seems right to me.

Your comment does make me think that maybe the retry queue should be FILO, instead of FIFO, but doing that means that there's a chance that some data would never be persisted (assuming the right combination of the rates of new data and retryable failures), instead of "just" significantly delayed in when it is persisted.

jhelwig:
Great comments, thank you! Working on changes for the things you mentioned aside from the select biasing, and left a comment on that one.

PR #7517, opened 2025-10-14
feat(luminork): default to overlay actions and mgmt funcs for builtins

zacharyhamm:
When an action or management function is created in luminork, default to creating an "overlay" (schema parented) function and prototype instead of a variant-level function. This will allow the custom functions to persist between upgrades, and does not require locking the variant.

In order to make this work with our materialized views, adds a post-processing step to change detection that adds SchemaVariant changes to the change set to force the SchemaVariant related materialized views to rebuild so that they reflect these new functions.

Tests:

1. Create a new management function for a builtin via luminork. Observe that it is available in the management functions dropdown in the UI. Try to run it.

2. See that the function exists in the list of functions when you fetch the schema variant in luminork.

3. Create an action for a builtin via luminork. Confirm you cannot create a second action of that type. Add a component and enqueue that action. Apply to head. See that the action that ran was the new one, and not the default one for that variant.

4. See that the action function exists in the func list when you fetch the variant via luminork.

Additional tests:

  Use luminork to create a new create action for one change set, and a
  new create action in another change set, both for the same schema.
  Apply one of the change sets to head. Ensure that create actions work
  when applying the second change set after the first has been applied.

### DIFF
--- a/lib/rebaser-server/src/rebase.rs
+++ b/lib/rebaser-server/src/rebase.rs
@@ -3,6 +3,7 @@ use std::time::Duration;
 use audit_logs_stream::AuditLogsStreamError;
 use dal::{
     DalContext,
+    SchemaError,
     TransactionsError,
     Workspace,
     WorkspaceError,
@@ -446,6 +447,8 @@ async fn rebase_legacy(
 pub(crate) enum SendEddaUpdatesError {
     #[error("edda client error: {0}")]
     EddaClient(#[from] edda_client::ClientError),
+    #[error("schema error: {0}")]
+    Schema(#[from] SchemaError),
     #[error("transactions error: {0}")]
     Transactions(#[from] TransactionsError),
     #[error("workspace snapshot error: {0}")]
@@ -467,6 +470,7 @@ pub(crate) async fn send_updates_to_edda_split_snapshot(
             "rebaser.perform_rebase.detect_changes_for_edda_request"
         ))
         .await?;
+
     let change_batch_address = ctx.write_change_batch(changes).await?;
     let edda_update_request_id = edda
         .update_from_workspace_snapshot(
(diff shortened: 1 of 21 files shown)


zacharyhamm:
Corrections are required before this can merge. Working on them now.

zacharyhamm:
Corrections in with a test. Also going to click test it

lib/rebaser-server/src/rebase.rs
```diff
@@ -510,6 +525,63 @@ pub(crate) async fn send_updates_to_edda_legacy_snapshot(
     Ok(())
 }
 
+/// In order to have correct materialized views, sometimes a change in one thing
+/// means we should report a change in some other thing, even though that thing
+/// has not actually changed. For example, if an overlay function has been added
+/// to a schema, we need to recalculate the materialized views for the schema
+/// variants under that schema.
+#[instrument(level = "info", name = "post_process_changes", skip_all)]
+async fn post_process_changes(
```
zacharyhamm: @jhelwig this was my dead-stupid solution to the schema -> variant rebuild issue

jhelwig: Curious about the decision to put this in the rebaser, instead of in edda.

zacharyhamm: hmm good question. I'll look at moving it to edda, since that will mean it doesn't have an impact on the rebaser round trip

zacharyhamm: Ok i've updated this to do the post processing in edda

jhelwig (approved):

lib/rebaser-server/src/rebase.rs
zacharyhamm: Sorry for the back & forth. I moved it to edda but did not remove it from rebaser. Now it's correct.

PR #7519, opened 2025-10-14
chore(si-data-pg): Add error events to narrow down connection closed errors

britmyerss:
We’re still seeing transient `Connection closed` errors across services and across dbs but we haven’t narrowed down where/why/when. This adds error events across a number of function calls to help us narrow down where/when/why it’s happening.

<div><br/>via <a href="https://giphy.com/agt/">America&#x27;s Got Talent</a> on <a href="https://giphy.com/gifs/agt-nbc-americas-got-talent-IgYHnGwwDHPVwXMakx">GIPHY</a></div>

### DIFF
--- a/lib/si-data-pg/src/lib.rs
+++ b/lib/si-data-pg/src/lib.rs
@@ -474,7 +474,9 @@ impl PgPool {
         span.record("db.pool.size", pool_status.size);
         span.record("db.pool.available", pool_status.available);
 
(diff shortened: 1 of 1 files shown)


fnichol (dismissed):
This looks good!

I'm wondering if you might consider adding the error lines to `pg_transaction.query` and `pg_transaction.query_one`? Unless you're tracking some call sites that aren't using a transaction in which case we're all good.

PR #7522, opened 2025-10-15
componentCreate using working copy

wendybujalski:

This PR adjusts the `componentCreate` tool to allow for users to create components either using the default variant or their working copy of the component's schema.

#### Out of Scope:

This PR deliberately does not allow the `createComponent` tool to select a schema variant by ID, since users should not have to think about schema variants at all. Instead, the AI Agent can select either the default variant or the working copy.

## How was it tested?

The following manual tests were run via the AI Agent -

- The demo listed in [ENG-3260](https://linear.app/system-initiative/issue/ENG-3260/spike-on-issues-between-mcp-modeling-and-authoring-tools-specifically) done multiple times with wide variations on wording. All successful!
- Alternating adding a prop and function to a schema, creating a component after each edit - successful!
- Creating a brand new schema and iterating on props and functions
- Editing an existing schema and iterating on props and functions

### DIFF
--- a/bin/si-mcp-server/src/data/components.ts
+++ b/bin/si-mcp-server/src/data/components.ts
@@ -29,5 +29,5 @@ export const AttributesSchema = z.record(
     z.number().describe("a number value"),
   ]),
 ).describe(
-  "the attributes of the component; the desired state of the resource",
+  "the attributes of the component; the desired state of the resource. do not start attributes with /root",
 );
--- a/bin/si-mcp-server/src/tools/componentCreate.ts
+++ b/bin/si-mcp-server/src/tools/componentCreate.ts
@@ -23,6 +23,7 @@ const CreateComponentInputSchemaRaw = {
   schemaName: z.string().describe("the schema name of the component to create"),
   componentName: z.string().describe("the name of the component to create"),
   attributes: AttributesSchema,
+  useWorkingCopy: z.boolean().optional().describe("Set this boolean to true in order to create a component using the current working copy of the schema. If the user has been editing the schema of this component, use the working copy. Do not mention schema variants or locked/unlocked to the user. Instead, refer to the unlocked schema variant as the current working copy of the schema."),
 };
 
 const CreateComponentOutputSchemaRaw = {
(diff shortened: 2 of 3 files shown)


bin/si-mcp-server/src/tools/componentCreate.ts
```diff
@@ -34,6 +35,19 @@ const CreateComponentOutputSchemaRaw = {
     componentId: z.string().describe("the component id"),
     componentName: z.string().describe("the components name"),
     schemaName: z.string().describe("the schema for the component"),
+    schemaVariant: z.object({
```
stack72: I don't actually think we need to output any of this schema info - it's not of any use for a user

bin/si-mcp-server/src/data/components.ts
```diff
@@ -29,5 +29,5 @@ export const AttributesSchema = z.record(
     z.number().describe("a number value"),
   ]),
 ).describe(
-  "the attributes of the component; the desired state of the resource",
+  "the attributes of the component; the desired state of the resource. do not start attributes with /root",
```
stack72, line 32: weird you had to write this but 🤷

bin/si-mcp-server/src/tools/componentCreate.ts
```diff
@@ -23,6 +23,7 @@ const CreateComponentInputSchemaRaw = {
   schemaName: z.string().describe("the schema name of the component to create"),
   componentName: z.string().describe("the name of the component to create"),
   attributes: AttributesSchema,
+  useWorkingCopy: z.boolean().optional().describe("Set this boolean to true in order to create a component using the current unlocked schema variant. If the user has been editing the schema of this component, use the working copy. Do not mention schema variants or locked/unlocked schemas to the user. Instead, refer to the unlocked variant as the current working copy of the schema."),
```
stack72: you talk about unlocked / locked, editing and working copy :D I think making it easier to read is better

bin/si-mcp-server/src/data/components.ts
wendybujalski: Yeah, this tiny addition of context fixed this annoying issue that was causing about half of all createComponent tool calls to fail. Idk why it kept wanting to put "/root" on the front of every attribute but this made it stop!

bin/si-mcp-server/src/tools/componentCreate.ts
wendybujalski: Adjusted the language just a little bit.

PR #7527, opened 2025-10-16
feat(luminork): Add ability to detach funcs from variants

stack72:


lib/luminork-server/src/service/v1/schemas/detach_action_binding.rs
```diff
@@ -0,0 +1,109 @@
...
+        schema_variant_id,
+        func_id,
+    }): Path<SchemaVariantFuncV1RequestPath>,
+) -> SchemaResult<Json<DetachFuncBindingV1Response>> {
+    if ctx.change_set_id() == ctx.get_workspace_default_change_set_id().await? {
+        return Err(SchemaError::NotPermittedOnHead);
+    }
+
+    let schema_variant = SchemaVariant::get_by_id(ctx, schema_variant_id).await?;
+    if schema_variant.is_locked() {
```
britmyerss: We do this same check in each of the `delete_*_binding` calls, so no need to do it here too. You'll probably want to convert that error kind into this one to get the response shape correct.

PR #7528, opened 2025-10-16
feat(si-conduit): Implement v0.1

vbustamante:

Adds si-conduit, the executable that allows to push and (in the future) pull schema definitions between your workspace and your local file system. Usage is defined in the [README](https://github.com/systeminit/si/blob/victor/si-conduit/bin/si-conduit/README.md) file for the project.

#### Screenshots:

image 

#### Out of Scope:

Right now it can only create the scaffold of an asset, with an example for each kind of func (excluding attribute and auth funcs), and push the assets on a folder to a single workspace. It will fail if the asset already exists.

## How was it tested?

- [X] Build it with ts checking
- [X] Ran scaffold and export-assets in dev mode
- [X] Ran scaffold and export-assets with a compiled executable


bin/si-conduit/deno.json
```diff
@@ -0,0 +1,12 @@
+{
```
stack72, line 1: You will need this: "nodeModulesDir": "auto",

stack72, line 1: You will also need "workspace": [],

bin/si-conduit/src/analytics.ts
```diff
@@ -0,0 +1,83 @@
...
+      distinctId: this.userId,
+      properties: {
+        userId: this.userId,
+      },
+    });
+  }
+
+  private initializePostHog(): PostHog {
+    const apiKey = Deno.env.get("POSTHOG_API_KEY") ||
+      "phc_KpehlXOqtU44B2MeW6WjqR09NxRJCYEiUReA58QcAYK"; // Prod Posthog
```
stack72, line 33: We should post that this is a publicly used API Key or we will get reports from users

fnichol:
We need to double check the Buck2 (local/remote) and the remote build instructions in the package's readme. Then we'll be good.

bin/si-conduit/deno.json
fnichol: Yeah, we'll add, good call for the time being

bin/si-conduit/src/analytics.ts
fnichol: Good plan

PR #7529, opened 2025-10-16
Using deployment mvs for "add component" list

jobelenus:

Now that we have `DefaultSchemaVariant` MVs we do not need the `SchemaVariantCategories` MV, we can just use all the `SchemaVariant` MVs we have.

We can delete the ch0nky `SchemaVariantCategories` MV entirely!

## How was it tested?
NOTE: my [bracket] additions there are just a sanity check to make sure it wasn't showing dupes and showed installed over defaults... that won't be in the finished product.

image

1. Install a "default" schema, make sure it creates a component
2. Open the list, see that one you installed be labelled as "installed"
3. Install a "installed" schema, make sure it creates a component
4. Unlock an installed schema, make sure you see installed and editing

Less ch0nky!!!

### DIFF
--- a/app/web/src/newhotness/AddComponentModal.vue
+++ b/app/web/src/newhotness/AddComponentModal.vue
@@ -235,16 +235,19 @@ import { debounce } from "lodash-es";
 import { FzfResultItem } from "fzf";
 import EditingPill from "@/components/EditingPill.vue";
 import {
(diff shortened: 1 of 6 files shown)


stack72 (approved):
good on my side! The web worker parts are where I'd be less confident in reviewing

PR #7530, opened 2025-10-16
feat(clover): add azure mgmt funcs

sprutton1:

Adds import and discover funcs for Azure. I believe that the resource_value/domains are currently incorrect, but we are at least populating what we have here correctly.

Not that we needed to add back normalizing the only props to ensure we could build the PropUsageMap correctly.

This includes some flashy stuff like walking pagination and refinements, lifted without shame from AWS.

#### Screenshots:

image

One discovered, one imported.

image

## How was it tested?

Checkout them screenshots. This will not generate any changes for other providers.

- [X] Integration tests pass

### DIFF
--- a/bin/clover/src/pipelines/azure/funcs/management/discover.ts
+++ b/bin/clover/src/pipelines/azure/funcs/management/discover.ts
@@ -1,0 +1,264 @@
+async function main({
+  thisComponent,
+}: Input): Promise<Output> {
+  const component = thisComponent;
+  const tenantId = requestStorage.getEnv("AZURE_TENANT_ID");
+  const clientId = requestStorage.getEnv("AZURE_CLIENT_ID");
+  const clientSecret = requestStorage.getEnv("AZURE_CLIENT_SECRET");
+
+  if (!tenantId || !clientId || !clientSecret) {
+    throw new Error("Azure credentials not found");
+  }
+
+  const subscriptionId = _.get(
+    component.properties,
+    ["domain", "subscriptionId"],
+    "",
+  );
+  const resourceGroup = _.get(
+    component.properties,
+    ["domain", "resourceGroup"],
+    "",
+  );
+  const resourceType = _.get(
+    component.properties,
+    ["domain", "extra", "AzureResourceType"],
+    "",
+  );
+  const apiVersion = _.get(
+    component.properties,
+    ["domain", "extra", "apiVersion"],
+    "2023-01-01",
+  );
+  const propUsageMapJson = _.get(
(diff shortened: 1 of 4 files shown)


bin/clover/src/pipelines/azure/funcs/management/discover.ts
```diff
@@ -0,0 +1,273 @@
...
+  }
+
+  if (!resourceType) {
+    return {
+      status: "error",
+      message: "AzureResourceType not found in domain.extra",
+    };
+  }
+
+  // Convert Azure::Service::Resource to Microsoft.Service/resources format
```
sprutton1, line 53: pretty big smell that we should change to use the `Microsoft.Service/resource` pattern

bin/clover/src/pipelines/azure/funcs/management/discover.ts
```diff
@@ -0,0 +1,273 @@
...
+        value,
+        (v) => !_.isEmpty(v) || _.isNumber(v) || _.isBoolean(v),
+      );
+      if (_.isEmpty(refinement[key])) {
+        delete refinement[key];
+      }
+    }
+  }
+
+  // Build the list URL based on whether resourceGroup is specified
```
sprutton1: Maybe it would make more sense to just store this on the asset in the extra prop?

bin/clover/src/pipelines/azure/funcs/management/discover.ts
jkeiser: Yeah, I'm pretty certain we'll want to do that!

bin/clover/src/pipelines/azure/funcs/management/discover.ts
```diff
@@ -0,0 +1,266 @@
...
+
+    // Build domain by only including writable properties from the resource
+    const domainProperties: Record<string, any> = {
+      subscriptionId,
+      resourceGroup,
+    };
+
+    // Copy writable properties from the resource that match the schema
+    for (const [key, value] of Object.entries(fullResource)) {
+      if (key === "id" || key === "type") continue;
```
jkeiser: Are id and type writeable? Why wouldn't we set them?

jkeiser (approved):
Just one comment

bin/clover/src/pipelines/azure/funcs/management/discover.ts
sprutton1: I removed this. I was parsing the id and type manually for a bit, but it's no longer necessary

PR #7532, opened 2025-10-16
Azure: generate network resources with circular types

jkeiser:
We're not generating the loadBalancers schema for Azure right now, among several other critical network ones, due to some issues with circular types.

This PR generates many more valid Azure schemas, and removes some non-valid ones, by changing some rules:

* Process circular schemas (network schemas in particular have a lot of this, including loadBalancers)
* Heavily restrict the endpoints we process to avoid bringing in non-resources (e.g. `".../providers/Microsoft.Network/loadBalancers/{loadBalancerName}"`)
* Don't use "list" endpoints as a source for "read" (they almost always have arrays, not objects)
* Don't process readonly schemas (if it only has GET it's probably a query, search or list)

This PR also brings in real OpenAPI types, so we can stop maintaining them. It does *not* fix up Hetzner or AWS's use of these types.

DRAFT: this could affect Hetzner and AWS and I want to make sure it doesn't change any schemas before checking in

#### Out of Scope:

* loadBalancers are now generating, but are still too big (>4MB). That's next.
* `x-ms-client-flatten` support isn't in, so we are generating a ton of unnecessary indirect properties named `properties`. This is OK, as `properties` is in fact part of the schema, but we may want to support this as users will not be expecting to see it (no other generated Azure clients force you to pass a properties object). 

## How was it tested?

- [X] Saw more Azure resources!!!

### DIFF
--- a/bin/clover/src/pipelines/azure/spec.ts
+++ b/bin/clover/src/pipelines/azure/spec.ts
@@ -7,10 +7,10 @@ import {
   AzureSchema,
   PropertySet,
   isAzureObjectProperty,
-  AZURE_HTTP_METHODS,
-  AzureOpenApiOperation,
   AzureProperty,
-  AzureOpenApiSpec,
+  AzureOpenApiDocument,
+  AzureOpenApiOperation,
+  AZURE_HTTP_METHODS,
 } from "./schema.ts";
 import { JSONSchema } from "../draft_07.ts";
 import { ExpandedPkgSpec } from "../../spec/pkgs.ts";
@@ -26,9 +26,7 @@ export function extractPropertiesFromRequestBody(
   assert(bodyParams.length <= 1, "Expected at most one body parameter");
 
   const { schema } = bodyParams[0];
-  if (!isAzureObjectProperty(schema))
-    console.log(`schema: ${JSON.stringify(schema, null, 2)}`);
-  assert(isAzureObjectProperty(schema), "Expected body schema to be an object");
+  assert(isAzureObjectProperty(schema), "Body schema is not an object");
 
   return {
     properties: schema.properties ?? {},
@@ -39,8 +37,12 @@ export function extractPropertiesFromRequestBody(
 export function extractPropertiesFromResponseBody(
   operation: AzureOpenApiOperation | null,
 ) {
-  const schema = operation?.responses["200"]?.schema;
+  const schema = operation?.responses?.["200"]?.schema;
   if (!schema) return { properties: {}, required: [] };
+  assert(
+    isAzureObjectProperty(schema),
+    `Response schema is not an object: ${operation.operationId}`,
+  );
 
   return {
     properties: schema.properties ?? {},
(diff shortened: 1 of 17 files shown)


sprutton1:
> Don't process readonly schemas (if it only has GET it's probably a query, search or list)

We should enumerate these. I would say things like `Location` fit this description. We have a custom asset for this already, but I'm not sure what else might be useful from this category.

bin/clover/src/pipelines/azure/spec.ts
```diff
@@ -7,16 +7,25 @@ import {
...
 
+// Fix typescript issue where Array.isArray doesn't work for readonly arrays
+// https://github.com/microsoft/TypeScript/issues/17002#issuecomment-2781717755
+declare global {
+  interface ArrayConstructor {
+    // deno-lint-ignore no-explicit-any
+    isArray(arg: ReadonlyArray<any> | any): arg is ReadonlyArray<any>;
+  }
+}
+
```
sprutton1: :shaking_face:

bin/clover/src/pipelines/azure/spec.ts
```diff
@@ -413,71 +494,59 @@ export function mergeAzureResourceOperations(
...
-  }
-  return null;
+  // Form: /subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Network/applicationGateways/{applicationGatewayName}
+  const match = path.match(/\/providers\/([^/]+)\/([^/{]+)\/\{[^/}]+\}$/);
+  if (!match) return null;
+  const [_, providerNamespace, resourceType] = match;
+  const serviceName = providerNamespace.split(".").pop() || providerNamespace;
+  const capitalizedResource =
+    resourceType.charAt(0).toUpperCase() + resourceType.slice(1);
+  return `Azure::${serviceName}::${capitalizedResource}`;
```
sprutton1, line 496: Nothing to do here, but we really should follow Microsoft's convention in naming

bin/clover/src/spec/props.ts
```diff
@@ -139,59 +123,24 @@ export function findPropByName(
 // Create top-level prop such as domain, resource_value, secrets, etc.
 export function createDefaultPropFromCf(
```
sprutton1: nit: maybe we should change the `FromCf` bits to `FromJsonSchema`

bin/clover/src/spec/props.ts
```diff
@@ -139,59 +123,24 @@ export function findPropByName(
...
-      queue,
-      docFn,
-      requiredFn,
-      providerConfig,
-    );
-    if (!prop) continue;
-    if (propArgs.addTo) propArgs.addTo(prop);
-  }
+  // Recursively create prop and all its children
+  const rootProp = createPropFromCf(
```
sprutton1: same nit here

bin/clover/src/spec/props.ts
```diff
@@ -202,346 +151,357 @@ export function createDefaultPropFromCf(
...
+          "/",
+        )}`,
+      );
+    }
 
-  if (!cfProp.title) {
-    cfProp.title = name;
-  }
+    if (propsBeingCreated.includes(schemaProp)) {
+      console.debug(`Circular property definition: ${propPath.join("/")}`);
```
sprutton1: We do have a logger in `src/logger.ts`. We use it kind of sporadically.

sprutton1 (dismissed):
Amazing work here!

bin/clover/src/spec/props.ts
jkeiser: yeah! Everything in generic should work off either JSONSchema or something without a `Cf` prefix :) I'd like to make a pass on it soon.

bin/clover/src/spec/props.ts
jkeiser: Changed all logging in this file

sprutton1 (approved):

PR #7533, opened 2025-10-17
fix(web): Fix an issue wjth Workspace integrations

stack72:
The update endpoint accepted a user-controlled integration_id in the URL path and validated it AFTER lookup, allowing attackers to modify integrations in other workspaces by supplying a different integration ID.

Eliminated integration IDs from the API surface entirely:

1. Removed ID from URL - Changed POST /workspaces/{id}/integrations/{integration_id} to POST /workspaces/{id}/integrations
2. Server-side lookup - Integration is now looked up by workspace context from auth token, not by user-supplied ID
3. Sanitized responses - Created IntegrationResponse type that excludes the internal database ID from API responses

Why is this better?
-  Each workspace has exactly one integration row (enforced by UNIQUE INDEX on workspace_pk)
- No user-controllable identifiers means no opportunity for ID manipulation
- Defense in depth: ID never leaves the server, making attacks architecturally impossible

### DIFF
--- a/app/web/src/components/WorkspaceIntegrationsModal.vue
+++ b/app/web/src/components/WorkspaceIntegrationsModal.vue
@@ -39,9 +39,8 @@ function open() {
 const webhookUrl = ref("");
 
 const updateIntegrations = () => {
-  if (!webhookUrl.value || webhookUrl.value === "" || !integration.value?.pk)
-    return;
-  workspacesStore.UPDATE_INTEGRATION(integration.value?.pk, webhookUrl.value);
+  if (!webhookUrl.value || webhookUrl.value === "") return;
+  workspacesStore.UPDATE_INTEGRATION(webhookUrl.value);
 
   modalRef.value?.close();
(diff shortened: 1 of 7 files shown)


stack72:
When creating a new integration - we get a response of:

```
{
    "integration": {
        "workspacePk": "01HPJ3K7RC486W9RPR3T0B6HS2",
        "slackWebhookUrl": "https://testing-slack.com"
    }
}
```

Notice no integrationId in there!!

When I  update it, the API call I make it:

```
http://localhost:8080/api/v2/workspaces/01HPJ3K7RC486W9RPR3T0B6HS2/integrations <---- where that ID is the workspaceID
```

and the response I get is:

```
{
    "integration": {
        "workspacePk": "01HPJ3K7RC486W9RPR3T0B6HS2",
        "slackWebhookUrl": "https://testing-slack-updated.com"
    }
}
```

No IDs are returned anywhere to the web app now!

PR #7537, opened 2025-10-17
Bust deployment atoms, refactor to common pattern

jobelenus:

1. Deployment atoms weren't getting cache busts sent to tanstack (b/c nothing was reading them in the last PR!)
2. Handle global atoms with `getKind`

## How was it tested?

Same as the last PR 🤪 (add component modal)

### DIFF
--- a/app/web/src/store/realtime/heimdall.ts
+++ b/app/web/src/store/realtime/heimdall.ts
@@ -31,9 +31,9 @@ import {
   ConnStatusFn,
 } from "@/workers/types/dbinterface";
 import {
-  CachedDefaultVariant,
   Connection,
   EntityKind,
+  GLOBAL_IDENTIFIER,
   GlobalEntity,
   SchemaMembers,
 } from "@/workers/types/entity_kind_types";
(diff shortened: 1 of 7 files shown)


app/web/src/store/realtime/heimdall.ts
```diff
@@ -237,13 +235,24 @@ const bustTanStackCache: BustCacheFn = (
...
-  queryClient?.invalidateQueries({ queryKey });
-  if (!noBroadcast) {
-    db.broadcastMessage({
-      messageKind: "cacheBust",
-      arguments: { workspaceId, changeSetId, kind, id },
-    });
+  if (workspaceId === "-" && changeSetId === "-") {
+    const queryKey = ["global", kind, id];
+    queryClient?.invalidateQueries({ queryKey });
+    if (!noBroadcast) {
```
zacharyhamm: fwiw, this `if` could be moved below and occur only once

app/web/src/workers/webworker.ts
```diff
@@ -1189,6 +1196,19 @@ const handleDeploymentPatchMessage = async (
...
+    [...modifications, ...inserts, ...removals]
+      .filter((atom): atom is GlobalAtom =>
+        GLOBAL_ENTITIES.includes(atom.kind as GlobalEntity),
+      )
+      .forEach((atom) => {
+        kinds.push(atom.kind);
+        bustDeployment(atom.kind, atom.id);
+      });
+    kinds.forEach((k) => {
+      bustDeployment(k, "-");
```
zacharyhamm: These magic "-" are not making me ecstatic. Can we stash this in a `const` somewhere with a name?

jobelenus: Absolutely!

PR #7547, opened 2025-10-18
feat(auth-api): Allow workspace routes to be used via automation tokens

stack72:
With this change, I can do the following using our workspace api tokens:

```
curl -X 'GET' \
  'http://localhost:9001/workspaces' \
  -H 'accept: */*' \
  -H 'Authorization: Bearer <redacted>’
```

The response is:

```
[
  {
    "id": "01K7TBG7KSSFAFQQ8AKY4WRE2B",
    "instanceEnvType": "SI",
    "instanceUrl": "https://app.systeminit.com",
    "displayName": "paul's  Production Workspace",
    "creatorUserId": "01K7TBG7KFKKG7CH3KK11126GH",
    "deletedAt": null,
    "token": "01K7TBG7KSQY5HB5P42GMW69SJ",
    "isDefault": true,
    "quarantinedAt": null,
    "description": "",
    "isFavourite": false,
    "isHidden": false,
    "approvalsEnabled": false,
    "creatorUser": {
      "firstName": "Paul",
      "lastName": "Stack"
    },
    "role": "OWNER",
    "invitedAt": "2025-10-18T00:28:51.220Z"
  }
]
```

Before this change:

```
{
  "kind": "AutomationToken",
  "message": "Automation tokens may not access the auth api"
}
```

### DIFF
--- a/bin/auth-api/src/routes/workspace.routes.ts
+++ b/bin/auth-api/src/routes/workspace.routes.ts
@@ -39,7 +39,7 @@ import { tracker } from "../lib/tracker";
 import { findLatestTosForUser } from "../services/tos.service";
 import { automationApiRouter, extractAuthUser, router } from ".";
 
-router.get("/workspaces", async (ctx) => {
+automationApiRouter.get("/workspaces", async (ctx) => {
   const authUser = extractAuthUser(ctx);
   ctx.body = await getUserWorkspaces(authUser.id);
 });
@@ -98,7 +98,7 @@ automationApiRouter.get("/workspaces/:workspaceId", async (ctx) => {
   ctx.body = workspace;
 });
 
-router.delete("/workspaces/:workspaceId", async (ctx) => {
+automationApiRouter.delete("/workspaces/:workspaceId", async (ctx) => {
   extractAuthUser(ctx, true);
   const { authUser, workspaceId } = await authorizeWorkspaceRoute(ctx, [RoleType.OWNER]);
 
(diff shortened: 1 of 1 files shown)


bin/auth-api/src/routes/workspace.routes.ts
```diff
@@ -98,7 +98,7 @@ automationApiRouter.get("/workspaces/:workspaceId", async (ctx) => {
   ctx.body = workspace;
 });
 
-router.delete("/workspaces/:workspaceId", async (ctx) => {
+automationApiRouter.delete("/workspaces/:workspaceId", async (ctx) => {
```
jkeiser, line 101: Just checking, this is on purpose, yes?

bin/auth-api/src/routes/workspace.routes.ts
```diff
@@ -211,7 +211,7 @@ router.get("/workspace/:workspaceId/members", async (ctx) => {
   ctx.body = members;
 });
 
-router.post("/workspace/:workspaceId/membership", async (ctx) => {
+automationApiRouter.post("/workspace/:workspaceId/membership", async (ctx) => {
   const { authUser, workspace } = await authorizeWorkspaceRoute(ctx, [
```
jkeiser, line 215: I believe this will work, but we should make certain API tokens from non-owners can't take action here--i don't think we've had occasion to test that before.

jkeiser (approved):

bin/auth-api/src/routes/workspace.routes.ts
britmyerss: yeah, based on feedback :)

PR #7548, opened 2025-10-18
feat(luminork): Add API endpoint to add components to view

keeb:
Add new POST /v1/w/{workspace}/cs/{change_set}/components/add_to_view endpoint that accepts a view name and list of component IDs. The endpoint:

- Creates the view automatically if it doesn't exist
- Adds all specified components to the view with default geometry
- Follows sdf's error handling: succeeds if at least one component is added, skips components already in the view
- Tracks operations via PostHog analytics
- Includes full OpenAPI documentation

Logic lifted from lib/sdf-server/src/service/v2/view/add_components.rs


lib/luminork-server/src/service/v1/components/add_to_view.rs
```diff
@@ -0,0 +1,109 @@
...
+        (status = 422, description = "Validation error - Invalid request data", body = crate::service::v1::common::ApiError),
+        (status = 500, description = "Internal server error", body = crate::service::v1::common::ApiError)
+    )
+)]
+pub async fn add_to_view(
+    ChangeSetDalContext(ref ctx): ChangeSetDalContext,
+    tracker: PosthogEventTracker,
+    payload: Result<Json<AddToViewV1Request>, axum::extract::rejection::JsonRejection>,
+) -> Result<StatusCode, ComponentsError> {
+    let Json(payload) = payload?;
```
britmyerss, line 55: Do we want this?

lib/luminork-server/src/service/v1/components/add_to_view.rs
```diff
@@ -0,0 +1,109 @@
...
+
+    // Prevent mutations on HEAD change set
+    if ctx.change_set_id() == ctx.get_workspace_default_change_set_id().await? {
+        return Err(ComponentsError::NotPermittedOnHead);
+    }
+
+    // Look up view by name, create if it doesn't exist
+    let view = match View::find_by_name(ctx, &payload.view_name).await? {
+        Some(view) => view,
+        None => View::new(ctx, &payload.view_name).await?,
```
britmyerss, line 65: I don't remember if this will prevent duplicate names or not, worth a check

lib/luminork-server/src/service/v1/components/add_to_view.rs
```diff
@@ -0,0 +1,109 @@
...
+    };
+    let view_id = view.id();
+
+    // Add components to view - match sdf behavior:
+    // If at least one component succeeds, don't blow up if errors happen
+    let mut at_least_one_succeeded = false;
+    let mut latest_error = None;
+
+    for component_id in payload.component_ids.iter() {
+        match Component::add_to_view(ctx, *component_id, view_id, RawGeometry::default()).await {
```
britmyerss, line 75: Reviewing on thumbs, is this the right logic? Looks like we won't proceed if we get any other error than it already existing....

lib/luminork-server/src/service/v1/components/add_to_view.rs
stack72: The find_by_name will check for that view name so no dupe views :D

lib/luminork-server/src/service/v1/components/add_to_view.rs
stack72: yes, this checks for validation issues - we do this in most of hte Luminork routes

PR #7552, opened 2025-10-19
Fix SDF API tests and Re-enable queueing the Edda Task for Deployment MVs on Migration

britmyerss:
With the removal of the SchemaVariantCategories MV, these tests need to be updated to not rely on them.  We were previously using this MV to determine whether various Schemas used in tests were installed or not, as SDF still requires the caller to pass in this information.  

The following updates were made: 
- Add support for fetching the deployment index and deployment MVs from SDF
- Create a test helper that creates the correct component payload, based on whether the schema in question is installed or not
- Update all tests to use the new helper
- Update 8-check_mjolnir, which was directly relying on SchemaVariantCategories to determine that a variant was installed when the component was created for it. Replace this with checking for the LuminorkDefaultVariant 
- Update 2-create_and_apply_across_change_sets to hardcode the `SchemaId` for the `TestResourceActions` Schema, which is a special asset we use only for that test that exists in the module index

Bonus:
We had disabled enqueuing the Edda task that would build the Deployment MVs, when the builtins are cached during migration as the task was taking 4ish hours. Now that this task completes rather quickly, let’s do this on migration again. This mostly improves the local dev experience so we don’t have to manually enqueue it to see the builtins available for component creation on startup.

#### Out of Scope:

We should really update SDF and Web to work like Luminork when creating new components. A job for another day. 

## How was it tested?
All tests pass locally, including various permutations of having variants installed or not. 

<div><br/>via <a href="https://giphy.com/billions/">Billions</a> on <a href="https://giphy.com/gifs/billions-season-4-episode-10-410-ibvAMs6KEaesONKsGS">GIPHY</a></div>

### DIFF
--- a/lib/dal/src/cached_module.rs
+++ b/lib/dal/src/cached_module.rs
@@ -310,7 +310,7 @@ impl CachedModule {
         ctx: &DalContext,
         modules: &HashMap<String, ModuleDetailsResponse>,
         module_index_client: ModuleIndexClient,
-        _edda_client: EddaClient,
+        edda_client: EddaClient,
     ) -> CachedModuleResult<Vec<CachedModule>> {
         let hashes = modules.keys().map(ToOwned::to_owned).collect_vec();
         let uncached_hashes = CachedModule::find_missing_entries(ctx, hashes).await?;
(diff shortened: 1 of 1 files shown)


britmyerss:
Successful run here in tools: https://github.com/systeminit/si/actions/runs/18625982473

PR #7561, opened 2025-10-20
Graceful failure for changeset mvIndex

jobelenus:

If an `MvIndex` fails to build, the endpoint to retrieve them will throw a 500. This will not let a user out of the lobby.

We ought to let people out of the lobby. With these changes, we mark 500s separately, and display an error for that change set after the user exits the lobby.

NOTE: we didn't need the retry logic in the main browser thread anymore.

## How was it tested?

First, lets test a 500. I put this code into the `get_change_set_index` route in SDF:
```
    if change_set_id.to_string() == "01K81JTR58GKFG0MKTEZY3MZT7" {
        let e = Error::missing_field("foo")
        return Err(IndexError::DeserializingMvIndexData(e));
    }
```
When you load that change set:
1. you're in the lobby
2. you see 5 attempts in the network panel which all 500
3. after that you come out of the lobby and you see this: 
image
You can move change sets to `HEAD`, and make new ones no problem

Second, lets test a 404 that ought to indicate "index is rebuilding" to make sure I didn't break anything. Replace my `get_change_set_index` changes with this:
```
    if change_set_id.to_string() == "01K81JTR58GKFG0MKTEZY3MZT7" {
        return Err(IndexError::IndexNotFound(workspace_pk, change_set_id));
    }
```

The user _will_ be stuck in the lobby based on my artificial test case. In the real world, the system would fire an `IndexUpdate` message, which we re-fire the cold start that would succeed and the user would leave the lobby.

Third;
1. put my initial status code 500 changes back
2. load the page, get past the lobby
3. remove my SDF changes entirely
5. DO NOT REFRESH THE PAGE
6. select the head change set
7. select the change set i was special casing
8. it will coldstart, and succeed, and the user will see the change set.\

Fourth;
1. nuke my frigg
2. cold start
3. watch the index re-build and get out of the lobby
4. make a new change set, see that it works / no errors

### DIFF
--- a/app/web/src/newhotness/Workspace.vue
+++ b/app/web/src/newhotness/Workspace.vue
@@ -127,6 +127,23 @@
         />
         <Onboarding v-else @completed="onboardingCompleted = true" />
       </template>
+      <main
+        v-else-if="indexFailedToLoad"
+        :class="
+          clsx(
+            'grow min-h-0 m-sm border flex flex-row items-center justify-center',
+            themeClasses('border-neutral-400', 'border-neutral-600'),
+          )
+        "
+      >
+        <EmptyState
+          icon="logo-si"
+          text="Unable to Load Change Set Data"
+          secondaryText="We failed to load the data associated with this change set. This is our problem, and we have been alerted. You may select another change set."
+          finalLinkText="Click here to see your other workspaces"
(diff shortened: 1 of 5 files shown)


britmyerss (approved):
well done! The code comments throughout are extremely helpful too

PR #7562, opened 2025-10-21
Add userPk to the bifrost observability data

jobelenus:
## How was it tested?

1. See the `userPk` in the console
2. Will see data in tools prod honeycomb


zacharyhamm (approved):
🚀

PR #7570, opened 2025-10-21
adding search functionality to the views dropdown

wendybujalski:

Just like the workspaces and change sets dropdowns, the views dropdown now has a search available when more than the number of views exceeds the `DEFAULT_DROPDOWN_SEARCH_THRESHOLD` (which is 10)

#### Screenshots:

Screenshot 2025-10-21 at 5 39 45 PM

#### Out of Scope:

There may be other dropdowns in the UI we want to also add search to! I just did this one very quickly because it was bothering me and was only a few minutes of work to add 😅

## How was it tested?

Double checked the search functionality with a variety of cases after making sure the flow is identical to the working flow in the change sets dropdown.

### DIFF
--- a/app/web/src/newhotness/Explore.vue
+++ b/app/web/src/newhotness/Explore.vue
@@ -36,8 +36,10 @@
       <template v-else>
         <div class="flex-none flex flex-row items-start gap-xs">
           <DropdownMenuButton
+            ref="viewsDropdownRef"
             class="rounded min-w-[128px] h-[2.5rem]"
-            :options="viewListOptions"
+            :options="filteredViewListOptions"
+            :search="viewListOptions.length > DEFAULT_DROPDOWN_SEARCH_THRESHOLD"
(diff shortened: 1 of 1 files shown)


app/web/src/newhotness/Explore.vue
```diff
@@ -838,6 +841,20 @@ const viewListOptions = computed(() => {
     .map((l) => ({ value: l.id, label: l.name }))
     .sort((a, b) => a.label.localeCompare(b.label));
 });
+const viewsDropdownRef = ref<InstanceType<typeof DropdownMenuButton>>();
+const filteredViewListOptions = computed(() => {
+  const searchString = viewsDropdownRef.value?.searchString;
+
+  if (!searchString || searchString === "") {
+    return viewListOptions.value;
+  }
```
nickgerace, line 850: Good idea to default to showing everything

nickgerace (approved):
This looks good to me!

PR #7571, opened 2025-10-21
Prevent incomplete subscriptions

nickgerace:
## Description

This change prevents incomplete subscriptions in edge case scenarios by removing new node updates for attribute prototype arguments conditionally. This condition is triggered when there is a set difference between the set of all destinations that no longer exist in the graph (i.e. the "incomplete" subscriptions) and the set of all new node updates.

This change fixes a bug where the erasure of a subscription component (not the subscriber) in one change set would result in subscriber components containing incomplete subscriptions when HEAD also contained changes to the subscription itself. This case is tested in an integration test.

## Test Steps in the UI

1. In a new change set, create an `AWS Credential` and an `AWS::EC2::Instance` component
2. Subscribe to the credential from the EC2 instance component
3. Apply the change set to HEAD (the create action does not need to pass)
4. Create a change set and erase the credential component (do not apply this change set)
5. Create another change set and remove the subscription to the credential
6. In that same change set, re-add the subscription and then apply the change set
7. Switch back to the erasure change set and observe that both the attributes panel and debug panel now work (without this PR, there would be an incomplete subscription)

## Integration Test Screenshots

Without the fix:

Screenshot 2025-10-22 at 5 58 51 PM

With the fix:

Screenshot 2025-10-22 at 6 30 41 PM

### DIFF
--- a/lib/dal/src/workspace_snapshot/node_weight/attribute_prototype_argument_node_weight.rs
+++ b/lib/dal/src/workspace_snapshot/node_weight/attribute_prototype_argument_node_weight.rs
@@ -1,3 +1,5 @@
+use std::collections::HashSet;
+
 use serde::{
     Deserialize,
     Serialize,
@@ -9,11 +11,15 @@ use si_events::{
     ulid::Ulid,
 };
 use si_id::ComponentId;
+use telemetry::prelude::*;
 
 use crate::{
     EdgeWeightKindDiscriminants,
     workspace_snapshot::{
-        graph::LineageId,
+        graph::{
+            LineageId,
+            detector::Update,
+        },
         node_weight::traits::CorrectTransforms,
     },
 };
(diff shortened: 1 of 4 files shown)


lib/dal/src/workspace_snapshot/node_weight/attribute_prototype_argument_node_weight.rs
```diff
@@ -118,4 +123,67 @@ impl std::fmt::Debug for AttributePrototypeArgumentNodeWeight {
...
+            return Ok(updates);
+        }
+
+        // Setup our caches needed to both determine which updates to remove and to perform the
+        // removals.
+        let mut destinations_that_do_not_exist: HashSet<Ulid> = HashSet::new();
+        let mut all_new_nodes = HashSet::new();
+        let mut indices_for_new_node_updates_for_ourself = Vec::new();
+
+        // For each update, we need to collect all destinations that no do not exist in the current
```
jhelwig: `that now do not exist`? `that do not exist`? I don't think the `no` was intentional.

lib/dal/src/workspace_snapshot/node_weight/attribute_prototype_argument_node_weight.rs
```diff
@@ -118,4 +123,67 @@ impl std::fmt::Debug for AttributePrototypeArgumentNodeWeight {
...
+        // If there is a difference between the sets, we need to check that there are updates to
+        // remove. There should be exactly one: the new node update for ourself.
+        if destinations_that_do_not_exist
+            .difference(&all_new_nodes)
+            .next()
+            .is_some()
+        {
+            for idx in indices_for_new_node_updates_for_ourself {
+                updates.remove(idx);
+            }
```
jhelwig, line 195: After the first `updates.remove(idx)` all subsequent indices will be wrong, unless you process them in reverse order. (Everything shifts forward after each removal.)

Since multiple indices really shouldn't happen, it might be worth adding a warning if `indices_for_new_node_updates_for_ourself.len() > 1`.

jhelwig (changes requested):
One small, but important change in the index handling, but otherwise looks good. 👍🏻

lib/dal/src/workspace_snapshot/node_weight/attribute_prototype_argument_node_weight.rs
zacharyhamm: ah that's a good call. Yes there should only ever be one

nickgerace: I'll fix this!

PR #7573, opened 2025-10-21
Workspace Snapshot Eviction

jhelwig:
## Problem

The workspace snapshot eviction system had a critical race condition causing hard failures during concurrent change set operations:

**Race condition timeline:**
```
T0: ChangeSet A points to Snapshot S1 (only reference)
T1: User creates ChangeSet B
    - Transaction starts, reads A's pointer (S1)
    - In-memory: Will create B pointing to S1
    - Transaction NOT committed yet
T2: Rebaser processes ChangeSet A
    - Updates A: S1 → S2
    - Immediate eviction check: "Does any change set point to S1?"
    - Result: NO (B's transaction not visible yet)
    - Decision: EVICT S1
T3: ChangeSet B transaction commits
    - Database now has: B points to S1
    - But S1 no longer exists in durable storage
    - Result: HARD FAILURE when B tries to use S1
```

**Window of vulnerability:** Milliseconds to ~1 second (typical transaction duration)

### Previous Attempt: Creation-Time Grace Period

A prior attempt used a grace period based on when snapshots were created, but had fundamental flaws:

1. **Still raceable with old snapshots** - If a base snapshot was created 10 minutes ago, a new change set creation happening now gets no protection from the grace period
2. **Missed evictions** - Eviction check only happened once when the change set moved away. If the snapshot was within the grace period at that moment, it never got evicted even after the period expired
3. **Wrong metric** - Grace period based on creation time, not when the snapshot was last used

## Solution

Redesigned eviction using **recency-based tracking with polling** to eliminate the race condition:

**Core approach:**
1. Track when snapshots were LAST USED (not created)
2. Grace period (300s default) vastly exceeds transaction duration (<1s)
3. Continuous polling service checks for candidates every 60s
4. Atomic database query validates both age and active references

**Why this eliminates the race:**

By the time a snapshot becomes eligible for eviction (300 seconds after last use), any in-flight transaction will have committed 299+ seconds ago. The grace period provides a 300x safety margin over typical transaction duration.

Even in extreme cases (10-second transaction due to load), there's still a 290-second buffer.

### Architecture

New `lib/snapshot-eviction` library with zero generic type parameters:

```rust
pub struct SnapshotEvictor {
    si_db_pool: PgPool,                    // For queries + metadata
    layer_cache_pool: PgPool,              // For workspace_snapshots deletion
    layered_event_client: LayeredEventClient, // For NATS events
    config: SnapshotEvictionConfig,
}
```

#### 1. Memory-Only Eviction (Rebaser)
When a change set is rebased to a new snapshot, the rebaser immediately triggers memory-only eviction:
- Removes snapshot from local in-memory foyer cache
- Publishes NATS eviction event to invalidate all service instances' caches
- **Does NOT delete from PostgreSQL**
- Fire-and-forget with debug-level error logging
- Goal: Immediate memory pressure relief across the cluster

#### 2. Full Eviction (Forklift)
Forklift's polling-based eviction is the **only** path that deletes from PostgreSQL:
- Polls for eviction candidates using grace period (5 minutes default)
- Deletes from PostgreSQL durable storage
- Publishes NATS eviction event
- Removes from tracking table

This controlled approach with grace periods prevents race conditions where in-flight transactions might reference recently-evicted snapshots.

### Database Tracking

- `snapshot_last_used` table tracks when snapshots were last used
- Updated automatically when change sets update their pointers
- Migration U3913 creates the tracking table
- Migration U3914 backfills existing snapshots

**Tracking update (via `ChangeSet::update_pointer`):**
```sql
UPDATE change_set_pointers SET workspace_snapshot_address = $new WHERE id = $id;

INSERT INTO snapshot_last_used (snapshot_id, last_used_at, created_at)
VALUES ($old, CLOCK_TIMESTAMP(), CLOCK_TIMESTAMP())
ON CONFLICT (snapshot_id) DO UPDATE SET last_used_at = CLOCK_TIMESTAMP();
```

### Eviction Library

**Query for candidates:**
```sql
SELECT s.snapshot_id, s.last_used_at
FROM snapshot_last_used s
LEFT JOIN change_set_pointers cs ON cs.workspace_snapshot_address = s.snapshot_id
WHERE s.last_used_at < (NOW() - $grace_period * INTERVAL '1 second')
  AND cs.id IS NULL
ORDER BY s.last_used_at ASC
LIMIT $batch_size
FOR UPDATE OF s SKIP LOCKED
```

**Three-step eviction:**
1. Delete from `workspace_snapshots` table (layer-cache DB) via PgLayer
2. Publish NATS event (Actor::System, Tenancy with empty ULIDs) for cache invalidation
3. Delete metadata from `snapshot_last_used` (si-db)

All operations autocommit. Failures logged but don't stop processing.

### Forklift Integration

Runs as tokio task spawned in forklift's `try_run` method:
```rust
let snapshot_evictor = SnapshotEvictor::new(si_db_pool, layer_cache_pool, event_client, config);
let eviction_task = tokio::spawn(async move {
    snapshot_evictor.run(eviction_shutdown).await
});
```

Graceful shutdown coordination with existing forklift services (audit logs, billing events).

### Metrics Instrumentation

**Per-cycle batch counters:**
- `monotonic_counter.snapshot_eviction_cycles_completed` - Total eviction cycles completed
- `monotonic_counter.snapshot_eviction_successes` - Total successful evictions (recorded per-cycle as batch)
- `monotonic_counter.snapshot_eviction_failures` - Total failed evictions (recorded per-cycle as batch) with `error_type` label

**Distributions:**
- `histogram.snapshot_eviction_candidates_found` - Distribution of candidates found per cycle (tracks workload characteristics)
- `histogram.snapshot_eviction_cycle_duration_seconds` - Complete cycle execution time in seconds
- `histogram.snapshot_eviction_duration_seconds` - Eviction duration distribution in seconds with `status` and `error_type` labels
- `histogram.snapshot_candidate_queue_latency_seconds` - Time from eligibility to processing in seconds

Success and failure counters are emitted once per cycle with batch totals, avoiding partial window artifacts in Prometheus queries.

Duration and queue latency are recorded in seconds to work with default OpenTelemetry histogram buckets (0-10000 range), providing good granularity for multi-second operations.

### Grafana Dashboard

Comprehensive observability dashboard with organized panels:

**Health Section:**
- **Success Rate** - Percentage with thresholds (red <95%, yellow 95-99%, green >99%)

**Workload Section:**
- **Candidates Found** - Distribution of eviction candidates per cycle
- **Candidate Queue Latency P50/P95/P99** - How long candidates wait from eligibility to processing
- **Cycle Duration P50/P95/P99** - End-to-end cycle execution time

**Throughput Section:**
- **Eviction Results** - Time series showing successes and failures over time

**Reliability Section:**
- **Failures by Error Type** - Breakdown of failure modes (database, pg_pool, layer_cache, nats)

**Performance Section:**
- **Eviction Duration P50/P95/P99 (Success)** - Latency percentiles for successful evictions
- **Failure Duration by Error Type P50/P95/P99** - Latency analysis for different failure types

## Configuration

Requires forklift configuration with eviction settings and two separate database pool configurations:

```toml
[snapshot_eviction]
grace_period_seconds = 300    # 5 minutes (min: 30s)
poll_interval_seconds = 60    # 1 minute (min: 30s)
batch_size = 100

  [snapshot_eviction.si_db]
  dbname = "si"
  hostname = "localhost"
  port = 5432
  # ... standard PgPoolConfig fields

  [snapshot_eviction.layer_cache_pg]
  dbname = "si_layer_cache"
  hostname = "localhost"
  port = 5432
  # ... standard PgPoolConfig fields
```

The eviction task requires separate database connections:
- **si_db**: For querying `snapshot_last_used` and `change_set_pointers` tables
- **layer_cache_pg**: For deleting snapshots from `workspace_snapshots` table

Values below minimum thresholds (30 seconds) are automatically clamped with warning logs.

## Testing

- Integration test scaffolding validates eviction workflow
- Test for snapshot tracking on pointer updates
- Manual load testing confirmed proper behavior under concurrent operations

## Dashboard Metrics - Local Testing

image

## Migration Path

1. **U3913**: Creates `snapshot_last_used` table with appropriate indexes
2. **U3914**: Backfills table with existing snapshots, setting `last_used_at` to current timestamp
3. **Rebaser integration**: Automatic tracking on all pointer updates via `ChangeSet::update_pointer`
4. **Forklift deployment**: Spawns eviction task as part of normal server lifecycle

Existing snapshots get a grace period from deployment time, preventing immediate eviction of long-lived snapshots.

### DIFF
--- a/lib/dal/tests/integration_test/component/delete.rs
+++ b/lib/dal/tests/integration_test/component/delete.rs
@@ -199,6 +199,13 @@ async fn delete_multiple_components(ctx: &mut DalContext) -> Result<()> {
     )
     .await?;
 
+    // Remove any pending actions before applying to base to prevent
+    // component_still_on_head from getting a resource via Create action
+    let action_ids = Action::all_ids(ctx).await?;
+    for action_id in action_ids {
+        Action::remove_by_id(ctx, action_id).await?;
+    }
+
     ChangeSetTestHelpers::apply_change_set_to_base(ctx).await?;
 
     ChangeSetTestHelpers::fork_from_head_change_set(ctx).await?;


lib/snapshot-eviction/Cargo.toml
```diff
@@ -0,0 +1,32 @@
+[package]
```
fnichol, line 1: Seeing the formatting here has led me to discover that a `:check-format-toml` rule isn't being added via the `rust_library` macro, which is my fault. This `Cargo.toml` is fine given the current CI checks so I can update when the rule is in place (it's mostly around sorting deps lexically and using the workspace-provided metadata from the root)

lib/snapshot-eviction/src/config.rs
```diff
@@ -0,0 +1,168 @@
...
+
+        if self.poll_interval_seconds < MIN_POLL_INTERVAL_SECONDS {
+            warn!(
+                configured = self.poll_interval_seconds,
+                minimum = MIN_POLL_INTERVAL_SECONDS,
+                "poll_interval_seconds below minimum, clamping to minimum"
+            );
+            self.poll_interval_seconds = MIN_POLL_INTERVAL_SECONDS;
+        }
+    }
```
fnichol, line 95: 👍🏽

lib/dal/src/change_set.rs
```diff
@@ -472,6 +475,19 @@ impl ChangeSet {
...
+            .await?
+            .pg()
+            .query_none(
+                "INSERT INTO snapshot_last_used (snapshot_id, last_used_at, created_at)
+                 VALUES ($1, CLOCK_TIMESTAMP(), CLOCK_TIMESTAMP())
+                 ON CONFLICT (snapshot_id)
+                 DO UPDATE SET last_used_at = CLOCK_TIMESTAMP()",
+                &[&old_snapshot_address.to_string()],
+            )
+            .await?;
```
fnichol, line 489: Here's one of the rare times where it's nice that we're in a db txn.

lib/dal/tests/integration_test/change_set.rs
```diff
@@ -588,3 +588,47 @@ async fn change_set_approval_flow(ctx: &mut DalContext) {
         .collect_vec();
     assert_eq!(components.len(), 2);
 }
+
+#[test]
+async fn update_pointer_tracks_last_used(ctx: &mut DalContext) {
```
fnichol, line 593: Great idea!

lib/rebaser-server/src/config.rs
```diff
@@ -275,14 +257,6 @@ fn default_quiescent_period_secs() -> u64 {
     DEFAULT_QUIESCENT_PERIOD_SECS
 }
 
-fn default_snapshot_eviction_grace_period() -> Duration {
-    DEFAULT_SNAPSHOT_EVICTION_GRACE_PERIOD
-}
-
-fn default_snapshot_eviction_grace_period_secs() -> u64 {
-    DEFAULT_SNAPSHOT_EVICTION_GRACE_PERIOD_SECS
-}
```
fnichol, line 284: It's always about Configuration Management at the end of the day, isn't it 😄

lib/rebaser-server/src/rebase.rs
```diff
@@ -147,20 +144,17 @@ pub(crate) async fn perform_rebase(
         }
     }
 
+    // Spawn fire-and-forget task to evict old snapshot from memory caches
+    // This provides immediate memory pressure relief across all service instances
+    // PostgreSQL deletion is handled separately by forklift's polling eviction
```
fnichol, line 149: Yeah, good call out

lib/rebaser-server/src/rebase.rs
```diff
@@ -514,32 +508,20 @@ pub(crate) async fn send_updates_to_edda_legacy_snapshot(
...
-    );
-    if !in_use && !is_recent {
-        ctx.layer_db().workspace_snapshot().evict(
-            workspace_snapshot_address,
-            ctx.events_tenancy(),
-            ctx.events_actor(),
-        )?;
-    }
-
+    ctx.layer_db().workspace_snapshot().evict_memory_only(
```
fnichol, line 519: I like the intention-revealing method name here

lib/snapshot-eviction/src/evictor.rs
```diff
@@ -0,0 +1,276 @@
...
+    pub async fn run(&self, shutdown: CancellationToken) -> SnapshotEvictionResult<()> {
+        info!(
+            grace_period_seconds = self.config.grace_period_seconds,
+            poll_interval_seconds = self.config.poll_interval_seconds,
+            "Snapshot eviction task starting"
+        );
+
+        loop {
+            tokio::select! {
+                biased;
```
fnichol, line 73: After I learned about `biased;`, I do love me some `biased` when it's called for, and I agree it's called for here!

lib/forklift-server/src/server.rs
```diff
@@ -172,25 +216,101 @@ impl Server {
...
+        // Extract fields to avoid partial move issues
+        let snapshot_evictor = self.snapshot_evictor;
+        let inner_audit_logs = self.inner_audit_logs;
+        let inner_billing_events = self.inner_billing_events;
+        let shutdown_token = self.shutdown_token;
+
+        // Spawn snapshot eviction background task
+        let eviction_shutdown = shutdown_token.clone();
+        let eviction_task =
+            tokio::spawn(async move { snapshot_evictor.run(eviction_shutdown).await });
```
fnichol, line 228: The `tokio::spawn` is okay assuming that the Evictor is ok being cancelled at any point is whatever it's doing. If some form of "gracefully wrapping up work" is needed, then wiring that spawn into a `TaskTracker` would help. I'll double check the evictor's core loop, so this might by an FYI for others and not a factor in this case 😉

lib/snapshot-eviction/src/evictor.rs
```diff
@@ -0,0 +1,276 @@
...
+                // Extract error type for labeling
+                let error_type = match &e {
+                    crate::SnapshotEvictionError::Database(_) => "database",
+                    crate::SnapshotEvictionError::PgPool(_) => "pg_pool",
+                    crate::SnapshotEvictionError::LayerCache(_) => "layer_cache",
+                    crate::SnapshotEvictionError::Nats(_) => "nats",
+                };
+
+                // Record failure counter with error type
+                monotonic!(snapshot_eviction_failures = 1, error_type = error_type);
```
fnichol, line 217: That's a really nice looking pattern. After we do this 3,4,5 times I bet there's some kind of pattern extraction (even if it's a macro) to make this almost a one-liner

fnichol (dismissed):
Yeah, fantastic work! I'm a touch curious on the shutdown semantics of the evictor task but I can follow up to make sure I understand. It looks to me like the detection and removal work could be aborted at any moment in a clean/transactional way and would naturally resume on process restart (implying to work is duplicated nor lost).

lib/forklift-server/src/server.rs
jhelwig: It is intended that it is completely safe to un-gracefully shut down the task at any point. Everything should be idempotent, and the metadata is the last thing cleaned up, so if it manages to delete the actual snapshot from PG, and send the "evict from memory" NATS message before it is killed, then it should find the metadata record that is pointing at a snapshot that's also not pointed at by any change sets and re-run the delete query, re-send the evict NATS message, and finally delete the metadata record.

fnichol (approved):

PR #7579, opened 2025-10-22
[feat] Azure: bring in publicIPAddress.id, other type fixes

jkeiser:
Right now you can't add a publicIPAddress to a networkInterface in Azure, because we aren't generating the ID. This fixes that, and also brings in the 13 schemas we couldn't bring in because of boolean schema types.

Because Azure reuses a *lot* of common definitions with allOf and $ref, 

* Deeper allOf handling:
  * treats `allOf` as `extend`, allowing { allOf: { ... }, properties { ... } } in the same. Recursively merges properties
  * Makes allOf, oneOf and anyOf merging strict
  * Merges `required` and `enum` fields
  * Brings in more child values with `allOf`, bringing in descriptions, readOnly information and merging properties
  * Delays normalization of type/format/etc. until after all allOf/oneOf/etc. alternatives have been processed
* Handles `true` (any) and `false` (never) JSON schemas consistently and more correctly
* Fixes situation where empty `properties` property in Azure was considered string instead of object

## How was it tested?

- [X] Integration tests pass

### DIFF
--- a/bin/clover/src/commands/generateSiSpecs.ts
+++ b/bin/clover/src/commands/generateSiSpecs.ts
@@ -18,6 +18,7 @@ import "../pipelines/dummy/spec.ts";
 
 const logger = _logger.ns("siSpecs").seal();
 const SI_SPEC_DIR = "si-specs";
+const MAX_SPEC_SIZE_MB = 20;
 
 export async function generateSiSpecs(options: PipelineOptions) {
   const specs: ExpandedPkgSpec[] = [];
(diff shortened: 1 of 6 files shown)


bin/clover/src/commands/generateSiSpecs.ts
```diff
@@ -38,10 +38,10 @@ export async function generateSiSpecs(options: PipelineOptions) {
...
       logger.debug(`Writing ${name}.json`);
       const blob = new Blob([specJson]);
-      if (blob.size > 4 * 1024 * 1024) {
-        logger.warn(`${spec.name} is bigger than 4MBs. Skipping.`);
-        continue;
-      }
+      // if (blob.size > 4 * 1024 * 1024) {
+      //   logger.warn(`${spec.name} is bigger than 4MBs. Skipping.`);
+      //   continue;
+      // }
```
sprutton1: Should we just delete this?

bin/clover/src/pipelines/azure/schema.ts
```diff
@@ -8,34 +8,46 @@ import type {
...
->[string] & {
-  schema?: JSONSchema;
-};
-
-/// Remove { $ref } from schema (which is what happens when you call SwaggerParser.Dereference)
-type Dereference<T> = DereferenceChildren<Exclude<T, { $ref: string }>>;
-/// Remove { $ref } recursively from all children of a schema (used by Dereference)
-type DereferenceChildren<T> = {
-  [K in keyof T]: Dereference<T[K]>;
-};
```
sprutton1, line 43: Curious why we no longer needs this?

bin/clover/src/pipelines/azure/spec.ts
```diff
@@ -82,42 +82,35 @@ export function parseAzureSpec(
...
+///
+/// Recursively normalizes child properties as well.
+export function normalizeAzureSchema(
+  prop: JSONSchema | undefined,
+  processing: JSONSchema[],
+): NormalizedAzureSchema | undefined {
+  if (prop === undefined) return undefined;
+  if (processing.includes(prop)) return undefined;
+  processing.push(prop);
+  // console.log(`Normalizing schema: ${util.inspect(processing, { depth: 3 })}`);
```
sprutton1: nit: cleanup

bin/clover/src/commands/generateSiSpecs.ts
jkeiser: I'm OK with it!

bin/clover/src/pipelines/azure/schema.ts
jkeiser: Now we have a direct declaration of the schema, which is the only place we *really* care about the $refs. There's a couple of other places where $refs can occur in a Swagger spec, and we assert and move on in those places (it doesn't happen in current stuff).

We might bring this back in the future, but when we *can* avoid a generic recursive type like this we may as well! They are awesome but a little mind-bendy :)

sprutton1 (approved):

PR #7581, opened 2025-10-22
Adding more instrumentation to mv_build_error

jobelenus:
## How was it tested?

Locally, I hit "rebuild index" I actually got an error
```
edda_server::materialized_view: name="mv_build_error" si.error.message="materialized views error: schema variant error: secret defining schema variant (01K6BNPVVK336M6V2QFEQZ6ZJ6) has no output sockets and needs one for the secret corresponding to its secret definition"
kind=SchemaVariant
id=01K6BNPVVK336M6V2QFEQZ6ZJ6
entity_id=01K6BNPVVK336M6V2QFEQZ6ZJ6
entity_kind=SchemaVariant
change_set_id=01JYPTEC5JM3T1Y4ECEPT9560J
workspace_id=01HRFEV0S23R1G23RP75QQDCA7
```

### DIFF
--- a/lib/edda-server/src/materialized_view/deployment.rs
+++ b/lib/edda-server/src/materialized_view/deployment.rs
@@ -29,7 +29,7 @@ use crate::{
     materialized_view::{
         BuildMvInnerReturn,
         BuildMvOp,
-        BuildMvTaskResult,
+        DeploymentBuildMvTaskResult,
         MaterializedViewError,
         MvBuilderResult,
         build_mv,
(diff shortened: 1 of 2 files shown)


lib/edda-server/src/materialized_view/deployment.rs
```diff
@@ -592,7 +593,7 @@ where
     )
     .await;
 
-    (mv_kind, mv_id, start.elapsed(), result)
+    (mv_kind, mv_id, start.elapsed(), None, None, result)
```
jhelwig: Curious about why these changes on the deployment MV side are here. Doesn't look like it actually changes anything aside from passing a couple of `None` back from the future that are completely unused?

jobelenus: It uses the same type that I added the entity_kind and entity_id to. I can just make them separate types if that's preferred / more expected?

jhelwig: Ah! Yeah, these look like they'll likely be diverging enough that having separate change set & deployment MV build task results probably makes sense.

jhelwig (approved):

PR #7582, opened 2025-10-22
feat(conduit): Allow push to update assets and funcs

vbustamante:

Allow for `si-conduit push` to detect existing schemas (same name), unlock them and modify them, along with unlocking and updating, deleting or adding any functions on the filesystem schema. 

Also adds auth func support to conduit 

#### Out of Scope:

Even if there are no changes to the schema or functions, schemas described on the filesystem will be unlocked on the changeset. This will be fixed in a further PR. Funcs that are exactly the same get skipped.

## How was it tested?

- [X] Regression test: Validate the we can still push net new assets and funcs
- [X] Manual test: Create an asset will all the scaffold funcs. Deleted an action, added a credential, updated codegen then ran push and all of it worked 
- [X] Manual test: Create a component from a schema, then update it and merge. Than, update it again to make sure that it unlocks the default schema, not the old one.

### DIFF
--- a/bin/si-conduit/README.md
+++ b/bin/si-conduit/README.md
@@ -326,7 +326,7 @@ Enable verbose logging to debug issues:
 
 ```bash
 # Maximum verbosity (trace level)
-si-conduit -vvvv schema scaffold generate MySchema
+si-conduit -v4 schema scaffold generate MySchema
 
 # Or specify a numeric level (0-4)
 si-conduit --verbose 4 schema scaffold generate MySchema
@@ -343,8 +343,8 @@ si-conduit --verbose 4 schema scaffold generate MySchema
 - [x] Push code generators
 - [x] Add PostHog analytics events
 - [x] Write comprehensive README
-- [ ] Handle existing function bindings for asset updates
-- [ ] Handle existing function names for new and updating assets
+- [x] Handle existing function bindings for asset updates
+- [x] Handle existing function names for new and updating assets
 
 ### Code Style
 
@@ -356,5 +356,5 @@ si-conduit --verbose 4 schema scaffold generate MySchema
 
 ## License
 
-See the main System Initiative repository for license information.
+See the root of the System Initiative repository for license information.
 
--- a/bin/si-conduit/deno.json
+++ b/bin/si-conduit/deno.json
@@ -1,7 +1,7 @@
 {
   "workspace": [],
   "tasks": {
-    "dev": "deno run --allow-net --allow-env --allow-read --allow-write main.ts",
+    "dev": "deno run --allow-net --allow-env --allow-read --env-file=.env.local --allow-write main.ts",
     "build": "deno compile --allow-net --allow-env --allow-read --allow-write main.ts",
     "lint": "deno lint",
     "test": "deno test"
@@ -15,7 +15,7 @@
     "@logtape/pretty": "jsr:@logtape/pretty@^1.1.0",
     "@logtape/redaction": "jsr:@logtape/redaction@^1.1.1",
     "@std/path": "https://deno.land/std@0.224.0/path/mod.ts",
-    "@systeminit/api-client": "jsr:@systeminit/api-client@^1.3.3",
+    "@systeminit/api-client": "jsr:@systeminit/api-client@^1.5.0",
     "axios": "npm:axios@1.11.0",
     "posthog-node": "npm:posthog-node@^4.2.1"
   },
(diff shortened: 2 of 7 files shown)


bin/si-conduit/src/cli/push-assets.ts
```diff
@@ -486,50 +528,143 @@ export async function pushAssets(
...
+              changeSetId,
+              funcId: unlockedFuncId,
+              updateFuncV1Request: filesystemFunc,
+            });
+          }
+        } else {
+          log.debug(`creating ${filesystemFuncKind} ${filesystemFunc.name}...`);
+          createdFuncs += 1;
+
+          switch (filesystemFuncKind) {
```
stack72, line 639: Are you missing Authentication in this switch?

vbustamante: Yeah - I skipped it since we were not creating  it on scaffold but it does make sense to add it

stack72 (dismissed):

bin/si-conduit/deno.json
```diff
@@ -6,7 +6,7 @@
   },
   "imports": {
     "@cliffy/command": "jsr:@cliffy/command@^1.0.0-rc.7",
-    "@systeminit/api-client": "jsr:@systeminit/api-client@^1.3.3",
+    "@systeminit/api-client": "jsr:@systeminit/api-client@^1.5.0",
```
fnichol, line 18: Good call!

bin/si-conduit/src/cli.ts
```diff
@@ -121,15 +121,12 @@ export async function run() {
...
     throw new Error("Workspace ID is required");
   }
 
   const authApi = new AuthApiClient(apiToken, workspaceId);
 
-  let workspace: WorkspaceDetails;
-  workspace = await authApi.getWorkspaceDetails();
-
-  return workspace;
+  return authApi.getWorkspaceDetails();
```
fnichol: I'm a little confused by this one, isn't [`AuthApiClient.getWorkspaceDetails()`](https://github.com/systeminit/si/blob/main/bin/si-conduit/src/auth-api-client.ts#L17) async?

bin/si-conduit/src/cli/push-assets.ts
fnichol: I can't lie, I know it's not the most fashionable to use `case` statements in TypeScript/JavaScript, but I still love 'em!

fnichol (approved):
This is looking pretty great! I think the only remaining task is to rebase on top of main, and sort out the merge conflicts. From a quick look, it seems like `push-assets.ts` isn't as bad of a merge conflict as I expected, so yay!

bin/si-conduit/src/cli.ts
vbustamante: @fnichol If a function returns a promise, it does not need the async keyword - that's only for functions that call await inside of them. In this case we can just call the promise generating method, which already executes the http request, and the caller can decide whether to await on it or not

fnichol: Ah that's right, you didn't change the return type, derp!

bin/si-conduit/README.md
```diff
@@ -326,7 +326,7 @@ Enable verbose logging to debug issues:
 
 ```bash
 # Maximum verbosity (trace level)
-si-conduit -vvvv schema scaffold generate MySchema
+si-conduit -v4 schema scaffold generate MySchema
```
fnichol, line 329: 👍🏽

bin/si-conduit/src/cli.ts
```diff
@@ -209,13 +209,14 @@ function buildRemoteCommand() {
...
-        .action(async ({ root }) => {
+        .option("-s, --skip-confirmation", "Skip confirmation prompt")
+        .action(async ({ root, skipConfirmation }) => {
           const project = createProject(root);
 
           const ctx = Context.instance();
           const cliContext = await initializeCliContextWithAuth({ ctx });
 
-          await pushAssets(cliContext, project);
+          await pushAssets(cliContext, project, skipConfirmation);
```
fnichol, line 219: Yeah..that was me forgetting, good catch

bin/si-conduit/src/cli/push-assets.ts
```diff
@@ -499,50 +583,139 @@ export async function pushAssets(
...
+                ...baseVariantPayload,
+                createVariantManagementFuncV1Request: filesystemFunc,
+              });
+              break;
+            case "Authentication":
+              await siSchemasApi.createVariantAuthentication({
+                ...baseVariantPayload,
+                createVariantAuthenticationFuncV1Request: filesystemFunc,
+              });
+              break;
```
fnichol, line 674: 👍🏽

PR #7590, opened 2025-10-23
better error handling for AttributePanel, fixed truncation bug

wendybujalski:

This PR improves the error handling in the `AttributePanel` in the following two ways -

- When a subscription fails due to incompatible types (for example, trying to subscribe an Array prop to a String value) the attributes endpoint now returns a 400 and the frontend displays a user-friendly message.
- When an error occurs in the `AttributePanel` the error is now cleared out when the user reselects that AV input.

Also included in this PR is a fix to a truncation bug in the headers of the `AttributePanel` found while working  on it. Part of that bug fix is an improvement to TruncateWithTooltip which makes it reactive to the browser window changing size.

#### Screenshots:

Screenshot 2025-10-22 at 8 01 57 PM

Truncation bug before being fixed -

https://github.com/user-attachments/assets/32596d85-9a08-49cf-8619-9b0735340fd8

#### Out of Scope:

All other errors in the `AttributePanel` besides a `SubscriptionTypeMismatch` still display the same. We may want to have more detailed error handling in the future.

We may want to revisit the styling of the error display or adjust the copy of the user-friendly error message for the 400.

## How was it tested?

For the changes to error handling, I exercised a variety of scenarios which both triggered the 400 error and ones that should not to ensure it behaved as expected.

For the truncation bug I tested every possible combination of elements in the header at a variety of window widths to ensure it displays as expected in all scenarios.

### DIFF
--- a/app/web/src/newhotness/AttributePanel.vue
+++ b/app/web/src/newhotness/AttributePanel.vue
@@ -338,12 +338,22 @@ const save = async (
   value = value?.toString() ?? "";
   const payload = makeSavePayload(path, value, propKind, connectingComponentId);
 
-  const { req, newChangeSetId } =
+  const { req, newChangeSetId, errorMessage } =
     await call.put<componentTypes.UpdateComponentAttributesArgs>(payload);
 
   const key = `${props.component.id}-${path}`;
   if (!saveApi.ok(req)) {
-    saveErrors.value[key] = value;
+    if (
+      req.status === 400 &&
+      errorMessage?.includes("Type mismatch on subscription")
+    ) {
+      // Bad request error due to type mismatch
+      saveErrors.value[key] = "Subscription failed due to type mismatch.";
+    } else {
+      // All other errors go here
+      saveErrors.value[key] = `\`${value}\` failed to save`;
+    }
+    removeSubscriptionMutation.mutate(path);
   } else {
     delete saveErrors.value[key];
   }
--- a/app/web/src/newhotness/layout_components/AttributeChildLayout.vue
+++ b/app/web/src/newhotness/layout_components/AttributeChildLayout.vue
@@ -30,7 +30,7 @@
     >
       <Icon
         :name="open ? 'chevron--down' : 'chevron--right'"
-        class="group-hover/header:scale-125"
+        class="group-hover/header:scale-125 flex-none"
       />
       <slot name="header" />
     </dt>
--- a/app/web/src/newhotness/layout_components/AttributeInput.vue
+++ b/app/web/src/newhotness/layout_components/AttributeInput.vue
@@ -244,7 +244,9 @@
         <span v-if="props.validation?.message">
           {{ props.validation.message }}
         </span>
-        <span v-else-if="hasError"> `{{ errorValue }}` failed to save </span>
+        <span v-else-if="hasError">
+          {{ errorValue }}
+        </span>
       </div>
 
       <!-- socket connections incompatibility message -->
@@ -367,7 +369,7 @@
               )
             "
           >
-            `{{ errorValue }}` failed to save
+            {{ errorValue }}
           </div>
 
           <!-- raw value selection area -->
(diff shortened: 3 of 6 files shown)


app/web/src/newhotness/layout_components/AttributeChildLayout.vue
```diff
@@ -30,7 +30,7 @@
     >
       <Icon
         :name="open ? 'chevron--down' : 'chevron--right'"
-        class="group-hover/header:scale-125"
+        class="group-hover/header:scale-125 flex-none"
```
wendybujalski, line 33: Prevents the chevron Icon from being squished by the flex container.

app/web/src/newhotness/layout_components/AttributeInput.vue
```diff
@@ -244,7 +244,9 @@
         <span v-if="props.validation?.message">
           {{ props.validation.message }}
         </span>
-        <span v-else-if="hasError"> `{{ errorValue }}` failed to save </span>
+        <span v-else-if="hasError">
+          {{ errorValue }}
```
wendybujalski, line 248: Moved this copy up to where the errorValue is built.

app/web/src/newhotness/layout_components/AttributeInput.vue
```diff
@@ -1255,6 +1257,9 @@ const resetEverything = () => {
   inputTouched.value = false;
   showAllPossibleConnections.value = false;
   cancelTabBehavior.value = false;
+  if (errorValue.value) {
+    delete errorContext.value.saveErrors.value[errorKey.value];
+  }
```
wendybujalski, line 1262: Clears the error when the input window is opened again.

app/web/src/newhotness/layout_components/ComponentAttribute.vue
wendybujalski, line 1: All of the changes in this file are just fixing the truncation bug on the headers.

app/web/src/newhotness/AttributePanel.vue
```diff
@@ -338,12 +338,22 @@ const save = async (
...
 
   const key = `${props.component.id}-${path}`;
   if (!saveApi.ok(req)) {
-    saveErrors.value[key] = value;
+    if (
+      req.status === 400 &&
+      errorMessage?.includes("Type mismatch on subscription")
+    ) {
+      // Bad request error due to type mismatch
+      saveErrors.value[key] = "Subscription failed due to type mismatch.";
```
wendybujalski, line 351: We may want to adjust this copy @britmyerss ?

lib/vue-lib/src/design-system/general/TruncateWithTooltip.vue
wendybujalski, line 1: This minor change makes the TruncateWithTooltip component respond appropriately (such as enabling/disabling the tooltip if necessary) when the browser window changes size.

lib/sdf-server/src/service/v2/component.rs
```diff
@@ -86,12 +81,13 @@ pub type Result<T> = std::result::Result<T, Error>;
...
 impl IntoResponse for Error {
     fn into_response(self) -> Response {
+        dbg!(&self);
         let status_code = match self {
-            Error::NoValueToSet(..) => StatusCode::BAD_REQUEST,
             Error::SchemaVariantUpgradeSkipped | Error::UpgradeSkippedDueToActions => {
                 StatusCode::NOT_MODIFIED
             }
             Error::AttributeValueNotFound(_, _) => StatusCode::NOT_FOUND,
+            Error::Attributes(AttributesError::AttributeValue(AttributeValueError::SubscriptionTypeMismatch { .. })) => StatusCode::BAD_REQUEST,
```
wendybujalski: This sets the status code for the error which occurs when a subscription fails due to a type mismatch to be 400.

lib/sdf-server/src/service/v2/component.rs
```diff
@@ -87,11 +89,13 @@ pub type Result<T> = std::result::Result<T, Error>;
 impl IntoResponse for Error {
     fn into_response(self) -> Response {
         let status_code = match self {
-            Error::NoValueToSet(..) => StatusCode::BAD_REQUEST,
```
britmyerss, line 90: is this case not being used? If it is, we can use 
```
match self {
...
Error::NoValueToSet(..) | Error::Attributes(........) => StatusCode::BAD_REQUEST
}
```

app/web/src/newhotness/AttributePanel.vue
britmyerss: let's check with @VictoriaMolgado here!

lib/sdf-server/src/service/v2/component.rs
jobelenus: Yea, it's not being used, we checked!

wendybujalski: Error::NoValueToSet is no longer an error used anywhere at all, so it was removed entirely!

PR #7591, opened 2025-10-23
feat(si-conduit): implement remote schema pull

fnichol:
This change extracts file system write operations from `generators.ts` into a new `materialize.ts` module and implements the remote schema pull command. The generator functions were handling both workflow orchestration and file I/O, which made them harder to test and reuse.

The new `materialize.ts` module provides functions for creating schema directories and writing metadata and code files to disk. Each function checks if files already exist before writing to prevent accidental overwrites, throwing a `FileExistsError` if a file is already present. The functions handle actions, code generators, management functions, and qualifications with a consistent pattern: check for existing files, create directories as needed, write content, and log what was created using project-relative paths.

The `remote schema pull` command in `src/command/remote/schema/pull.ts` uses the System Initiative API client to fetch schemas from a remote workspace and materialize them to the local file system. The command accepts schema names as arguments, retrieves the schema and default variant data from the API, fetches all attached function code and metadata, and writes everything to disk using the new materialization functions. A `SchemaVariantFuncs` helper class filters functions by kind since the API represents function kinds in a way that needs interpretation.

Generator functions in `src/command/schema/` have been refactored to use the materialization module rather than handling file writes directly. This removes duplication where each generator was implementing the same "create directory, check file exists, write, log" pattern. The generators now focus on preparing content and delegate file operations to the materialize module.

The `Project` class gained helper methods for constructing function metadata and code file paths, providing a single place where path construction logic lives. The `relativeToStr` method returns project-relative path strings for cleaner log output.

The CLI integration adds the `remote schema pull` command structure and ensures users are authenticated with a workspace selected before attempting to pull. Analytics tracking captures which schemas were pulled and which weren't found remotely.

### DIFF
--- a/bin/si-conduit/README.md
+++ b/bin/si-conduit/README.md
@@ -13,6 +13,7 @@ SI Conduit is a Deno-based CLI application that provides a structured workflow f
 - **Project Module** (`src/project.ts`): Project structure management and path utilities for working with schemas and their functions.
 - **Authentication** (`src/auth-api-client.ts`, `src/jwt.ts`): API authentication and JWT token handling for secure communication with System Initiative services.
 - **Generators** (`src/generators.ts`): Code generation utilities for scaffolding schemas and functions.
+- **Materialize Module** (`src/materialize.ts`): File system operations for creating and writing schema files, metadata, and function code to disk.
 
 ### Command Structure
 
@@ -27,7 +28,9 @@ si-conduit
 │   ├── qualification generate - Generate qualification functions
 │   └── scaffold generate    - Scaffold a complete schema structure
 ├── remote
-│   └── push                 - Push schemas to remote workspace
+│   └── schema
+│       ├── pull             - Pull schemas from remote workspace
+│       └── push             - Push schemas to remote workspace
 └── whoami                   - Display authenticated user information
 ```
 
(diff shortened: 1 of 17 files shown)


bin/si-conduit/src/command/remote/schema/pull.ts
```diff
@@ -0,0 +1,395 @@
...
+    }
+  }
+
+  ctx.analytics.trackEvent("remote_schema_pull", {
+    schemaNames: pulledSchemaNames,
+    notFoundSchemaNames,
+  });
+
+  // TODO(fnichol): If some schemas were not found, should we error out? At the
+  // moment we aren't
```
nickgerace, line 60: Perhaps, we should log the error for now.

bin/si-conduit/src/command/remote/schema/pull.ts
```diff
@@ -0,0 +1,395 @@
...
+
+  const data = await getSchemaAndVariantBySchemaName(
+    api,
+    changeSetCoord,
+    schemaName,
+  );
+
+  if (!data) {
+    // TODO(fnichol): log error, throw, or other here?
+    return undefined;
```
nickgerace, line 123: Throwing is probably the solid motion here.

bin/si-conduit/README.md
```diff
@@ -27,7 +28,9 @@ si-conduit
 │   ├── qualification generate - Generate qualification functions
 │   └── scaffold generate    - Scaffold a complete schema structure
 ├── remote
-│   └── push                 - Push schemas to remote workspace
+│   └── schema
+│       ├── pull             - Pull schemas from remote workspace
+│       └── push             - Push schemas to remote workspace
```
nickgerace, line 33: Love the README!

vbustamante (approved):

PR #7592, opened 2025-10-23
fix(dal): Send final ws event and update func run state on failure

britmyerss:
This change does a few things:
- On failure, ensure we send the WSEvent that the function was executed, so clients know (namely the front end, which was otherwise stuck thinking the function was still running)
- On failure, update the state of the func run, mimicking actions (and continuing to reuse the action status as we do in the successful case)
- Don’t swallow all errors and trace accordingly in honeycomb

<div><br/>via <a href="https://giphy.com/gifs/moodman-WnUfmomqsqmxpU4cJv">GIPHY</a></div>

### DIFF
--- a/lib/dal/src/management/prototype.rs
+++ b/lib/dal/src/management/prototype.rs
@@ -690,17 +690,33 @@ impl ManagementPrototype {
         manager_component_geometry: HashMap<String, ManagementGeometry>,
         managed_component_placeholders: HashMap<String, ComponentId>,
         func_run_channel: FuncRunnerValueChannel,
+        func_run_id: FuncRunId,
     ) -> ManagementPrototypeResult<ManagementPrototypeExecution> {
         let run_value = match func_run_channel.await {
             Ok(Err(FuncRunnerError::ResultFailure {
-                kind: _,
+                kind,
                 message,
                 backend: _,
-            })) => return Err(ManagementPrototypeError::FuncExecutionFailure(message)),
-            other => other.map_err(|_| ManagementPrototypeError::FuncRunnerRecvError)??,
+            })) => {
+                handle_failed_management_func(ctx, manager_component_id, id, func_run_id).await?;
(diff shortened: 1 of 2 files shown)


lib/dal/src/management/prototype.rs
```diff
@@ -690,17 +690,33 @@ impl ManagementPrototype {
...
+                handle_failed_management_func(ctx, manager_component_id, id, func_run_id).await?;
+                warn!(error=true,si.error.message=?other, si.func_run.id=?func_run_id, si.manager_component_id=?manager_component_id);
+                return Err(ManagementPrototypeError::FuncRunner(Box::new(other)));
+            }
+            Err(err) => {
+                handle_failed_management_func(ctx, manager_component_id, id, func_run_id).await?;
+                warn!(error=true, si.error.message=?err, si.func_run.id=?func_run_id, si.manager_component_id=?manager_component_id);
+                return Err(ManagementPrototypeError::FuncRunnerRecvError);
+            }
+            Ok(Ok(success)) => success,
```
nickgerace, line 717: This is an aside but I wonder how honeycomb interprets `?func_run_id` vs. `%func_run_id` since we can provide either or

lib/dal/src/management/prototype.rs
```diff
@@ -1032,6 +1049,23 @@ impl ManagementPrototype {
...
+    ctx: &DalContext,
+    manager_component_id: ComponentId,
+    id: ManagementPrototypeId,
+    func_run_id: FuncRunId,
+) -> Result<(), ManagementPrototypeError> {
+    FuncRunner::update_run(ctx, func_run_id, |func_run| {
+        func_run.set_action_result_state(Some(si_events::ActionResultState::Failure));
+    })
+    .await?;
+    WsEvent::management_func_executed(ctx, id, manager_component_id, func_run_id)
```
nickgerace, line 1062: Should we also use one of the `WsEvent::management_operations*` three events too?

zacharyhamm: This is before operations have begun so i think no

PR #7600, opened 2025-10-24
feat: add create overlay action and management funcs on builtins

aaron-dernley:

Gives our MCP server the ability to create overlay functions for action and management types on builtin schemas.

#### Screenshots:

Screenshot 2025-10-24 at 16 35 56

#### Out of Scope:

Agreed out of scope was the returning the contents of a function when requested via MCP that is listed in the Linear - https://linear.app/system-initiative/issue/ENG-3267/allow-users-to-author-action-mgmt-overlays-using-the-ai-agent

## How was it tested?

- [X] Integration tests pass
- [X] Manual test: new functionality works via claude - create new overlay management func, create new overlay action func, cannot add an extra overlay func for either action or management funcs
- [X] Manual test: (regression check) authoring still works as expected following the previously implemented path.

### DIFF
--- a/bin/si-mcp-server/deno.json
+++ b/bin/si-mcp-server/deno.json
@@ -12,7 +12,7 @@
     "@onjara/optic": "jsr:@onjara/optic@^2.0.3",
     "@std/assert": "jsr:@std/assert@1",
     "@std/encoding": "jsr:@std/encoding@^1.0.10",
-    "@systeminit/api-client": "jsr:@systeminit/api-client@^1.4.0",
+    "@systeminit/api-client": "jsr:@systeminit/api-client@^1.6.1",
     "axios": "npm:axios@^1.6.1",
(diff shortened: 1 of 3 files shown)


wendybujalski (dismissed):

bin/si-mcp-server/deno.json
```diff
@@ -12,7 +12,7 @@
     "@onjara/optic": "jsr:@onjara/optic@^2.0.3",
     "@std/assert": "jsr:@std/assert@1",
     "@std/encoding": "jsr:@std/encoding@^1.0.10",
-    "@systeminit/api-client": "jsr:@systeminit/api-client@^1.4.0",
+    "@systeminit/api-client": "jsr:@systeminit/api-client@^1.6.1",
```
wendybujalski, line 15: This is the version which adds the `installSchema` endpoint, which we need to be able to ensure the schema the function will be on is installed without unlocking that schema.

PR #7604, opened 2025-10-24
Fallback to plain text in func run viewer when the logs are too large

jobelenus:

Rendering the CodeViewer with such large logs is freezing the main thread of the browser

#### Screenshots:

image

### DIFF
--- a/app/web/src/newhotness/FuncRunDetails.vue
+++ b/app/web/src/newhotness/FuncRunDetails.vue
@@ -7,6 +7,7 @@
     :funcRun="funcRun"
     :status="funcRunStatus(funcRun, managementFuncJobState?.state) || ''"
     :logText="logText"
+    :logsTooLarge="logsTooLarge"
     :errorHint="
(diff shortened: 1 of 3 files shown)


app/web/src/newhotness/FuncRunDetails.vue
```diff
@@ -281,6 +282,10 @@ const { data: funcRunLogsQuery } = useQuery<FuncRunLog | undefined>({
 
 const funcRunLogs = computed(() => funcRunLogsQuery.value);
 
+const logsTooLarge = computed(
+  () => (funcRunLogs.value?.logs.length ?? 0) > 175,
```
britmyerss, line 286: any science behind this number?

jobelenus: Nope! The one breaking in prod is 209. Just gonna have to iterate a bit here

britmyerss (approved):

PR #7605, opened 2025-10-24
feat(clover): discover discriminants and put them in the domain

sprutton1:

Azure schemas can contain `discriminants` where an enum prop may map to a complicated object. 

```
discriminant: "kind",
properties : {
  "kind": {
    "type": "string",
    "description": "Type of the script.",
    "enum": [
      "AzurePowerShell",
      "AzureCLI"
    ]
  },
}
```
Elsewhere, in the definitions, you'll find the matching discriminators.

```
 "AzurePowerShellScript": {
      "type": "object",
      "x-ms-discriminator-value": "AzurePowerShell",
      "description": "Object model for the Azure PowerShell script.",
      "properties": {
        "properties": {
          "description": "Properties of the Azure PowerShell script object.",
          "$ref": "#/definitions/AzurePowerShellScriptProperties",
          "x-ms-client-flatten": true
        }
      },
    }
...
```

So we need to detect when discriminants exist, gather all of the relevant objects, and put them in the domain. Because the UI does not support either/or style displays, all of them will be present.

It's important to note that discriminators may be of entirely different shapes or have the same props of different types, so we can't just smash the entirety of them into the prop tree or attempt to gather them using `anyOf`.

After we add the objects to the domain, we also keep a record of `dicscriminant:discriminators` in the `PropUsageMap` so we can:

1. Detect if multiple are filled out. This should error
2. Merge the properties of the filled out ones into the request payload

#### Screenshots:

Successful create with domain before and after merging
```
[Fri, 24 Oct 2025 19:13:14 GMT] [info] Running CLI command: "az 'login' '--service-principal' '--username' '[redacted]' '--password' '[redacted]' '--tenant' '[redacted]'"
[Fri, 24 Oct 2025 19:13:17 GMT] [info] Registering environment variable AZURE_TENANT_ID
[Fri, 24 Oct 2025 19:13:17 GMT] [info] Registering environment variable AZURE_CLIENT_ID
[Fri, 24 Oct 2025 19:13:17 GMT] [info] Registering environment variable AZURE_CLIENT_SECRET
[Fri, 24 Oct 2025 19:13:18 GMT] [info] {
[Fri, 24 Oct 2025 19:13:18 GMT] [info]   kind: "AzurePowerShell",
[Fri, 24 Oct 2025 19:13:18 GMT] [info]   location: "eastus",
[Fri, 24 Oct 2025 19:13:18 GMT] [info]   name: "test",
[Fri, 24 Oct 2025 19:13:18 GMT] [info]   subscriptionId: "2cd357b4-5d4f-471b-ac9c-cd4f2a4ccd07",
[Fri, 24 Oct 2025 19:13:18 GMT] [info]   resourceGroup: "clover",
[Fri, 24 Oct 2025 19:13:18 GMT] [info]   AzurePowerShellScript: {
[Fri, 24 Oct 2025 19:13:18 GMT] [info]     kind: "AzurePowerShell",
[Fri, 24 Oct 2025 19:13:18 GMT] [info]     properties: {
[Fri, 24 Oct 2025 19:13:18 GMT] [info]       azPowerShellVersion: "7.4",
[Fri, 24 Oct 2025 19:13:18 GMT] [info]       retentionInterval: "P1D",
[Fri, 24 Oct 2025 19:13:18 GMT] [info]       scriptContent: "test"
[Fri, 24 Oct 2025 19:13:18 GMT] [info]     }
[Fri, 24 Oct 2025 19:13:18 GMT] [info]   },
[Fri, 24 Oct 2025 19:13:18 GMT] [info]   extra: {
[Fri, 24 Oct 2025 19:13:18 GMT] [info]     apiVersion: "2023-08-01",
[Fri, 24 Oct 2025 19:13:18 GMT] [info]     AzureResourceType: "Azure::Resources::DeploymentScripts",
[Fri, 24 Oct 2025 19:13:18 GMT] [info]     PropUsageMap: '{"createOnly":["subscriptionId","resourceGroup","name"],"updatable":["AzureCliScript","AzurePowerShellScript","kind","tags","location","identity","identity","location","tags","kind","properties","identity","location","tags","kind","properties","type","userAssignedIdentities","type","userAssignedIdentities","containerSettings","storageAccountSettings","cleanupPreference","primaryScriptUri","supportingScriptUris","scriptContent","arguments","environmentVariables","forceUpdateTag","retentionInterval","timeout","azCliVersion","type","userAssignedIdentities","containerSettings","storageAccountSettings","cleanupPreference","primaryScriptUri","supportingScriptUris","scriptContent","arguments","environmentVariables","forceUpdateTag","retentionInterval","timeout","azPowerShellVersion","containerGroupName","subnetIds","storageAccountName","storageAccountKey","containerGroupName","subnetIds","storageAccountName","storageAccountKey"],"discriminators":{"kind":["AzurePowerShellScript","AzureCliScript"]}}'
[Fri, 24 Oct 2025 19:13:18 GMT] [info]   }
[Fri, 24 Oct 2025 19:13:18 GMT] [info] }
[Fri, 24 Oct 2025 19:13:18 GMT] [info] {
[Fri, 24 Oct 2025 19:13:18 GMT] [info]   kind: "AzurePowerShell",
[Fri, 24 Oct 2025 19:13:18 GMT] [info]   location: "eastus",
[Fri, 24 Oct 2025 19:13:18 GMT] [info]   properties: {
[Fri, 24 Oct 2025 19:13:18 GMT] [info]     azPowerShellVersion: "7.4",
[Fri, 24 Oct 2025 19:13:18 GMT] [info]     retentionInterval: "P1D",
[Fri, 24 Oct 2025 19:13:18 GMT] [info]     scriptContent: "test"
[Fri, 24 Oct 2025 19:13:18 GMT] [info]   }
[Fri, 24 Oct 2025 19:13:18 GMT] [info] }
[Fri, 24 Oct 2025 19:13:19 GMT] [info] Output: {
  "protocol": "result",
  "status": "success",
  "executionId": "01K8BT7BQ4GXZR21VANTNHFXG6",
  "resourceId": "/subscriptions/2cd357b4-5d4f-471b-ac9c-cd4f2a4ccd07/resourceGroups/clover/providers/Microsoft.Resources/deploymentScripts/test",
  "payload": {
    "kind": "AzurePowerShell",
    "location": "eastus",
    "properties": {
      "provisioningState": "Creating",
      "azPowerShellVersion": "7.4",
      "scriptContent": "test",
      "retentionInterval": "P1D",
      "timeout": "P1D",
      "containerSettings": {},
      "status": {},
      "cleanupPreference": "Always"
    },
    "systemData": {
      "createdBy": "[redacted]",
      "createdByType": "Application",
      "createdAt": "2025-10-24T19:13:19.4392268Z",
      "lastModifiedBy": "[redacted]",
      "lastModifiedByType": "Application",
      "lastModifiedAt": "2025-10-24T19:13:19.4392268Z"
    },
    "id": "/subscriptions/2cd357b4-5d4f-471b-ac9c-cd4f2a4ccd07/resourceGroups/clover/providers/Microsoft.Resources/deploymentScripts/test",
    "type": "Microsoft.Resources/deploymentScripts",
    "name": "test"
  },
```

Showing the failure case if multiples are filled out: 

```
[Fri, 24 Oct 2025 19:24:00 GMT] [info] Output: {
  "protocol": "result",
  "status": "success",
  "executionId": "01K8BTTZ3FBDKF3KQ7CRA5ZYRQ",
  "payload": null,
  "health": "error",
  "message": "Multiple discriminator subtypes filled for \"kind\": AzurePowerShellScript, AzureCliScript. Only one should be filled."
}
```

## How was it tested?

- [X] Integration tests pass

### DIFF
--- a/bin/clover/src/pipelines/azure/pipeline-steps/addDefaultProps.ts
+++ b/bin/clover/src/pipelines/azure/pipeline-steps/addDefaultProps.ts
@@ -2,6 +2,7 @@ import _ from "lodash";
 import { ExpandedPkgSpec } from "../../../spec/pkgs.ts";
 import {
   addPropSuggestSource,
+  bfsPropTree,
   createObjectProp,
   createScalarProp,
   ExpandedPropSpec,
@@ -12,6 +13,14 @@ import { AzureSchema } from "../schema.ts";
 export interface PropUsageMap {
   createOnly: string[];
   updatable: string[];
+  // Maps discriminator property -> subtype definition name -> enum value
+  // e.g., { "kind": { "AzurePowerShellScript": "AzurePowerShell", "AzureCliScript": "AzureCLI" } }
+  discriminators: Record<string, Record<string, string>> | undefined;
+  // Maps discriminator property -> subtype definition name -> list of properties defined by that subtype
+  // e.g., { "kind": { "AzurePowerShellScript": ["properties"], "AzureCliScript": ["properties"] } }
+  discriminatorSubtypeProps:
+    | Record<string, Record<string, string[]>>
+    | undefined;
 }
 
 export function addDefaultProps(
@@ -143,7 +152,49 @@ export function addDefaultProps(
       const propUsageMap: PropUsageMap = {
         createOnly: [],
         updatable: [],
+        discriminators: {},
+        discriminatorSubtypeProps: {},
       };
+
+      const azureSchema = schemaVariant.superSchema as AzureSchema;
+      propUsageMap.discriminators = azureSchema.discriminators;
+
+      // Derive discriminatorSubtypeProps by walking the domain schema structure
+      const discriminatorSubtypePropsMap: Record<
+        string,
+        Record<string, string[]>
+      > = {};
+
+      if (azureSchema.discriminators) {
+        for (
+          const [discriminatorProp, subtypeMap] of Object.entries(
+            azureSchema.discriminators,
+          )
+        ) {
+          const domainProp = findPropByName(domain, discriminatorProp);
+
+          if (domainProp && domainProp.kind === "object") {
+            discriminatorSubtypePropsMap[discriminatorProp] = {};
+
+            for (const subtypeName of Object.keys(subtypeMap)) {
+              const subtypeProp = domainProp.entries.find((e) =>
+                e.name === subtypeName
+              );
+
+              if (subtypeProp && subtypeProp.kind === "object") {
+                const allPropNames: string[] = [];
+                bfsPropTree(subtypeProp.entries, (prop) => {
+                  allPropNames.push(prop.name);
+                });
+                discriminatorSubtypePropsMap[discriminatorProp][subtypeName] =
+                  allPropNames;
+              }
+            }
+          }
+        }
+      }
+
+      propUsageMap.discriminatorSubtypeProps = discriminatorSubtypePropsMap;
 
       const queue: ExpandedPropSpec[] = _.cloneDeep(domain.entries);
 
--- a/bin/clover/src/pipelines/azure/spec.ts
+++ b/bin/clover/src/pipelines/azure/spec.ts
@@ -3,10 +3,12 @@ import util from "node:util";
 import logger from "../../logger.ts";
 import { CfHandler, CfHandlerKind } from "../types.ts";
 import {
-  AzureSchema,
-  AzureProperty,
+  AzureDefinitions,
   AzureOpenApiDocument,
   AzureOpenApiOperation,
+  AzureProperty,
+  AzureSchema,
+  AzureSchemaDefinition,
   NormalizedAzureSchema,
 } from "./schema.ts";
 import { JSONSchema } from "../draft_07.ts";
@@ -22,8 +24,11 @@ const MAX_RESOURCE_DEPTH = 3;
 /// Make sure you also add a comment explaining why--all schemas should be supported!
 const IGNORE_RESOURCE_TYPES = new Set<string>([
   // GET endpoint returns an array
-  "Microsoft.PowerBI/privateLinkServicesForPowerBI"
-
+  "Microsoft.PowerBI/privateLinkServicesForPowerBI",
+  // Discriminator subtypes have properties with no type field (fluentdConfigUrl has description but no type)
+  "Azure::ServiceFabricMesh::Applications",
+  // Discriminator subtypes have properties with no type field (jobStageDetails has no type and no allOf)
+  "Azure::DataBox::Jobs",
 ]);
 
 export function parseAzureSpec(
@@ -50,10 +55,13 @@ export function parseAzureSpec(
   for (const [path, methods] of Object.entries(openApiDoc.paths)) {
     const pathInfo = parseEndpointPath(path);
     if (!pathInfo || !methods) continue;
-    const resourceType = `${pathInfo.resourceProvider}/${pathInfo.resourceType}`;
+    const resourceType =
+      `${pathInfo.resourceProvider}/${pathInfo.resourceType}`;
 
     // Presently we only support Microsoft. providers
-    if (!pathInfo.resourceProvider.toLowerCase().startsWith("microsoft.")) continue;
+    if (!pathInfo.resourceProvider.toLowerCase().startsWith("microsoft.")) {
+      continue;
+    }
     // Ignore certain problematic resource types (temporarily until we fix them)
     if (IGNORE_RESOURCE_TYPES.has(resourceType)) continue;
 
@@ -100,6 +108,7 @@ export function parseAzureSpec(
       resource.putOperation,
       resource.handlers,
       openApiDoc.info.version,
+      openApiDoc.definitions,
     );
     if (spec) {
       specs.push(spec);
@@ -121,15 +130,23 @@ export function parseAzureSpec(
 ///
 /// Recursively normalizes child properties as well.
 export function normalizeAzureSchema(
-  prop: JSONSchema | undefined,
-  processing: JSONSchema[],
+  prop: AzureSchemaDefinition | undefined,
+  processing: AzureSchemaDefinition[],
+  definitions?: AzureDefinitions,
+  discriminatorCollector?: Record<string, Record<string, string>>,
 ): NormalizedAzureSchema | undefined {
   if (prop === undefined) return undefined;
   if (processing.includes(prop)) return undefined;
   processing.push(prop);
   try {
     // Flatten the schema, merging allOf props and such, before we normalize type/format
-    const normalized = flattenAzureSchema(prop, {}, processing);
+    const normalized = flattenAzureSchema(
+      prop,
+      {},
+      processing,
+      definitions,
+      discriminatorCollector,
+    );
     if (normalized === undefined) return undefined;
 
     // Infer type from format / properties / items if missing
@@ -179,6 +196,13 @@ export function normalizeAzureSchema(
       }
     }
 
+    // At the root level, attach collected discriminators
+    if (
+      discriminatorCollector && Object.keys(discriminatorCollector).length > 0
+    ) {
+      normalized.discriminators = discriminatorCollector;
+    }
+
     return normalized;
   } finally {
     processing.pop();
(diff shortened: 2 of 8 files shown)


bin/clover/src/pipelines/azure/spec.ts
```diff
@@ -600,17 +625,107 @@ function extractPropertiesFromResponseBody(
...
+): AzureSchemaDefinition | undefined {
+  if (!definitions || typeof schema !== "object") {
+    return schema;
+  }
+  if (processing.includes(schema)) return schema;
+  processing.push(schema);
+
+  try {
+    // Build discriminator map
+    if (discriminatorMap.length === 0) {
```
jkeiser: Can we build the discriminator map outside? Especially if it's a schema with no discriminators, this is going to iterate over every definition on every single recursive property. `O(# props * # definitions)` -- n^2 kinda

bin/clover/src/pipelines/azure/spec.ts
```diff
@@ -600,17 +625,107 @@ function extractPropertiesFromResponseBody(
...
+        )
+      }`,
     );
   }
   return result;
 }
 
+/// Applies discriminator handling by finding subtypes and adding them as new sub-objects.
+/// Recursively processes the schema tree to find and expand all discriminated types.
+function applyDiscriminators(
```
jkeiser: I bet most of this method would go away if we did this inside the normalizer--populate the definition map before doing any normalizing, then let the existing recursion do its work and check for `discriminator` as we normalize a prop.

Not a necessity even if it's 100% true though--I think this will work, and we don't necessarily need access to the more advanced stuff the normalizer does (like unions) because you're nesting each object into its own sub-object, so merging (union/intersect) isn't an issue.

bin/clover/src/pipelines/azure/spec.ts
```diff
@@ -600,17 +625,107 @@ function extractPropertiesFromResponseBody(
...
+      const discriminatorField = schema.properties?.[discriminatorProp];
+      const enumValues = discriminatorField?.enum;
+
+      if (enumValues) {
+        // Filter discriminatorMap to only include subtypes whose x-ms-discriminator-value
+        // matches one of the enum values
+        for (const [defName, defSchema] of discriminatorMap) {
+          const discriminatorValue = defSchema["x-ms-discriminator-value"];
+          if (discriminatorValue && enumValues.includes(discriminatorValue)) {
+            subtypeEntries.push([defName, defSchema]);
```
jkeiser: Would it be easier for the user if the sub-prop had the same name as the discriminant? It feels like it might, though I haven't seen as many examples yet as you have :)

bin/clover/src/pipelines/azure/spec.ts
```diff
@@ -600,17 +625,107 @@ function extractPropertiesFromResponseBody(
...
   return result;
 }
 
+/// Applies discriminator handling by finding subtypes and adding them as new sub-objects.
+/// Recursively processes the schema tree to find and expand all discriminated types.
+function applyDiscriminators(
+  schema: AzureSchemaDefinition,
+  definitions: AzureDefinitions | undefined,
+  processing: AzureSchemaDefinition[],
+  discriminatorMap: Array<[string, AzureSchemaDefinition]>,
```
jkeiser: If this were Record<string, AzureSchemaDefinition> (keyed by x-ms-discriminator-value) you could do faster lookup instead of having to iterate over the array every time.

bin/clover/src/pipelines/azure/spec.ts
sprutton1: You are probably right, but I was wary of further complicating the logic there. If it's the right call to do that, we can.

bin/clover/src/pipelines/azure/spec.ts
sprutton1: we pass the map through and only rebuild it if it's empty, so it should only happen once per schema

bin/clover/src/pipelines/azure/spec.ts
```diff
@@ -600,17 +625,107 @@ function extractPropertiesFromResponseBody(
...
+    }
+
+    // If we found subtypes, add them as separate properties so the user can fill in one
+    if (subtypeEntries.length > 0 && schema.discriminator) {
+      const discriminatorProp = schema.discriminator;
+      const newProperties = { ...schema.properties };
+      const subtypeNames: string[] = [];
+
+      for (const [defName, subtype] of subtypeEntries) {
+        newProperties[defName] = subtype;
```
jkeiser: We should check whether newProperties already has something with that name and throw if so--we don't want to silently overwrite one just because there was a real prop with a similar name to what we chose.

bin/clover/src/pipelines/azure/spec.ts
sprutton1: it ends up pretty clean in the ui as-is. 
image

bin/clover/src/pipelines/azure/spec.ts
jkeiser: If there are no discriminators in that openapi spec, it'll be empty the whole time, right? Then on each prop we'll re-run through all the /definitions again unnecessarily.

bin/clover/src/pipelines/azure/spec.ts
jkeiser: Nah, I think it's not something to block on unless you were already rebuilding the logic (the normalizer has a lot of subtle constraints and takes time to grok, as much as I tried to make it simple).

bin/clover/src/pipelines/azure/spec.ts
jkeiser: Yeah, it seems good in this case. It may be worth looking at other discriminators and deciding whether the pattern works there as well; since MS doesn't expose the definition name to its actual API, some of them might be less nice and normal :)

bin/clover/src/pipelines/azure/spec.ts
sprutton1: ah, good point.

bin/clover/src/pipelines/azure/pipeline-steps/addDefaultProps.ts
```diff
@@ -143,8 +151,46 @@ export function addDefaultProps(
...
+        Record<string, string[]>
+      > = {};
+
+      if (azureSchema.discriminators) {
+        for (
+          const [discriminatorProp, subtypeMap] of Object.entries(
+            azureSchema.discriminators,
+          )
+        ) {
+          const domainProp = findPropByName(domain, discriminatorProp);
```
jkeiser, line 174: Are 100% of these things direct children of the top-level `/domain`? Or can they be nested under sub-objects or arrays/maps?

jkeiser: (I guess as long as we only support the top-level ones for now, this is fine!)

bin/clover/src/pipelines/azure/spec.ts
```diff
@@ -125,15 +134,23 @@ export function parseAzureSpec(
...
   if (prop === undefined) return undefined;
   if (processing.includes(prop)) return undefined;
   processing.push(prop);
   try {
     // Flatten the schema, merging allOf props and such, before we normalize type/format
-    const normalized = flattenAzureSchema(prop, {}, processing);
+    const normalized = flattenAzureSchema(
+      prop,
+      {},
+      processing,
```
jkeiser, line 146: oy, at some point we need to make the flattener an object with all this crap in it :) (Not right now)

bin/clover/src/pipelines/azure/spec.ts
```diff
@@ -690,8 +950,8 @@ function parseEndpointPath(path: string) {
   if (!!resourceGroupParam !== !!resourceNameParam) return undefined;
 
   const serviceName = providerNamespace.split(".").pop() || providerNamespace;
-  const capitalizedResource =
-    resourceType.charAt(0).toUpperCase() + resourceType.slice(1);
+  const capitalizedResource = resourceType.charAt(0).toUpperCase() +
```
jkeiser: I'm not sure, but I'm starting to worry you and I have dueling typescript formatters :)

bin/clover/src/pipelines/azure/pipeline-steps/addDefaultProps.ts
sprutton1: No, they can be nested. Just an oversight, I'll walk the tree here.

bin/clover/src/pipelines/azure/spec.ts
sprutton1: Yep. It's a format-off

PR #7607, opened 2025-10-24
fix: can't update any schema due to an unrelated subscription cycle

jkeiser:
When there is a self-subscription or a cycle of subscriptions, you can't update any attribute function bindings anywhere. There are multiple other endpoints (many of them schema endpoints) that also fail.

The problem is these endpoints enable cycle checking, which will throw if there is *any* cycle in the graph, not just if the edges being added create cycles. But we allow certain graph cycles due to subscriptions: for example, Foo:/domain/A -> Bar:/domain/B -> Foo:/domain/C is not a cycle in practice, but because each subscription points at the component root and has a string path, there is a cycle in the *graph* (Foo:/domain/A -> Bar (+ "/domain/B" string path), Bar:/domain/B -> Foo (+ "/domain/C" string path)).

This will get worse in Azure, where self-subscriptions can actually be desirable.

This PR narrows down our cycle checks, so they only throw if *the edge you are adding* would create a cycle.

There are ways we could eliminate these cycles entirely, but narrowly-focused errors (and code that doesn't topologically sort the whole graph) are probably desirable either way.

## How was it tested?

- [X] Integration tests pass
- [X] New tests
- [ ] Manual test: create subscription cycle and update an asset

### DIFF
--- a/lib/dal/src/workspace_snapshot/graph/v4.rs
+++ b/lib/dal/src/workspace_snapshot/graph/v4.rs
@@ -305,15 +305,66 @@ impl WorkspaceSnapshotGraphV4 {
         self.get_node_index_by_id(node_id)
     }
 
-    fn add_edge_inner(
+    fn check_would_create_cycle(
+        &self,
+        from_node_index: NodeIndex,
+        to_node_index: NodeIndex,
+    ) -> WorkspaceSnapshotGraphResult<()> {
+        // If there is already an edge from a to b, it's impossible to create a new cycle
+        if self
+            .graph
+            .find_edge(from_node_index, to_node_index)
+            .is_some()
+        {
+            return Ok(());
+        }
+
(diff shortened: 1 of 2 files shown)


jkeiser:
@zacharyhamm I think you're right that the O() is the same, since the worst case is you have to hit every edge on the graph ... but unless it's implemented in a silly way, has_path_connecting(from, to) should iterate through many fewer edges than toposort(entire graph) which *must* hit every edge.

zacharyhamm:
In the normal case, namely that there is *no* path connecting, you can't learn that until you've checked every single edge :)

jkeiser:
Depends what you mean by "all edges," and how it's implemented :) For example, a depth-first search starting at A has to follow all edges/paths *from A,* but won't look at all the edges in the *graph.*

PR #7608, opened 2025-10-24
fix: Thread overlays through to components

britmyerss:
This change does the following: 
- When building the ActionPrototypeViewList for a variant, return overlays if they exist (this allows the front end to see them!)
- When fetching components from luminork, also return the overlays if they exist (so the agent knows about them)
- When upgrading a component, if there is an existing variant scoped action enqueued (Create/Update/Refresh/Destroy), swap it for the overlay if it exists

#### Out of Scope:

I didn't go as far as hydrating which functions are overlays or not - this will come later
## How was it tested?

- [X] New Integration test for this use case, including component upgrade

<div><br/>via <a href="https://giphy.com/heyarnold/">Hey Arnold</a> on <a href="https://giphy.com/gifs/heyarnold-hey-arnold-nicksplat-l1JoihoOA6VxIQiD6">GIPHY</a></div>

### DIFF
--- a/lib/dal/src/action/prototype.rs
+++ b/lib/dal/src/action/prototype.rs
@@ -563,6 +563,43 @@ impl ActionPrototype {
         Ok(prototypes)
     }
 
+    /// Find the all action prototypes for a given variant. If a prototype is one of the
+    /// unique ones (`ActionKind::Create`, `ActionKind::Update`,
+    /// `ActionKind::Refresh`, `ActionKind::Destroy`), then look first for the
+    /// version defined at the schema level, and only look for the variant level
+    /// if no schema level prototype is found. For `ActionKind::Manual`, find
+    /// prototype at both levels.
+    /// Return all valid prototypes for the variant
+    pub async fn list_for_schema_and_variant_id(
+        ctx: &DalContext,
+        schema_variant_id: SchemaVariantId,
+    ) -> ActionPrototypeResult<Vec<Self>> {
+        let mut prototypes = vec![];
+        // grab all the different kinds, handling overlays if needed
+        let create_actions =
+            Self::find_by_kind_for_schema_or_variant(ctx, ActionKind::Create, schema_variant_id)
+                .await?;
+        prototypes.extend(create_actions);
+        let update_actions =
+            Self::find_by_kind_for_schema_or_variant(ctx, ActionKind::Update, schema_variant_id)
+                .await?;
+        prototypes.extend(update_actions);
+
+        let refresh_actions =
+            Self::find_by_kind_for_schema_or_variant(ctx, ActionKind::Refresh, schema_variant_id)
+                .await?;
+        prototypes.extend(refresh_actions);
+        let destroy_actions =
+            Self::find_by_kind_for_schema_or_variant(ctx, ActionKind::Destroy, schema_variant_id)
+                .await?;
+        prototypes.extend(destroy_actions);
+        let manual_actions =
+            Self::find_by_kind_for_schema_or_variant(ctx, ActionKind::Manual, schema_variant_id)
+                .await?;
+        prototypes.extend(manual_actions);
+        Ok(prototypes)
+    }
+
     pub async fn run(
         ctx: &DalContext,
         id: ActionPrototypeId,
--- a/lib/dal/src/component.rs
+++ b/lib/dal/src/component.rs
@@ -2546,10 +2546,12 @@ impl Component {
 
         let queued_for_old_component = Action::find_for_component_id(ctx, old_component_id).await?;
         let available_for_new_component =
-            ActionPrototype::for_variant(ctx, new_schema_variant_id).await?;
+            ActionPrototype::list_for_schema_and_variant_id(ctx, new_schema_variant_id).await?;
         for existing_queued in queued_for_old_component {
             let action = Action::get_by_id(ctx, existing_queued).await?;
             let action_prototype_id = Action::prototype_id(ctx, existing_queued).await?;
+            let existing_queued_prototype =
+                ActionPrototype::get_by_id(ctx, action_prototype_id).await?;
             // what do we do about the various states?
             // maybe you shouldn't upgrade a component if an action
             // is dispatched or running for the current?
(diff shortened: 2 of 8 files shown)


lib/dal/tests/integration_test/materialized_views.rs
```diff
@@ -57,11 +57,12 @@ async fn actions(ctx: &DalContext) -> Result<()> {
     let schema_variant_id = Component::schema_variant_id(ctx, component.id()).await?;
 
     // Gather what we need for the assertions after the component has been created.
-    let create_action_prototype = ActionPrototype::for_variant(ctx, schema_variant_id)
-        .await?
-        .into_iter()
-        .find(|ap| ap.kind == dal::action::prototype::ActionKind::Create)
-        .ok_or_eyre("could not find action prototype")?;
+    let create_action_prototype =
+        ActionPrototype::list_for_schema_and_variant_id(ctx, schema_variant_id)
```
jobelenus, line 61: ❤️ how tidy this change is! Are there any other spots where we need to swap these fn calls?

lib/dal/src/action/prototype.rs
```diff
@@ -563,6 +563,43 @@ impl ActionPrototype {
...
     }
 
+    /// Find the all action prototypes for a given variant. If a prototype is one of the
+    /// unique ones (`ActionKind::Create`, `ActionKind::Update`,
+    /// `ActionKind::Refresh`, `ActionKind::Destroy`), then look first for the
+    /// version defined at the schema level, and only look for the variant level
+    /// if no schema level prototype is found. For `ActionKind::Manual`, find
+    /// prototype at both levels.
+    /// Return all valid prototypes for the variant
+    pub async fn list_for_schema_and_variant_id(
```
nickgerace, line 573: My one nit here is that usually our pattern for method naming is `<verb>_for_<parameters>`, but I think this is arguable that the resulting list returns the items corresponding to both the schema and the variant, despite "only" the variant being the parameter.

Maybe we can iterate on this naming? Just an idea.

Also perhaps we should trace this too? Not sure how "loud" it would be.

lib/dal/src/component.rs
```diff
@@ -2563,8 +2565,14 @@ impl Component {
...
-                        if available_func.name == queued_func.name
-                            && available_func.kind == queued_func.kind
+                        // check if the create/update/refresh/delete action
+                        // should be replaced by an overlay
+                        // otherwise enqueue the same func with the same name and kind
+                        if (available_func.kind == queued_func.kind
+                            && existing_queued_prototype.kind == available_action_prototype.kind
+                            && existing_queued_prototype.kind != ActionKind::Manual)
+                            || (available_func.name == queued_func.name
+                                && available_func.kind == queued_func.kind)
```
nickgerace, line 2575: Okay I can buy this. Smells like this should be another method with an integration test, but not blocking...

lib/dal/tests/integration_test/component/upgrade.rs
```diff
@@ -398,6 +401,86 @@ async fn upgrade_component_type_after_explicit_set(ctx: &mut DalContext) -> Resu
     Ok(())
 }
 
+#[test]
+async fn upgrade_component_attaches_overlay_func(ctx: &mut DalContext) -> Result<()> {
+    let swifty = component::create(ctx, "swifty", "swifty").await?;
+    let actions = dal_materialized_views::action::action_view_list::assemble(ctx.clone()).await?;
+    // ensure there's one create action
+    assert!(actions.actions.len() == 1);
+    let initial_create_action = actions.actions.first().expect("has exactly one action");
```
nickgerace, line 410: Good use of `first()` and interesting idea to use `assert!` with the length check inside rather than `assert_eq!` with `1` being the expected value... haven't thought about doing that in a long time. Might be easier to read.

lib/dal/tests/integration_test/materialized_views.rs
britmyerss: yeah it's in a few spots. Right now we're still returning BOTH scopes in luminork and we're not ready to wrangle that just yet.

lib/dal/tests/integration_test/component/upgrade.rs
```diff
@@ -398,6 +401,86 @@ async fn upgrade_component_type_after_explicit_set(ctx: &mut DalContext) -> Resu
...
+        default_variant_id,
+    )
+    .await?;
+    assert!(
+        actions_pre_overlay
+            .action_prototypes
+            .iter()
+            .any(|action| action.name == initial_create_action.name
+                && action.kind == initial_create_action.kind)
+    );
```
nickgerace, line 428: I _think_ this is fine assuming you are okay with [any](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.any) masking that there could be more than one prototype that matches the condition.

lib/dal/tests/integration_test/component/upgrade.rs
```diff
@@ -398,6 +401,86 @@ async fn upgrade_component_type_after_explicit_set(ctx: &mut DalContext) -> Resu
...
+            .iter()
+            .any(|action| action.name == initial_create_action.name)
+    );
+    // now unlock the variant so we have an upgrade available
+    let unlocked =
+        VariantAuthoringClient::create_unlocked_variant_copy(ctx, default_variant_id).await?;
+    ChangeSetTestHelpers::commit_and_update_snapshot_to_visibility(ctx).await?;
+
+    // now upgrade the component
+    Component::upgrade_to_new_variant(ctx, swifty, unlocked.id()).await?;
```
nickgerace, line 467: Very nice test case

nickgerace (approved):
Conditionally approved given the nits depending on how important you think they are @britmyerss :)

Great work!!

lib/dal/src/component.rs
britmyerss: this is actually what's covered in the integration test!

lib/dal/src/action/prototype.rs
britmyerss: yeah I expect this to change when we add support for overlays for conduit, so I kept it the same as `ManagementPrototype::list_for_schema_and_variant_id` for now

PR #7612, opened 2025-10-25
fix(dal): Only block duplicated func names if they're bound to the same schema/variant

vbustamante:
Allow us to create duplicated func names if they're not bound to the same schema variant. Also updates integrations tests for the new rules.

As an extra , removes fs endpoints from SDF


stack72 (approved):
Love the addition of the test here!

PR #7616, opened 2025-10-27
Toggle button for all the AV groups to help exploration and searching

jobelenus:

On searching, open all the groups, since we're filtering things out in that implementation

The Azure components have lots of components and it can be hard to navigate when they're all open.

#### Screenshots:

image
image
image

## How was it tested?

1. Hit the toggle button
2. See that things close
3. Hit the toggle button
4. See that things open
5. Hit the toggle buttojn
6. Search for an AV
7. See that things are open & filtered


wendybujalski (dismissed):

stack72:
I actually don't think this makes it any easier - it now just creates a dense bunch of headers that we need to search through

PR #7617, opened 2025-10-27
Prevent orphaned update actions

nickgerace:
## Description

This change prevents orphaned update actions in a non-HEAD change set by adding a correct transform. The transform ensures that the action has a corresponding component in either the graph or the list of updates.

This change also refactors the correct transforms in the action node weight module to separate scenarios where the node exists on the graph or not. The split graph implementation remains knowingly untested (checked with @zacharyhamm), but has been ported.

Finally, this change uses an "is subset" check rather than a "difference" check for both the attribute prototype argument node (PR https://github.com/systeminit/si/pull/7571) to ensure we do not remove updates when "valid additions" occur. This is the same motion used for action nodes in this PR.

## Testing

1. Have a component with a resource on HEAD
2. Create two change sets, "A" and "B"
3. In change set "B", erase the component
4. In change set "A", change one field that will trigger an update action, put that action on hold, and then apply the change set
5. Switch to change set "B", and see that no actions exists (without this change an "orphaned" action would exist)

### DIFF
--- a/lib/dal/src/workspace_snapshot/node_weight/action_node_weight.rs
+++ b/lib/dal/src/workspace_snapshot/node_weight/action_node_weight.rs
@@ -14,6 +14,7 @@ use si_events::{
     ulid::Ulid,
 };
 use si_split_graph::SplitGraphNodeId;
+use telemetry::prelude::*;
 
 use super::traits::CorrectTransformsResult;
 use crate::{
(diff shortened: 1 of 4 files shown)


lib/dal/src/workspace_snapshot/node_weight/action_node_weight.rs
```diff
@@ -128,104 +129,174 @@ impl CorrectTransforms for ActionNodeWeight {
...
+            }
+            Update::NewNode { node_weight } => {
+                all_new_nodes.insert(node_weight.id());
+
+                // This should happen one or zero times.
+                if node_weight.id() == id {
+                    indices_for_new_node_updates_for_ourself.push(idx);
                 }
-                _ => {}
             }
```
zacharyhamm, line 271: Don't we have to check for new nodes that match the two node weight criteria here? Like won't this remove the update if there is some *other* new node in this update set, unrelated to these changes?

lib/dal/src/workspace_snapshot/node_weight/action_node_weight.rs
```diff
@@ -128,104 +129,174 @@ impl CorrectTransforms for ActionNodeWeight {
...
-                        )
-                    }),
-            )
+    // If there is a difference between the sets, we need to check that there are updates to
+    // remove. If the conditional passes, there should be exactly one: the new node update for
+    // ourself.
+    if destinations_that_do_not_exist
+        .difference(&all_new_nodes)
+        .next()
+        .is_some()
```
zacharyhamm: so suppose a new component is also part of the update set, for example. then we'll remove this update despite it potentially being valid.

zacharyhamm: A quick fix. `if !destinations_that_do_not_exist.is_subset_of(&all_new_nodes)`

nickgerace: Good idea!

lib/dal/src/workspace_snapshot/node_weight/action_node_weight.rs
nickgerace: Yes and I think your suggestion covers off on this!

zacharyhamm (approved):

PR #7621, opened 2025-10-27
fix(luminork): Update set_ai_agent_executed to return a 204 status code

stack72:
Supercedes #7308


lib/luminork-server/src/service/v1/user.rs
vbustamante: This should probably be changed too, right?

stack72: good catch!!

PR #7623, opened 2025-10-27
feat(si-conduit): add file overwrite and orphan cleanup to remote pull

fnichol:
This change enhances the remote schema pull command to safely overwrite
existing local files and delete orphaned functions that no longer exist in the
remote workspace. Prior to this change, the pull command would fail if any
target files already existed locally, and there was no mechanism to clean up
functions that had been deleted remotely, leading to stale artifacts
accumulating in the local project over time.

The pull command now passes an `overwrite: true` option to all materialize
functions when writing schema metadata, function code, and format version
files. The materialize module has been updated to accept this option and skip
the file existence check when overwriting is enabled, allowing repeated pulls
to update local files with remote changes.

To handle orphaned functions, the pull logic now tracks which function names
exist locally before fetching remote data, compares them against the set of
functions pulled from the remote workspace, and deletes any local functions
that weren't present in the remote response. This applies to all function
types: actions, code generators, management functions, and qualifications.
Empty function directories are also cleaned up automatically after all their
contents have been removed.

The materialization code now skips writing when file content is unchanged,
reducing unnecessary disk operations and making log output more informative.
The `createFileWithLogging` function now checks existing file contents before
writing and logs "Created", "Updated", or "Skipped (unchanged)" as appropriate.

File writing operations are now atomic through a write-to-temporary-
then-rename strategy in the `AbsoluteFilePath.writeTextFile` method. This
ensures that files are never left in a partially written state if the process
is interrupted, which is particularly important when overwriting existing
function code and metadata files. A temporary file with a random suffix is
created in the same directory, written completely, then atomically renamed to
the target filename.

### DIFF
--- a/bin/si-conduit/src/command/remote/schema/pull.ts
+++ b/bin/si-conduit/src/command/remote/schema/pull.ts
@@ -9,7 +9,12 @@ import { Context } from "../../../context.ts";
 import { FunctionMetadata, SchemaMetadata } from "../../../generators.ts";
 import * as materialize from "../../../materialize.ts";
 import { getLogger } from "../../../logger.ts";
-import { AbsoluteFilePath, Project } from "../../../project.ts";
+import {
+  AbsoluteDirectoryPath,
+  AbsoluteFilePath,
+  normalizeFsName,
+  Project,
+} from "../../../project.ts";
 
 const logger = getLogger();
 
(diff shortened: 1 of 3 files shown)


bin/si-conduit/src/command/remote/schema/pull.ts
```diff
@@ -165,15 +177,42 @@ async function pullSchemaByName(
...
+        localActionName,
+      );
+      const codePath = project.actionFuncCodePath(schemaName, localActionName);
+
+      await deleteFunctionFiles(project, metadataPath, codePath);
+      deletedActionPaths.push({ metadataPath, codePath });
+    }
+  }
+
+  // Clean up actions directory if it's now empty
```
vbustamante, line 201: Good touch

vbustamante (approved):

PR #7626, opened 2025-10-28
feat: Azure: Use resource types as schema names

jkeiser:
Azure resources are generally identified like `Microsoft.Compute/virtualMachines`. To make them more like others, currentl we generate schemas like `Azure::Compute::VirtualMachines`, but users (and Claude) are not going to recognize that format.

This PR makes Azure schema names like `Microsoft.Compute/virtualMachines` and categories like `Microsoft.Compute`.

It also prevents generation of schemas outside `Microsoft.` for now, because we don't know the support story for those.

## How was it tested?

- [X] Integration tests pass
- [X] Spotcheck generated schemas
- [X] Manual test: can create components from new schemas
- [x] Manual test: can CRUD still! (Important)

### DIFF
--- a/bin/hoist/src/main.rs
+++ b/bin/hoist/src/main.rs
@@ -1,5 +1,8 @@
 use std::{
-    collections::HashMap,
+    collections::{
+        HashMap,
+        HashSet,
+    },
     fs::{
         self,
         DirEntry,
@@ -389,7 +392,7 @@ async fn diff_summaries_with_module_index(
                         total_added.fetch_add(1, Ordering::SeqCst);
                         let provider = metadata
                             .name()
-                            .split("::")
+                            .split(|c: char| !c.is_alphanumeric())
                             .next()
                             .unwrap_or("Unknown")
                             .to_string();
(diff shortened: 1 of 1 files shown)


bin/clover/src/commands/generateSiSpecs.ts
```diff
@@ -34,14 +34,14 @@ export async function generateSiSpecs(options: PipelineOptions) {
   let imported = 0;
   for (const spec of specs) {
     const specJson = JSON.stringify(unexpandPackageSpec(spec), null, 2);
-    const name = spec.name;
+    const name = spec.name.replaceAll("/", ".");
```
sprutton1, line 37: What's happening here?

jkeiser: The schemas have slash in the name, which requires subdirectories ... I could have created the directory here but wasn't ready to go fix hoist and any other tools to recurse subdirectories. (And unsure if that works with people's workflow either, sometimes I've created subdirectories to hold old versions of the specs and I'm not sure if any tools do that.)

PR #7629, opened 2025-10-28
feat(layer-cache): skip NATS publishing on retry attempts

jhelwig:
Add is_retry parameter to try_write_layers to conditionally skip NATS publishing when processing events from the retry queue.

- Initial attempts (is_retry=false): Publish to NATS + write to PostgreSQL in parallel
- Retry attempts (is_retry=true): Write to PostgreSQL only, skip NATS entirely

This prevents duplicate NATS events from filling NATS disk, polluting caches, cluttering observability, and wasting bandwidth during heavy retry periods. PostgreSQL is the source of truth; remote services fetch from PG on cache miss.

### DIFF
--- a/lib/si-layer-cache/src/persister.rs
+++ b/lib/si-layer-cache/src/persister.rs
@@ -286,7 +286,7 @@ impl PersisterTask {
                 );
 
                 self.tracker.spawn(async move {
-                    match task.try_write_layers(event.clone()).await {
+                    match task.try_write_layers(event.clone(), false).await {
                         Ok(_) => {
                             metric!(
                                 counter.layer_cache_persister_write_success = 1,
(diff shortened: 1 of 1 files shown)


lib/si-layer-cache/src/persister.rs
```diff
@@ -614,64 +614,76 @@ impl PersistEventTask {
...
+                    operation = "write"
+                );
+            } else if nats_error.is_some() {
+                info!(
+                    metrics = true,
+                    counter.layer_cache_persister_nats_error = 1,
+                    cache_name = &cache_name,
+                    operation = "write"
+                );
             }
```
nickgerace, line 678: I know this is just the same code moved, but this is neat! Good to see the code from the dashboards you made.

nickgerace (approved):
Arguably more readable too!

PR #7630, opened 2025-10-28
feat: leaf func overlays

zacharyhamm:
Introduces LeafPrototypes, which are direct children of schemas. LeafPrototypes allow users to create Qualifications and CodeGeneration functions that are connected to the Schema and not the SchemaVariant. This means they will persist across variant upgrades and be available for all components for that Schema whether or not they are the latest version. They also do not require the schema variant to be "unlocked" to be created, so they allow users to extend existing Schemas without missing out on updates provided by the module index. 

LeafPrototypes point to a LeafKind function (Qualification or CodeGen) and are executed in the dependent values update. Since they are not part of the schema variant, they have to create their destination attribute values on the fly (the key/value pair for the qualificaton or code maps). Their inputs are defined as AttributePaths (but currently restricted to LeafInputLocations) and executed "on the fly" instead of via the normal AttributePrototype execution pattern.

These destination values receive a normal AttributePrototype that points to the leaf function, but it has no attribute prototype arguments and so won't execute in the "normal" fashion. However it will be discovered as a LeafOverlay destination and treated accordingly if it is encountered in the construction of the DependentValueGraph.

Automated tests ensure the leaf overlays work as expected, and should cover off on whether the DVU continues to function as expected.


lib/dal/src/attribute/value/dependent_value_graph.rs
```diff
@@ -46,11 +54,52 @@ use crate::{
...
-    component_ids_for_av: BTreeMap<AttributeValueId, ComponentId>,
+    inner: DependencyGraph<DependentValue>,
+    values_that_need_to_execute_from_prototype_function: BTreeSet<DependentValue>,
+    component_id_for_av: BTreeMap<AttributeValueId, ComponentId>,
+    root_attribute_values_for_component: BTreeMap<ComponentId, AttributeValueId>,
+    schema_id_for_component: BTreeMap<ComponentId, SchemaId>,
+    leaf_prototypes_for_schema: BTreeMap<SchemaId, Arc<Vec<(LeafPrototype, String)>>>,
+    leaf_overlays_for_component: BTreeMap<ComponentId, Arc<Vec<ComponentLeaf>>>,
+    leaf_element_ids: BTreeMap<AttributeValueId, ComponentLeaf>,
 }
```
nickgerace, line 103: Do we have to start worrying about how big this could become in memory? I'm guessing no.

lib/dal/src/attribute/value/dependent_value_graph.rs
```diff
@@ -46,11 +54,52 @@ use crate::{
...
+
+impl DependentValue {
+    pub fn attribute_value_id(&self) -> AttributeValueId {
+        match self {
+            DependentValue::AttributeValue(attribute_value_id) => *attribute_value_id,
+            DependentValue::OverlayDestination {
+                destination_map_id,
+                destination_element_id,
+                ..
+            } => (*destination_element_id).unwrap_or(*destination_map_id),
```
nickgerace, line 76: What is going on here? Is this "give me the element or the entire map"?

lib/dal/src/attribute/value/dependent_value_graph.rs
```diff
@@ -270,11 +357,32 @@ impl DependentValueGraph {
...
+                .root_attribute_value_for_component(ctx, current_component_id)
+                .await?;
 
-            if AttributeValue::parent_id(ctx, current_attribute_value.id())
-                .await?
-                .is_none()
-            {
+            // Gather up the leaf overlays (if any) for this component to see if
+            // the current attribute value is an input to them, then add it to
+            // the graph if so
```
nickgerace, line 366: Got it

lib/dal/src/attribute/value/dependent_value_graph.rs
```diff
@@ -270,11 +357,32 @@ impl DependentValueGraph {
...
+                    let destination_map_value = DependentValue::OverlayDestination {
+                        leaf_prototype_id: leaf.prototype_id,
+                        destination_map_id: leaf.destination_map_id,
+                        destination_element_id: leaf.destination_element_id,
+                        root_attribute_value_id,
+                    };
+                    self.value_depends_on(
+                        destination_map_value,
+                        DependentValue::AttributeValue(current_attribute_value.id()),
+                    );
```
nickgerace, line 381: Okay so this is the sauce so to speak. The destination map value depends on the current value?

lib/dal/src/attribute/value/dependent_value_graph.rs
```diff
@@ -376,9 +484,11 @@ impl DependentValueGraph {
 
                     work_queue
                         .push_back(WorkQueueValue::Discovered(controlling_attribute_value_id));
-                    self.inner.id_depends_on(
-                        controlling_attribute_value_id,
-                        current_attribute_value_controlling_value_id,
+                    self.value_depends_on(
```
nickgerace, line 487: Just using the new enum because now overlays are in play 👍

lib/dal/src/attribute/value/dependent_value_graph.rs
```diff
@@ -401,23 +511,139 @@ impl DependentValueGraph {
...
         }
 
         Ok(())
     }
 
+    async fn leaf_overlays_for_component(
+        &mut self,
+        ctx: &DalContext,
+        component_id: ComponentId,
+    ) -> AttributeValueResult<Arc<Vec<ComponentLeaf>>> {
```
nickgerace, line 533: Why an Arc? who else has ownership over this?

lib/dal/src/attribute/value/dependent_value_graph.rs
```diff
@@ -401,23 +511,139 @@ impl DependentValueGraph {
...
+                for proto in prototypes {
+                    let func_id = LeafPrototype::func_id(ctx, proto.id())
+                        .await
+                        .map_err(Box::new)?;
+
+                    let func = Func::get_by_id(ctx, func_id).await?;
+                    prototypes_and_func_names.push((proto, func.name));
+                }
+                let arcvec = Arc::new(prototypes_and_func_names);
+                vacant_entry.insert(arcvec.clone());
```
nickgerace, line 601: Caching on self... good idea

lib/dal/src/job/definition/dependent_values_update.rs
```diff
@@ -420,57 +459,85 @@ impl DependentValuesUpdate {
...
+                                )
+                                .await;
 
-                        metric!(counter.dvu.function_execution = 1);
+                                dependency_graph.cycle_on_self(independent_value);
+                                spawned_ids.insert(independent_value);
 
+                                // Couldn't get the before value? skip!
+                                continue;
+                            }
```
nickgerace, line 502: What is going on here? Is it that you could have a "before value" that's uninitialized? Why not use `None` or json null?

lib/dal/src/schema/leaf.rs
nickgerace, line 1: GitHub wouldn't render this for me. Reviewers: https://github.com/systeminit/si/blob/zack/leaf-func-overlays/lib/dal/src/schema/leaf.rs

nickgerace:
@zacharyhamm I'm going to give the others a chance to review, but I did a first pass.

lib/dal/src/job/definition/dependent_values_update.rs
zacharyhamm: I'm not sure what you mean exactly! This is in the case that you get an error when attempting to fetch the "before value".

lib/dal/src/attribute/value/dependent_value_graph.rs
zacharyhamm: The destination map (`/qualification', '/code') will always exist. But the *element* will not exist since it's not defined as part of the variant. We scaffold leaf functions on component creation and create the map key/value attributes with the right prototype. But these overlay leaves are not scaffolded, since they aren't part of the variant, so we have to create the element on demand. But, if the element already exists, it's more useful to as the "attribute value id" than the map.

lib/dal/src/attribute/value/dependent_value_graph.rs
zacharyhamm: It does seem like a lot of maps but I don't think there's a lot of overhead here on the whole

lib/dal/src/attribute/value/dependent_value_graph.rs
zacharyhamm: I did the Arc for two reasons: (1) I didn't want to clone the whole vec to get it out of the map (maps return a ref, but that ref would break some borrow checker rules for the way we use it and (2) it has to be an Arc, not an Rc, because of async. Arc good! Fast! (on x86 systems the "Relaxed" ordering atomics are the same instructions as the "non" atomic instructions so an Rc and Arc are going to be very close to equivalent in machine code, although the compiler will do less reordering of instructions)

lib/dal/src/job/definition/dependent_values_update.rs
britmyerss: I think Nick's asking why we wouldn't default the before value and proceed with the new after value.  what errors are you expecting/seeing?

jhelwig (approved):

PR #7637, opened 2025-10-29
feat(layer-cache): fix retry queue metric types

jhelwig:
Replace incorrect counter metrics with correct monotonic counters and gauges to align with proper OpenTelemetry semantics:

- Event counters (enqueued, success, failed, permanent_failure) now use monotonic!() for cumulative totals, enabling proper rate() queries in Grafana

- Queue depth tracking now uses gauge!() with absolute values instead of increment/decrement counters, providing accurate point-in-time queue size visualization

- Added initial depth gauge emission on startup when scanning existing queues to ensure metrics reflect reality immediately

- Added depth gauge to mark_failure for consistency in Prometheus data point timing across all retry queue operations

This change ensures Grafana dashboards correctly display retry queue metrics without requiring dashboard query modifications.

### DIFF
--- a/lib/si-layer-cache/src/retry_queue.rs
+++ b/lib/si-layer-cache/src/retry_queue.rs
@@ -56,6 +56,10 @@ use std::{
     },
 };
 
+use telemetry_utils::{
+    gauge,
+    monotonic,
+};
 use tokio::sync::mpsc;
 use tokio_util::sync::CancellationToken;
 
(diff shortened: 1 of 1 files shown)


zacharyhamm (approved):

fnichol (approved):
Ah yeah, I'm really looking forward to more of these changes across the codebase. Should help us a lot once we visualize in dashboards.

PR #7639, opened 2025-10-29
feat(si-conduit): add authentication function support to remote pull

fnichol:
This change introduces support for pulling authentication functions when executing remote schema pulls. Prior to this change, the `si-conduit` tool would pull action, codegen, management, and qualification functions but would skip authentication functions entirely, leaving schemas incomplete when synchronized from remote workspaces.

With this change, authentication functions are now handled consistently with other function types. The implementation follows the established pattern used for actions, codegens, and other function categories:

- Authentication functions are fetched from the remote API using their function IDs
- Both metadata JSON and TypeScript code files are materialized to disk
- Local authentication functions that no longer exist remotely are automatically deleted (orphan cleanup)
- Empty authentication directories are removed when all auth functions are deleted

### DIFF
--- a/bin/si-conduit/src/command/remote/schema/pull.ts
+++ b/bin/si-conduit/src/command/remote/schema/pull.ts
@@ -94,6 +94,7 @@ type SchemaAndVariantData = {
   func: {
     schemaId: string;
     actionIds: string[];
+    authIds: string[];
     codegenIds: string[];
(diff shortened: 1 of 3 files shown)


stack72 (approved):
Love seeing the support for Auth funcs here! Definite delighter

PR #7640, opened 2025-10-29
Show previous attribute input value, and highlight subscription on focus

jobelenus:
More Quality of Life!! When you focus on an attribute input, the incoming subscription will open and highlight itself for ease of navigation!

#### Screenshots:

image

image

## How was it tested?

Click around!


wendybujalski (changes requested):
Love these improvements, but it looks like you'll need to test around long values and truncation -

Image

wendybujalski (approved):

PR #7649, opened 2025-10-30
feat: Create subresource schemas for Azure

jkeiser:
VM extensions are only identifiable as *sub-resources*--`/providers/Microsoft.Compute/virtualMachines/myMachine/extensions/myExtension`.

This PR:
* Adds two-level subresources like extensions
* Supports arbitrary depth subresources in create/update/refresh/delete using a new `resourceId` domain prop that mirrors the path (i.e. `/subscriptions/{subscriptionId}/providers/Microsoft.Compute/virtualMachines/{vmName}/extensions/{vmExtensionName}`)
* Uses the parameter name in the schema for subscriptionId, resourceGroupName, and name
* Merges the parameters from the schema, and uses the parameter *name* in the schema (generally means `name` and `resourceGroup` are now things like `vmName` and `resourceGroupName`).

## How was it tested?

- [X] Integration tests pass
- [X] Manual test: extensions is created, CRUD works

### DIFF
--- a/bin/clover/src/pipelines/aws/spec.ts
+++ b/bin/clover/src/pipelines/aws/spec.ts
@@ -118,11 +118,25 @@ export function pkgSpecFromCf(cfSchema: CfSchema): ExpandedPkgSpec {
   );
 }
 
-export function parseSchema(rawSchema: unknown): ExpandedPkgSpec[] {
-  const cfDb = rawSchema as CfDb;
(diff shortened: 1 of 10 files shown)


aaron-dernley:
How come some schemas show up in the diff with lowercase microsoft? @jkeiser or @sprutton1

PR #7652, opened 2025-10-30
Evict closed connections when checking for long-lived and idle connections

jhelwig:
The pool eviction loop now checks if connections are already closed (e.g., due to server disconnection or timeout) and removes them immediately. Previously, closed connections would only be evicted when they exceeded the max idle time, potentially leaving dead connections in the pool during the check interval.

Connection state isn't checked until we try to get a connection from the pool, so proactively checking during eviction means we should run into fewer "zombie" connections. While the idle and max age checks should prevent the remote end from closing connections for those reasons, there are always other reasons a connection could be closed.

### DIFF
--- a/lib/si-data-pg/src/lib.rs
+++ b/lib/si-data-pg/src/lib.rs
@@ -3013,12 +3013,13 @@ async fn evict_old_and_idle_pgpool_connections(
     loop {
         tokio::time::sleep(check_interval).await;
         let connection_count = Rc::new(RefCell::new(0));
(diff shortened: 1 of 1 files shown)


fnichol (approved):
Oh yeah, that's a fantastic tweak, allowing the code to proactively reap those connections. Ideally this helps lower our "time-to-live-connection" when we call `pg_pool.get().await` as there's a higher chance that all pool connections are young/recently active/live

PR #7655, opened 2025-10-30
Add one prop suggestion to Azure LB schema

nickgerace:
## Description

This change adds one prop suggestion to the Azure LB schema. This is mainly for me, Nick, to learn how to do this, but is a valid change on its own.

### DIFF
--- a/bin/clover/src/pipelines/azure/provider.ts
+++ b/bin/clover/src/pipelines/azure/provider.ts
@@ -1,5 +1,6 @@
 import {
   FetchSchemaOptions,
+  PropOverrideFn,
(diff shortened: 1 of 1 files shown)


sprutton1 (approved):

bin/clover/src/pipelines/azure/provider.ts
```diff
@@ -61,6 +63,22 @@ function azureIsChildRequired(
   return schema.requiredProperties.has(childName);
 }
 
+// TODO(nick): move to file.
```
nickgerace, line 66: This is intentionally staying for now

PR #7656, opened 2025-10-30
feat: create overlay qualifications and code gens with luminork

zacharyhamm:
The create qualification and create codegen endpoints in luminork will
now default to creating overlay functions if the schema is a builtin.

Tested by creating a qualification and codegen in luminork for a security group. Then creating a working workspace (cred + region and some vpcs imported) and creating security groups. Ensuring DVU dependent features (actions and quals/code gens) still worked correctly. Ensured qualification overlays re-run when button clicked on component.

### DIFF
--- a/lib/dal/src/func/binding/leaf.rs
+++ b/lib/dal/src/func/binding/leaf.rs
@@ -2,6 +2,8 @@ use serde::{
     Deserialize,
     Serialize,
 };
+use si_frontend_types::LeafBindingPrototype;
+use si_id::LeafPrototypeId;
 use telemetry::prelude::*;
 
 use super::{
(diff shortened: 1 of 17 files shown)


lib/dal/src/func/binding/leaf.rs
```diff
@@ -137,8 +159,23 @@ impl LeafBinding {
                 // let attribute_prototype_id =
                 //     Component::upsert_leaf_function(ctx, component_id, leaf_kind, inputs, &func).await?;
             }
-            EventualParent::Schemas(_) => {
-                // zack: how to handle?
```
fnichol, line 141: 😄

fnichol (approved):
Nice work!

PR #7657, opened 2025-10-30
fix: add prop suggestion for ou parentid

aaron-dernley:
Adds a prop suggestion to the ParentId prop on the AWS::Organization::OrganizationalUnit Schema.

#### Screenshots:

Screenshot 2025-10-30 at 22 27 11

## How was it tested?

- [X] Manual test: added suggestion to Component via the UI to test before implementing here.

### DIFF
--- a/bin/clover/src/pipelines/aws/overrides.ts
+++ b/bin/clover/src/pipelines/aws/overrides.ts
@@ -77,6 +77,14 @@ export const AWS_PROP_OVERRIDES: Record<
     DataProtectionPolicy: policyDocumentProp,
   },
 
+  // AWS::Organizations
+  "AWS::Organizations::OrganizationalUnit": {
+    ParentId: [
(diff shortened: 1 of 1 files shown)


bin/clover/src/pipelines/aws/overrides.ts
```diff
@@ -77,6 +77,11 @@ export const AWS_PROP_OVERRIDES: Record<
     DataProtectionPolicy: policyDocumentProp,
   },
 
+  // AWS::Organizations
+  "AWS::Organizations::OrganizationalUnit": {
+    ParentId: suggest("AWS::Organizations::Organization", "RootId"),
```
stack72: We should be able to make a second suggestion to another OU Id as well

aaron-dernley: Just pushed that change up @stack72

PR #7659, opened 2025-10-30
Add prop suggestions for Azure pluralized matches

nickgerace:
## Description

This change adds prop suggestions for all Azure assets where the parent prop of an "id" prop matches the pluralized name of another Azure asset.

An example: say you have a "Microsoft.Network/todds" asset and a "Microsoft.Network/howards" asset. The "todds" asset has a deeply nested prop called "id" where its parent is "howard" (pluralized). By convention, this implies that the "id" expected corresponds to a resource ID for the "howard" asset.

That being said, this change does not cover every scenario. For example, if the parent of the "id" prop is "pools", but an "ipamPool" ID is expected, it will not suggested. That is for a future iteration.


nickgerace:
Just amended to ignore `/root/domain`.

jkeiser (approved):

PR #7660, opened 2025-10-30
fix(conduit): Don't unlock unnecessary schemas or create unnecessary change sets

vbustamante:

Change the order in which we did operations on push so we can avoid write unnecessary unlocks and write operations, by checking the filesystem data against head as the first step. 

Extras: 
- Added the link  in schemas when pushing
- Make sure we fail very early if we have duplicated schema names

#### Screenshots:

image

## How was it tested?

- [X] Manual test: Validated that only having unchanged schemas does not create a changeset
- [X] Manual test: Created a new schema, pulled an 2 existing schemas and modified one. Validated that only the necessary schemas and functions would be unlocked, and also that funcs are being erased.

### DIFF
--- a/bin/si-conduit/src/command/remote/schema/push.ts
+++ b/bin/si-conduit/src/command/remote/schema/push.ts
@@ -1,4 +1,10 @@
-import { ChangeSetsApi, FuncsApi, SchemasApi } from "@systeminit/api-client";
+import {
+  ChangeSetsApi,
+  FindSchemaV1Response,
+  FuncsApi,
+  GetFuncV1Response,
+  SchemasApi,
+} from "@systeminit/api-client";
 import { AxiosError } from "axios";
 import { Context } from "../../../context.ts";
 import { AuthenticatedCliContext } from "../../../cli/helpers.ts";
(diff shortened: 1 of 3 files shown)


bin/si-conduit/src/generators.ts
```diff
@@ -329,7 +329,7 @@ function schemaMetadata(schemaName: string): SchemaMetadata {
     name: schemaName,
     category: "",
     description: "optional",
-    documentation: "optional, should be a link",
+    documentation: null, // "optional, should be a link",
```
fnichol, line 382: 👍🏽

bin/si-conduit/src/command/remote/schema/push.ts
```diff
@@ -435,25 +453,217 @@ export async function callRemoteSchemaPush(
...
+  // Pre check: Compare schema and funcs to head
+  // ==================================
+  const changeSets = (await changeSetsApi.listChangeSets({ workspaceId }))
+    .data.changeSets as { id: string; isHead: boolean }[]; // Mocking this type since the client lib does not have it right now
+
+  const headChangesetId = changeSets.find((cs) => cs.isHead)?.id;
+
+  if (!headChangesetId) {
+    throw new Error("No head changeset found");
+  }
```
fnichol, line 489: Interesting! Yes all these checks I didn't have to consider with the `pull` code

bin/si-conduit/src/command/remote/schema/push.ts
```diff
@@ -619,47 +771,61 @@ export async function callRemoteSchemaPush(
...
+            await siSchemasApi.detachCodegenFuncBinding(detachPayload);
+            break;
+          case "Management":
+            await siSchemasApi.detachManagementFuncBinding(detachPayload);
+            break;
+          case "Authentication":
+            await siSchemasApi.detachAuthenticationFuncBinding(detachPayload);
+            break;
+          default:
+            console.error(`Unknown func kind for func ${funcId}: ${kind}`);
```
fnichol, line 844: Not a bad idea, for anything in the future

PR #7667, opened 2025-10-31
fix(dal): ensure deletes and upgrades cannot break components with subs

zacharyhamm:
Suppose you have a component A on HEAD. And, in another change set, you have B, which has an outgoing ValueSubscription edge to the root attribute value of A (subscribe a value on B to a value on A). If you either: upgrade A to a new schema variant and apply to HEAD, or delete A and apply to HEAD, we will not remove the root attribute value or the incoming ValueSubscription edges for that deleted component without this correction. The result is that the component B will be broken in the change set.

This correction ensures that any time we either delete a component, or detect that its schema variant is being upgraded, we esnure the root attribute value is removed and we ensure that the attribute prototype arguments that are the source for the ValueSubscription edges are also removed, so that the subscribing (destination) component does not have broken attribute values.

Should close BUG-1123.

This one deserves manual testing, first just deleting and appying and seeing how it goes. Second, replicate the scenario in the description above and ensure that the change sets remain in a good state.

### DIFF
--- a/lib/dal/src/workspace_snapshot/node_weight/component_node_weight.rs
+++ b/lib/dal/src/workspace_snapshot/node_weight/component_node_weight.rs
@@ -168,6 +168,7 @@ impl CorrectTransforms for ComponentNodeWeight {
         let is_self = |node: &NodeInformation| node.id == self.id.into();
 
         let mut component_will_be_deleted = false;
+        let mut schema_variant_changed = false;
         let mut remove_edges = vec![];
         for update in &updates {
             match update {
@@ -216,6 +217,7 @@ impl CorrectTransforms for ComponentNodeWeight {
                     // just in case.
                     remove_edges
                         .extend(graph.outgoing_edges(component_node_idx, EdgeWeightKind::Root));
+                    schema_variant_changed = true;
                     remove_edges.extend(
                         graph.outgoing_edges(component_node_idx, EdgeWeightKind::SocketValue),
                     );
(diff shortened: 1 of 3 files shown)


zacharyhamm:
I wouldn't merge this one without really putting it through its paces. I'll be doing more manual testing monday morning.

lib/dal/src/workspace_snapshot/node_weight/component_node_weight.rs
```diff
@@ -226,6 +228,12 @@ impl CorrectTransforms for ComponentNodeWeight {
         }
 
         if component_will_be_deleted {
+            // Also remove any incoming edges to the component in case there
+            // is a frame contains in another change set, or management edges
+            remove_edges.extend(graph.edges_directed(component_node_idx, Incoming));
+        }
+
+        if component_will_be_deleted || schema_variant_changed {
```
nickgerace, line 236: Ah I see. The problem is that the upgrade needs an AV refresh in a similar motion to deleted components.

lib/dal/src/workspace_snapshot/node_weight/component_node_weight.rs
```diff
@@ -272,6 +309,7 @@ fn category_kind(
 }
 
 /// Creates an Update::RemoveEdge from an EdgeReference by looking up the nodes.
+#[inline(always)]
```
nickgerace, line 312: For my own learning, what made you strongly suggest inlining?

lib/dal/tests/integration_test/node_weight/component.rs
```diff
@@ -40,6 +52,61 @@ async fn deleting_a_component_deletes_component_in_other_change_sets(ctx: &mut D
     );
 }
 
+#[test]
+async fn deleting_a_value_source_component_does_not_break_other_change_sets(
+    ctx: &mut DalContext,
+) -> Result<()> {
```
nickgerace, line 58: Lovely

lib/dal/src/workspace_snapshot/node_weight/component_node_weight.rs
zacharyhamm: Tiny function called lots :)

PR #7669, opened 2025-10-31
Virtualize the component creation modal

jobelenus:
Thank you Victor for getting this started!!

Virtualize the list of thousands of assets in the add component modal.

This means no longer using the `<Treeform>` components because it creates a nested hierarchy that won't work with the virtual approach because it needs a flat list.

## How was it tested?

- [X] typing doesn't suck!!!
- [X] Integration tests pass
- [X] Manual test: new functionality works in UI (all the keyboard controls, scrolling, etc etc)
- [X] Manual test: (regression check) creating a component still works with click and enter
- [X] Manual test: unlock an asset, ensure you see `editing` as well as the default in the list
- [X] Here is a fun one:
- open the modal
- press up a few times
- scroll to the middle of the list
- open a category
- click on a schema
- ensure the list is scrolled to the one you clicked on

### DIFF
--- a/app/web/src/newhotness/AddComponentModal.vue
+++ b/app/web/src/newhotness/AddComponentModal.vue
@@ -7,6 +7,7 @@
     size="max"
     class="w-[max(800px,66vw)]"
     @click="onClick"
+    @close="close"
   >
     <div class="h-[45svh]">
       <div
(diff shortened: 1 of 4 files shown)


wendybujalski (changes requested):
It looks like you lost the hover state for the headers when you switched from TreeNode to AddComponentModalListRow.

The scroll isn't quite right for the keyboard controls too, looking into that now.

But once we have those issues fixed this should be good to go!

PR #7670, opened 2025-10-31
create dynamic templates

adamhjk:
feat(si-tmpl): CLI tool for declarative System Initiative templates

si-tmpl is a CLI tool that runs declarative templates to manage System Initiative infrastructure components. Templates are TypeScript files that define what components should exist and how they should be configured, and si-tmpl handles the convergence between desired state (working set) and actual state (existing components).

The execution flow follows a pipeline: First, the template configures search queries to find baseline components in the workspace. The baseline is deep-cloned into a working set, which can be transformed through name patterns (with regex/EJS support) and custom transform functions. This working set represents the desired end state.

Convergence is the core operation: si-tmpl queries for existing components created by previous template runs (identified by template tags), diffs the working set against existing components to build pending changes (creates, updates, deletes), and topologically sorts these changes to handle dependencies. Components in the working set are matched to existing components first by ID, then by name for dynamically-created components, enabling idempotent template execution.

The TemplateContext API provides rich component manipulation: setAttribute(), deleteAttribute(), and
setSiblingAttribute() for attribute operations; setSubscription() for connecting components; copyComponent() forduplicating existing components; and newComponent() for creating from scratch. Schema helpers like
getSchemaIdByName() and getSchemaName() enable working with human-readable schema names instead of IDs.

All changes are executed via the System Initiative API with proper error handling, caching (for schemas,
searches, and components), and structured logging. Templates can accept validated input data using Zod schemas, cache baselines for faster iteration, and support dry-run mode for testing transformations before applying them.

It provides an external dynamic templating mechanism.

#### Out of Scope:

This probably will also work for a lot more use cases, or at least the basics of the structure will (in particular the convergence part). But it does not implement them.

Also, this codes long-term home should be si-conduit.

## How was it tested?
- [x] Lots of unit tests
- [x] Manual test and iteration

### DIFF
--- a/bin/si-tmpl/BUCK
+++ b/bin/si-tmpl/BUCK
@@ -1,0 +1,53 @@
+load(
+    "@prelude-si//:deno.bzl",
+    "deno_compile",
+    "deno_format",
+    "deno_test",
+)
+load(
+    "@prelude-si//:macros.bzl",
+    "alias",
+    "export_file",
+)
+
+deno_compile(
+    name = "si-tmpl",
+    main = "main.ts",
+    out = "si-tmpl",
+    srcs = glob([
+        "src/**/*.ts",
+        "main.ts",
+    ]),
+    permissions = [
+        "allow-env",
+        "allow-net",
+        "allow-read",
+        "allow-write",
+    ],
+    visibility = ["PUBLIC"],
+)
+
+deno_format(
+    name = "fix-format",
+    srcs = glob(["**/*.ts"]),
+    ignore = ["node_modules"],
+)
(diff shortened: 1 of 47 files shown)


adamhjk:
You can test this yourself with the existing template (which is based off of our si-demo):

Set your SI_API_TOKEN environment variable, and then:

```
deno run -A ./main.ts run ./tmpl/test.ts -k whatever --input ./tmpl/test-prod-input.yaml --baseline /tmp/foo.yaml
```

And it will create a new si-logo demo environment for you.

PR #7674, opened 2025-11-01
fix(dal): Ensure we carry default subscriptions as part of component upgrades

stack72:
Fixes BUG-1121

When a component's attribute value was marked as a default subscription source (allowing other components to automatically subscribe to it), that marking was lost during component upgrades. After upgrading to a new schema variant, users would need to manually re-mark the attribute values as default subscription sources.

### DIFF
--- a/lib/dal/src/component.rs
+++ b/lib/dal/src/component.rs
@@ -2396,6 +2396,18 @@ impl Component {
             Component::root_attribute_value_id(ctx, original_component_id).await?;
         let original_subscriber_apas = AttributeValue::subscribers(ctx, original_root_id).await?;
 
+        // Save which attribute values are marked as default subscription sources
+        let mut original_default_sub_source_paths = vec![];
+        let mut av_work_queue = VecDeque::from([original_root_id]);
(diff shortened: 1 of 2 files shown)


lib/dal/src/component.rs
```diff
@@ -2527,6 +2539,19 @@ impl Component {
...
+            while let Some(new_av_id) = new_av_work_queue.pop_front() {
+                if let Some(new_path) = AttributeValue::get_path_for_id(ctx, new_av_id).await? {
+                    if original_default_sub_source_paths.contains(&new_path) {
+                        AttributeValue::set_as_default_subscription_source(ctx, new_av_id).await?;
+                    }
+                }
+                new_av_work_queue.extend(AttributeValue::child_av_ids(ctx, new_av_id).await?);
+            }
+        }
+
```
zacharyhamm, line 2554: Instead of looping over every attribute value in the component and asking if it matches the given paths, let's just loop over the paths, use `resolve` against the new component, and then call set_as_default_subscription_source. Should be more efficient (there are likely only 1 or 2 default sub sources per component)

PR #7677, opened 2025-11-03
chore(dal): remove double write on new change set

zacharyhamm:
Eviction races required us to write a new snapshot for every new change set (to get a new address). This should no longer be necessary.


jhelwig (approved):

jhelwig:
Nice to see the eviction changes resulting in code deletion elsewhere. 👍🏻

zacharyhamm:
integration tests failed for a 10 second timeout in an action test.

PR #7678, opened 2025-11-03
feat(si-mcp-server): support overlays for codegen and qualification

zacharyhamm:
Adds codegens and qualifications to the functions that we create overlays for if the schema is a builtin


zacharyhamm:
diff looks bigger than it actually is. I'm blaming `deno fmt`

stack72 (dismissed):

wendybujalski (dismissed):

stack72 (approved):

PR #7682, opened 2025-11-04
fix(web): Send correct { $source: null } format when clearing input fields

stack72:
Fixes: BUG-1036

When users cleared an input field in the AttributePanel, the frontend was sending `{ "/path": null }` instead of `{ "$source": null }`. This caused attribute functions not to re-run after clearing manual overrides, leaving stale values in place.

Example scenario:
1. AMI component has an attribute function that computes a value for prop ImageId
2. User manually sets prop ImageId to "manual-override"
3. User clears the input field in AttributePanel
4. Expected: Attribute function re-runs and restores computed value
5. Actual: Value remains as "manual-override" - function never re-runs

## Root Cause

The frontend's makeSavePayload() function in attribute_tree.ts treated empty string inputs as plain null values instead of using the proper `{ $source: null }` format. This bypassed the correct code path that removes component prototypes and triggers attribute function re-execution.

### DIFF
--- a/app/web/src/newhotness/logic_composables/attribute_tree.ts
+++ b/app/web/src/newhotness/logic_composables/attribute_tree.ts
@@ -114,13 +114,16 @@ export const makeSavePayload: MakePayload = (
   // TODO - Paul there's a better way to handle this for sure!
   let coercedVal: string | boolean | number | null = value;
 
+  const payload: componentTypes.UpdateComponentAttributesArgs = {};
+
   // We don't want to coerce a prop path when connecting via a subscription, so skip it (e.g. prop
   // kind is "integer", but the value is the prop path, which is a "string").
   if (!connectingComponentId) {
     if (value === "") {
-      // For now, we don't allow a user to enter an empty string becuase passing an empty string is
-      // effectively an unset operation.
-      coercedVal = null;
+      // When clearing an input field, send $source: null to properly unset the value
+      // and trigger attribute functions to re-run with the default/schema variant prototype
+      payload[path] = { $source: null };
+      return payload;
     } else if (propKind === PropKind.Boolean) {
       coercedVal = value.toLowerCase() === "true" || value === "1";
     } else if (propKind === PropKind.Integer) {
@@ -130,7 +133,6 @@ export const makeSavePayload: MakePayload = (
     }
   }
 
-  const payload: componentTypes.UpdateComponentAttributesArgs = {};
   payload[path] = coercedVal;
   if (connectingComponentId) {
     payload[path] = {
--- a/lib/dal/src/attribute/attributes.rs
+++ b/lib/dal/src/attribute/attributes.rs
@@ -435,13 +435,10 @@ async fn update_attributes_inner(
                         // If the parent is a map or array, remove the value
                         AttributeValue::remove(ctx, target_av_id).await?;
                     } else {
-                        // Otherwise, just set it to its default value
-                        if AttributeValue::component_prototype_id(ctx, target_av_id)
-                            .await?
-                            .is_some()
-                        {
-                            AttributeValue::use_default_prototype(ctx, target_av_id).await?;
-                        }
+                        // Otherwise, use the default prototype (either by removing an existing
+                        // component prototype to revert to the schema variant prototype, or by
+                        // creating a si:Unset component prototype to explicitly unset the value)
+                        AttributeValue::use_default_prototype(ctx, target_av_id).await?;
                     }
 
                     ctx.write_audit_log(
--- a/lib/dal/src/attribute/value.rs
+++ b/lib/dal/src/attribute/value.rs
@@ -1717,26 +1717,35 @@ impl AttributeValue {
         ctx: &DalContext,
         attribute_value_id: AttributeValueId,
     ) -> AttributeValueResult<()> {
-        Self::update(ctx, attribute_value_id, None).await?;
-        let prototype_id = Self::component_prototype_id(ctx, attribute_value_id)
-            .await?
-            .ok_or(AttributeValueError::NoComponentPrototype(
-                attribute_value_id,
-            ))?;
+        // Check if there's a component prototype to remove
+        if let Some(_prototype_id) = Self::component_prototype_id(ctx, attribute_value_id).await? {
+            // There was an existing component prototype (e.g., from a manual override)
+            // Clear the value by calling update(None) - this creates a si:Unset component prototype
+            Self::update(ctx, attribute_value_id, None).await?;
 
-        ctx.workspace_snapshot()?
-            .remove_edge(
-                attribute_value_id,
-                prototype_id,
-                EdgeWeightKindDiscriminants::Prototype,
-            )
-            .await?;
+            // Remove the si:Unset prototype so the attribute value falls back to the schema variant prototype
+            let current_prototype_id = Self::component_prototype_id(ctx, attribute_value_id)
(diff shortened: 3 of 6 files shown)


jkeiser:
Seeing the value to null is not the same as clearing it; seeing it to null should actually set it to null (not use the default value).

jkeiser:
I.e. I think `{ "/path": null }` should call update (null), otherwise there's no way to set the value to null. Clearing the value would use the default value (which might not be null) instead of null.

Clearing the input field should use `{ "/path": { "source": null } }`. Unless we decide you should never be able to explicitly set a field's value to null!

lib/dal/src/attribute/attributes.rs
```diff
@@ -343,16 +343,26 @@ async fn update_attributes_inner(
...
-                        if before_value.as_ref() != Some(&value) {
-                            // Build the Audit Log entry!
-                            after_value_source =
-                                Some(PropValueSource::Value(serde_json::to_value(value)?));
-
-                            // If the values have changed then we should enqueue an update action
-                            // if the values haven't changed then we can skip this update action as it is usually a no-op
-                            Component::enqueue_update_action_if_applicable(ctx, target_av_id)
+
+                        // If the value is null, treat this as "unsetting" the value rather than
```
jkeiser: This means that the following PUT would also unset the value rather than setting it to null, which is undesirable:

```
PUT /attributes
{
  "/domain/Foo": { $source: { $value: null } }
}
```

This form is supposed to mean "set the source to a static value of null," and should either do that (which is what I think it currently does, at least when I initially wrote this) or yield an error saying it's currently unsupported.

I understand we're trying to make `"/domain/Foo": null` unset the value. In my opinion, we should leave it the same as now (`"/domain/Foo": <any value>` should just set the value), but reasonable minds can disagree here :) If that's the decision we take I can point you at the code that would make it happen (it's in the JSON structs we use to deserialize).

TL;DR I think we should leave this code the same and just change the attributes panel.

lib/dal/src/attribute/attributes.rs
```diff
@@ -435,13 +445,10 @@ async fn update_attributes_inner(
...
-                        if AttributeValue::component_prototype_id(ctx, target_av_id)
-                            .await?
-                            .is_some()
-                        {
-                            AttributeValue::use_default_prototype(ctx, target_av_id).await?;
-                        }
+                        // Otherwise, use the default prototype (either by removing an existing
+                        // component prototype to revert to the schema variant prototype, or by
+                        // creating a si:Unset component prototype to explicitly unset the value)
+                        AttributeValue::use_default_prototype(ctx, target_av_id).await?;
```
jkeiser, line 441: I like this change! use_default_prototype should do all the checking itself.

lib/dal/src/attribute/value.rs
```diff
@@ -1717,26 +1717,38 @@ impl AttributeValue {
         ctx: &DalContext,
         attribute_value_id: AttributeValueId,
     ) -> AttributeValueResult<()> {
-        Self::update(ctx, attribute_value_id, None).await?;
-        let prototype_id = Self::component_prototype_id(ctx, attribute_value_id)
-            .await?
-            .ok_or(AttributeValueError::NoComponentPrototype(
-                attribute_value_id,
-            ))?;
+        // Check if there's a component prototype to remove
```
jkeiser, line 1720: Just a Rust thing, take or leave if it strikes your fancy: we have been moving towards a new Rust syntax `let else { return }` for cases where the `else` is really small and easily digestible, to sort of "get it out of the way" before the real `Some()` case you care about:

```rust
// 1. let else { return }; is a new Rust thing that lets you handle None quickly and return
let Some(prototype_id) = Self::component_prototype_id(ctx, attribute_value_id).await? else {
  // Put the else case in here and return!
  return Self::update(ctx, attribute_value_id, None).await;
};

// Now you can put the normal case (remove the prototype) with no nesting!
```

lib/dal/tests/integration_test/attributes.rs
```diff
@@ -639,3 +642,169 @@ async fn update_attribute_child_of_subscription(ctx: &mut DalContext) -> Result<
...
+    assert_eq!(
+        json!({
+            "Value": "manual-override",
+        }),
+        component::domain(ctx, "test").await?
+    );
+
+    // Verify the value is marked as overridden (has a component prototype)
+    let av_id = value::id(ctx, ("test", "/domain/Value")).await?;
+    let component_prototype = AttributeValue::component_prototype_id(ctx, av_id).await?;
```
jkeiser, line 696: Note: you can use `value::has_value()` for this if you want to use the fluent test syntax (which would mean you don't have to catch the av_id).

lib/dal/tests/integration_test/attributes.rs
```diff
@@ -639,3 +642,169 @@ async fn update_attribute_child_of_subscription(ctx: &mut DalContext) -> Result<
...
+    assert!(
+        component_prototype.is_some(),
+        "Value should have a component prototype after manual override"
+    );
+
+    // Delete the manual override with { $source: null }
+    attributes::update_attributes(
```
jkeiser, line 703: Thank you for adding this test! I am really surprised I didn't add this. Bad John!

app/web/src/newhotness/logic_composables/attribute_tree.ts
```diff
@@ -114,13 +114,16 @@ export const makeSavePayload: MakePayload = (
   // TODO - Paul there's a better way to handle this for sure!
   let coercedVal: string | boolean | number | null = value;
 
+  const payload: componentTypes.UpdateComponentAttributesArgs = {};
```
jkeiser, line 117: Comment for the future: since this is only being used in one place, may as well just move this to line 126 and do `return { [path]: { $source: null } };`

lib/dal/src/attribute/value.rs
```diff
@@ -1717,26 +1717,34 @@ impl AttributeValue {
...
-                EdgeWeightKindDiscriminants::Prototype,
-            )
-            .await?;
+        if let Some(_prototype_id) = maybe_prototype_id {
+            // There was an existing component prototype (e.g., from a manual override)
+            // Remove it so the attribute value falls back to the schema variant prototype
+            // The si:Unset prototype created by update(None) will also be removed
+            let current_prototype_id = Self::component_prototype_id(ctx, attribute_value_id)
+                .await?
+                .expect("update(None) should have created a component prototype");
```
jkeiser: We need to preserve the `.ok_or()` here; `.expect()` will panic and stop the whole server.

Sorry I didn't catch this in the earlier review!

lib/dal/tests/integration_test/property_editor.rs
```diff
@@ -409,6 +409,115 @@ async fn override_value_then_reset(ctx: &mut DalContext) {
...
+        json!(original_pirate_name),
+        "After first reset, should be back to default"
+    );
+    assert_eq!(
+        prop_view["overridden"],
+        json!(false),
+        "Should not be overridden"
+    );
+
+    // Second call: should explicitly unset (create si:Unset prototype)
```
jkeiser, line 470: For reference, I think `use_default_prototype()` should not explicitly unset, ever. It should always use the default prototype. I get why you are doing it--you are making something possible that is currently impossible!--but I think we'll want to do it a different way. I'll make a proposal later, though; we should merge this and move on for now.

jkeiser (approved):

PR #8331, opened 2026-01-22
feat(web) - helpful hiding functionality for Map view

wendybujalski:

This PR adds two hiding features to the Map view to make it easier to navigate and read. One hiding feature hides all credential components from the Map, while the other hides all components which have no connections to any other components. Both features are on by default, with a button to toggle them off.

Both features are behind a feature flag so that we can have our team try them out before rolling anything out to users.

This PR also includes some minor refactoring improvements and style fixes to the area the buttons were added to.

#### Screenshots:

https://github.com/user-attachments/assets/003172c7-5015-4b9b-a610-96ba20793724

#### Out of Scope:

These hiding features may be better implemented via a different UI than individual buttons, and we may decide that one or both should be disabled by default instead of enabled by default.

This PR also does not make any other changes to the Map besides these two hiding features.

## How was it tested?

Aggressive manual testing with a wide variety of components and configurations.

Not yet, but if we roll out this feature we will want to update the docs to explain it to users.

### DIFF
--- a/app/web/src/newhotness/Explore.vue
+++ b/app/web/src/newhotness/Explore.vue
@@ -146,7 +146,14 @@
             </Transition>
           </div>
         </div>
-        <div class="flex-none flex flex-row flex-wrap items-center gap-xs">
+        <div :class="
+          clsx(
+            'flex-none flex flex-row flex-wrap items-center gap-xs',
+            !showGrid && (mapRef?.selectedComponents.size ?? 0) > 0
+            ? 'pr-[440px]' // right padding to accommodate the ExploreGridTile in the Map view
+            : 'max-w-[100vw]', // prevents the buttons from going offscreen on smaller window sizes
+          )"
+        >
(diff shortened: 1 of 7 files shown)


app/web/src/newhotness/Explore.vue
```diff
@@ -156,48 +163,47 @@
...
-                    'border-neutral-400 hover:border-action-500',
-                    'border-neutral-600 hover:border-action-300',
-                  ),
-                  queryOnlyDiff
-                    ? themeClasses('bg-action-200', 'bg-action-900')
-                    : themeClasses('bg-neutral-100', 'bg-neutral-900'),
-                )
-              "
+            <NewButton 
+              tone="toggle"
```
wendybujalski, line 167: These custom buttons have been replaced with a new `toggle` Button tone.

jobelenus (dismissed):

jobelenus (approved):

PR #8332, opened 2026-01-22
chore: improve dal integration test performance

zacharyhamm:
Two changes here:

Adds arguments to the dal test macro that enable and disable specific
"servers" per test. By default, only the rebaser and pinga will be
started, and you can enable the rest with:

  enable_veritech
  enable_edda
  enable_forklift

Note that I have not enabled edda for any test, and all tests pass, so
it looks like we don't have any real integration testing for edda at the
moment.

The one you'll want the most is enable_veritech, which you should be
sure to enable any time you will have to execute functions.

This took about 10-30 seconds off my test runs.

And

In read_wait_for_memory we spin checking if the requested object has
landed in the memory cache (via the events stream) and then we look at
disk and durable storage. This adds a 2+ second overhead for many
fetches, which either are not about to land in the memory cache because
they are for old objects which have been evicted, or because we just
don't propagate data fast enough across NATS for it to land.

Checking the normal read path before trying to spin has a huge
performance result for the integration tests, easily 100 seconds faster
in CI, and will likely have a positive impact throughout the stack.


lib/dal/tests/integration_test/action/schema_level.rs
```diff
@@ -22,7 +22,7 @@ use dal_test::{
 };
 use pretty_assertions_sorted::assert_eq;
 
-#[test]
+#[test(enable_veritech)]
```
nickgerace, line 25: Okay I've wanted this idea for like ~3-4 years and I am SO happy to finally see it! Yes yes yes!

lib/si-test-macros/src/dal_test.rs
```diff
@@ -37,19 +37,30 @@ use crate::{
...
+    }
+    if args.should_start_server("pinga") {
+        expander.setup_start_pinga_server();
+    }
+    if args.should_start_server("edda") {
+        expander.setup_start_edda_server();
+    }
+    if args.should_start_server("rebaser") {
+        expander.setup_start_rebaser_server();
+    }
```
nickgerace, line 63: Ended up being simple! Nice.

lib/si-test-macros/src/lib.rs
```diff
@@ -26,12 +26,34 @@ const SPAN_EVENTS_ENV_VAR: &str = "SI_TEST_LOG_SPAN_EVENTS";
...
+
+    /// Check if a specific server should be started
+    pub(crate) fn should_start_server(&self, server_name: &str) -> bool {
+        if DEFAULT_SERVERS.contains(&server_name) {
+            !self.should_skip_server(server_name)
+        } else {
+            self.should_enable_server(server_name)
+        }
+    }
+}
```
nickgerace, line 55: Ah _this_ is it. I never took the time to get deep into the idea, but I'm glad it was relatively small in the end.

nickgerace (approved):

PR #8333, opened 2026-01-22
Fix: nav bar doesn't turn to light mode in light mode!

jobelenus:
So we need an "exception" tone for "dont do themeClass"
image


wendybujalski (approved):
While I'd personally prefer `useThemeContainer("dark")` as the solution to this problem to avoid adding another button tone, this solution works just fine 👍

jobelenus:
Yea, I'm not sure about a reasonable way to drill that down into a "force" for the themeClass evaluation 

> While I'd personally prefer `useThemeContainer("dark")` as the solution to this problem to avoid adding another button tone, this solution works just fine 👍

PR #8335, opened 2026-01-23
fix(dal): import func arguments during module update and fix PropPath…

aaron-dernley:
… conversion

Two fixes for module import:

  1. import_funcs_for_module_update was not importing func arguments after creating/updating funcs, causing MissingFuncArgument errors when attribute functions referenced arguments that didn't exist.

  2. create_attr_proto_arg was converting prop_path strings directly via .into() which stored the slash-separated path as a single part. PropPath uses \x0B (vertical tab) as the internal separator, so as_parts() would return the whole path as one element, causing ChildPropNotFoundByName errors. Now properly splits by '/' and constructs PropPath with the parts.

## How was it tested?

- [X] Integration tests pass
- [X] Manual test: Importing a clover generated schema with a new attribute func now works

- [X] No
- [ ] Yes, and this PR includes it
- [ ] Yes, and this PR does not include it (reasoning below)

### DIFF
--- a/lib/dal/src/pkg/import.rs
+++ b/lib/dal/src/pkg/import.rs
@@ -591,7 +591,7 @@ pub async fn import_funcs_for_module_update(
             }
         }
 
-        import_func(
+        let func = import_func(
             ctx,
             &func_spec,
             None,
@@ -600,6 +600,11 @@ pub async fn import_funcs_for_module_update(
             super::UpdateMode::UpdateExisting,
         )
         .await?;
+
+        let args: Vec<_> = func_spec.arguments()?.into_iter().collect();
+        if !args.is_empty() {
+            import_func_arguments(ctx, func.id, &args).await?;
+        }
     }
 
     Ok(thing_map)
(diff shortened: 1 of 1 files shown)


lib/dal/src/pkg/import.rs
```diff
@@ -600,6 +600,12 @@ pub async fn import_funcs_for_module_update(
             super::UpdateMode::UpdateExisting,
         )
         .await?;
+
+        // Import func arguments (this was missing and caused MissingFuncArgument errors)
```
sprutton1: nit: we shouldn't need this comment. The git blame and history will give us this context if we need to go find it

lib/dal/src/pkg/import.rs
```diff
@@ -1297,7 +1303,8 @@ pub async fn import_only_new_funcs(
             };
 
             // Find or create the func arguments for the provided spec.
-            for argument in func_spec.arguments()? {
+            let args: Vec<_> = func_spec.arguments()?.into_iter().collect();
+            for argument in args {
```
sprutton1: nit: curious if this change is necessary?

sprutton1 (dismissed):
Couple of nits, but otherwise lgtm

aaron-dernley:
Made those changes based on your feedback @sprutton1 - good catches, thank you.

stack72 (approved):

PR #8336, opened 2026-01-23
fix(clover): cloudformation only asset overhaul

aaron-dernley:
This PR restructures how CloudFormation-only AWS assets (those without a Cloud Control read handler) are generated in Clover. Instead of directly exposing CloudFormation properties at the domain root, they are now nested under a properties object, with a computed CloudFormationResourceBody output that generates the full CF resource structure.

  Changes

  New Files:
  - funcs/attribute/awsCfOnlyAttr.ts - Attribute function that combines properties and extra into a CloudFormation resource body JSON
  - funcs/code-gen/awsCfOnlyCodeGen.ts - Codegen function that wraps the resource body in a full CF template
  - funcs/qualifications/awsCfOnlyLint.ts - Qualification function that validates CF structure and optionally runs cfn-lint

  Modified Files:
  - funcs.ts - Added awsCfOnlyCodeGen to CODE_GENERATION_FUNC_SPECS and new QUALIFICATION_FUNC_SPECS with awsCfOnlyLint
  - pruneCfAssets.ts - Complete overhaul to:
    - Restructure domain props: wrap CF props under properties, keep extra as sibling
    - Add CloudFormationResourceBody JSON prop with attribute function binding
    - Add AwsResourceType hidden prop to extra with the schema name
    - Replace leaf functions with new CF-only codegen and qualification functions
  - overrides.ts - Updated Route53::RecordSet function hashes

  Route53::RecordSet Overrides (updated for new domain structure):
  - actions/create.ts - Read from domain.properties
  - actions/refresh.ts - Read from domain.properties, return payload with nested properties
  - management/import.ts - Output domain: { properties: {...} }
  - management/discover.ts - Output domain: { properties: {...} }

  Domain Structure (Before → After)

  Before:
```
  domain/
    Name
    Type
    TTL
    ...
    extra/
      Region
```
  After:
```
  domain/
    properties/
      Name
      Type
      TTL
      ...
    extra/
      Region
      AwsResourceType (hidden)
      CloudFormationOnly (hidden)
    CloudFormationResourceBody (computed)
```
  Impact

  Affects ~200 CloudFormation-only AWS assets. Route53::RecordSet has custom overrides that were updated to work with the new structure.

## How was it tested?

- [X] Integration tests pass
- [X] Imported some assets manually and verified the shape, attribute func, codgen and qualification funcs.

- [X] No
- [ ] Yes, and this PR includes it
- [ ] Yes, and this PR does not include it (reasoning below)


aaron-dernley:
https://github.com/systeminit/si/pull/8336#issuecomment-3790335916

Seems to have picked up the same changes as the nightly run for some reason, that PR was merged a few hours ago.

bin/clover/src/pipelines/aws/funcs.ts
```diff
@@ -66,6 +66,25 @@ export const CODE_GENERATION_FUNC_SPECS = {
...
+  },
+  awsCfOnlyLint: {
+    id: "dba872348ce39bfa8c15002037263b2528d81833c6c1f24fef5c9bc4115145b4",
+    backendKind: "jsAttribute",
+    responseType: "qualification",
+    displayName: "CloudFormation Lint Validation",
+    path: "./src/pipelines/aws/funcs/qualifications/awsCfOnlyLint.ts",
+  },
+} as const satisfies Record<string, FuncSpecInfo>;
+
```
sprutton1, line 87: It's weird to me that these are here but the attribute func is in the other file. They should all be in the same place, here or there.

bin/clover/src/pipelines/aws/pipeline-steps/pruneCfAssets.ts
```diff
@@ -88,7 +118,157 @@ function createAttributeFunc() {
...
+    ) {
+      p.metadata.propPath = [...newPrefix, ...oldPath.slice(oldPrefix.length)];
+    }
+  });
+}
+
+/**
+ * Creates a JSON prop with CodeEditor widget
+ */
+function createJsonProp(
```
sprutton1, line 238: Would this be better alongside the other generic "create this kind of prop" methods we have?

bin/clover/src/pipelines/aws/pipeline-steps/pruneCfAssets.ts
```diff
@@ -25,28 +34,56 @@ export function pruneCfAssets(specs: ExpandedPkgSpec[]): ExpandedPkgSpec[] {
...
-    }
+    const cfOnlyProp = createScalarProp(
+      "CloudFormationOnly",
+      "boolean",
+      extraProp.metadata.propPath,
+      false,
+    );
+    cfOnlyProp.data.defaultValue = "true";
+    cfOnlyProp.data.hidden = true;
+    extraProp.entries.push(cfOnlyProp);
```
sprutton1, line 86: Maybe just do this inside of restructure since you already create it there. I think we have a `getOrCreate()` method somewhere. We should, at least

